PR #21, opened 2020-02-26
Get Ethereum Relay to Compile

### LINKED ISSUE #17: Compile PoA relay
Seems that currently the PoA <-> Substrate relay code does not compile on the CI, we should get that sorted out,.

HCastano:
A few things still TODO:
- Check that I made the correct changes w.r.t `jsonrpsee`
- Add a node template
- Update Substrate dependencies once that stabilizes a bit more

Closes #17.


bin/node/node/Cargo.toml
```diff
@@ -15,6 +15,7 @@ futures = "0.3.1"
 log = "0.4.8"
 structopt = "0.3.8"
 bridge-node-runtime = { version = "0.1.0", path = "../runtime" }
+sp-bridge-eth-poa = { version = "0.1.0", path = "../../../primitives/ethereum-poa" }
```
tomusdrw: I don't think we should name crates `sp-`. It stands for "Substrate Primitives", and that's actually Bridge Primitive and not Substrate Primitive :)

HCastano: Yeap, I was thinking about that actually. Same with the leading `pallet-` since that's Substrate specific.

Issue #20, opened 2020-02-21
Update Substrate dependencies to use crates.io versions


HCastano:
With the v2.0 freeze that happened in Substrate today we'll hopefully be seeing a lot of crates get published to crates.io soon. We'll need to update our dependencies to use that instead of pulling from git.

tomusdrw:
we already do that I believe.

svyatonik:
> we already do that I believe.

I'm not sure - Wwe still have something like that all across the codebase:
```toml
[dependencies.frame-support]
version = "2.0.0-rc6"
tag = 'v2.0.0-rc6'
default-features = false
git = "https://github.com/paritytech/substrate/"
```

tomusdrw:
I've tried that and it seems that some crates are not really released. It failed for me at `substrate-test-runtime-client` (CC @gnunicorn )
Afaict it means we can't really use crates.io versions anywhere else because of conflicts.

svyatonik:
We may just remove this dependency - it isn't super-important. I guess the other missing dependency will be `node-primitives`, but we should avoid using it. I'm introducing `millau-primitives` and `rialto-primitives` in #376 anyway.

But generally, imo, we should switch to crates.io anyway (sooner or later). If we're unable to reuse our own crates from crates.io, then noone can't :)

gnunicorn:
> It failed for me at substrate-test-runtime-client (CC @gnunicorn )

For which crates? I have a todo to fix the releasing process to publish everything that is _not_ marked `publish = false` rather than matching specific names (as we do now). It seems that at least one or two matches were false and we didn't release a package or two - which I then added after. If you have any other misses, let me know.

> I guess the other missing dependency will be `node-primitives`

Yeah, we are currently not releasing the binaries stuff, mainly because it changes soo often that it feels futile to even try to keep that semver compliant. That said, I have the plan to release at least the primitives and runtimes for both node and polkadot at some point, too.

Either way, even with that dependency missing, if you were to only pull that in via git (using the `v2.0.0`-tag) and everything else from crates.io _should_ be working. Please let me know if not.

PR #22, opened 2020-02-28
Add a Node

HCastano:
Adds a node to use for testing out our bridge runtimes. It'll first be used by #21.

I have all of the Substrate dependencies pinned to `v2.0.0-alpha.2` at specific commit. I'm okay getting this merged in like that and updating it once everything ends up on crates.io.


HCastano:
I need to get the WASM toolchain in the CI to make Travis happy, but I'll do that tomorrow(ish).

bin/node/node/src/chain_spec.rs
```diff
@@ -0,0 +1,150 @@
...
+	AccountId, AuraConfig, BalancesConfig, GenesisConfig, GrandpaConfig,
+	SudoConfig, SystemConfig, WASM_BINARY, Signature
+};
+use sp_consensus_aura::sr25519::{AuthorityId as AuraId};
+use grandpa_primitives::{AuthorityId as GrandpaId};
+use sc_service;
+use sp_runtime::traits::{Verify, IdentifyAccount};
+
+// Note this is the URL for the telemetry server
+//const STAGING_TELEMETRY_URL: &str = "wss://telemetry.polkadot.io/submit/";
```
tomusdrw: Remove?

bin/node/node/src/chain_spec.rs
```diff
@@ -0,0 +1,150 @@
...
+
+/// The chain specification option. This is expected to come in from the CLI and
+/// is little more than one of a number of alternatives which can easily be converted
+/// from a string (`--chain=...`) into a `ChainSpec`.
+#[derive(Clone, Debug)]
+pub enum Alternative {
+	/// Whatever the current runtime is, with just Alice as an auth.
+	Development,
+	/// Whatever the current runtime is, with simple Alice/Bob auths.
+	LocalTestnet,
```
tomusdrw: Do we really need both initially?

bin/node/runtime/src/lib.rs
```diff
@@ -0,0 +1,353 @@
+//! The Substrate Node Template runtime. This can be compiled with `#[no_std]`, ready for Wasm.
```
tomusdrw: Could we extend `polkadot/kusama` runtime instead? Perhaps use common stuff from their repo?

One thing that might not work is that we will end up in git dependency hell if we don't solve #20 first.

HCastano: Is there anything in particular that you'd like to get out from their runtime?

bin/node/node/src/chain_spec.rs
HCastano: Guess not, I'll remove the `LocalTestnet` variant

bin/node/runtime/src/lib.rs
HCastano: Talked about this off Github. What we'll do is keep the template for now, and once we have made more substantial progress with the runtimes we'll add a Kusama/Polkadot node for testing.

PR #34, opened 2020-03-18
Reward submitters only when submitted block is finalized

svyatonik:
Before this PR, the assumption was that the submitter would be rewarded when he's submitting block and the block is valid AND unknown to the runtime. However, this could lead to situation when some authority generate multiple valid blocks at its slot and submits all these blocks to the runtime. All blocks are valid and unknown and authority gets rewarded => that's a breach.

The obvious fix would be to reject second block with the same number from the same authority, but that could lead to situation when we are rejecting block that will actually be finalized later => te synchronization will stuck.

The other option is to change reward scheme. My proposal is to reward only when submitted block is finalized. The authority could still submit these headers, but: (1) it'll pay fee for transactions (2) it won't get rewarded. In this scheme, submitters won't get rewards for providing uncles, but since uncles are actually useless for the runtime, I think this is fine.

I have also left previous callbacks - `on_valid_headers_submitted` and `on_invalid_headers_submitted`. The provider could still implement some scheme for (additionally) penalizing submitters that are providing 'good bad' headers. 

The fix doesn't change number of storage reads/writes - it only adds `Option<Submitter>` field to  data that has already been read/written to the storage.

### DIFF
--- a/modules/ethereum/src/lib.rs
+++ b/modules/ethereum/src/lib.rs
@@ -647,6 +647,8 @@ pub(crate) mod tests {
 	use primitives::{rlp_encode, H520};
 	use std::collections::{hash_map::Entry, HashMap};
 
+	type AccountId = u64;
+
 	pub fn genesis() -> Header {
 		Header {
 			seal: vec![vec![42].into(), vec![].into()],
@@ -702,7 +704,7 @@ pub(crate) mod tests {
 		best_block: (u64, H256, U256),
 		finalized_block: (u64, H256),
 		oldest_unpruned_block: u64,
-		headers: HashMap<H256, StoredHeader<u64>>,
+		headers: HashMap<H256, StoredHeader<AccountId>>,
 		headers_by_number: HashMap<u64, Vec<H256>>,
 		next_validators_set_id: u64,
 		validators_sets: HashMap<u64, (H256, Vec<Address>)>,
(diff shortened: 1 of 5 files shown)


modules/ethereum/src/lib.rs
```diff
@@ -179,18 +187,23 @@ impl ImportContext {
...
+	///
+	/// Returns header and its submitter (if known).
+	fn header(&self, hash: &H256) -> Option<(Header, Option<Self::Submitter>)>;
 	/// Get header import context by parent header hash.
 	fn import_context(&self, parent_hash: &H256) -> Option<ImportContext>;
 	/// Get new validators that are scheduled by given header.
 	fn scheduled_change(&self, hash: &H256) -> Option<Vec<Address>>;
 	/// Insert imported header.
-	fn insert_header(&mut self, header: HeaderToImport);
+	fn insert_header(&mut self, submitter: Option<Self::Submitter>, header: HeaderToImport);
```
tomusdrw: Why isn't `submitter` part of `HeaderToImport`? I feel that this would look a bit nicer.

modules/ethereum/src/lib.rs
```diff
@@ -651,7 +702,7 @@ pub(crate) mod tests {
 		best_block: (u64, H256, U256),
 		finalized_block: (u64, H256),
 		oldest_unpruned_block: u64,
-		headers: HashMap<H256, StoredHeader>,
+		headers: HashMap<H256, StoredHeader<u64>>,
```
tomusdrw: Could we get a typedef `type AccountId = u64`? Now it's a bit confusing to see `u64` for both block numbers and submitters. I know it's only test code, but still.

PR #37, opened 2020-03-18
Run RustFmt as part of the CI

HCastano:


.travis.yml
```diff
@@ -6,6 +6,9 @@ rust:
 before_install:
   - rustup install nightly
   - rustup target add wasm32-unknown-unknown --toolchain nightly
+  - rustup component add rustfmt
+script:
+  - cargo fmt --all -- --check
```
tomusdrw: When you override `script` you need to add `cargo test --all` as well, otherwise the build is running only `fmt` check (but it's blazingly fast, have to admit :D)

Alternatively we could run fmt at `before_script` or sth?

tomusdrw:
Thanks :bowing_man: :)

.travis.yml
HCastano: Whoops, didn't realize that, haha. I think `before_script` will be better as that way we won't have to manually add the `build` and `test` commands.

Issue #38, opened 2020-03-19
Only accept headers-in-unsigned-transactions when we know we're almost at the tip of PoA chain


svyatonik:
extracted point(4) from #26, because it'll need another round of research:
...
4) one other restriction that we may introduce: we know approximate block time of PoA (it is stepDuration...stepDuration * maximumEmptySteps) - what if we only accept unsigned transactions when we believe we're at the tip of PoA chain (i.e. when timestampModule.timestamp - BestKnownPoAHeader.timestamp < N * approximateBlockTime)?
...

Need to be caution with computing approximate block time, because (c): "there were some issues recently on network (running some super old PoA), where after upgrade the current step and "expected step" did not really fully match. I.e. due to some time drifts, validators being down, etc the step calculated from stepDuration did not match the actual step in the network. Maybe if we make it relative, i.e. using the step from the last known block as a reference point, this could work"

svyatonik:
Another possible optimization: https://github.com/paritytech/parity-bridges-common/pull/45#discussion_r402970795

tomusdrw:
Closing, since we don't plan to actively work on the PoA bridge in the near future. Happy to re-open if the outlook changes.

PR #45, opened 2020-03-23
Sync ethereum headers using unsigned (substrate) transactions

### LINKED ISSUE #26: Explore sync-headers-using-unsigned-transactions option in PoA bridge
What I learned from talks with Tomek: switching from signed transactions to unsigned would allow us to deduplicate on txpool/network (since all txs for block B will have the same hash H). This, however, brings some problems:
1) we can't reward submitters for providing correct blocks - this could be solved by switching to reward-for-messages scheme, because to sync messages submitters would need to sync headers first;
2) we can't submit several blocks in single transaction - the hashes would be different.

I'd like to explore this option && see:
1) how many sync-header transactions we could verify in a sec - unsigned transactions need to be verified before they accepted to tx pool;
2) whether using unsigned transactions would lead to slower sync (because we can't provide multiple blocks in single tx), or not.

svyatonik:
closes #26 

I've also updated relay to support 3 modes (to be able to test this + to be able to use it as 'backup' option):
`./ethereum-poa-relay --sub-tx-mode=signed`: default mode, sync using signed transactions.
`./ethereum-poa-relay --sub-tx-mode=unsigned`: sync using unsigned transactions (slow when compared to signed, because we allow max 10 transactions (read: eth headers) in pool)
`./ethereum-poa-relay --sub-tx-mode=backup`: sync using signed transactions, but only when we do not see any new eth headers in substrate chain for 5 minutes (it should work like burst: if no new headers => submit 128 eth headers to substrate, then wait again for 5min delay).

I've extracted #38 into separate issue - will do that next


svyatonik:
For those who want to test it, I'm using following script:
```bash
#!/bin/bash
set -e

rm -rf substrate.db
rm -rf parity.db

cargo build --manifest-path=../parity-bridges-common/Cargo.toml -p bridge-node
cp ../parity-bridges-common/target/debug/bridge-node .
cargo build --manifest-path=../parity-bridges-common/Cargo.toml -p ethereum-poa-relay
cp ../parity-bridges-common/target/debug/ethereum-poa-relay .

unbuffer ./parity -d parity.db --chain kovan --no-warp --no-persistent-txqueue --jsonrpc-apis=all 2>&1 | unbuffer -p gawk '{ print strftime("Parity: [%Y-%m-%d %H:%M:%S]"), $0 }' | unbuffer -p tee parity.log&
unbuffer ./bridge-node --dev -d parity.db 2>&1 | unbuffer -p gawk '{ print strftime("Substrate: [%Y-%m-%d %H:%M:%S]"), $0 }' | unbuffer -p tee substrate.log&
sleep 10
unbuffer ./ethereum-poa-relay --sub-tx-mode=unsigned 2>&1 | unbuffer -p gawk '{ print strftime("Bridge: [%Y-%m-%d %H:%M:%S]"), $0 }' | unbuffer -p tee bridge.log&

```

modules/ethereum/src/lib.rs
```diff
@@ -68,6 +74,16 @@ pub struct AuraConfiguration {
 	pub maximum_extra_data_size: u64,
 }
 
+/// Transaction pool configuration.
+///
+/// This is used to limit number of unsigned headers transactions in
+/// the pool. We never use it to verify signed transactions.
+pub struct PoolConfiguration {
+	/// Maximal difference between number of header from unsigned transaction
+	/// and current best block. This must be selected with caution - we can't
```
HCastano: We can't change it?

modules/ethereum/src/lib.rs
```diff
@@ -83,6 +99,33 @@ pub struct StoredHeader<Submitter> {
 	/// this is the set that has produced the block itself.
 	/// The hash is the hash of block where validators set has been enacted.
 	pub next_validators_set_id: u64,
+	/// Hash of the last block which has **SCHEDULED** validators set change.
+	/// Note that signal doesn't mean that the set has been (or ever will be) enacted.
+	/// Note that the header may already be pruned.
```
HCastano: What are the implications if the header has already been pruned?

modules/ethereum/src/lib.rs
```diff
@@ -83,6 +99,33 @@ pub struct StoredHeader<Submitter> {
...
+pub struct ValidatorsSet {
+	/// Validators of this set.
+	pub validators: Vec<Address>,
+	/// Hash of the block where this set has been signalled. None if this is the first set.
+	pub signal_block: Option<H256>,
+	/// Hash of the block where this set has been enacted.
+	pub enact_block: H256,
+}
+
+/// Validators set change as it is stored in the runtime storage.
```
HCastano: Looks like this was copied from the struct above :P

modules/ethereum/src/lib.rs
```diff
@@ -100,7 +143,7 @@ pub struct HeaderToImport<Submitter> {
 	/// Total chain difficulty at the header.
 	pub total_difficulty: U256,
 	/// Validators set enacted change, if happened at the header.
-	pub enacted_change: Option<Vec<Address>>,
+	pub enacted_change: Option<(Option<H256>, Vec<Address>)>,
```
HCastano: So is this tuple `(Header at which change was enacted, authority set enacted by this change)`?

modules/ethereum/src/lib.rs
```diff
@@ -112,28 +155,13 @@ pub struct ImportContext<Submitter> {
...
 	parent_total_difficulty: U256,
-	next_validators_set_id: u64,
-	next_validators_set: (H256, Vec<Address>),
+	parent_scheduled_change: Option<ScheduledChange>,
+	validators_set_id: u64,
+	validators_set: ValidatorsSet,
+	last_signal_block: Option<H256>,
 }
 
 impl<Submitter> ImportContext<Submitter> {
```
HCastano: Github won't let me comment on the struct since it's not part of the PR, but I'm not sure what the `ImportContext` represents. Can you clarify that for me?

modules/ethereum/src/lib.rs
```diff
@@ -149,19 +177,24 @@ impl<Submitter> ImportContext<Submitter> {
 		&self.parent_total_difficulty
 	}
 
+	/// Returns true if parent header has signalled validators set change.
```
HCastano:
```suggestion
	/// Returns the validator set change if the parent header has signaled a change.
```

modules/ethereum/src/lib.rs
```diff
@@ -251,9 +285,25 @@ pub trait Trait: frame_system::Trait {
 
 decl_module! {
 	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
-		/// Import Aura chain headers. Ignores non-fatal errors (like when known
-		/// header is provided), rewards for successful headers import and penalizes
-		/// for fatal errors.
+		/// Import single Aura header. Requires transaction to be **UNSIGNED**.
+		pub fn unsigned_import_header(origin, header: Header, receipts: Option<Vec<Receipt>>) {
```
HCastano: Why don't we have a function to import a chain of unsigned headers just like we do with signed headers? It seems like weird to me that the APIs don't totally match up

modules/ethereum/src/verification.rs
```diff
@@ -32,38 +32,183 @@
...
 use crate::error::Error;
 use crate::validators::step_validator;
-use crate::{AuraConfiguration, ImportContext, Storage};
+use crate::{AuraConfiguration, ImportContext, PoolConfiguration, ScheduledChange, Storage};
+use codec::Encode;
 use primitives::{public_to_address, Address, Header, SealedEmptyStep, H256, H520, U128, U256};
 use sp_io::crypto::secp256k1_ecdsa_recover;
+use sp_std::{vec, vec::Vec};
+
+/// Checks that we are able to ***try to** import this header.
```
HCastano:
```suggestion
/// Pre-check to see if should try and import this header.
```

modules/ethereum/src/verification.rs
```diff
@@ -32,38 +32,183 @@
...
 		.import_context(submitter, &header.parent_hash)
 		.ok_or(Error::MissingParentBlock)?;
-	let validators = context.validators();
+	let header_step = contextual_checks(config, &context, None, header)?;
+	validator_checks(config, &context.validators_set().validators, header, header_step)?;
+
+	Ok(context)
+}
+
+/// Perform basic checks that only require header iteself.
```
HCastano:
```suggestion
/// Perform basic checks that only require header itself.
```

modules/ethereum/src/verification.rs
```diff
@@ -32,38 +32,183 @@
...
+pub fn accept_aura_header_into_pool<S: Storage>(
+	storage: &S,
+	config: &AuraConfiguration,
+	pool_config: &PoolConfiguration,
+	header: &Header,
+) -> Result<(Vec<Vec<u8>>, Vec<Vec<u8>>), Error> {
+	// check if we can verify further
+	let (hash, _) = is_importable_header(storage, header)?;
+
+	// we can always do contextless checks
```
HCastano:
```suggestion
	// Perform basic checks that only require header
```

modules/ethereum/src/verification.rs
```diff
@@ -32,38 +32,183 @@
...
+	let (best_number, best_hash, _) = storage.best_block();
+	let difference = header.number.saturating_sub(best_number);
+	if difference > pool_config.max_future_number_difference {
+		return Err(Error::UnsignedTooFarInTheFuture);
+	}
+
+	// TODO: only accept new headers when we're at the tip of PoA chain
+	// https://github.com/paritytech/parity-bridges-common/issues/38
+
+	// we do not want to see more at most one header with given number from single authority
```
HCastano:
```suggestion
	// we do not want to see more than at most one header with given number from single authority
```

modules/ethereum/src/verification.rs
```diff
@@ -32,38 +32,183 @@
...
+	// => every header is providing tag (block_number + authority)
+	// => since only one tx in the pool can provide the same tag, they're auto-deduplicated
```
HCastano: This is only for the pool of unsigned transactions right? Also, are "tags" specific to unsigned transactions?

modules/ethereum/src/verification.rs
```diff
@@ -32,38 +32,183 @@
...
+	let prv_number_and_authority_tag = (header.number, header.author).encode();
```
HCastano: What's `prv`?

modules/ethereum/src/verification.rs
```diff
@@ -32,38 +32,183 @@
...
+			))
+		}
+		None => {
+			// we know nothing about parent header
+			// => the best thing we can do is to believe that there are no forks in
+			// PoA chain AND that the header is produced either by previous, or next
+			// scheduled validators set change
+			let header_step = header.step().ok_or(Error::MissingStep)?;
+			let best_context = storage.import_context(None, &best_hash).expect(
+				"import context is None only when header is missing from the storage;\
```
HCastano: What if there is no best header because this is the first block we're importing?

modules/ethereum/src/verification.rs
HCastano: Oh, is it previous? I would at least put `prev`.

modules/ethereum/src/lib.rs
svyatonik: It depends where config comes from - if it is hardcoded (like now), then it can be changed by changing runtime. We an put it to storage, to allow changing it without updating runtime. Could you, please, elaborate - why and when do you want to change it?

modules/ethereum/src/lib.rs
svyatonik: You can't read header by the hash :)

modules/ethereum/src/lib.rs
svyatonik: It is `Validators set change` and struct above is `Validators set`.

modules/ethereum/src/lib.rs
svyatonik: There are 2 groups of header verification rules. Rules from first group can be checked just by looking at the header itself, without knowing its parent and blockchain state (not the regular blockchain state - just something that is associated with parent header). 2nd group requires this context. `ImportContext` meant to store everything we need for checking 2nd group rules.

modules/ethereum/src/lib.rs
svyatonik: One of good things that we're getting with unsigned transactions is that they're auto-deduplicated by the txpool/network. I.e. now unsigned transaction is just a `Call(unsigned_import_header(Header, Option<Vec<Receipt>>)).encode` without any additional data. So if we have that transaction in the pool AND someone sends us the same transaction again, we'll see that the hash is the same => won't bother with checks, etc. If we would allow to submit several header at once (without any strict rules on how many/when/...), then we'll get a case when actor#1 which submit tx1 with headers (101, 102, 103), actor#2 which submits tx2 with headers (101, 102), actor#3 which submits tx3 with headers (101). All these transactions have different hashes && `validate_unsigned()` would be called. Then we have a problems with `validate_unsigned()` - which transactions to accept, and which to reject - if we'll accept all transactions, then we would allow to spam blocks without any payment (header 101 is provided by every transaction). If we'll accept single transaction, then we may miss these additional headers.

That said, having multiple headers in single unsigned transaction is possible if we'll enforce some rules that will reintroduce deduplication - like we only accept 16 headers (starting with number N so that N % 16 == 0) in single unsigned transaction when we know we are behind the chain tip and only 1 header otherwise. But we won't have the same API anyway. And I have a feeling that this optimization will reveal another bunch of problems that we'll need to solve. So let's not introduce it here for now.

While writing this, I've realized that I'm not checking if `receipts` are provided (when required) in `validate_unsigned()`. And I need to, because otherwise: (1) we can have two transactions for the same header with different hashes (2) transaction that is missing required `receipts` could push good transaction out of pool (2) this allow submitter to spam blocks with unneeded data. So please do not resolve this comment - I'll fix it

modules/ethereum/src/verification.rs
svyatonik: Yes - this function is called only for unsigned transactions. I think tags are also valid for signed transactions, but they're provided at some other level (probably when runtime checks if we're able to dispatch call?). Better ask @tomusdrw .

And another question for @tomusdrw  - how's pool/runtime deals with tags collisions? Or it is the runtime author problem (then we probably want to include method name in the tag)? Like if I have two completely unrelated methods (one is `endow_account()` and the second one is `update_timestamp()`). Both are providing/require tags which are encoded `u32` => could it be that (by occasion) I'll move future `endow_account()` transaction to ready queue when some `update_timestamp()` transaction is mined (because it provides required tag (`42u32.encode()`)? TBH I wasn't thinking about that yet, so my tags are simply something like `(number, hash).encode()`.

modules/ethereum/src/verification.rs
svyatonik: There's always `initial header` (which points to Kovan genesis in our node). Initial header is the best header...initially :)

modules/ethereum/src/lib.rs
svyatonik: OK - so in addition to verifying that receipts are provided when required, we also need to verify that receipts are valid. Hopefully, this wouldn't happen too often - normally it only comes with validators set change. But not sure how often we'll see false positives.

modules/ethereum/src/lib.rs
svyatonik: No :) The change is enacted by this (importing) header. So we need pointer to the block where this change has been scheduled. I've updated comment. Could replace with struct if you feel it would be better.

modules/ethereum/src/verification.rs
svyatonik: I've changed `req_` to `requires_` and `prv_` to `provides_`

modules/ethereum/src/lib.rs
HCastano: I left the comment because the sentence seemed incomplete.

modules/ethereum/src/lib.rs
HCastano: I think typedef-ing it would be cleaner

modules/ethereum/src/lib.rs
HCastano: Okay, so to be clear the context is  extra information needed to validate that the current header is valid. In this case the context needed is information associated with the current header's parent, yes?

I know it's not part of the PR, but I think having a more descriptive comment as part of the `ImportContext` struct would be helpful :)

svyatonik: Looks true :)

Could you suggest documentation that you'd like to see? I do not want something like "it is extra information for..." or "it has everything required to...". But I can't suggest something other than "It is used to verify header within blockchain context", which I believe isn't something you wanted to see there :)

HCastano: Sure. What do you think of this?

```
/// Header import context.
///
/// The import context contains information needed by the header verification
/// pipeline which is not directly part of the header being imported. This includes
/// information relating to its parent, and the current validator set (which
/// provide _context_ for the current header).
```

modules/ethereum/src/error.rs
```diff
@@ -33,45 +33,49 @@
 use sp_runtime::RuntimeDebug;
 
 /// Header import error.
-#[derive(RuntimeDebug)]
+#[derive(Clone, Copy, RuntimeDebug)]
 #[cfg_attr(feature = "std", derive(PartialEq))]
 pub enum Error {
 	/// The header is beyound last finalized and can not be imported.
```
HCastano:
```suggestion
	/// The header is beyond last finalized and can not be imported.
```

modules/ethereum/src/lib.rs
```diff
@@ -370,6 +442,38 @@ impl<T: Trait> Module<T> {
...
+					&kovan_aura_config(),
+					&kovan_validators_config(),
+					&pool_configuration(),
+					header,
+					receipts.as_ref(),
+				);
+
+				match accept_result {
+					Ok((requires, provides)) => Ok(ValidTransaction {
+						priority: TransactionPriority::max_value(),
```
tomusdrw, line 462: I'd be more conservative here so that it can still be tuned a bit in case there are multiple transactions in the pool.

I'm drafting an issue to Substrate to suggest that all unsigned transactions should always have a `BasePriority` configurable on the runtime level. That way when you have multiple pallets sending unsigned transactions you can make sure they are correctly ordered in the pool.

So I'd suggest here it should be:
```rust
trait Trait {
 ...
  type UnsignedTransactionBasePriority: Get<TransactionPriority>;
  ...
}
...
ValidTransaction {
   priority: T::UnsignedTransactionBasePriority::get() + height,
   ...
}
```

And then we should set the BasePriority to something like `1 << 20`.

modules/ethereum/src/lib.rs
```diff
@@ -370,6 +442,38 @@ impl<T: Trait> Module<T> {
...
+						requires,
+						provides,
+						longevity: TransactionLongevity::max_value(),
```
tomusdrw, line 465: Let's make it configurable from outside. I think it should have some more sane validity that depends on the block times of both blockchains.

modules/ethereum/src/lib.rs
```diff
@@ -370,6 +442,38 @@ impl<T: Trait> Module<T> {
...
+						propagate: true,
+					}),
+					Err(error::Error::UnsignedTooFarInTheFuture) => UnknownTransaction::CannotLookup.into(),
```
tomusdrw, line 468: I think you can use `Custom` variant here. Also it is not going to be banned - I think it's worth a comment why this behaviour is desired here.

modules/ethereum/src/lib.rs
```diff
@@ -742,17 +850,87 @@ pub(crate) mod tests {
...
+				ValidatorsSet {
+					validators: finalized_set,
+					signal_block: None,
+					enact_block: self.headers_by_number[&0][0],
+				},
+			);
+
+			self.headers
+				.get_mut(&self.headers_by_number[&number][0])
+				.unwrap()
```
tomusdrw, line 908: Should be `expect`?

modules/ethereum/src/lib.rs
```diff
@@ -742,17 +850,87 @@ pub(crate) mod tests {
...
+				.next_validators_set_id = set_id;
+			if let Some(signalled_set) = signalled_set {
+				self.headers
+					.get_mut(&self.headers_by_number[&number][0])
+					.unwrap()
```
tomusdrw, line 913: Extracting for the second time, and another `unwrap`. Maybe lift to a variable?

modules/ethereum/src/lib.rs
```diff
@@ -742,17 +850,87 @@ pub(crate) mod tests {
...
+				set_id,
+				ValidatorsSet {
+					validators: finalized_set,
+					signal_block: None,
+					enact_block: self.headers_by_number[&0][0],
+				},
+			);
+
+			self.headers
+				.get_mut(&self.headers_by_number[&number][0])
```
tomusdrw, line 907: Why do we do that only for block `0` on the list? Is it guaranteed to be canonical? Shouldn't we do that for every element on that list?

modules/ethereum/src/lib.rs
```diff
@@ -742,17 +850,87 @@ pub(crate) mod tests {
...
+				.unwrap()
+				.next_validators_set_id = set_id;
+			if let Some(signalled_set) = signalled_set {
+				self.headers
+					.get_mut(&self.headers_by_number[&number][0])
+					.unwrap()
+					.last_signal_block = Some(self.headers_by_number[&(number - 1)][0]);
```
tomusdrw, line 914: Possible overflow. I suppose a simple `ensure!` at the beginning to make sure this function is not called for genesis block would prevent errors and help debugging the issue (shall it happen :))

modules/ethereum/src/lib.rs
```diff
@@ -785,19 +963,25 @@ pub(crate) mod tests {
 			parent_hash: &H256,
 		) -> Option<ImportContext<Self::Submitter>> {
 			self.headers.get(parent_hash).map(|parent_header| {
-				let (next_validators_set_start, next_validators) =
-					self.validators_sets.get(&parent_header.next_validators_set_id).unwrap();
+				let validators_set = self
+					.validators_sets
+					.get(&parent_header.next_validators_set_id)
+					.unwrap()
```
tomusdrw, line 969: Another `unwrap`. I just realised that it's all only test code so don't mind the previous comments. Deduplication already logged in https://github.com/paritytech/parity-bridges-common/issues/46

modules/ethereum/src/verification.rs
```diff
@@ -31,39 +31,205 @@
...
+		_ => (),
+	}
+
+	// we do not want to have all future headers in the pool at once
+	// => if we see header with number > maximal ever seen header number + LIMIT,
+	// => we consider this transaction invalid, but only at this moment (we do not want to ban it)
+	// => let's mark it as Unknown transaction
+	let (best_number, best_hash, _) = storage.best_block();
+	let difference = header.number.saturating_sub(best_number);
+	if difference > pool_config.max_future_number_difference {
```
tomusdrw, line 90: We should use `difference` to prioritise transactions in the pool. I.e. the ones with higher difference should have priority bump.

The resoning behind it would be, that if we ever have multiple bridges included in one runtime, we want the bridges that are the most behind to be processed earlier.

modules/ethereum/src/verification.rs
```diff
@@ -31,39 +31,205 @@
...
+	// we want to see at most one header with given number from single authority
+	// => every header is providing tag (block_number + authority)
+	// => since only one tx in the pool can provide the same tag, they're auto-deduplicated
+	let provides_number_and_authority_tag = (header.number, header.author).encode();
+
+	// we want to see several 'future' headers in the pool at once, but we may not have access to
+	// previous headers here
+	// => we can at least 'verify' that headers comprise a chain by providing and requiring
+	// tag (header.number, header.hash)
+	let provides_header_number_and_hash_tag = (header.number, hash).encode();
```
tomusdrw, line 106: Please add some prefix (module name) to all tags to avoid conflicts with other pallets, like so:
```
(b"EthereumBridge", header.number, hash).encode()
```

modules/ethereum/src/verification.rs
tomusdrw: Re 1: The tags for signed transactions are generated by `frame_system` based on the sender and nonce (see `CheckNonce` `SignedExtension`).

Re 2: Indeed, we need to guarantee uniqueness of `provides` tags to make sure they don't conflict between pallets and function calls. Simply including some pallet-specific prefix as string should do. Suggested in another comment.

modules/ethereum/src/verification.rs
```diff
@@ -31,39 +31,205 @@
...
+	if header.gas_limit > config.max_gas_limit {
+		return Err(Error::InvalidGasLimit);
+	}
+	if header.number != 0 && header.extra_data.len() as u64 > config.maximum_extra_data_size {
+		return Err(Error::ExtraDataOutOfBounds);
+	}
+
+	// we can't detect if block is from future in runtime
+	// => let's only do an overflow check
+	if header.timestamp > i32::max_value() as u64 {
```
tomusdrw, line 205: We could actually check the `timestamp` module if available? I think it's worth adding as an (optional) feature via associated type. Logged as #51

modules/ethereum/src/verification.rs
```diff
@@ -176,13 +321,55 @@ fn verify_signature(expected_validator: &Address, signature: &H520, message: &H2
...
+	header: &Header,
+) -> Option<Vec<Address>> {
+	// that's the earliest block number we may met in following loop
+	// it may be None if that's the first set
+	let best_set_signal_block = context.validators_set().signal_block;
+
+	// if parent schedules validators set change, then it may be our set
+	// else we'll start with last known change
+	let parent_schedules_change = context.parent_scheduled_change();
+	let mut current_set_signal_block = match parent_schedules_change {
```
tomusdrw, line 337: Second time I see that logic, maybe it's worth to make it a function on `context`?

relays/ethereum/src/ethereum_sync.rs
```diff
@@ -111,7 +111,14 @@ impl HeadersSync {
 	}
 
 	/// Select headers that need to be submitted to the Substrate node.
-	pub fn select_headers_to_submit(&self) -> Option<Vec<&QueuedHeader>> {
+	pub fn select_headers_to_submit(&self, stalled: bool) -> Option<Vec<&QueuedHeader>> {
+		// if we operate in backup mode, we only submit headers when sync has stalled
+		if !stalled {
```
tomusdrw, line 116: Maybe merge the two conditions? This way they read more like the comment sentence above.
```suggestion
		if self.params.sub_tx_mode == SubstrateTransactionMode::Backup && !stalled {
```

relays/ethereum/src/ethereum_sync_loop.rs
```diff
@@ -53,6 +53,9 @@ const SUBSTRATE_TICK_INTERVAL_MS: u64 = 5_000;
 /// the subscriber will receive every best header (2) reorg won't always lead to sync
 /// stall and restart is a heavy operation (we forget all in-memory headers).
 const STALL_SYNC_TIMEOUT_MS: u64 = 30_000;
+/// Delay (in milliseconds) after we have seen update of best Ethereum header in Substrate,
+/// for us to treat sync stalled. ONLY when relay operates in backup mode.
+const BACKUP_STALL_SYNC_TIMEOUT_MS: u64 = 5 * 60_000;
```
tomusdrw, line 58: This stuff should probably be configurable as it might be chain-specific no? It's fine for now though.

modules/ethereum/src/lib.rs
```diff
@@ -99,41 +142,43 @@ pub struct HeaderToImport<Submitter> {
...
 pub struct ImportContext<Submitter> {
 	submitter: Option<Submitter>,
 	parent_header: Header,
 	parent_total_difficulty: U256,
-	next_validators_set_id: u64,
-	next_validators_set: (H256, Vec<Address>),
+	parent_scheduled_change: Option<ScheduledChange>,
+	validators_set_id: u64,
+	validators_set: ValidatorsSet,
+	last_signal_block: Option<H256>,
```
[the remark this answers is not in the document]
svyatonik: `parent_scheduled_change` is `Some()` if parent header has scheduled change. `last_signal_block` is `Some()` if there has ever been block (excluding parent) that has scheduled validators set change and it points to last of such blocks.

modules/ethereum/src/lib.rs
```diff
@@ -251,9 +302,25 @@ pub trait Trait: frame_system::Trait {
...
 	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
-		/// Import Aura chain headers. Ignores non-fatal errors (like when known
-		/// header is provided), rewards for successful headers import and penalizes
-		/// for fatal errors.
+		/// Import single Aura header. Requires transaction to be **UNSIGNED**.
+		pub fn unsigned_import_header(origin, header: Header, receipts: Option<Vec<Receipt>>) {
+			frame_system::ensure_none(origin)?;
+
+			// just ignore error here, because we can't identify submitter
+			let _ = import_header(
```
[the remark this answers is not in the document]
svyatonik: Right - ignoring here is somewhat stupid. Can't even remember why I made this :/

I think returning error from the call should print it, right? (haven't checked)

modules/ethereum/src/lib.rs
```diff
@@ -370,6 +442,38 @@ impl<T: Trait> Module<T> {
 	}
 }
 
+impl<T: Trait> frame_support::unsigned::ValidateUnsigned for Module<T> {
+	type Call = Call<T>;
+
+	fn validate_unsigned(call: &Self::Call) -> TransactionValidity {
+		match *call {
```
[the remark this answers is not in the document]
svyatonik: I prefer match over `if let {} else {}` everywhere. Not sure why if should be better  here :)

modules/ethereum/src/lib.rs
svyatonik: Let's not justify unwraps in test code - afaik the requirement was only for production code. I could replace with something like `.expect("This is test code; you can actually see this expect because tests are able to break in-memory storage in any way they want; and sometimes that's what they want; so no qed")`, but I doubt it is ok.

P.S.: this test storage will probably be removed in #46

modules/ethereum/src/lib.rs
svyatonik: Because that's what I needed in tests - they are not about forks, they are about cases when there's change scheduled by ancestors (no matter if there are forks, or not).

modules/ethereum/src/lib.rs
svyatonik: It is used in 2 tests and not intended to be used outside of test code.

relays/ethereum/src/ethereum_sync_loop.rs
svyatonik: Yeah - there are some constants that are chain specific. We could either make them configurable through cli, or probably auto-adjust, based on what we are receiving from clients. I'll file an issue later

modules/ethereum/src/verification.rs
svyatonik: I don't think it'll work in that way. So you have two bridges. The first knows about block#100 and chain tip is at block#1000. The second knows about block#200 and tip is at block#205. So we are submitting tx to bridge#1 with block#101 and to bridge#2 with block #201. Difference is 1 for both transactions => we can't prioritize bridge#1 over bridge#2.

I think what you're talking about is more fits in #38 - after we'll make something like this, we will be able to determine how far beyond tip we are and prioritize accordingly.

svyatonik:
Re tag prefixes + configurable priority/longevity - let's extract this to separate issue, because there are also other things that needs to be extracted and configured from outside. Most important areaura configuration and validators configuration. Now we can only sync Kovan using this module.

modules/ethereum/src/lib.rs
svyatonik: See top level comment - let's extract everything about configuration to separate issue.

modules/ethereum/src/lib.rs
svyatonik: See top level comment - let's extract everything about configuration to separate issue.

modules/ethereum/src/verification.rs
svyatonik: OK, so the prefix should probably even be bridge+chain specific => let's make it configurable on runtime level => see top level comment.

modules/ethereum/src/lib.rs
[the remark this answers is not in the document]
svyatonik: (ok - need to check this, will do a bit later)

relays/ethereum/src/main.rs
```diff
@@ -115,5 +115,18 @@ fn ethereum_sync_params() -> Result<ethereum_sync_loop::EthereumSyncParams, Stri
 			sp_core::sr25519::Pair::from_string(sub_signer, sub_signer_password).map_err(|e| format!("{:?}", e))?;
 	}
 
+	match matches.value_of("sub-tx-mode") {
+		Some("signed") => eth_sync_params.sub_tx_mode = ethereum_sync_loop::SubstrateTransactionMode::Signed,
+		Some("unsigned") => {
+			eth_sync_params.sub_tx_mode = ethereum_sync_loop::SubstrateTransactionMode::Unsigned;
+
+			// tx pool won't accept too much unsigned transactions
+			eth_sync_params.max_headers_in_submitted_status = 10;
```
HCastano, line 124: Should this be configurable?

modules/ethereum/src/lib.rs
```diff
@@ -251,9 +308,24 @@ pub trait Trait: frame_system::Trait {
 
 decl_module! {
 	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
-		/// Import Aura chain headers. Ignores non-fatal errors (like when known
-		/// header is provided), rewards for successful headers import and penalizes
-		/// for fatal errors.
+		/// Import single Aura header. Requires transaction to be **UNSIGNED**.
+		pub fn unsigned_import_header(origin, header: Header, receipts: Option<Vec<Receipt>>) {
```
HCastano, line 312: I was reading through the code again and managed to confuse myself for a moment thinking that this function called `import_header`**s** instead of `import_header`. I think it'd be less confusing to rename the functions to:

`unsigned_import_header` => `import_unsigned_header`
`import_headers` => `import_signed_headers`

modules/ethereum/src/lib.rs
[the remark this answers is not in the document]
svyatonik: Error is printed when node is started with `RUST_LOG=runtime=debug`

relays/ethereum/src/main.rs
svyatonik: Yeah - as well as several others. Let's do that in separate PR, though (see top level comment).

PR #57, opened 2020-04-07
Add missing RPC required for relay

svyatonik:
Relay needs `AccountNonceApi` to build signed transactions.

### DIFF
--- a/bin/node/runtime/src/lib.rs
+++ b/bin/node/runtime/src/lib.rs
@@ -330,6 +330,12 @@ impl_runtime_apis! {
 		}
 	}
 
+	impl frame_system_rpc_runtime_api::AccountNonceApi<Block, AccountId, Index> for Runtime {
+		fn account_nonce(account: AccountId) -> Index {
+			System::account_nonce(account)
+		}
+	}
+
 	impl sp_bridge_eth_poa::EthereumHeadersApi<Block> for Runtime {
 		fn best_block() -> (u64, sp_bridge_eth_poa::H256) {
 			BridgeEthPoA::best_block()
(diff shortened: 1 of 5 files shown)


bin/node/runtime/src/lib.rs
```diff
@@ -330,6 +330,12 @@ impl_runtime_apis! {
 		}
 	}
 
+	impl frame_system_rpc_runtime_api::AccountNonceApi<Block, AccountId, Index> for Runtime {
+		fn account_nonce(account: AccountId) -> Index {
+			System::account_nonce(account)
```
tomusdrw, line 335: This will not take the transaction pool into account (where is substrate it does). Why don't use the extension defined in substrate repo that contains this logic?

Another option would be to have relay detect if this method is available (by calling `rpc_methods` first) and falling back to `System::account_nonce` state read instead).

svyatonik: > This will not take the transaction pool into account (where is substrate it does). Why don't use the extension defined in substrate repo that contains this logic?

You're looking at runtime code now :) Of course it won't take transaction pool into account. And substrate runtime can't do this either - it can't access pool.

> Another option would be to have relay detect if this method is available (by calling rpc_methods first) and falling back to System::account_nonce state read instead).

Our node provides required RPC - please look above, at the changes to node code (`with_rpc_extensions() { ... FullSystem::new() ... }`)

svyatonik:
No - it isn't temporary workaround. What's correct implementation you're talking about?

bin/node/runtime/src/lib.rs
tomusdrw: :man_facepalming: Sorry, I was sure that this is RPC implementation, sorry about that!

Issue #50, opened 2020-03-31
Solidity contract that'll accept Substrate headers


svyatonik:
(all other accompanying stuff, like: passing messages to/from Substrate, GRANDPA builtin is outside of scope of this issue)

svyatonik:
closed by #65 && #69

PR #58, opened 2020-04-07
Make relay generic over source/target chains

svyatonik:
This is required to implement Substrate -> Ethereum headers synchronization, which is required to test solidity contract (#50) in action. The synchronization process would be almost the same (the biggest difference would be the extra header data - tx receipts vs justification) as the `eth-to-sub`, so this PR mostly just moves code around + makes it generic over chains (i.e. instead of `ethereum_headers.rs` there's now generic `headers.rs`).

Also: there's now no default mode for the relay - you should specify subcommand. Currently the only subcommand is `eth-to-sub`. Next PRs will introduce `sub-to-eth` mode.


relays/ethereum/src/ethereum_types.rs
```diff
@@ -30,84 +34,48 @@ pub type Header = web3::types::Block<H256>;
...
+	type Header = Header;
+	type Extra = Vec<Receipt>;
 
-	/// Returns ID of parent header.
-	pub fn parent_id(&self) -> HeaderId {
-		HeaderId(
-			self.header.number.expect(HEADER_ID_PROOF).as_u64() - 1,
-			self.header.parent_hash,
-		)
+	fn source_name() -> &'static str {
```
tomusdrw, line 54: Could be simply a `const` on a trait?

relays/ethereum/src/ethereum_types.rs
```diff
@@ -30,84 +34,48 @@ pub type Header = web3::types::Block<H256>;
...
+		"Ethereum"
 	}
 
-	/// Returns reference to header.
-	pub fn header(&self) -> &Header {
-		&self.header
+	fn target_name() -> &'static str {
+		"Substrate"
```
tomusdrw, line 59: ditto

relays/ethereum/src/substrate_client.rs
```diff
@@ -131,6 +128,7 @@ pub async fn ethereum_header_known(
 /// Submits Ethereum header to Substrate runtime.
 pub async fn submit_ethereum_headers(
 	client: Client,
+	signer: sp_core::sr25519::Pair,
```
tomusdrw, line 131: That should probably be configurable, but not in this PR. Could you log an issue?

svyatonik: It is already configurable (`--sub-signer=//Alice --sub-signer-password=HelloWorld`)

tomusdrw: I meant the crypto type (either `sr` or `ed`). I don't think it's required for all chains to use `MultiSigner` and support both crypto.

svyatonik: Oh, ok - I see now, thanks. But relay directly references (our) runtime - I mean to encode `Call`s, generate signatures, etc. And our runtime specifies that it is using `MultiSignature`. So I'm not sure it'll be easy to make it generic over runtime. Wdyt?

I had similar problem in secret-store and I ended up with moving everything runtime-specific to separate file ([see here](https://github.com/paritytech/secret-store/blob/master/bin/substrate/src/runtime.rs#L17)). So to reconfigure SS for different chain, you just need to redefine what's require in this file. We could do something like that here, or do you think there's a better way?

tomusdrw: Yeah, I think that's fine, it's super low-priority though. I'm almost certain that nearly all of the chains will be using `MultiSigner` so we should be fine with single, hardcoded crypto.

svyatonik: #61

Issue #62, opened 2020-04-10
Prune old headers with separate transaction


svyatonik:
And potentially reward submitter for pruning (if prune transaction is signed? or should we only accept signed pruning transactions?)

**PART 1: Problem with unsigned transactions**
1) authority generates multiple valid headers at its slot;
2) (most of) headers are 'mined' and occupy storage;
3) we have potentially unlimited number of entries in the [`HeadersByNumber`](https://github.com/paritytech/parity-bridges-common/blob/878d17a3ec99fe0776d926873242de2807818f63/modules/ethereum/src/lib.rs#L364) map;
4) later, when one of these headers is finalized, we may end up exceeding block (weight? fee?) limit because we need to prune these headers [here](https://github.com/paritytech/parity-bridges-common/blob/878d17a3ec99fe0776d926873242de2807818f63/modules/ethereum/src/lib.rs#L589) => transaction with **valid header** submitted by **good validator** is not accepted => the sync is stuck.

**PART 2: Problem with signed transactions**
The problem is the same, though malicious authority will pay for its actions. But the sync will stuck anyway.

**PART 3: Potential solution**
Do not intermix headers sync with headers pruning - there should be another API entry for pruning obsolete headers. Probably we need to reward for calling this method?

Another solution would be to disable pruning at all.

tomusdrw:
Weights do not depend on storage size currently. I think we only support fixed-size weights, so probably because of this we probably won't run into that issue.
However this is not really a solution, and the problem might still be there when we introduce more sophisticated weighting mechanism. I like the idea of splitting pruning, but would prefer to have that configurable. I.e. in some chains where we can safely limit the number of forks, we could perform pruning anyway. We could also perform pruning if the number of forks is within limits.

Issue #60, opened 2020-04-07
Get rid of hardcoded constants in relay


svyatonik:
folow up of #45 

Like [this](https://github.com/paritytech/parity-bridges-common/pull/45#pullrequestreview-387761441) and [this](https://github.com/paritytech/parity-bridges-common/pull/45/files/9cc72e3312c13b0dd23b7c624a0c4e7a8a256d30#diff-6a9ed349350a62b0d01e2e8e51eb0e19). Also, since there's quite a few constants there, we may think of supporting configuration files.

PR #65, opened 2020-04-13
Solidity contract that accepts unverified substrate headers

svyatonik:
part of #50 

This PR introduces:
1) ethereum-contract-builtin crate, which will be used by bridge contract to import headers and verify finality proofs. Right now it has `fn parse_substrate_header()`, which is (almost) completed and `fn verify_substrate_finality_proof()` which is unimplemented. This combination allows contract to import Substrate headers, but not to import finality proofs;
2) solidity contract that is able to import unverified Substrate headers and (in theory) import and verify finality proofs;
3) `sub-to-eth` subcommand for the relay to be able to test that headers sync. The pipeline is almost the same as in `eth-to-sub`, but we have no extra data to be submitted along with header (like transactions receipts in eth-to-sub). Also just realized that there are some parts of future `AsyncExtra` data sync (which is finality proof in our case) - can remove on demand, just wanted to make separate PRs for headers import and finality proof import;
4) `eth-deploy-contract` subcommand for the relay which deploys bridge contract on eth-like chain. We may throw it later, or make a separate binary, but right now I find it much more convenient to have it as a subcommand in relay.

Some details:
1) in the first version of contract, it has been able to import only finalized headers - i.e. it had `function importHeaders(bytes[] calldata rawHeaders, bytes calldata rawFinalityProof)` instead of two functions (`function importHeader()` and `function importFinalityProof()`) now. But it might be impossible to fit this data in single transaction in some cases (like when GRANDA validators were unable to finalize headers for some long time);
2) there is no pruning in bridge contract now, mostly for simplicity - we may add support later. But we need to take #62 into account;
3) relay code is mostly the same, though there are some new RPC calls for both Substrate and Ethereum nodes. So most new lines are utility lines describing sub-to-eth sync, code reorgs and the contract itself. I was thinking of introducing tests for the contract, but then realized that I won't be able to do that without mocking builtins. And I do not know any way to mock builtins. So the contract comes without any unit tests - please advice if you know any way to solve it.


svyatonik:
To test:
1) build parity from https://github.com/svyatonik/parity/tree/substrate-builtins-stubs ;
2) start syncing:
```bash
#!/bin/bash
set -e

rm -rf substrate.db
rm -rf parity.db

cargo build --manifest-path=../parity-bridges-common/Cargo.toml -p bridge-node
cp ../parity-bridges-common/target/debug/bridge-node .
cargo build --manifest-path=../parity-bridges-common/Cargo.toml -p ethereum-poa-relay
cp ../parity-bridges-common/target/debug/ethereum-poa-relay .

# Start Substrate and Parity nodes
RUST_LOG=runtime=debug unbuffer ./bridge-node --dev -d parity.db 2>&1 | unbuffer -p gawk '{ print strftime("Substrate: [%Y-%m-%d %H:%M:%S]"), $0 }' | unbuffer -p tee substrate.log&
#RUST_LOG=miner=trace,own_tx=trace,executive=trace
unbuffer ./parity -d parity.db --chain dev --force-sealing --no-warp --no-persistent-txqueue --jsonrpc-apis=all --jsonrpc-cors=all 2>&1 | unbuffer -p gawk '{ print strftime("Parity: [%Y-%m-%d %H:%M:%S]"), $0 }' | unbuffer -p tee parity.log&


# Deploy bridge contract
sleep 10
RUST_LOG=bridge=trace ./ethereum-poa-relay eth-deploy-contract  2>&1 | unbuffer -p gawk '{ print strftime("Substrate: [%Y-%m-%d %H:%M:%S]"), $0 }' | unbuffer -p tee deploy.log&

# Start syncing Substrate -> Ethereum
sleep 10
RUST_LOG=bridge=trace unbuffer ./ethereum-poa-relay sub-to-eth 2>&1 | unbuffer -p gawk '{ print strftime("Bridge: [%Y-%m-%d %H:%M:%S]"), $0 }' | unbuffer -p tee bridge.log&
```

relays/ethereum/res/substrate-bridge.json
```diff
@@ -0,0 +1,119 @@
+[
```
HCastano, line 1: Can you rename this file to indicate that it's a smart contract ABI?

relays/ethereum/src/utils.rs
```diff
@@ -0,0 +1,37 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+/// Macro that returns (client, Err(error)) tuple from function if result is Err(error).
+#[macro_export]
+macro_rules! bail_on_error {
```
HCastano, line 19: I'm pretty confused as to why we have so man functions that take a `Client` as an argument and also return one

relays/ethereum/src/ethereum_client.rs
```diff
@@ -150,6 +214,172 @@ async fn transaction_receipt(client: Client, hash: H256) -> (Client, Result<Rece
...
+	}
+}
+
+/// Submits Substrate headers to Ethereum contract.
+pub async fn submit_substrate_headers(
+	client: Client,
+	params: EthereumSigningParams,
+	contract_address: Address,
+	headers: Vec<QueuedSubstrateHeader>,
+) -> (Client, Result<(Vec<H256>, Vec<SubstrateHeaderId>), Error>) {
```
HCastano, line 278: What do you think about:

```
type EthereumTransactionHash = H256;
```

and using that here (and all the functions that follow)?

relays/ethereum/src/ethereum_client.rs
```diff
@@ -150,6 +214,172 @@ async fn transaction_receipt(client: Client, hash: H256) -> (Client, Result<Rece
...
+	.await
+}
+
+/// Submit ethereum transaction.
+async fn submit_ethereum_transaction(
+	client: Client,
+	params: &EthereumSigningParams,
+	contract_address: Option<Address>,
+	nonce: Option<U256>,
+	double_gas: bool,
```
HCastano, line 331: How would I know when this should be used?

relays/ethereum/src/ethereum_deploy_contract.rs
```diff
@@ -0,0 +1,136 @@
...
...516020890160208301848184846011600019fa61070057600080fd5b5050505b6040805160a081018252968752602087019590955293850192909252606084015250608082015292915050565b6002549392505050565b6040518060a0016040528060008019168152602001600080191681526020016000815260200160008152602001606081525090565b828054600181600116156101000203166002900490600052602060002090601f016020900481019282601f106107b157805160ff19168380011785556107de565b828001600101855582156107de579182015b828111156107de5782518255916020019190600101906107c3565b506107ea929150610863565b5090565b828054600181600116156101000203166002900490600052602060002090601f016020900481019282601f1061082757805485556107de565b828001600101855582156107de57600052602060002091601f016020900482015b828111156107de578254825591600101919060010190610848565b61087d91905b808211156107ea5760008155600101610869565b9056fe4d697373696e6720706172656e74206865616465722066726f6d207468652073746f726167654d697373696e672066696e616c69747920746172676574206865616465722066726f6d207468652073746f726167654d697373696e672072657175697265642066696e616c6974792070726f6f6620666f7220706172656e7420686561646572a2646970667358221220920ca01b59940884722a766076b4f1d1d825eba446e530980bd310f776703dad64736f6c63430006060033")
```
HCastano, line 48: Given that this is going to change every time we update the contract we should probably read it from a file, no?

relays/ethereum/src/utils.rs
svyatonik: Because they need to consume `Client`? It was the requirement of the old `jsonrpsee` version (that has been used in the very beginning). On the last version(s), the client could be shared between several futures. So it can be changed. PRs are welcome :)

relays/ethereum/src/ethereum_client.rs
svyatonik: Not sure, what do you mean by this question? :) The direct answer is when you want to provide `estimated_gas * 2` instead of `estimated_gas`.

svyatonik: Probably [this comment](https://github.com/paritytech/parity-bridges-common/pull/65/files#diff-440221e3b161f2af0600683d4f2ab86aR290) could be helpful?

HCastano: My question was more: how would I know in what conditions I need to double the gas? That comment is helpful, but in my mind only covers one case

modules/ethereum-contract/builtin/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/// Parse Substrate header.
+pub fn parse_substrate_header(raw_header: &[u8]) -> Result<Header, Error> {
+	RuntimeHeader::decode(&mut &raw_header[..])
+		.map(|header| Header {
+			hash: header.hash(),
+			parent_hash: header.parent_hash,
+			number: header.number,
+			signal: None, // TODO: parse me
```
HCastano, line 72: Is this going be be done as part of this PR?

modules/ethereum-contract/builtin/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+pub fn verify_substrate_finality_proof(
+	_best_set_id: u64,
+	_raw_best_voters: &[u8],
+	_raw_best_header: &[u8],
+	_raw_headers: &[&[u8]],
+	_raw_finality_proof: &[u8],
+) -> Result<(usize, usize), Error> {
+	Err(Error::JustificationVerify(ClientError::Msg(
+		"Not yet implemented".into(),
+	))) // TODO: implement me
```
HCastano, line 87: I think it's fine to leave this unimplemented while we figure out the details of proof verification, but we should attach the TODO to an issue

relays/ethereum/res/substrate-bridge-bytecode.hex
```diff
@@ -0,0 +1 @@
...87516020890160208301848184846011600019fa61076d57600080fd5b5050505b6040805160a081018252968752602087019590955293850192909252606084015250608082015292915050565b6002549392505050565b6040518060a0016040528060008019168152602001600080191681526020016000815260200160008152602001606081525090565b828054600181600116156101000203166002900490600052602060002090601f016020900481019282601f1061081e57805160ff191683800117855561084b565b8280016001018555821561084b579182015b8281111561084b578251825591602001919060010190610830565b506108579291506108d0565b5090565b828054600181600116156101000203166002900490600052602060002090601f016020900481019282601f10610894578054855561084b565b8280016001018555821561084b57600052602060002091601f016020900482015b8281111561084b5782548255916001019190600101906108b5565b6108ea91905b8082111561085757600081556001016108d6565b9056fe4d697373696e6720706172656e74206865616465722066726f6d207468652073746f726167654d697373696e672066696e616c69747920746172676574206865616465722066726f6d207468652073746f726167654d697373696e672072657175697265642066696e616c6974792070726f6f6620666f7220706172656e7420686561646572a26469706673582212206cc35a10288e85e37e14250b2d4af37dc4fbcd59635965d9ac0421cb8db1f66c64736f6c63430006060033
```
HCastano, line 1: I wonder, should this be in version control or could we get away with having a custom build.rs to make sure this gets generated? What do you guys think?

relays/ethereum/src/ethereum_client.rs
```diff
@@ -150,6 +214,173 @@ async fn transaction_receipt(client: Client, hash: H256) -> (Client, Result<Rece
...
+		Some(nonce) => (client, nonce),
+		None => bail_on_error!(account_nonce(client, params.signer.address().as_fixed_bytes().into()).await),
+	};
+	let (client, gas) = bail_on_error!(
+		estimate_gas(
+			client,
+			CallRequest {
+				to: contract_address,
+				data: Some(encoded_call.clone().into()),
+				..Default::default()
```
HCastano, line 342: This is more of a dumb Rust question, but why's this here? `CallRequest` has two fields which are already set, so it seems unnecessary to me - but I might be missing something

relays/ethereum/src/sync.rs
```diff
@@ -228,9 +215,20 @@ mod tests {
 		H256::from_low_u64_le(1000 + number)
 	}
 
+	fn default_sync_params() -> HeadersSyncParams {
+		HeadersSyncParams {
+			max_future_headers_to_download: 128,
+			max_headers_in_submitted_status: 128,
+			max_headers_in_single_submit: 32,
+			max_headers_size_in_single_submit: 131_072,
```
HCastano, line 223: Where did this number come from?

relays/ethereum/src/ethereum_client.rs
svyatonik: That comments is the only place where `double_gas` is `true`. So it covers one case, but it is the only case :) So the answer is - when you expect that the gas requirements may change.

Just looking at the contract code again, and I've realized that the loop has (that has been in `importHeader` initially) been removed now && header import is now contextless. I'll leave `double_gas`, though, because it'll be required for next PR.

modules/ethereum-contract/builtin/src/lib.rs
svyatonik: Nope. since it only affects finality. And finality is the next PR.

relays/ethereum/res/substrate-bridge-bytecode.hex
svyatonik: I do not have `solc` or `solcjs` on my machine. And I do not want to. So imo not

relays/ethereum/src/sync.rs
svyatonik: This is inside test code && originally comes from Ethereum -> Substrate sync. It is 32 * 4096 (`max_headers * max_header_and_receipts_size_we_expect`). For ethereum it gives ~6_500_000 gas / transaction (without execution cost). IMO should be ok too, given current limits.

Also just noticed that there values in configurations were wrong (sub2eth were used in eth2sub and vice versa) - fixed.

relays/ethereum/src/ethereum_sync_loop.rs
```diff
@@ -30,41 +30,30 @@ const SUBSTRATE_TICK_INTERVAL_MS: u64 = 5_000;
...
-			sub_signer: sp_keyring::AccountKeyring::Alice.pair(),
-			sync_params: Default::default(),
+			eth: Default::default(),
+			sub: Default::default(),
+			sub_sign: Default::default(),
+			sync_params: HeadersSyncParams {
+				max_future_headers_to_download: 128,
+				max_headers_in_submitted_status: 128,
+				max_headers_in_single_submit: 32,
+				max_headers_size_in_single_submit: 131_072,
```
HCastano, line 53: We should make this a `const` and have the explanation you gave me for how to get to this number as a comment

relays/ethereum/src/sync.rs
HCastano: Same thing I mentioned earlier with the other max header const

relays/ethereum/src/substrate_sync_loop.rs
```diff
@@ -0,0 +1,194 @@
...
+			// of the contract that is deployed by default signer and 0 nonce
+			eth_contract_address: "731a10897d267e19b34503ad902d0a29173ba4b1"
+				.parse()
+				.expect("address is hardcoded, thus valid; qed"),
+			sub: Default::default(),
+			sync_params: HeadersSyncParams {
+				max_future_headers_to_download: 8,
+				max_headers_in_submitted_status: 4,
+				max_headers_in_single_submit: 4,
+				max_headers_size_in_single_submit: 131_072,
```
HCastano, line 64: Const-ify, lol

relays/ethereum/src/substrate_sync_loop.rs
```diff
@@ -0,0 +1,194 @@
...
+				prune_depth: 256,
+				target_tx_mode: TargetTransactionMode::Signed,
+			},
+		}
+	}
+}
+
+/// Substrate client as headers source.
+struct SubstrateHeadersSource {
```
HCastano, line 73: nit
```suggestion
struct SubstrateHeaderSource {
```

relays/ethereum/src/substrate_sync_loop.rs
```diff
@@ -0,0 +1,194 @@
...
+			.boxed()
+	}
+
+	fn header_extra(self, id: SubstrateHeaderId, _header: &Header) -> Self::HeaderExtraFuture {
+		ready((self, Ok((id, ()))))
+	}
+}
+
+/// Ethereum client as Substrate headers target.
+struct EthereumHeadersTarget {
```
HCastano, line 109: nit
```suggestion
struct EthereumHeaderTarget {
```

relays/ethereum/src/substrate_types.rs
```diff
@@ -14,16 +14,61 @@
...
+/// Substrate header ID.
+pub type SubstrateHeaderId = HeaderId<bridge_node_runtime::Hash, bridge_node_runtime::BlockNumber>;
+
+/// Queued substrate header ID.
+pub type QueuedSubstrateHeader = QueuedHeader<SubstrateHeadersSyncPipeline>;
+
+/// Substrate synchronization pipeline.
+#[derive(Clone, Copy, Debug)]
+#[cfg_attr(test, derive(PartialEq))]
+pub struct SubstrateHeadersSyncPipeline;
```
HCastano, line 46: nit
```suggestion
pub struct SubstrateHeaderSyncPipeline;
```

relays/ethereum/src/substrate_client.rs
```diff
@@ -341,3 +413,20 @@ fn create_unsigned_submit_transaction(header: QueuedEthereumHeader) -> bridge_no
 
 	bridge_node_runtime::UncheckedExtrinsic::new_unsigned(function)
 }
+
+/// When RPC method returns encoded value.
+fn rpc_returns_encoded_value<T: Decode>(value: Value) -> Result<T, Error> {
+	let encoded_response: Bytes = from_value(value).map_err(|_| Error::ResponseParseFailed)?;
```
HCastano, line 419: When a request fails to decode, would there be any indication about _which_ request failed?

relays/ethereum/src/sync.rs
svyatonik: On a second thought, in sub2eth, max_**_in_single_submit aren't that important now - there's single ethereum transaction per single substrate header. And these constants only make sense when several headers are submitted in single tx (like in eth2sub signed transactions). I've made conts with appropriate comments in `ethereum_sync_loop.rs`

relays/ethereum/src/substrate_client.rs
svyatonik: You'll get something like "Error retrieving best header number from Substrate: <...>" in logs. We always have at most 1 request at a time with a single exception - when we're submitting multiple headers in multiple transactions. But this would lead to single error anyway - "Error submitting headers to Substrate node: <...>".

relays/ethereum/src/cli.yml
```diff
@@ -43,3 +43,82 @@ subcommands:
...
+                help: Hex-encoded secret to use when transactions are submitted to the Ethereum node.
+            - eth-contract-code:
+                long: eth-contract-code
+                value_name: ETH_CONTRACT_CODE
+                help: Code of bridge contract.
+                takes_value: true
+            - sub-host:
+                long: sub-host
+                value_name: SUB_HOST
+                help: Connect to Substrate node at given host.
```
tomusdrw, line 103: Why does it need connecting to substrate node when deploying?

relays/ethereum/src/ethereum_client.rs
```diff
@@ -122,11 +190,7 @@ pub async fn transactions_receipts(
 ) -> (Client, Result<(EthereumHeaderId, Vec<Receipt>), Error>) {
 	let mut transactions_receipts = Vec::with_capacity(transacactions.len());
 	for transacaction in transacactions {
```
tomusdrw, line 192: There are typos here: 
```suggestion
	for transaction in transactions {
```
but it's a minor thing :)

relays/ethereum/src/ethereum_client.rs
```diff
@@ -150,6 +214,172 @@ async fn transaction_receipt(client: Client, hash: H256) -> (Client, Result<Rece
...
+			data: Some(encoded_call.into()),
+		})
+		.map_err(|e| Error::RequestSerialization(e)),
+		client
+	);
+	let (client, call_result) =
+		bail_on_error!(call_rpc::<Bytes>(client, "eth_call", Params::Array(vec![call_request]),).await);
+	let (number, raw_hash) = match call_decoder.decode(&call_result.0) {
+		Ok((raw_number, raw_hash)) => (raw_number, raw_hash),
+		Err(error) => return (client, Err(Error::ResponseParseFailed(format!("{}", error)))),
```
tomusdrw, line 235: It's a bit out of scope now, but this pattern of returning `Client` is a bit annoying. Why do we need to own it?

relays/ethereum/src/ethereum_client.rs
```diff
@@ -150,6 +214,172 @@ async fn transaction_receipt(client: Client, hash: H256) -> (Client, Result<Rece
...
+	params: EthereumSigningParams,
+	contract_address: Address,
+	headers: Vec<QueuedSubstrateHeader>,
+) -> (Client, Result<Vec<SubstrateHeaderId>, Error>) {
+	let (mut client, mut nonce) =
+		bail_on_error!(account_nonce(client, params.signer.address().as_fixed_bytes().into()).await);
+
+	let ids = headers.iter().map(|header| header.id()).collect();
+	for header in headers {
+		let (ret_client, _) = bail_on_error!(
```
tomusdrw, line 284:
```suggestion
		client = bail_on_error!(
		  ...
		).0;
```

relays/ethereum/src/ethereum_client.rs
```diff
@@ -150,6 +214,172 @@ async fn transaction_receipt(client: Client, hash: H256) -> (Client, Result<Rece
...
+		gas_price: params.gas_price,
+		data: encoded_call,
+	}
+	.sign(&params.signer.secret().as_fixed_bytes().into(), &params.chain_id);
+	let transaction = bail_on_arg_error!(
+		to_value(Bytes(raw_transaction)).map_err(|e| Error::RequestSerialization(e)),
+		client
+	);
+	let (client, _) = bail_on_error!(
+		call_rpc::<TransactionHash>(client, "eth_submitTransaction", Params::Array(vec![transaction])).await
```
tomusdrw, line 360: To be honest I'd really prefer to enclose low-level non-bridge-sepcific (so no `EthereumSigningParams`) calls into a separate module and have it properly typed:
```rust
mod ethereum_rpc {
  pub async fn submitTransaction(client: _, transaction: RawTransaction, secret: _) -> (Client, Result<_>) {
   ...
  }
  pub async fn call(client: _, CallRequest) -> .. { }
}
```
Right now it feels quite error prone to duplicate the same logic in multiple places (especially eth_call) and use stringified methods.

The higher-level (bridge-specific), like: `deploy_bridge`, `best_substrate_block` should be exposed from this module, but I'd keep the low-level ones encapsulated, so that it's easier to distinguish bridge-specific logic from RPC call boilerplate (like encoding the `CallRequest`).

relays/ethereum/src/ethereum_deploy_contract.rs
```diff
@@ -0,0 +1,136 @@
...
+	/// Initial header.
+	pub sub_initial_header: Option<Vec<u8>>,
+}
+
+impl Default for EthereumDeployContractParams {
+	fn default() -> Self {
+		EthereumDeployContractParams {
+			eth: Default::default(),
+			eth_sign: Default::default(),
+			// compiler: 0.6.6+commit.6c089d02 + optimization
```
tomusdrw, line 47: Please move that to metadata.

relays/ethereum/src/ethereum_deploy_contract.rs
```diff
@@ -0,0 +1,136 @@
...
+			sub_client,
+			initial_header_hash,
+			params.sub_initial_authorities_set,
+		).await;
+		let initial_set = initial_set?;
+
+		log::info!(
+			target: "bridge",
+			"Deploying Ethereum contract.\r\n\tInitial header: {}\r\n\tInitial authorities set ID: {}\r\n\tInitial authorities set: {}",
+			hex::encode(&initial_header),
```
tomusdrw, line 79: It'd be worth displaying decoded (human-readable) values as well.

relays/ethereum/src/ethereum_deploy_contract.rs
```diff
@@ -0,0 +1,136 @@
...
+	}
+}
+
+/// Prepare initial GRANDPA authorities set.
+async fn prepare_initial_authorities_set(
+	sub_client: substrate_client::Client,
+	sub_initial_header_hash: SubstrateHash,
+	sub_initial_authorities_set: Option<Vec<u8>>,
+) -> (substrate_client::Client, Result<Vec<u8>, String>) {
+	let (sub_client, initial_authorities_set) = match sub_initial_authorities_set {
```
tomusdrw, line 127: The `Client` handling is so cumbersome that I'm even inclined to pass it via `thread_local` or `lazy_static`. Something like this:
```rust
static mut CLIENT: Option<Client> = None;

struct WithClient;
impl WithClient {
    pub fn new(s: Client) -> Self {
        unsafe {
            if CLIENT.is_some() {
                panic!("There should be only one instance of client");
            }
            CLIENT = Some(s);
        }
        WithClient
    }

    pub fn with_client<R>(&mut self, call: impl FnMut(Client) -> (Client, R)) -> R {
        let client = unsafe { CLIENT.take().expect("...") };
        let (client, result) = call(client);
        unsafe { CLIENT = Some(client); }
        result
    }
}
```

You can then keep passing `WithClient` mutable reference (or agregate all functions in a single struct) and return regular `Result<>` object (so `?` would work), but only in places where you really need to own the client you would fall back to this tuple-return format.


Or simply the functions assume the client is initialized in `lazy_static!` and we don't pass any extra objects at all (this will require Mutex though)

relays/ethereum/src/ethereum_sync_loop.rs
```diff
@@ -14,57 +14,52 @@
...
-			eth_port: 8545,
-			sub_host: "localhost".into(),
-			sub_port: 9933,
-			sub_signer: sp_keyring::AccountKeyring::Alice.pair(),
-			sync_params: Default::default(),
+			eth: Default::default(),
+			sub: Default::default(),
+			sub_sign: Default::default(),
+			sync_params: HeadersSyncParams {
+				max_future_headers_to_download: 128,
```
tomusdrw, line 56: Can we move these to contstants too? best to have them in a single place.

relays/ethereum/src/ethereum_sync_loop.rs
```diff
@@ -177,16 +172,16 @@ impl TargetClient<EthereumHeadersSyncPipeline> for SubstrateHeadersTarget {
 	}
 
 	fn submit_headers(self, headers: Vec<QueuedEthereumHeader>) -> Self::SubmitHeadersFuture {
-		let (signer, sign_transactions) = (self.signer, self.sign_transactions);
-		substrate_client::submit_ethereum_headers(self.client, signer.clone(), headers, sign_transactions)
+		let (sign_transactions, sign_params) = (self.sign_transactions, self.sign_params);
```
tomusdrw, line 175:
```suggestion
		let { sign_transactions, sign_params, client, ... } = self;
```
This should work too.

relays/ethereum/src/substrate_client.rs
```diff
@@ -56,23 +102,42 @@ impl MaybeConnectionError for Error {
...
+/// Returns Substrate header by number.
+pub async fn header_by_number(client: Client, number: Number) -> (Client, Result<SubstrateHeader, Error>) {
+	let (client, hash) = bail_on_error!(block_hash_by_number(client, number).await);
+	header_by_hash(client, hash).await
+}
+
 /// Returns best Ethereum block that Substrate runtime knows of.
 pub async fn best_ethereum_block(client: Client) -> (Client, Result<EthereumHeaderId, Error>) {
-	let (client, result) = call_rpc::<(u64, H256)>(
+	let (client, result) = call_rpc(
```
tomusdrw, line 133: Same as wtih ethereum_client, can we extract this low-level calls (including serialization to some encapsulated module)? IMHO it would be way more readable if we avoided mixing bridge logic with rpc payload boilerplate.

relays/ethereum/src/substrate_client.rs
```diff
@@ -128,91 +201,102 @@ pub async fn ethereum_header_known(
...
+}
+
+/// Get GRANDPA authorities set at given block.
+pub async fn grandpa_authorities_set(client: Client, block: Hash) -> (Client, Result<Vec<u8>, Error>) {
+	let block = bail_on_arg_error!(to_value(block).map_err(|e| Error::RequestSerialization(e)), client);
+	call_rpc(
+		client,
+		"state_call",
+		Params::Array(vec![
+			serde_json::Value::String("GrandpaApi_grandpa_authorities".into()),
```
tomusdrw, line 284: Let's also please move all strings to constants placed in the top part of the module.

relays/ethereum/src/substrate_client.rs
```diff
@@ -224,42 +308,30 @@ async fn next_account_index(
...
-async fn call_rpc_u64(mut client: Client, method: &'static str, params: Params) -> (Client, Result<u64, Error>) {
-	async fn do_call_rpc(client: &mut Client, method: &'static str, params: Params) -> Result<u64, Error> {
+async fn call_rpc<T>(
+	mut client: Client,
+	method: &'static str,
+	params: Params,
+	decode_value: impl Fn(Value) -> Result<T, Error>,
+) -> (Client, Result<T, Error>) {
+	async fn do_call_rpc<T>(
+		client: &mut Client,
```
tomusdrw, line 330: Why do we keep passing owned Client if a mutable reference is enough? It's because we then spawn the futures on an executor? Maybe best to just put these bare functions to a struct where `Client` is owned?

relays/ethereum/src/substrate_sync_loop.rs
```diff
@@ -0,0 +1,196 @@
...
+			eth: Default::default(),
+			eth_sign: Default::default(),
+			// the address 0x731a10897d267e19b34503ad902d0a29173ba4b1 is the address
+			// of the contract that is deployed by default signer and 0 nonce
+			eth_contract_address: "731a10897d267e19b34503ad902d0a29173ba4b1"
+				.parse()
+				.expect("address is hardcoded, thus valid; qed"),
+			sub: Default::default(),
+			sync_params: HeadersSyncParams {
+				max_future_headers_to_download: 8,
```
tomusdrw, line 61: Please move hardcoded values to constants. I think these kind of parameters could be even grouped in a separate module to have a nice overview of hardcoded configuration.

relays/ethereum/src/utils.rs
tomusdrw: Oh, I see an explanation of the `Client` thing. Fair enough, I've created an issue to rework this: https://github.com/paritytech/parity-bridges-common/issues/72

relays/ethereum/src/ethereum_client.rs
tomusdrw: NVM, found why.

relays/ethereum/src/ethereum_deploy_contract.rs
tomusdrw: NVM, the whole thing. Should be reworked as per: https://github.com/paritytech/parity-bridges-common/issues/72

relays/ethereum/src/substrate_sync_loop.rs
tomusdrw: See: https://github.com/paritytech/parity-bridges-common/issues/75

relays/ethereum/src/substrate_client.rs
tomusdrw: Logger here: https://github.com/paritytech/parity-bridges-common/issues/74

relays/ethereum/src/substrate_client.rs
tomusdrw: See: https://github.com/paritytech/parity-bridges-common/issues/75

relays/ethereum/src/ethereum_client.rs
tomusdrw: Logged here: https://github.com/paritytech/parity-bridges-common/issues/74

relays/ethereum/src/ethereum_client.rs
```diff
@@ -14,39 +14,106 @@
...
+			chain_id: 0x11, // Parity dev chain
+			// account that has a lot of ether when we run instant seal engine
+			// address: 0x00a329c0648769a73afac7f9381e08fb43dbea72
+			// secret: 0x4d5db4107d237df6a3d58ee5f70ae63d73d7658d4026f2eefd2f204c81682cb7
+			signer: KeyPair::from_secret_slice(
+				&hex::decode("4d5db4107d237df6a3d58ee5f70ae63d73d7658d4026f2eefd2f204c81682cb7")
+					.expect("secret is hardcoded, thus valid; qed"),
+			)
+			.expect("secret is hardcoded, thus valid; qed"),
+			gas_price: 8_000_000_000u64.into(), // 8 Gwei
```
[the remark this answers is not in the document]
tomusdrw: See: https://github.com/paritytech/parity-bridges-common/issues/73

modules/ethereum-contract/contracts/substrate-bridge.sol
```diff
@@ -0,0 +1,327 @@
...
+
+	/// Import finality proof.
+	function importFinalityProof(
+		uint256 finalityTargetNumber,
+		bytes32 finalityTargetHash,
+		bytes memory rawFinalityProof
+	) public {
+		// check that header that we're going to finalize is already imported
+		require(
+			headerByHash[finalityTargetHash].number == finalityTargetNumber,
```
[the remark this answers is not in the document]
svyatonik: [SOLIDITY] I'm not against that (will consider that when preparing final version of the contract), but I see nothing criminal in re-finalizing finalized block (genesis in this case) - it just should be a noop. Do you have any other concerns to mention?

modules/ethereum-contract/contracts/substrate-bridge.sol
```diff
@@ -0,0 +1,327 @@
...
+	/// Import header.
+	function importHeader(
+		bytes memory rawHeader
+	) public {
+		// parse header
+		ParsedHeader memory header = parseSubstrateHeader(rawHeader);
+		require(
+			!headerByHash[header.hash].isKnown,
+			"Header is already known"
+		);
```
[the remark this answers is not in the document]
svyatonik: [SOLIDITY] This is checked by the `parentHeader.isKnown && parentHeader.number == header.number - 1,` later. So unless we'll have 2^32 headers (which will break contract assumptions anyway), we won't allow to import such headers.

modules/ethereum-contract/contracts/substrate-bridge.sol
```diff
@@ -0,0 +1,327 @@
...
+			);
+			prevSignalTargetNumber = header.number + header.signalDelay;
+		}
+
+		// check if parent header has emitted validators set change signal
+		uint64 validatorsSetId = parentHeader.validatorsSetId;
+		bytes32 prevSignalHeaderHash = parentHeader.prevSignalHeaderHash;
+		if (parentHeader.signal.length != 0) {
+			prevSignalHeaderHash = header.parentHash;
+			validatorsSetId = validatorsSetId + 1;
```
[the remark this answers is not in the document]
svyatonik: [SOLIDITY] Noted :)

modules/ethereum-contract/contracts/substrate-bridge.sol
```diff
@@ -0,0 +1,327 @@
...
+		bytes memory rawHeader
+	) public {
+		// parse header
+		ParsedHeader memory header = parseSubstrateHeader(rawHeader);
+		require(
+			!headerByHash[header.hash].isKnown,
+			"Header is already known"
+		);
+
+		// check if we're able to coninue chain with this header
```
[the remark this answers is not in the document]
svyatonik: [SOLIDITY] Noted

modules/ethereum-contract/contracts/substrate-bridge.sol
```diff
@@ -0,0 +1,327 @@
...
+		}
+
+		// check if parent header has emitted validators set change signal
+		uint64 validatorsSetId = parentHeader.validatorsSetId;
+		bytes32 prevSignalHeaderHash = parentHeader.prevSignalHeaderHash;
+		if (parentHeader.signal.length != 0) {
+			prevSignalHeaderHash = header.parentHash;
+			validatorsSetId = validatorsSetId + 1;
+		}
+
```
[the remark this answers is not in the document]
svyatonik: [SOLIDITY] Yeah, good point! Noted

relays/ethereum/src/ethereum_client.rs
```diff
@@ -14,39 +14,106 @@
...
+	/// Ethereum transactions signer.
+	pub signer: KeyPair,
+	/// Gas price we agree to pay.
+	pub gas_price: U256,
+}
+
+impl Default for EthereumSigningParams {
+	fn default() -> Self {
+		EthereumSigningParams {
+			chain_id: 0x11, // Parity dev chain
```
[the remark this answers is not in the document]
svyatonik: Yes, see #60

relays/ethereum/src/cli.yml
svyatonik: To get genesis header + initial validators set, if it isn't specified. It was the main reason why I needed this option. Otherwise I could just have specified the contract code in the eth chainspec.

relays/ethereum/src/ethereum_deploy_contract.rs
svyatonik: Added header logging. Authorities are opaque to relay.

relays/ethereum/res/substrate-bridge-metadata.txt
```diff
@@ -0,0 +1,5 @@
+Last Change Date: 2020-04-28
+Solc version: 0.6.6+commit.6c089d02
+Source hash (keccak256): 0xdc46aff04e37129265223e507d17f1407a70cb1ecea3230e1eaa77a17586724d
+Source gist: https://gist.github.com/svyatonik/876b388f9507a8de242cb2db9547c4f0
+Compiler flags used (command to produce the file): `docker run -i ethereum/solc:0.6.6 --optimize --bin - < substrate-bridge.sol`
```
tomusdrw, line 5: :bowing_man:

modules/ethereum-contract/contracts/substrate-bridge.sol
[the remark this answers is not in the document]
tomusdrw: Since `0` is used for initialized storage as well, I'd prefer to keep it on the safe side here. i.e. with `finalityTargetNumber ==0` this check would pass with arbitrarty `finalityTargetHash` (i.e. non-existent).
So nothing very specific for now, but just my internal warning side lightning up :)

relays/ethereum/res/substrate-bridge-metadata.txt
```diff
@@ -0,0 +1,5 @@
+Last Change Date: 2020-04-28
```
HCastano, line 1: Is this date necessary? Could we not get this from the commit info?

tomusdrw: Commits sometimes get lost (when you move to a new repo) or we might move stuff around. I'd prefer to keep it explicitly.

relays/ethereum/src/ethereum_sync_loop.rs
```diff
@@ -14,57 +14,58 @@
...
+const MAX_HEADERS_IN_SINGLE_SUBMIT: usize = 32;
+/// Max total size of headers in single submit transaction. This only affects signed
+/// submissions, when several headers are submitted at once. 4096 is the maximal **expected**
+/// size of the Ethereum header + transactions receipts (if they're required).
+const MAX_HEADERS_SIZE_IN_SINGLE_SUBMIT: usize = MAX_HEADERS_IN_SINGLE_SUBMIT * 4096;
+/// Max Ethereum headers we want to have in all 'before-submitted' states.
+const MAX_FUTURE_HEADERS_TO_DOWNLOAD: usize = 128;
+/// Max Ethereum headers count we want to have in 'submitted' state.
+const MAX_SUBMITTED_HEADERS: usize = 128;
+/// Max depth of in-memory headers in all states, before we'll forget about them.
```
HCastano, line 40:
```suggestion
/// Max depth of in-memory headers in all states. Past this depth they will be forgotten (pruned).
```

relays/ethereum/src/substrate_sync_loop.rs
```diff
@@ -0,0 +1,202 @@
...
+
+/// Interval (in ms) at which we check new Substrate headers when we are synced/almost synced.
+const SUBSTRATE_TICK_INTERVAL_MS: u64 = 10_000;
+/// Interval (in ms) at which we check new Ethereum blocks.
+const ETHEREUM_TICK_INTERVAL_MS: u64 = 5_000;
+/// Max Ethereum headers we want to have in all 'before-submitted' states.
+const MAX_FUTURE_HEADERS_TO_DOWNLOAD: usize = 8;
+/// Max Ethereum headers count we want to have in 'submitted' state.
+const MAX_SUBMITTED_HEADERS: usize = 4;
+/// Max depth of in-memory headers in all states, before we'll forget about them.
```
HCastano, line 37: I'd replace this with the same comment suggestion I made earlier

relays/ethereum/src/substrate_sync_loop.rs
```diff
@@ -0,0 +1,202 @@
...
+			eth_contract_address: "731a10897d267e19b34503ad902d0a29173ba4b1"
+				.parse()
+				.expect("address is hardcoded, thus valid; qed"),
+			sub: Default::default(),
+			sync_params: HeadersSyncParams {
+				max_future_headers_to_download: MAX_FUTURE_HEADERS_TO_DOWNLOAD,
+				max_headers_in_submitted_status: MAX_SUBMITTED_HEADERS,
+				// since we always have single Substrate header in separate Ethereum transaction,
+				// all max_**_in_single_submit aren't important here
+				max_headers_in_single_submit: 4,
```
HCastano, line 71: Would this also not be `MAX_SUBMITTED_HEADERS`?

svyatonik: It isn't that important in sub2eth, as comment above states.

HCastano:
> Looks good for me. There are couple of rename suggestions by @HCastano - I don't really have a strong opinion, cause the plural sounds ok to me, but I'm not a native speaker.

If the structs were some sort of collection I'd be okay with the plurals. Since they're not the naming is a bit awkward. While I would prefer to see them changed, I also don't have a strong enough opinion that I'm going to block this over the naming, lol

relays/ethereum/Cargo.toml
```diff
@@ -11,11 +11,17 @@ async-stream = "0.2.0"
 clap = { version = "2.33.0", features = ["yaml"] }
 codec = { package = "parity-scale-codec", version = "1.0.0" }
 env_logger = "0.7.0"
+ethabi = "12.0"
+ethabi-contract = "11.0"
+ethabi-derive = "11.0"
+ethereum-tx-sign = { git = "https://github.com/svyatonik/ethereum-tx-sign.git", branch = "up-ethereum-types" }
```
svyatonik, line 17: The biggest change of latest merge-with-master - PR is already opened in original repo, but let's refer to my branch for now

PR #69, opened 2020-04-20
Process finality proofs in solidity PoA -> Substrate contract

svyatonik:
Based on top of #65. Right now this PR has [the single commit](https://github.com/paritytech/parity-bridges-common/commit/0bc574307135be9e42c1607bce228c8e42175af9).

There's a lot of things to do - I'm opening it early to discuss things that must be discussed.

A brief overview of TODOs:
- [x] Discuss: how should we verify grandpa justifications - write custom code in builtin? Extract `GrandpaJustification::decode_and_verify()` in separate crate? Make Substrate RPC for that?
- [x] Discuss: do we want to support other kinds of finality proofs? Like `Commit`? Do we need custom RPCs for that? Create issue, if required.
- [x] Extract runtime changes to separate PR;
- [x] Support forced validators set changes (or at least create issue for that - logged #76 )
- [x] Provide tests, where possible


modules/ethereum-contract/builtin/src/lib.rs
```diff
@@ -0,0 +1,264 @@
...
+	)
+	.map_err(Error::JustificationVerify)
+	.map(|_| ())
+}
+
+/// ==================================================================================================================
+/// === Everything below should be (?) exposed by Substrate ==========================================================
+/// ==================================================================================================================
+
+mod substrate {
```
svyatonik, line 127: @HCastano @tomusdrw This is the most important question about this PR - how's better to organize this check? This code is now well isolated in single [Substrate file](https://github.com/paritytech/substrate/blob/master/client/finality-grandpa/src/justification.rs), but all methods are private. I think it may be helpful for other developers as well.

IMO it is tightly linked to what you're working on, @HCastano ? I mean that finality subscription RPC - what's going on with justifications that are read using this subcription? Are they blindly trusted, or verified somewhere else? Just want to synchronize here, so we won't make similar changes to Substrate code :)

For using in builtin, we have 2 options. The worst is to duplicate code here (this is what you're seeing now) - it depends on Substrate internals => let's not use it. So the only option (imo) is to make `GrandpaJustification::decode_and_verify()` public (probably extracting this whole struct in separate crate before). Wdyt?

modules/ethereum-contract/builtin/src/lib.rs
```diff
@@ -0,0 +1,264 @@
...
+		.map_err(Error::HeaderDecode)
+}
+
+/// Verify GRANDPA finality proof.
+pub fn verify_substrate_finality_proof(
+	finality_target_number: BlockNumber,
+	finality_target_hash: Hash,
+	best_set_id: u64,
+	raw_best_set: &[u8],
+	raw_finality_proof: &[u8],
```
svyatonik, line 108: @HCastano @tomusdrw TIL (not actually today, but...) that we may listen to `Commit` messages on full nodes && finalize headers using these messages. Let me refresh your memory - `Justifications` are **rare** and if we'd depend on justifications, then we may only finalize headers in huge bunches (like ~30000 blocks - see https://github.com/paritytech/substrate/issues/4728 for example). That is the problem for native light client as well.

We may also want to support finalization by `Commit` messages here. The problem is that (afaiu) these messages are not persistent && we can't access these from RPCs. So if we would choose to use it, then we need some RPCs to deal with that. I'm not sure if that's a part of that finality subscription RPC, Hernando is working on? Wdyt - do we need it now, or it could be added later?

svyatonik: RPC that @HCastano is working on, will emit non-persistent justifications, made of intermediate commits. So the only thing we need is to use that RPC in relay, when it'll be ready.

svyatonik: logged: #71

modules/ethereum-contract/builtin/src/lib.rs
svyatonik: https://github.com/paritytech/substrate/pull/5746

svyatonik:
re forced changes - extracted to #76

relays/ethereum/src/substrate_sync_loop.rs
```diff
@@ -89,6 +91,16 @@ impl SourceClient<SubstrateHeadersSyncPipeline> for SubstrateHeadersSource {
...
 	type HeaderExtraFuture = Ready<(Self, Result<(SubstrateHeaderId, ()), Self::Error>)>;
+	type HeaderCompletionFuture = Pin<
+		Box<
+			dyn Future<
+				Output = (
+					Self,
+					Result<(SubstrateHeaderId, Option<GrandpaJustification>), Self::Error>,
+				),
+			>,
+		>,
```
tomusdrw, line 102: That's `rustfmt`, right? Can we try to squeeze it into less lines? Maybe introducing:
```
type ClientOutput<X, T> = (X, Result<T, X::Error>);
```
would help here?

relays/ethereum/src/sync_loop.rs
```diff
@@ -30,13 +34,13 @@ use std::future::Future;
 /// direct child of previous best header. But: (1) subscription doesn't guarantee that
 /// the subscriber will receive every best header (2) reorg won't always lead to sync
 /// stall and restart is a heavy operation (we forget all in-memory headers).
-const STALL_SYNC_TIMEOUT_MS: u64 = 30_000;
+const STALL_SYNC_TIMEOUT: Duration = Duration::from_millis(5 * 60 * 1_000);
```
tomusdrw, line 37: ditto `from_secs` here and below

relays/ethereum/src/sync_loop.rs
```diff
@@ -50,6 +54,13 @@ pub trait SourceClient<P: HeadersSyncPipeline>: Sized {
 	type HeaderByNumberFuture: Future<Output = (Self, Result<P::Header, Self::Error>)>;
 	/// Future that returns extra data associated with header.
 	type HeaderExtraFuture: Future<Output = (Self, Result<(HeaderId<P::Hash, P::Number>, P::Extra), Self::Error>)>;
+	/// Future that returns data required to 'complete' header.
+	type HeaderCompletionFuture: Future<
+		Output = (
```
tomusdrw, line 59: `ClientOutput` helpful here as well? In all associated types methods?

relays/ethereum/src/headers.rs
```diff
@@ -379,10 +484,81 @@ impl<P: HeadersSyncPipeline> QueuedHeaders<P> {
...
+		let status = self.status(id);
+		let header = match status {
+			HeaderStatus::MaybeOrphan => header(&self.maybe_orphan, id),
+			HeaderStatus::Orphan => header(&self.orphan, id),
+			HeaderStatus::MaybeExtra => header(&self.maybe_extra, id),
+			HeaderStatus::Extra => header(&self.extra, id),
+			HeaderStatus::Ready => header(&self.ready, id),
+			HeaderStatus::Incomplete => header(&self.incomplete, id),
+			HeaderStatus::Submitted => header(&self.submitted, id),
+			HeaderStatus::Unknown => return false,
```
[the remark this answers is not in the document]
svyatonik: (1) it only affects `Unknown` case - `Synced` should mean false (2) I think it is documented well enough - I mean when true is returned (3) it is private method that is only called twice from within this module (4) I find it much harder to understand what `Option<bool>` means (especially if all calls end up with `.unwrap_or(false)`) vs understanding `true`/`false`

relays/ethereum/src/headers.rs
```diff
@@ -331,17 +307,42 @@ impl<P: HeadersSyncPipeline> QueuedHeaders<P> {
...
 	}
 
+	/// Receive completion response from source node.
+	pub fn completion_response(&mut self, id: &HeaderId<P::Hash, P::Number>, completion: Option<P::Completion>) {
+		let completion = match completion {
+			Some(completion) => completion,
+			None => return, // we'll try refetch later
+		};
+
+		if let Some(_) = self.incomplete_headers.remove(id) {
```
HCastano, line 334:
```suggestion
		if self.incomplete_headers.remove(id).is_some() {
```

relays/ethereum/src/headers.rs
```diff
@@ -356,6 +357,109 @@ impl<P: HeadersSyncPipeline> QueuedHeaders<P> {
 		}
 	}
 
+	/// When header completion data is sent to target node.
+	pub fn header_completed(&mut self, id: &HeaderId<P::Hash, P::Number>) {
+		if let Some(_) = self.completion_data.remove(id) {
```
HCastano, line 362:
```suggestion
		if self.completion_data.remove(id).is_some() {
```

relays/ethereum/src/substrate_sync_loop.rs
```diff
@@ -14,22 +14,24 @@
...
-use crate::sync_loop::{SourceClient, TargetClient};
+use crate::sync_loop::{OwnedSourceFutureOutput, OwnedTargetFutureOutput, SourceClient, TargetClient};
 use crate::sync_types::SourceHeader;
 use futures::future::{ready, FutureExt, Ready};
-use std::{future::Future, pin::Pin};
+use std::{collections::HashSet, future::Future, pin::Pin, time::Duration};
 
 /// Interval (in ms) at which we check new Substrate headers when we are synced/almost synced.
-const SUBSTRATE_TICK_INTERVAL_MS: u64 = 10_000;
+const SUBSTRATE_TICK_INTERVAL: Duration = Duration::from_millis(10 * 1_000);
```
HCastano, line 32:
```suggestion
const SUBSTRATE_TICK_INTERVAL: Duration = Duration::from_secs(10);
```

relays/ethereum/src/substrate_sync_loop.rs
```diff
@@ -14,22 +14,24 @@
...
 };
 use crate::sync::{HeadersSyncParams, TargetTransactionMode};
-use crate::sync_loop::{SourceClient, TargetClient};
+use crate::sync_loop::{OwnedSourceFutureOutput, OwnedTargetFutureOutput, SourceClient, TargetClient};
 use crate::sync_types::SourceHeader;
 use futures::future::{ready, FutureExt, Ready};
-use std::{future::Future, pin::Pin};
+use std::{collections::HashSet, future::Future, pin::Pin, time::Duration};
 
 /// Interval (in ms) at which we check new Substrate headers when we are synced/almost synced.
```
HCastano, line 31:
```suggestion
/// Interval at which we check new Substrate headers when we are synced/almost synced.
```

relays/ethereum/src/substrate_sync_loop.rs
```diff
@@ -14,22 +14,24 @@
...
-const SUBSTRATE_TICK_INTERVAL_MS: u64 = 10_000;
+const SUBSTRATE_TICK_INTERVAL: Duration = Duration::from_millis(10 * 1_000);
 /// Interval (in ms) at which we check new Ethereum blocks.
-const ETHEREUM_TICK_INTERVAL_MS: u64 = 5_000;
+const ETHEREUM_TICK_INTERVAL: Duration = Duration::from_millis(5 * 1_000);
```
HCastano, line 34:
```suggestion
const ETHEREUM_TICK_INTERVAL: Duration = Duration::from_secs(5);
```

relays/ethereum/src/substrate_sync_loop.rs
```diff
@@ -14,22 +14,24 @@
...
+use crate::sync_loop::{OwnedSourceFutureOutput, OwnedTargetFutureOutput, SourceClient, TargetClient};
 use crate::sync_types::SourceHeader;
 use futures::future::{ready, FutureExt, Ready};
-use std::{future::Future, pin::Pin};
+use std::{collections::HashSet, future::Future, pin::Pin, time::Duration};
 
 /// Interval (in ms) at which we check new Substrate headers when we are synced/almost synced.
-const SUBSTRATE_TICK_INTERVAL_MS: u64 = 10_000;
+const SUBSTRATE_TICK_INTERVAL: Duration = Duration::from_millis(10 * 1_000);
 /// Interval (in ms) at which we check new Ethereum blocks.
```
HCastano, line 33:
```suggestion
/// Interval at which we check new Ethereum blocks.
```

HCastano:
Aside from the changes in `header.rs` it looks fine. I need to sit down again and go through the new changes in `headers` because I don't fully understand them quite yet.

modules/ethereum-contract/builtin/src/lib.rs
```diff
@@ -45,35 +49,266 @@ pub struct Header {
...
+			ConsensusLog::ScheduledChange(ScheduledChange {
+				next_authorities: authorities.clone(),
+				delay: 8,
+			})
+			.encode(),
+		));
+
+		assert_eq!(
+			parse_substrate_header(
+				&RuntimeHeader {
```
tomusdrw, line 203: Can you please add one more comparison against hex to make sure we never break backward compatibility?

modules/ethereum-contract/builtin/src/lib.rs
```diff
@@ -45,35 +49,266 @@ pub struct Header {
...
+	const EXAMPLE_JUSTIFICATION: &'static str = "2600000000000000a2f45892db86b2ad133ce57d81b7e4375bb7035ce9883e6b68c358164f3437750800000010a2f45892db86b2ad133ce57d81b7e4375bb7035ce9883e6b68c358164f34377508000000d66b4ceb57ef8bcbc955071b597c8c5d2adcfdbb009c73f8438d342670fdeca9ac60686cbd58105b10f51d0a64a8e73b2e5829b2eab3248a008c472852130b00439660b36c6c03afafca027b910b4fecf99801834c62a5e6006f27d978de234fa2f45892db86b2ad133ce57d81b7e4375bb7035ce9883e6b68c358164f34377508000000f5730c14d3cd22b7661e2f5fcb3139dd5fef37f946314a441d01b40ce1200ef70d810525f23fd278b588cd67473c200bda83c338c407b479386aa83798e5970b5e639b43e0052c47447dac87d6fd2b6ec50bdd4d0f614e4299c665249bbd09d9a2f45892db86b2ad133ce57d81b7e4375bb7035ce9883e6b68c358164f34377508000000c78d6ec463f476461a695b4791d30e7626d16fdf72d7c252c2cad387495a97e8c2827ed4d5af853d6e05d31cb6fb7438c9481a7e9c6990d60a9bfaf6a6e1930988dc3417d5058ec4b4503e0c12ea1a0a89be200fe98922423d4334014fa6b0eea2f45892db86b2ad133ce57d81b7e4375bb7035ce9883e6b68c358164f3437750800000052b4fc52d430286b3e2d650aa6e01b6ff4fae8b968893a62be789209eb97ee6e23780d3f5af7042d85bb48f1b202890b22724dfebce138826f66a5e00324320fd17c2d7823ebf260fd138f2d7e27d114c0145d968b5ff5006125f2414fadae6900";
```
tomusdrw, line 248: Thanks for the explanations!

Is there a way to actually generate that as well? How involved would it be?

relays/ethereum/src/headers.rs
```diff
@@ -544,6 +709,44 @@ fn prune_known_headers<P: HeadersSyncPipeline>(known_headers: &mut KnownHeaders<
...
+	);
+	*known_headers.entry(id.0).or_default().entry(id.1).or_insert(status) = status;
+}
+
+/// Returns queued imcomplete header with maximal elapsed time since last update.
+fn queued_incomplete_header<Id: Clone + Eq + std::hash::Hash, T>(
+	map: &mut LinkedHashMap<Id, T>,
+	filter: impl FnMut(&mut T) -> bool,
+) -> Option<(Id, &T)> {
+	// TODO: headers that have been just appended to the end of the queue would have to wait until
```
tomusdrw, line 733: Can we make that into an issue?

relays/ethereum/src/headers.rs
```diff
@@ -356,6 +357,109 @@ impl<P: HeadersSyncPipeline> QueuedHeaders<P> {
...
+			//
+			// the other option would be to track emitted transactions at least on target node,
+			// but it won't give us 100% guarantee anyway
+			//
+			// => we're just dropping completion data just after it has been submitted
+		}
+	}
+
+	/// When incomplete headers ids are receved from target node.
+	pub fn incomplete_headers_response(&mut self, ids: HashSet<HeaderId<P::Hash, P::Number>>) {
```
svyatonik, line 385: Need a test for case when two relays are running and relay#2 completes some header, making some descendant incomplete => make sure that relay#1 will also mark it incomplete, without submitting children.

modules/ethereum-contract/builtin/src/lib.rs
svyatonik: There's no way - it's encoded `GrandpaJustification`, struct with private fields.

relays/ethereum/src/headers.rs
svyatonik: #84

modules/ethereum-contract/builtin/src/lib.rs
tomusdrw: Could we just copy paste the code on our side? Sorry for pushing this so hard, but I remember the experience of debugging smart contracts and JSON tests in Ethereum, where such encoded hexes where pretty common. Within few months I bet we will completely forget the knowledge where this hexes come from and what were the assumptions and having a way to alter the parameters easily is going to be crucial.
Also if we want to add more tests in the future, we need a way to easily generate different edge cases that we need - with hexes the path of least resistance is to not add a test at all, cause the cost of generating the blob is super high if you don't know how to do it.

svyatonik: OK, I think we may split your comment into two parts.

>  I bet we will completely forget the knowledge where this hexes come from and what were the assumptions and having a way to alter the parameters easily is going to be crucial

I agree that simple hex literals weren't good enough - now we have a comments on how to update this. So if anything breaks, the only thing you need is to get block (hash and number) and justification from the node - that is well commented (I hope). Justification is currently opaque to the builtin and I'm not testing if **justification generation ** and **justification verification** works correctly here. I don't care whether block hash is simply encoded `int` and justification is simply two encoded `int`s which sum is the block hash, or it is some multisig structure. The only thing I'm testing here is that if the contract stores valid (from `sc_finality_grandpa` point of view) values of **header**, **validators set** and **justification**, then the block is finalized. And that it isn't finalized if contract stores some garbage.

> cause the cost of generating the blob is super high if you don't know how to do it

Please see comments above constants to see how to get these blobs. I don't think that cost of calling two RPC methods on first block with justification is that high. I'd be happy to fix comments if you have anything to add.

modules/ethereum-contract/builtin/src/lib.rs
```diff
@@ -45,35 +49,272 @@ pub struct Header {
...
+#[derive(Debug, PartialEq)]
 pub struct ValidatorsSetSignal {
 	/// Signal delay.
 	pub delay: BlockNumber,
 	/// New validators set.
 	pub validators: Vec<u8>,
 }
 
+/// Convert from U256 to BlockNumber. Fails if `U256` value isn't fitting within `BlockNumber`
+/// limits (the runtime referenced by this module uses u32 as `BlockNumber`).
```
HCastano, line 61: What would happen if we changed the runtime to use `u64` for `BlockNumber` without updating this function?

relays/ethereum/src/sync_loop.rs
```diff
@@ -18,7 +18,11 @@ use crate::sync::HeadersSyncParams;
...
 use num_traits::{Saturating, Zero};
-use std::future::Future;
+use std::{
+	collections::HashSet,
+	future::Future,
+	time::{Duration, Instant},
+};
 
 /// When we submit headers to target node, but see no updates of best
 /// source block known to target node during STALL_SYNC_TIMEOUT_MS milliseconds,
```
HCastano, line 28:
```suggestion
/// source block known to target node during STALL_SYNC_TIMEOUT milliseconds,
```

HCastano:
I've gone through this again, and I'm still struggling to fully understand the changes in `headers.rs`. I realize that the root of my problem is that I don't understand what it means for a header to be `complete` or `incomplete`. Can you clarify this for me Slava?

svyatonik:
> I've gone through this again, and I'm still struggling to fully understand the changes in headers.rs. I realize that the root of my problem is that I don't understand what it means for a header to be complete or incomplete. Can you clarify this for me Slava?

Most of Substrate headers are "complete". The only case when header is "incomplete", is when we it requires justification (i.e. it is the header that enacts new GRANDPA authorities set). This requirement is tracked by the contract. So when contract accepts "incomplete" header, it stops accepting its descendants, until justification is submitted.

modules/ethereum-contract/builtin/src/lib.rs
svyatonik: Nothing good would happen. This code references specific runtime, so whenever you change the runtime, it must be updated

modules/ethereum-contract/builtin/src/lib.rs
tomusdrw: > Justification is currently opaque to the builtin and I'm not testing if **justification generation ** and justification verification works correctly her

I fully understand that. We should definitely make sure that we are able to parse not the things that are generated by our code, but rather whatever substrate chain generates.

> The only thing I'm testing here is that if the contract stores valid (from sc_finality_grandpa point of view) values of header, validators set and justification, then the block is finalized. And that it isn't finalized if contract stores some garbage.

Totally agree with this too. However my main point with high cost of generating this is that it's not easy to control it. Like imagine you want to add more tests with some specific validator set (say one validator or even number of validators), or some peculiarity with Justification that has to be handled differently. In such case you would need to run a specific substrate chain, wait until it generates a particular justification you need and only then retrieve it over the RPC.

That said, I'm really far from blocking the PR and pushing you to implement that. I'm just seeing this from a bystander perspective, didn't really have a chance to work on the code directly to see if this is a really issue or just my imagination :)

relays/ethereum/src/headers.rs
```diff
@@ -356,6 +357,109 @@ impl<P: HeadersSyncPipeline> QueuedHeaders<P> {
...
+				&mut [&mut self.incomplete],
+				&mut self.ready,
+				&mut self.known_headers,
+				HeaderStatus::Ready,
+				&just_completed_header,
+			);
+
+			log::debug!(
+				target: "bridge",
+				"Completion data is no more required for header: {:?}",
```
HCastano, line 432:
```suggestion
				"Completion data is no longer required for header: {:?}",
```

relays/ethereum/src/headers.rs
```diff
@@ -356,6 +357,109 @@ impl<P: HeadersSyncPipeline> QueuedHeaders<P> {
...
+		})
+		.map(|(id, _)| id)
+	}
+
+	/// Returns header completion data to upload to target node.
+	pub fn header_to_complete(&mut self) -> Option<(HeaderId<P::Hash, P::Number>, &P::Completion)> {
+		queued_incomplete_header(&mut self.completion_data, |_| true)
+	}
+
 	/// Prune and never accep headers before this block.
```
HCastano, line 463:
```suggestion
	/// Prune and never accept headers before this block.
```

relays/ethereum/src/headers.rs
```diff
@@ -411,6 +587,14 @@ fn remove_header<P: HeadersSyncPipeline>(
 	header
 }
 
+/// Get header from the qeueue.
```
HCastano, line 590:
```suggestion
/// Get header from the queue.
```

relays/ethereum/src/headers.rs
```diff
@@ -411,6 +587,14 @@ fn remove_header<P: HeadersSyncPipeline>(
...
+fn header<'a, P: HeadersSyncPipeline>(
```
HCastano, line 591:
```suggestion
fn get_header_from_queue<'a, P: HeadersSyncPipeline>(
```

HCastano: I'm suggesting this since when I was reading some of the code that was like `HeaderStatus::MaybeOrphan => header(...)` it wasn't clear at first glance that `header()` was fetching stuff from the queues in the same way that seeing `queue.get(id)` would be.

relays/ethereum/src/headers.rs
```diff
@@ -544,6 +709,44 @@ fn prune_known_headers<P: HeadersSyncPipeline>(known_headers: &mut KnownHeaders<
...
+		target: "bridge",
+		"{} header {:?} is now {:?}",
+		P::SOURCE_NAME,
+		id,
+		status,
+	);
+	*known_headers.entry(id.0).or_default().entry(id.1).or_insert(status) = status;
+}
+
+/// Returns queued imcomplete header with maximal elapsed time since last update.
```
HCastano, line 728:
```suggestion
/// Returns queued incomplete header with maximal elapsed time since last update.
```

relays/ethereum/src/headers.rs
HCastano, line 733:
```suggestion
	// TODO (#84): headers that have been just appended to the end of the queue would have to wait until
```

relays/ethereum/src/headers.rs
```diff
@@ -1050,4 +1456,51 @@ pub(crate) mod tests {
 		queue.header_response(header(110).header().clone());
 		assert_eq!(queue.known_headers.len(), 1);
 	}
+
+	#[test]
+	fn incomplete_headers_are_still_incomplete_after_advance() {
+		let mut queue = QueuedHeaders::<EthereumHeadersSyncPipeline>::new();
+
+		// relay#1 knows that header#100 is incomplete && it has headers 101..104 in inomplete queue
```
HCastano, line 1464:
```suggestion
		// relay#1 knows that header#100 is incomplete && it has headers 101..104 in incomplete queue
```

relays/ethereum/src/headers.rs
```diff
@@ -1050,4 +1456,51 @@ pub(crate) mod tests {
...
+			.or_default()
+			.insert(hash(103), HeaderStatus::Incomplete);
+		queue
+			.known_headers
+			.entry(104)
+			.or_default()
+			.insert(hash(104), HeaderStatus::Incomplete);
+
+		// let's say relay#2 completes header#100 and then submits header#101+header#102 and it turns
+		// that header#102 is also incomplete
```
HCastano, line 1497:
```suggestion
		// out that header#102 is also incomplete
```

HCastano:
> > I've gone through this again, and I'm still struggling to fully understand the changes in headers.rs. I realize that the root of my problem is that I don't understand what it means for a header to be complete or incomplete. Can you clarify this for me Slava?
> 
> Most of Substrate headers are "complete". The only case when header is "incomplete", is when we it requires justification (i.e. it is the header that enacts new GRANDPA authorities set). This requirement is tracked by the contract. So when contract accepts "incomplete" header, it stops accepting its descendants, until justification is submitted.

Thanks for the clarification. I think you should add this (or some version of it) as a comment to the `HeaderCompletion` struct. I also think that the associated `Completion` types could have a short explanation (e.g the Ethereum `Completion` type is `()` because it doesn't not require any sort of `Justification`).

HCastano:
I've left a few comments, mainly just typo fixes and re-naming suggestions, nothing major :)

PR #70, opened 2020-04-22
Local testnet mode for substrate node

svyatonik:
This PR:
1) adds session pallet/frame/whatever to our node runtime, so that there are more than 1 session in Aura/Grandpa;
2) sessions are changing every 4 blocks;
3) Aura/Grandpa validators are changing on every session;
4) you can now use `--chain=local` option when starting node to start it with Alice/Bob/Charlie/Dave/Eve initial validators. Starting with session 2, validators will start changing.

This is all to be able to test our Substrate -> Ethereum sync out of the box. Previous node version had only `--dev` option, so we always had single session with single validator (Alice). I wasn't sure if solidity contract works well with changing GRANDPA validators sets. And now it's the only way to test it.


bin/node/node/src/chain_spec.rs
```diff
@@ -110,13 +157,19 @@ fn testnet_genesis(
 			balances: endowed_accounts.iter().cloned().map(|k| (k, 1 << 60)).collect(),
 		}),
 		pallet_aura: Some(AuraConfig {
-			authorities: initial_authorities.iter().map(|x| (x.0.clone())).collect(),
+			authorities: Vec::new(),
```
tomusdrw, line 160: It's not necessary, cause afair internally in the runtime there is a callback anyway to initialize genesis keys, right?

svyatonik: Yes. And it panics if authorities are set both in session and aura/grandpa config (even if they're the same)

PR #79, opened 2020-04-28
Migrate to new substrate

svyatonik:
I need that to be able to continue #69 (especially `GrandpaJustification::decode_and_verify_finalizes`). It seems that the update that made that fn public isn't yet on crates.io, so I've changed suffix from `-alpha.2` to `-alpha.6` (that's the last published version) && updated commit to the latest commit of Substrate (not sure why we specify suffixes if we're pointing to the specific commit anyway? :) ).

So whenever actual 2.0 beta will be released on crates.io, we'll need another round of updates, but:
1) it would be much easier (probably just replacing these `[dependencies.*]` with simple crates references);
2) this PR would allow us to continue working without depending on Substrate release schedule.

P.S.: this PR also includes `cargo update`

### DIFF
--- a/modules/ethereum/src/lib.rs
+++ b/modules/ethereum/src/lib.rs
@@ -293,7 +293,7 @@ pub trait Trait: frame_system::Trait {
 decl_module! {
 	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
 		/// Import single Aura header. Requires transaction to be **UNSIGNED**.
-		#[weight = 0]
+		#[weight = 0] // TODO: update me (https://github.com/paritytech/parity-bridges-common/issues/78)
 		pub fn import_unsigned_header(origin, header: Header, receipts: Option<Vec<Receipt>>) {
 			frame_system::ensure_none(origin)?;
 
@@ -314,7 +314,7 @@ decl_module! {
 		///
 		/// This should be used with caution - passing too many headers could lead to
 		/// enormous block production/import time.
-		#[weight = 0]
+		#[weight = 0] // TODO: update me (https://github.com/paritytech/parity-bridges-common/issues/78)
 		pub fn import_signed_headers(origin, headers_with_receipts: Vec<(Header, Option<Vec<Receipt>>)>) {
 			let submitter = frame_system::ensure_signed(origin)?;
 			let mut finalized_headers = BTreeMap::new();
--- a/modules/ethereum/src/lib.rs
+++ b/modules/ethereum/src/lib.rs
@@ -21,8 +21,8 @@ use frame_support::{decl_module, decl_storage};
 use primitives::{Address, Header, Receipt, H256, U256};
 use sp_runtime::{
 	transaction_validity::{
-		InvalidTransaction, TransactionLongevity, TransactionPriority, TransactionSource,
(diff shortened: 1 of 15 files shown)


modules/ethereum/src/lib.rs
```diff
@@ -359,19 +361,19 @@ decl_storage! {
 		/// Oldest unpruned block(s) number.
 		OldestUnprunedBlock: u64;
 		/// Map of imported headers by hash.
-		Headers: map hasher(blake2_256) H256 => Option<StoredHeader<T::AccountId>>;
+		Headers: map hasher(blake2_128_concat) H256 => Option<StoredHeader<T::AccountId>>;
```
tomusdrw, line 364: that can probably be `identity`, no?

modules/ethereum/src/lib.rs
```diff
@@ -359,19 +361,19 @@ decl_storage! {
...
-		ValidatorsSets: map hasher(blake2_256) u64 => Option<ValidatorsSet>;
+		ValidatorsSets: map hasher(blake2_128_concat) u64 => Option<ValidatorsSet>;
 		/// Validators sets reference count. Each header that is authored by this set increases
 		/// the reference count. When we prune this header, we decrease the reference count.
 		/// When it reaches zero, we are free to prune validator set as well.
-		ValidatorsSetsRc: map hasher(blake2_256) u64 => Option<u64>;
+		ValidatorsSetsRc: map hasher(blake2_128_concat) u64 => Option<u64>;
 		/// Map of validators set changes scheduled by given header.
-		ScheduledChanges: map hasher(blake2_256) H256 => Option<ScheduledChange>;
+		ScheduledChanges: map hasher(blake2_128_concat) H256 => Option<ScheduledChange>;
```
tomusdrw, line 376: Same here.

modules/substrate/src/lib.rs
```diff
@@ -96,18 +96,19 @@ pub trait Trait: system::Trait {}
...
-		pub TrackedBridges get(tracked_bridges): map hasher(blake2_256) BridgeId => Option<BridgeInfo<T>>;
+		pub TrackedBridges get(fn tracked_bridges): map hasher(blake2_128_concat) BridgeId => Option<BridgeInfo<T>>;
 	}
 }
 
 decl_module! {
 	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
 		type Error = Error<T>;
 
+		#[weight = 0] // TODO: update me
```
tomusdrw, line 111: Can you add an issue number to all the todos?

modules/ethereum/src/lib.rs
svyatonik: Yeah, you're probably right - good point :) I've also changed `ValidatorsSets` and `ValidatorsSetsRc`, because keys there are `A (slowly) incrementing index` (c).

modules/substrate/src/lib.rs
svyatonik: The issue I've filed today is about PoA -> Substrate bridge. I was thinking about submitting second issue about Substrate -> Substrate bridge, but looking at the code - 1 of 2 calls there is a noop and I'm not sure about the future of the other one. So I've decided not to spam issues now - I do not know what's the status/future of that part.

Issue #74, opened 2020-04-24
Extract low-level RPC calls to a separate module


tomusdrw:
to have them properly typed and to avoid duplication and mixing rpc boilerplate with relayer logic, see:
https://github.com/paritytech/parity-bridges-common/pull/65#discussion_r414460970

PR #80, opened 2020-04-29
RPC Module for Relays

HCastano:
Takes a stab at closing #74.

The idea here is that using the `rpc_api!` macro provided by `jsonrpsee` we're able to define all the "raw" RPC calls for both Substrate and Ethereum. These raw calls can then be used by the `SubstrateRpcClient` and `EthereumRpcClient`, which would take place of what's being done in `substrate_client.rs` and `ethereum_client.rs` respectively. I kept the structs and the RPC API in the same module because I didn't want to expose the RPC API publicly.

I imagine the usage will be something like this:

```rust
let mut eth_rpc = EthereumRpcClient::new(...);
eth_rpc.estimate_gas(...)
```

which is nicer than having to pass `Client` around everywhere.

I also haven't implemented all the RPCs yet, since I want to check what you guys think of this approach before I commit to that.


relays/ethereum/src/rpc.rs
```diff
@@ -0,0 +1,114 @@
...
+		Self { client }
+	}
+}
+
+#[async_trait]
+impl SubstrateRpc for SubstrateRpcClient {
+	async fn best_header(&mut self) -> Result<SubstrateHeader, ()> {
+		Ok(Substrate::chain_getHeader(&mut self.client, Params::None)
+			.await
+			.unwrap())
```
tomusdrw, line 104: Can we turn `unwrap` into `expect` or handle the error?

relays/ethereum/src/rpc.rs
```diff
@@ -0,0 +1,114 @@
...
+			serde_json::to_value(2).unwrap(),
+			serde_json::to_value(3).unwrap(),
+		]);
+
+		let gas = Ethereum::eth_estimateGas(&mut self.client, call_request).await.unwrap();
+		Ok(gas)
+	}
+}
+
+#[async_trait]
```
tomusdrw, line 79: :heart_eyes:

relays/ethereum/src/rpc.rs
```diff
@@ -0,0 +1,114 @@
...
+pub trait SubstrateRpc {
+	async fn best_header(&mut self) -> Result<SubstrateHeader, ()>;
```
tomusdrw, line 81: Missing docs?

relays/ethereum/src/rpc.rs
```diff
@@ -0,0 +1,114 @@
...
+#[async_trait]
+impl SubstrateRpc for SubstrateRpcClient {
+	async fn best_header(&mut self) -> Result<SubstrateHeader, ()> {
+		Ok(Substrate::chain_getHeader(&mut self.client, Params::None)
+			.await
+			.unwrap())
+	}
+
+	async fn header_by_hash(&mut self, hash: Hash) -> Result<SubstrateHeader, ()> {
+		let hash = serde_json::to_value(hash).unwrap();
```
tomusdrw, line 108: Should be `expect` as well.

relays/ethereum/src/rpc.rs
HCastano: Yeah, errors will be properly handled. I just wanted to put something out there first

relays/ethereum/src/rpc.rs
HCastano: Yeah, that'll get done

HCastano:
A few things I'm still thinking about:

1. Level of abstraction for RpcClient implementations

I want to keep the logic as low as I can, while keeping some of the magic behind the scenes (e.g wrapping the call arguments into `Params` and decoding results

2. Error handling

Right now I'm just "handling" (although that's a loose use of the word, lol) errors that come back from the `HttpTransport` which are fairly generic. I think I need to integrate some of the more specific errors that Slava put for both Substrate and Ethereum, but this comes back to the level of abstraction for this module. Should I let the caller deal with these?

svyatonik:
Looks good :) Please keep in mind that the structure of `sync_loop.rs` should be left the same, even though you'll be able to issue multiple requests at a time. I mean - once one RPC call is issued, we should wait for completion && do not issue other requests in the meantime (the only exception is headers submission where we are submitting multiple transactions at once).

HCastano:
I've gotten to a point where I'm trying to update `sync_loop` for the new interface. I've run into a few problems since the RPCs don't return a `Client` when they're called anymore.

The `sync_loop` works with Futures from the `TargetClient` trait. Previously the return types from implementers of this trait were something like this:

```rust
type BestHeaderIdFuture = Pin<Box<dyn Future<Output = (Self, Result<EthereumHeaderId, Self::Error>)>>>;
```
but can now be changed to this since the RPCs don't return a `Client`.

```rust
type BestHeaderIdFuture = Pin<Box<dyn Future<Output = Result<EthereumHeaderId, Self::Error>>>>;
```

Note the lack of `Self` in `Output`. The methods from this trait will also require a `&mut self` instead of an owned `self` (keep this in mind).

Since I don't have a `Client` returned anymore, I want to update `sync_loop::process_future_result()` to use a `&mut Client` instead of an owned `Client`. Since there are a lot of futures sharing the same client, I figured I would need something like an `Arc<Mutex<Client>>`. However, I noticed that the `maybe_client` argument of `process_future_result()` already uses a `&mut` without any synchronization primitives and it works fine. Should I be able to do the same thing with `client` then?

If `process_future_result()` can simply take a `&mut Client`, is there any point in having `maybe_client` anymore?  Also with this change, does `sync_loop::delay()` need to take `client` as a parameter anymore? I don't think it makes sense to continue doing so since it's not doing anything with it.

The Futures used by `futures::select!` come from the methods of the `TargetClient` trait (e.g `source_best_block_number_future` comes from [here](https://github.com/paritytech/parity-bridges-common/blob/master/relays/ethereum/src/sync_loop.rs#L296), which in turn comes from [here](https://github.com/paritytech/parity-bridges-common/blob/master/relays/ethereum/src/sync_loop.rs#L55)), which will, as previously mentioned, now use a `&mut self` instead of `self`. So the question comes up again, do I need any sort of synchronization primitives to make sure these methods work correctly?

One last note. The more I work through this PR, the more it seems like it'll have to close #72 based off the changes I made to the underlying RPC interface.

svyatonik:
IMO the main thing to keep is the concept of `maybe_client` or some other flag which means that the client is currently busy. Right now `maybe_client` is `None` when some future is executing => this allows us serializing RPC requests. All other stuff can be refactored to whatever works/you think is better.

`Mutex`es aren't actually required, because loop uses single thread executor. So unless you're trying to migrate to thread pool executor, they are redundant.

tomusdrw:
> One last note. The more I work through this PR, the more it seems like it'll have to close #72 based off the changes I made to the underlying RPC interface.

I think it's fine, unless you want to avoid that then maybe don't migrate everything - just keep old code in `sync_loop` for now and it can be fixed in the next PR by whoever is addressing #72

HCastano:
> > One last note. The more I work through this PR, the more it seems like it'll have to close #72 based off the changes I made to the underlying RPC interface.
> 
> I think it's fine, unless you want to avoid that then maybe don't migrate everything - just keep old code in `sync_loop` for now and it can be fixed in the next PR by whoever is addressing #72

Yeah, I think I'm gonna split this into two. The integration of the RPCs is getting a little bit messy, and I think it'll be distracting for whoever reviews the new RPC interface.

HCastano:
This is ready for a review. As is, this PR is a bunch of code that's never called, and it'll stay that way until it's integrated with the codebase as part of a PR targeting #72.

We should also have some tests for the RPC, which I think can either be done alongside #72 or as a separate PR. I'll probably need help from Tomek to get those set up properly.

On that note, I haven't tried any of these out on a real node, so it could be the case that all these RPCs are broken in some way right now. However, I think the main goal of this PR is more about the interfaces involved, and the rest can be fixed during the integration/testing PRs.

relays/ethereum/src/rpc.rs
```diff
@@ -0,0 +1,338 @@
...
+/// Proof of bool serialization success.
+const BOOL_SERIALIZATION_PROOF: &'static str = "bool serialization never fails; qed";
+
+type Result<T> = result::Result<T, RpcError>;
+type GrandpaAuthorityList = Vec<u8>;
+
+jsonrpsee::rpc_api! {
+	Ethereum {
+		#[rpc(method = "eth_estimateGas")]
+		fn estimate_gas(call_request: Params) -> U256;
```
tomusdrw, line 56: Isn't it possible to have strictly typed arguments instead of `Params`?

relays/ethereum/src/rpc.rs
```diff
@@ -0,0 +1,338 @@
...
+	async fn eth_call(&mut self, call_transaction: CallRequest) -> Result<Bytes> {
+		let params = Params::Array(vec![serde_json::to_value(call_transaction)?]);
+		Ok(Ethereum::call(&mut self.client, params).await?)
+	}
+}
+
+/// The API for the supported Substrate RPC methods.
+#[async_trait]
+pub trait SubstrateRpc {
+	async fn best_header(&mut self) -> Result<SubstrateHeader>;
```
tomusdrw, line 202: missing docs?

relays/ethereum/src/rpc.rs
```diff
@@ -0,0 +1,338 @@
...
+		let transport = HttpTransportClient::new(&uri);
+		let client = RawClient::new(transport);
+
+		Self { client }
+	}
+}
+
+#[async_trait]
+impl EthereumRpc for EthereumRpcClient {
+	/// Estimate gas usage for the given call.
```
tomusdrw, line 118: the docs should be rather on trait definition, not here, shouldn't they?

relays/ethereum/src/rpc.rs
```diff
@@ -0,0 +1,338 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! RPC Module
+
+#![allow(dead_code)]
+#![allow(unused_variables)]
```
tomusdrw, line 20: why?

relays/ethereum/src/rpc.rs
```diff
@@ -0,0 +1,338 @@
...
+		Ok(Ethereum::block_number(&mut self.client).await?.as_u64())
+	}
+
+	/// Retrieve block header by its number from Ethereum node.
+	async fn header_by_number(&mut self, block_number: u64) -> Result<EthereumHeader> {
+		// Only want to get hashes back from the RPC
+		let return_full_tx_obj = false;
+
+		let params = Params::Array(vec![
+			serde_json::to_value(U64::from(block_number)).expect(INT_SERIALIZATION_PROOF),
```
tomusdrw, line 135: Why not simply use `?` here instead of `expect`? I feel the code would simpler and I see no advantages of using `expect`.

relays/ethereum/src/rpc.rs
```diff
@@ -0,0 +1,338 @@
...
+		use sp_core::crypto::Ss58Codec;
+		let account = serde_json::to_value(account.to_ss58check())?;
+		let params = Params::Array(vec![account]);
+
+		Ok(Substrate::system_account_next_index(&mut self.client, params).await?)
+	}
+
+	/// Returns best Ethereum block that Substrate runtime knows of.
+	async fn best_ethereum_block(&mut self) -> Result<EthereumHeaderId> {
+		let call = EthereumHeadersApiCalls::BestBlock.to_string();
```
tomusdrw, line 270: I feel like this is a bit higher level call and my initial idea was to only extract low-level methods (like `state_call`) into a separate module and keep the ones that are more bridge-specific in a separate place. I don't feel strong about it though, but we should pay attention to avoid putting to much bridge-specific logic in this module.

primitives/ethereum-poa/src/lib.rs
```diff
@@ -344,6 +344,40 @@ pub fn public_to_address(public: &[u8; 64]) -> Address {
...
+	fn to_string(&self) -> String {
+		match self {
+			RuntimeApiCalls::GrandpaAuthorities => "GrandpaApi_grandpa_authorities",
+		}
+		.into()
+	}
+}
+
+// I want to be able to generate something like this directly from the decl_runtime_apis! macro
+pub enum EthereumHeadersApiCalls {
```
tomusdrw, line 363: I think this should stay private as it's only used internally by SubstrateClient, no?

We could simplify this to a bunch of consts instead if it's not part of the public API. Seems that currently we never really utilize the strong typing of that enum.

relays/ethereum/src/rpc.rs
```diff
@@ -0,0 +1,338 @@
...
+
+		let best_header_id = HeaderId(decoded_response.0, decoded_response.1);
+		Ok(best_header_id)
+	}
+
+	/// Returns whether or not transactions receipts are required for Ethereum header submission.
+	// Should I work with a QueuedEthereumHeader or a SubstrateEthereumHeader since that's what'll
+	// actually get encoded and sent to the RPC?
+	//
+	// I'm leaning towards the SubstrateEthereumHeader since that's a bit "lower level"
```
tomusdrw, line 285: Sounds good!

relays/ethereum/src/rpc.rs
```diff
@@ -0,0 +1,338 @@
...
+		let params = Params::Array(vec![serde_json::Value::String(call), serde_json::to_value(data)?]);
+
+		let encoded_response = Substrate::state_call(&mut self.client, params).await?;
+		let receipts_required: bool = Decode::decode(&mut &encoded_response.0[..])?;
+
+		// Gonna make it the responsibility of the caller to return (receipts_required, id)
+		Ok(receipts_required)
+	}
+
+	/// Returns whether or not the given Ethereum header is known to the Substrate runtime.
```
tomusdrw, line 298:
```suggestion
	/// Returns whether or not the given Ethereum header is known to the Substrate runtime.
	///
```

relays/ethereum/src/rpc.rs
```diff
@@ -0,0 +1,338 @@
...
+		let is_known_block: bool = Decode::decode(&mut &encoded_response.0[..])?;
+
+		// Gonna make it the responsibility of the caller to return (is_known_block, id)
+		Ok(is_known_block)
+	}
+
+	/// Submit an extrinsic for inclusion in a block.
+	// Q: Should I move the UncheckedExtrinsic type elsewhere so I don't have to pull it in from
+	// the runtime?
+	async fn submit_extrinsic(&mut self, transaction: UncheckedExtrinsic) -> Result<SubstrateHash> {
```
tomusdrw, line 320: IMHO you can change the input here to `Bytes`. I think it's more important to keep the RPC access layer runtime agnostic than to make it super strictly typed.

relays/ethereum/src/rpc.rs
HCastano: Since the no-one is using the `SubstrateRpcClient` and `EthereumRpcClient` structs I get a ton of warnings about unused/dead code. This would be removed once integration happens as part of #72.

primitives/ethereum-poa/src/lib.rs
HCastano: That's a good point. I put it here in order to keep its proximity close to the `decl_runtime_apis!` call where these enums come from.

relays/ethereum/src/rpc.rs
HCastano: Yeah you're right, I'll update this

relays/ethereum/src/rpc.rs
HCastano: I thought it would be nice for the module to encapsulate the parameter encoding/decoding aspect of it so the caller wouldn't have to worry about that themselves. These mostly what I consider "base" calls in that they're a wrapper around a single RPC method. There will be other high level calls, like ones which interact with contracts, which will be elsewhere (like in `ethereum_client.rs`). I'm open to moving these elsewhere, but I think that can be done later unless you have a strong opinion on why it should be done now.

relays/ethereum/src/rpc_errors.rs
```diff
@@ -0,0 +1,84 @@
...
+
+// Parity Bridges Common is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+#![allow(dead_code)]
```
HCastano, line 17: This will also be removed with #72.

relays/ethereum/src/rpc.rs
```diff
@@ -0,0 +1,307 @@
...
+	async fn best_block_number(&mut self) -> Result<u64> {
+		Ok(Ethereum::block_number(&mut self.client).await?.as_u64())
+	}
+
+	async fn header_by_number(&mut self, block_number: u64) -> Result<EthereumHeader> {
+		// Only want to get hashes back from the RPC
+		let return_full_tx_obj = false;
+
+		let header = Ethereum::get_block_by_number(&mut self.client, block_number).await?;
+		match header.number.is_some() && header.hash.is_some() {
```
svyatonik, line 140:
```suggestion
		match header.number.is_some() && header.hash.is_some() && header.logs_bloom.is_some() {
```

relays/ethereum/src/rpc.rs
```diff
@@ -0,0 +1,307 @@
...
+			false => Err(RpcError::Ethereum(EthereumNodeError::IncompleteHeader)),
+		}
+	}
+
+	async fn header_by_hash(&mut self, hash: H256) -> Result<EthereumHeader> {
+		// Only want to get hashes back from the RPC
+		let return_full_tx_obj = false;
+
+		let header = Ethereum::get_block_by_hash(&mut self.client, hash).await?;
+		match header.number.is_some() && header.hash.is_some() {
```
svyatonik, line 151:
```suggestion
		match header.number.is_some() && header.hash.is_some() && header.logs_bloom.is_some() {
```

relays/ethereum/src/rpc.rs
```diff
@@ -0,0 +1,307 @@
...
+		Ok(Ethereum::estimate_gas(&mut self.client, call_request).await?)
+	}
+
+	async fn best_block_number(&mut self) -> Result<u64> {
+		Ok(Ethereum::block_number(&mut self.client).await?.as_u64())
+	}
+
+	async fn header_by_number(&mut self, block_number: u64) -> Result<EthereumHeader> {
+		// Only want to get hashes back from the RPC
+		let return_full_tx_obj = false;
```
svyatonik, line 137: Remove?

relays/ethereum/src/rpc.rs
```diff
@@ -0,0 +1,307 @@
...
+		let header = Ethereum::get_block_by_number(&mut self.client, block_number).await?;
+		match header.number.is_some() && header.hash.is_some() {
+			true => Ok(header),
+			false => Err(RpcError::Ethereum(EthereumNodeError::IncompleteHeader)),
+		}
+	}
+
+	async fn header_by_hash(&mut self, hash: H256) -> Result<EthereumHeader> {
+		// Only want to get hashes back from the RPC
+		let return_full_tx_obj = false;
```
svyatonik, line 148: Remove?

relays/ethereum/src/rpc.rs
```diff
@@ -0,0 +1,301 @@
...
+	}
+
+	async fn eth_call(&mut self, call_transaction: CallRequest) -> Result<Bytes> {
+		Ok(Ethereum::call(&mut self.client, call_transaction).await?)
+	}
+}
+
+/// The API for the supported Substrate RPC methods.
+#[async_trait]
+pub trait SubstrateRpc {
```
tomusdrw, line 176: I think it wouldn't be unreasonable to extract this to a separate module that is somewhere outside of `relays/ethereum`. I imagine that the same substrate RPCs will be used for Substrate<>Substrate bridge for instance.
I don't consider this in scope of this PR though, just stating the fact :)

relays/ethereum/src/rpc_errors.rs
```diff
@@ -0,0 +1,84 @@
...
+
+use jsonrpsee::raw::client::RawClientError;
+use jsonrpsee::transport::http::RequestError;
+use serde_json;
+
+type RpcHttpError = RawClientError<RequestError>;
+
+#[derive(Debug)]
+pub enum RpcError {
+	Serialization(serde_json::Error),
```
tomusdrw, line 27: Missing docs.

relays/ethereum/src/rpc_errors.rs
```diff
@@ -0,0 +1,84 @@
...
+	/// We have received header with missing number and hash fields.
+	IncompleteHeader,
+	/// We have received receipt with missing gas_used field.
+	IncompleteReceipt,
+	/// Invalid Substrate block number received from Ethereum node.
+	InvalidSubstrateBlockNumber,
+}
+
+#[derive(Debug)]
+pub enum SubstrateNodeError {
```
tomusdrw, line 77: Missing docs for both enums.

relays/ethereum/src/rpc.rs
HCastano: Agreed

PR #91, opened 2020-05-15
Claim funds on Substrate chain by providing proof of funds locking on PoA chain

svyatonik:
So there's new module - `modules/currency-exchange`, which has one entry point:
```Rust
		pub fn import_peer_transaction(
			origin,
			transaction: <<T as Trait>::PeerBlockchain as Blockchain>::Transaction,
			block: <<T as Trait>::PeerBlockchain as Blockchain>::BlockHash,
			proof: <<T as Trait>::PeerBlockchain as Blockchain>::TransactionInclusionProof,
		) -> DispatchResult;
```
It checks whether provided tx is known to the 'bridge' module and its block is finalized, then it 'parses' that tx (for PoA -> Substrate transactions that means decoding `sender`, `to`, `nonce`, `value` and `data` fields + verifies that it is the tx sent to pre-configured address) + maps from this transaction to local Substrate chain receiver (in current impl, the tx mush have encoded Substrate' account id in its `data` field, as Tomek has suggested) + maps between currencies (currently 1:1) + grants given amount to given receiver.


primitives/ethereum-poa/src/lib.rs
```diff
@@ -344,6 +414,27 @@ pub fn public_to_address(public: &[u8; 64]) -> Address {
 	result
 }
 
+/// Verify ethereum merkle proof.
+fn verify_merkle_proof<T: AsRef<[u8]>>(expected_root: H256, items: impl Iterator<Item = T>) -> bool {
```
svyatonik, line 418: For now proof = whole set of transactions. That's also true for tx receipts. We may try to migrate to minimal proof in future.

svyatonik:
FWIW I've been able to claim funds by providing proof of https://kovan.etherscan.io/tx/0x7669b7a77c103ed330b3416843fd05dd0c294f61e3b3de27c22e21b33fc9f668 + manually editing some code parts that are failing (or require too much time to test) for this tx - i.e. I've changed recipient id to `//Bob` + removed tx block' finality requirements). This should be easier when we'll have a separate chain.

HCastano:
Is this ready for review @svyatonik?

Also, there's a _ton_ of commits here that aren't from this PR, can you merge master?

svyatonik:
Yes, it's ready. There's ton of commits because we're squashing when merging to master and I've started this wok before earlier PRs have been merged. It has no conflicts with master => you could just look at the changes.

modules/currency-exchange/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+	>;
+	/// This blockchain currency amount type.
+	type Amount;
+	/// Converter from peer blockchain currency type into current blockchain currency type.
+	type CurrencyConverter: CurrencyConverter<
+		SourceAmount = <Self::PeerMaybeLockFundsTransaction as MaybeLockFundsTransaction>::Amount,
+		TargetAmount = Self::Amount,
+	>;
+	/// Something that could grant money.
+	type Airdrop: Airdrop<Recipient = Self::AccountId, Amount = Self::Amount>;
```
tomusdrw, line 65: We should just require `Currency` trait for both this and `Amount`. I think it has a nice (and commonly used within substrate) API to mint new tokens and is implemented for `balances` module already.

bin/node/runtime/src/lib.rs
```diff
@@ -208,6 +211,45 @@ impl pallet_bridge_eth_poa::Trait for Runtime {
...
+}
+
+pub struct Airdrop;
+
+impl sp_bridge_eth_poa::exchange::Airdrop for Airdrop {
+	type Recipient = AccountId;
+	type Amount = Balance;
+
+	fn drop(recipient: Self::Recipient, amount: Self::Amount) -> sp_bridge_eth_poa::exchange::Result<()> {
+		<pallet_balances::Module<Runtime> as Currency<AccountId>>::deposit_into_existing(&recipient, amount)
```
tomusdrw, line 230:
```suggestion
		<pallet_balances::Module<Runtime> as Currency<AccountId>>::deposit_creating(&recipient, amount)
```
IMHO we should not fail if the account does not exist. Sending bridged funds might be the first encounter with substrate blockchain, we should not require any prior interaction.

tomusdrw: On the second thought I think we should rather use `issue` + `resolve_creating` to make sure that the change in total issuance is reflected.

tomusdrw: On the third thought, if we plan to introduce a two-way bridge into existing networks that have a value already (like Kusama) we should not change (mint/bunr) the total issuance but rather have some fixed pool that we can use to transfer the funds from/to. On the other chain locked tokens could be redeemed by the ones that top-up that pool.

Overall it seems that this will rather need to be configurable, so my other comment about using `Currency` directly (instead of `Airdrop`) should be disregarded.

modules/currency-exchange/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+use sp_std::vec::Vec;
+
+/// Peer blockhain interface.
+pub trait Blockchain {
+	/// Block hash type.
+	type BlockHash: Parameter;
+	/// Transaction type.
+	type Transaction: Parameter;
+	/// Transaction inclusion proof type.
+	type TransactionInclusionProof: Parameter;
```
tomusdrw, line 34: I'd convert the `Blockchain` trait into something more generic, like so:
```
struct RawTransaction(Vec<u8>);
trait TransactionInclusionProof {
  type BlockHash;
  type Header;

  fn transaction(&self) -> &RawTransaction;

  fn block_hash(&self) -> &Self::BlockHash;

  fn verify(&self, header: Header) -> bool;
}
```
and then implement it for a tuple (or a struct like this):
```
(H256 /* hash */, usize /* index */, Vec<RawTransaction> /* list of transactions */)
```

This does not require sending the transaction body twice in the payload and in the future will allow us to use different (more minimal) proof formats without changing much of the code (we can even support both formats if needed).
The verification logic stays close to the proof data as well, which is something that I like too. We can keep the transaction parsing separate though. 

Not super strong on this though, since I feel that it might not be generic enough and might actually complicate some other stuff (namely UX) - so open to discussion.

modules/currency-exchange/src/lib.rs
tomusdrw: Nevermind this - probably the way we create tokens will differ depending on the chain. I'd suggest naming `Airdrop` something more generic like: `type TransferTokens` so that it doesn't imply the tokens are "created" out of thin air, but rather they might be transferred from some common pool.

modules/currency-exchange/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+
+decl_module! {
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+		/// Imports lock fund transaction of the peer blockchain.
+		#[weight = 0] // TODO: update me (https://github.com/paritytech/parity-bridges-common/issues/78)
+		pub fn import_peer_transaction(
+			origin,
+			transaction: <<T as Trait>::PeerBlockchain as Blockchain>::Transaction,
+			block: <<T as Trait>::PeerBlockchain as Blockchain>::BlockHash,
+			proof: <<T as Trait>::PeerBlockchain as Blockchain>::TransactionInclusionProof,
```
tomusdrw, line 97: Imho this should have just one parameter of type `TransactionInclusionProof`, which in turn contains all these required details in a more compact form (namely no duplication in transaction data).

modules/currency-exchange/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+			//
+			// leaving it here for now just for simplicity
+			let is_transaction_finalized = <T as Trait>::PeerBlockchain::verify_transaction_inclusion_proof(
+				&transaction,
+				block,
+				&proof,
+			);
+			if !is_transaction_finalized {
+				fail!(Error::<T>::UnfinalizedTransaction);
+			}
```
tomusdrw, line 115:
```suggestion
			ensure!(is_transaction_finalized, Error::UnfinalizedTransaction);
```

modules/currency-exchange/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+		/// Imports lock fund transaction of the peer blockchain.
+		#[weight = 0] // TODO: update me (https://github.com/paritytech/parity-bridges-common/issues/78)
+		pub fn import_peer_transaction(
+			origin,
+			transaction: <<T as Trait>::PeerBlockchain as Blockchain>::Transaction,
+			block: <<T as Trait>::PeerBlockchain as Blockchain>::BlockHash,
+			proof: <<T as Trait>::PeerBlockchain as Blockchain>::TransactionInclusionProof,
+		) -> DispatchResult {
+			frame_system::ensure_signed(origin)?;
```
tomusdrw, line 99: Long-term we should reward the ones how submit such transactions. Can we have something super simplistic for now (like a `OnTransactionImport(who)` hook in the trait definition) so that it can be plugged in when needed without changing the code?

modules/currency-exchange/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+
+			Ok(())
+		}
+	}
+}
+
+decl_storage! {
+	trait Store for Module<T: Trait> as Bridge {
+		/// All transfers that have already been claimed.
+		Transfers: map hasher(blake2_128_concat) Vec<u8> => ();
```
tomusdrw, line 142: Why `Vec<u8>` and not the `EthTransactionTag` stronger type? The result would be the same anyway, wouldn't it? But at least there would be some reference to the key format.

modules/currency-exchange/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+		/// Invalid peer blockchain transaction provided.
+		InvalidTransaction,
+		/// Peer transaction has invalid amount.
+		InvalidAmount,
+		/// Peer transaction has invalid recipient.
+		InvalidRecipient,
+		/// Cannot map from peer recipient to this blockchain recipient.
+		FailedToMapRecipients,
+		/// Failed to convert from peer blockchain currency to this blockhain currency.
+		FailedToCovertCurrency,
```
tomusdrw, line 79:
```suggestion
		FailedToConvertCurrency,
```

primitives/ethereum-poa/src/lib.rs
```diff
@@ -336,6 +347,65 @@ impl std::fmt::Debug for Bloom {
...
+	message.append(&nonce);
+	message.append_raw(gas_price.as_raw(), 1);
+	message.append_raw(gas.as_raw(), 1);
+	message.append_raw(action.as_raw(), 1);
+	message.append(&value);
+	message.append(&payload);
+	if let Some(chain_id) = chain_id {
+		message.append(&chain_id);
+		message.append(&0u8);
+		message.append(&0u8);
```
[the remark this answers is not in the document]
tomusdrw: I thought that the issue is serious, but it's not. The other blockchain should verify that - we should probably only care about inclusion proof.
Also the whole signature verification is unnecessary, but since there is no other way to recover the sender we must do it :man_shrugging:

primitives/ethereum-poa/src/exchange.rs
```diff
@@ -0,0 +1,125 @@
...
+
+/// Currency airdrop.
+pub trait Airdrop {
+	/// Recipient type.
+	type Recipient;
+	/// Currency amount type.
+	type Amount;
+
+	/// Grant some money to given account.
+	fn drop(recipient: Self::Recipient, amount: Self::Amount) -> Result<()>;
```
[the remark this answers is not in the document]
tomusdrw: I think we should rather call it `transfer_to` or `deposit_into` to be more generic regarding where the funds are coming from.

bin/node/runtime/src/exchange.rs
```diff
@@ -0,0 +1,200 @@
...
+use pallet_bridge_currency_exchange::Blockchain;
+use sp_bridge_eth_poa::{
+	exchange::{Error as ExchangeError, LockFundsTransaction, MaybeLockFundsTransaction, Result as ExchangeResult},
+	transaction_decode,
+};
+use sp_std::vec::Vec;
+
+/// Address where locked PoA funds must be sent to (0xDEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEF).
+const LOCK_FUNDS_ADDRESS: [u8; 20] = [
+	0xDE, 0xAD, 0xBE, 0xEF, 0xDE, 0xAD, 0xBE, 0xEF, 0xDE, 0xAD, 0xBE, 0xEF, 0xDE, 0xAD, 0xBE, 0xEF, 0xDE, 0xAD, 0xDE,
```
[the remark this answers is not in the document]
svyatonik: It's `no-std` and `hex-literal` has no `no_std` support, afaiu. Do you want to decode every time we need this, or am I missing something (maybe there's some other crate I'm not aware of)?

primitives/ethereum-poa/src/lib.rs
```diff
@@ -83,6 +88,21 @@ pub struct Header {
 	pub seal: Vec<Bytes>,
 }
 
+/// Parsed ethereum transaction.
+#[derive(Debug, PartialEq)]
+pub struct Transaction {
```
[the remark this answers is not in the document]
svyatonik: There's a lot of such crates, but most lack `no_std` support and others lack something we need :/ We can 'fix' some crates first, or make functional bridge, gather list of all changes we need to make to these crates, made these changes and update bridge to use these crates. I'd vote for bridge-first, everything other later. We can choose the opposite option, if you think it'd be better.

primitives/ethereum-poa/src/exchange.rs
```diff
@@ -0,0 +1,125 @@
...
+	type Recipient = AccountId;
+
+	fn map(peer_recipient: Self::PeerRecipient) -> Result<Self::Recipient> {
+		Ok(peer_recipient)
+	}
+}
+
+/// Currency converter which is used when currency is the same on both chains.
+#[derive(Debug)]
+pub struct AsIsCurrencyConverter<Amount>(PhantomData<Amount>);
```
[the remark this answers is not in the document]
svyatonik: I've left different traits - I wanted to avoid using and confusing with `sp_runtime::Identity*` in this case.

primitives/ethereum-poa/src/exchange.rs
```diff
@@ -0,0 +1,125 @@
...
+	/// Peer currency amount type.
+	type Amount;
+
+	/// Parse lock funds transaction of the peer blockchain. Returns None if
+	/// transaction format is unknown, or it isn't a lock funds transaction.
+	fn parse(tx: &Self::Transaction) -> Result<LockFundsTransaction<Self::Id, Self::Recipient, Self::Amount>>;
+}
+
+/// Map that maps recipients from peer blockchain to this blockchain recipients.
+pub trait RecipientsMap {
```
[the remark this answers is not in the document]
svyatonik: With `TryInto` you're unable to introduce logic for converting between same types - e.g. when accounts on both chains are identified by the same type `Public`, but for mapping you actually need to, e.g. multiply this `Public` and some const to get another `Public`. You'd need to introduce some wrappers, right? If somewhere else (not in exchange) , you'd need e.g. another multiplication const for the same types (e.g. when account on chain1 has two accounts on chain2), you'd need to introduce another wrapper. With separate trait, all logic is isolated in single reusable implementation. So yes, I think it is better to have separate trait here

primitives/ethereum-poa/src/exchange.rs
```diff
@@ -0,0 +1,125 @@
...
+	type PeerRecipient;
+	/// Current blockchain recipient type.
+	type Recipient;
+
+	/// Lookup current blockchain recipient by peer blockchain recipient.
+	fn map(peer_recipient: Self::PeerRecipient) -> Result<Self::Recipient>;
+}
+
+/// Conversion between two currencies.
+pub trait CurrencyConverter {
```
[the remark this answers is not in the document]
svyatonik: Answered above

modules/currency-exchange/src/lib.rs
svyatonik: Renamed to `trait DepositInto`

modules/ethereum/src/lib.rs
```diff
@@ -623,6 +642,30 @@ impl<T: Trait> Storage for BridgeStorage<T> {
...
 }
 
+/// Verify that transaction is included into given finalized block.
+pub fn verify_transaction_finalized<S: Storage>(
+	storage: &S,
+	tx: &RawTransaction,
+	block: H256,
+	proof: &Vec<RawTransaction>,
+) -> bool {
+	if !proof.contains(tx) {
```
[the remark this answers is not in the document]
svyatonik: Yeah - transaction itself is redundant when trie-proof is provided. I was thinking of cases when the proof itself can't provide any info about transaction, except for its finality. That's why tx is currently a separate parameter. So probably I'll try to change to something like `index + Vec<Transaction>` as you have suggested.

modules/currency-exchange/src/lib.rs
svyatonik: `trait OnTransactionSubmitted`

bin/node/runtime/src/exchange.rs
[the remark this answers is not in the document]
svyatonik: Ok, stupid me - hex-literal is a proc-macro, so I guess this could be done

bin/node/runtime/src/exchange.rs
```diff
@@ -0,0 +1,216 @@
...
+// (at your option) any later version.
+
+// Parity Bridges Common is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
```
HCastano, line 16: Can you add some docs here?

bin/node/runtime/src/exchange.rs
```diff
@@ -0,0 +1,216 @@
...
+use frame_support::RuntimeDebug;
+use hex_literal::hex;
+use pallet_bridge_currency_exchange::Blockchain;
+use sp_bridge_eth_poa::{
+	exchange::{Error as ExchangeError, LockFundsTransaction, MaybeLockFundsTransaction, Result as ExchangeResult},
+	transaction_decode,
+};
+use sp_std::vec::Vec;
+
+/// Address where locked PoA funds must be sent to.
```
HCastano, line 27:
```suggestion
/// Ethereum address where locked PoA funds must be sent to.
```

bin/node/runtime/src/exchange.rs
```diff
@@ -0,0 +1,216 @@
...
+/// Eth blockchain from runtime perspective.
+pub struct EthBlockchain;
+
+impl Blockchain for EthBlockchain {
+	type Transaction = Vec<u8>;
+	type TransactionInclusionProof = EthereumTransactionInclusionProof;
+
+	fn verify_transaction_inclusion_proof(proof: &Self::TransactionInclusionProof) -> Option<Self::Transaction> {
+		let is_transaction_finalized =
+			crate::BridgeEthPoA::verify_transaction_finalized(proof.block, proof.index, &proof.proof);
```
HCastano, line 64: Can we not just hand this function `proof` and have it destructure it?

modules/currency-exchange/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+// (at your option) any later version.
+
+// Parity Bridges Common is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
```
HCastano, line 16: Can we get some docs here as well?

modules/currency-exchange/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+		type OnTransactionSubmitted = DummyTransactionSubmissionHandler;
+		type PeerBlockchain = DummyBlockchain;
+		type PeerMaybeLockFundsTransaction = DummyTransaction;
+		type RecipientsMap = DummyRecipientsMap;
+		type Amount = u64;
+		type CurrencyConverter = DummyCurrencyConverter;
+		type DepositInto = DummyDepositInto;
+	}
+
+	type Exhange = Module<TestRuntime>;
```
HCastano, line 314:
```suggestion
	type Exchange = Module<TestRuntime>;
```

modules/currency-exchange/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+
+			// reward submitter for providing valid message
+			T::OnTransactionSubmitted::on_valid_transaction_submitted(submitter);
+
+			Ok(())
+		}
+	}
+}
+
+decl_storage! {
```
HCastano, line 137: nit: A lot of the frame pallets have this as the first macro they use, instead of `decl_error!`, we should try and match that informal convention

modules/currency-exchange/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+		/// Imports lock fund transaction of the peer blockchain.
+		#[weight = 0] // TODO: update me (https://github.com/paritytech/parity-bridges-common/issues/78)
+		pub fn import_peer_transaction(
+			origin,
+			proof: <<T as Trait>::PeerBlockchain as Blockchain>::TransactionInclusionProof,
+		) -> DispatchResult {
+			let submitter = frame_system::ensure_signed(origin)?;
+
+			// ensure that transaction is included in finalized block that we know of
```
HCastano, line 101: Since this module can be used for checking Ethereum transactions, it's not exactly correct to say finalized blocks here

modules/currency-exchange/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+			).ok_or_else(|| Error::<T>::UnfinalizedTransaction)?;
+
+			// parse transaction
+			let transaction = <T as Trait>::PeerMaybeLockFundsTransaction::parse(&transaction)
+				.map_err(Error::<T>::from)?;
+			let transfer_id = transaction.id;
+			ensure!(
+				!Transfers::<T>::contains_key(&transfer_id),
+				Error::<T>::AlreadyClaimed
+			);
```
HCastano, line 119: The exclamation mark is easy to miss, since we need to be extra careful in the runtime I suggest the following

```suggestion
                      let has_not_been_claimed = !Transfers::<T>::contains_key(&transfer_id);
			ensure!(
				has_not_been_claimed,
				Error::<T>::AlreadyClaimed
			);
```

modules/currency-exchange/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+		});
+	}
+
+	#[test]
+	fn valid_transaction_accepted() {
+		new_test_ext().execute_with(|| {
+			assert_ok!(Exhange::import_peer_transaction(
+				Origin::signed(SUBMITTER),
+				(true, transaction(0)),
+			),);
```
HCastano, line 410: Does this comma need to be here?

modules/currency-exchange/Cargo.toml
```diff
@@ -0,0 +1,58 @@
+[package]
+name = "pallet-bridge-currency-exchange"
+description = "A Substrate Runtime module that accepts 'lock funds' transactions from peer chain and grants equivalent amount to given Substrate account."
```
HCastano, line 3:
```suggestion
description = "A Substrate Runtime module that accepts 'lock funds' transactions from a peer chain and grants an equivalent amount to a the appropriate Substrate account."
```

bin/node/runtime/src/exchange.rs
```diff
@@ -0,0 +1,216 @@
...
+	pub account: [u8; 20],
+	/// Lock transaction nonce.
+	pub nonce: sp_core::U256,
+}
+
+/// Eth blockchain from runtime perspective.
+pub struct EthBlockchain;
+
+impl Blockchain for EthBlockchain {
+	type Transaction = Vec<u8>;
```
HCastano, line 59: We should typedef `Vec<u8>`

bin/node/runtime/src/exchange.rs
```diff
@@ -0,0 +1,216 @@
...
+		let ferdie_raw: &[u8; 32] = ferdie_id.as_ref();
+		let mut eth_tx = ethereum_tx_sign::RawTransaction {
+			nonce: 0.into(),
+			to: Some(LOCK_FUNDS_ADDRESS.into()),
+			value: 100.into(),
+			gas: 100_000.into(),
+			gas_price: 100_000.into(),
+			data: ferdie_raw.to_vec(),
+		};
+		editor(&mut eth_tx);
```
HCastano, line 160: It's not clear from the function signature what `editor` does, can you elaborate it for me?

bin/node/runtime/src/exchange.rs
```diff
@@ -0,0 +1,216 @@
...
+	#[test]
+	fn invalid_transaction_rejected() {
+		assert_eq!(
+			EthTransaction::parse(&Vec::new()),
+			Err(ExchangeError::InvalidTransaction),
+		);
+	}
+
+	#[test]
+	fn invalid_with_invalid_peer_recipient_rejected() {
```
HCastano, line 188: I'm confused at the naming of the following three tests. Specifically with the `invalid_with_invalid_...` part

primitives/ethereum-poa/src/lib.rs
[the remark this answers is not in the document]
HCastano: If it doesn't take a long time to make these crates `no_std` compatible I think we should try and do that from the get-go. Otherwise, I'd be okay updating the bridge at a later time.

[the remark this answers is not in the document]
tomusdrw: Im fine with ditching other libraries and having the code duplicated in the repo. Can we make it into it's own crate though? It's going to be useful not only for the PoA use case.

primitives/ethereum-poa/src/exchange.rs
```diff
@@ -0,0 +1,125 @@
...
+	/// Grant some money to given account.
+	fn deposit_into(recipient: Self::Recipient, amount: Self::Amount) -> Result<()>;
+}
+
+/// Recipients map which is used when accounts ids are the same on both chains.
+#[derive(Debug)]
+pub struct IdentityRecipients<AccountId>(PhantomData<AccountId>);
+
+impl<AccountId> RecipientsMap for IdentityRecipients<AccountId> {
+	type PeerRecipient = AccountId;
```
HCastano, line 106: I'm a bit confused with the naming here. Is a `PeerRecipient` the same as a `Sender` from another chain?

primitives/ethereum-poa/src/exchange.rs
```diff
@@ -0,0 +1,125 @@
...
+
+	/// Grant some money to given account.
+	fn deposit_into(recipient: Self::Recipient, amount: Self::Amount) -> Result<()>;
+}
+
+/// Recipients map which is used when accounts ids are the same on both chains.
+#[derive(Debug)]
+pub struct IdentityRecipients<AccountId>(PhantomData<AccountId>);
+
+impl<AccountId> RecipientsMap for IdentityRecipients<AccountId> {
```
HCastano, line 105: What would a different implementation of `RecipientsMap` look like? I'm still trying to understand the usage of `ReceipientsMap`

modules/currency-exchange/src/lib.rs
HCastano: Sorry, it slipped my mind that we're in PoA land, where blocks can indeed be finalized, my bad!

primitives/ethereum-poa/src/lib.rs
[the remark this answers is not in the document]
svyatonik: It is already in separate crate, but this PR adds exchange primitives to that crate => I've extracted it to separate crate. The name of eth/poa-related crate is still `sp-bridge-eth-poa`, though

primitives/ethereum-poa/src/exchange.rs
svyatonik: It is decided by the combination of `MaybeLockFundsTransaction` and `RecipientsMap`  implementations.

primitives/ethereum-poa/src/exchange.rs
svyatonik: Get associated recipient from storage map?

bin/node/runtime/src/exchange.rs
svyatonik: It can edit fields of passed tx

HCastano: That's not clear to me. I'd either change the name to be more representative, or add a comment explaining that

modules/currency-exchange/src/lib.rs
HCastano: I think a description similar to that in `bin/node/runtime/src/exchange.rs` would be helpful here

bin/node/runtime/src/exchange.rs
```diff
@@ -0,0 +1,229 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Support for PoA -> Substrate native tokens exchange.
+//!
+//! If you want to exchange native
+//! PoA tokens for Substrate chain native tokens, you need to:
```
HCastano, line 20:
```suggestion
//! If you want to exchange native
//! PoA tokens for Substrate chain native tokens, you need to:
```
```suggestion
//! If you want to exchange native PoA tokens for native Substrate
//! chain tokens, you need to:
```

bin/node/runtime/src/exchange.rs
```diff
@@ -0,0 +1,229 @@
...
+//! 1) send some PoA tokens to `LOCK_FUNDS_ADDRESS` address on PoA chain. Data field of
+//!    the transaction must be SCALE-encoded id of Substrate account that will receive
+//!    funds on Substrate chain;
+//! 2) wait until that 'lock funds' transaction is mined on PoA chain;
```
HCastano, line 24:
```suggestion
//! 2) wait until the 'lock funds' transaction is mined on PoA chain;
```

bin/node/runtime/src/exchange.rs
```diff
@@ -0,0 +1,229 @@
...
+//! 3) wait until block with 'lock funds' transaction is finalized on PoA chain;
```
HCastano, line 25:
```suggestion
//! 3) wait until the block containing the 'lock funds' transaction is finalized on PoA chain;
```

bin/node/runtime/src/exchange.rs
```diff
@@ -0,0 +1,229 @@
...
+//! 4) wait until (or provide) required PoA header and its finality will be provided
```
HCastano, line 26:
```suggestion
//! 4) wait until the required PoA header and its finality are provided
```

bin/node/runtime/src/exchange.rs
```diff
@@ -0,0 +1,229 @@
...
+//!    to the PoA -> Substrate bridge module;
```
HCastano, line 27:
```suggestion
//!    to the PoA -> Substrate bridge module (it can be provided by you);
```

modules/currency-exchange/src/lib.rs
```diff
@@ -0,0 +1,414 @@
...
+				Error::<T>::AlreadyClaimed
+			);
+
+			// grant recipient
+			let recipient = T::RecipientsMap::map(transaction.recipient).map_err(Error::<T>::from)?;
+			let amount = T::CurrencyConverter::convert(transaction.amount).map_err(Error::<T>::from)?;
+			T::DepositInto::deposit_into(recipient, amount).map_err(Error::<T>::from)?;
+
+			// remember that we have accepted this transfer
+			Transfers::<T>::insert(transfer_id, ());
```
HCastano, line 123: Should this come before the `deposit_into` step? I'm not sure how this would happen, but I'm imagining a scenario in which we deposit funds and then fail to reach here, which would lead to us being able to claim stuff multiple times

modules/currency-exchange/src/lib.rs
```diff
@@ -0,0 +1,414 @@
...
+
+			// reward submitter for providing valid message
+			T::OnTransactionSubmitted::on_valid_transaction_submitted(submitter);
```
HCastano, line 126: Where does this reward come from?

modules/ethereum/src/lib.rs
```diff
@@ -623,6 +642,30 @@ impl<T: Trait> Storage for BridgeStorage<T> {
 	}
 }
 
+/// Verify that transaction is included into given finalized block.
+pub fn verify_transaction_finalized<S: Storage>(
+	storage: &S,
+	block: H256,
```
HCastano, line 648:
```suggestion
	block_hash: H256,
```

modules/ethereum/src/lib.rs
```diff
@@ -623,6 +642,30 @@ impl<T: Trait> Storage for BridgeStorage<T> {
...
+		return false;
+	}
+
+	let header = match storage.header(&block) {
+		Some((header, _)) => header,
+		None => return false,
+	};
+	let (finalized_number, finalized_hash) = storage.finalized_block();
+
+	if header.number > finalized_number || (header.number == finalized_number && block != finalized_hash) {
```
HCastano, line 662: I think something like this makes it a little easier to understand.

```rust
let is_finalized = header.number == finalized_number && block == finalized_hash;
if header.number > finalized_number || !is_finalized {
    ...
}
```

Also, what if `header.number` is less than `finalized_number`? How do we verify that the block has been finalized?

modules/currency-exchange/src/lib.rs
svyatonik: That's a good point - I'm also not sure how this could happen. The only case I could think of, should be impossible in Substrate - i.e. if implementation of `RecipientsMap`, `CurrencyConverter` or `DepositInto` will be calling into some untrusted (contract?) code and this code will be able to call back into 'system code' (here: submit same proof again). But if contract is able to call system code, then it also will be able to change any storage value (balance, ...). So this change won't change anything.

So the proper fix imo, is not to allow untrusted code to call into system code at all. Afaik it's already how it works. However, if you still think it'd be better to protect against that, I could change code to something like:
```
Transfers::insert();
if let Err(error) = deposit_into() {
  Transfers::remove();
  return Err(error);
}
```

It is something against regular "updating state after all checks are completed" pattern, though.

modules/currency-exchange/src/lib.rs
svyatonik: It is up to runtime to decide. Treasury? Locked funds?

modules/ethereum/src/lib.rs
```diff
@@ -659,7 +659,21 @@ pub fn verify_transaction_finalized<S: Storage>(
...
 	let (finalized_number, finalized_hash) = storage.finalized_block();
 
-	if header.number > finalized_number || (header.number == finalized_number && block != finalized_hash) {
+	// if header is not yet finalized => return
+	if header.number > finalized_number {
+		return false;
+	}
+
+	// check if header is actually finalized
+	let is_finalized = match header.number < finalized_number {
```
HCastano, line 674: Clippy would be mad if I didn't say anything :P
 https://rust-lang.github.io/rust-clippy/master/index.html#match_bool

modules/currency-exchange/src/lib.rs
HCastano: I see that's set to `()` in the runtime right now as well, so that clears it up.

modules/currency-exchange/src/lib.rs
HCastano: I took a look at the code again and I still don't know how it might happen, so no need to change the code to that. It wouldn't hurt for @tomusdrw should take a look and see if there's anything we missed

tomusdrw: Yeah, I was thinking about this when reviewing earlier, but I think the code is correct. To make it more future proof I'd suggest to group the two statements that are related into one block, like so:
```rust
let recipient = T::RecipientsMap::map(transaction.recipient).map_err(Error::<T>::from)?;
let amount = T::CurrencyConverter::convert(transaction.amount).map_err(Error::<T>::from)?;

// make sure to update the mapping if we deposit successfully to avoid double spending,
// i.e. whenever `deposit_into` is successful we MUST update `Transfers`.
{
    T::DepositInto::deposit_into(recipient, amount).map_err(Error::<T>::from)?;
    Transfers::<T>::insert(transfer_id, ())
}
```

This will make it harder to introduce `?` statements in between `deposit_into` and `Transfers::insert`.

svyatonik:
One last problem that affects `bridge` and `exchange` modules combination is that now `bridge` may prune unfinalized headers. And since `exchange` only accepts proofs on finalized blocks, some claims may never be accepted (because headers may already be pruned and we don't have CHT). There are 3 possible solutions that came up:
1) give some basic guarantees about claims - i.e. we only guarantee to accept claims that happened within last `N` finalized blocks. Pros: easy to implement. Cons: submitters have additional responsibilities + pruning window increases.
2) allow claims on unfinalized blocks - i.e. if lock funds tx has been mined in PoA block `N` and current finalized PoA block is `N - 10_000`, then anyone can submit proof that tx has been included in block `N` (if headers is already accepted by the `bridge`). And then, when block `N` will be finalized by `bridge`, actually make the transfer. If some sibling of `N` is finalized, drop this claim. Pros: quite easy to implement + all required transactions may be submitted (almost) altogether. Cons: ???;
3) construct CHTs when pruning + make required headers a part of proof-of-tx-inclusion. Pros: ???. Cons: large(r) and complex proofs + additional data that needs to be stored in `bridge` module.

I've implemented (1) option [here](https://github.com/paritytech/parity-bridges-common/tree/do-not-prune-unfinalized-headers) by moving pruning strategy from `bridge` module to runtime level (that needs to be done anyway). So the solution is that we are not pruning ~24 hours of finalized blocks => the claimant have at least 24 hours to submit transaction. I think that is enough for initial launch, right?

What needs to be done in long term? I'm not sure. Option (2) seems almost ideal, but I remember that Tomek has mentioned CHTs in some other context. So if you have any thoughts on that - please share, so that we may file an issue about that.

tomusdrw:
> I think that is enough for initial launch, right?
Yup, I think so.

> allow claims on unfinalized blocks - i.e. if lock funds tx has been mined in PoA block N and current finalized PoA block is N - 10_000, then anyone can submit proof that tx has been included in block N.

I feel that claims on unfinalized blocks should be more expensive (in terms of weight cost), even exponentially expensive the more claims for unfinalized block there is. Otherwise we are risking claim spam on a block that potentially (or even most likely) is a fork block, but simply finality has stalled for a bit. Also this complicates cost of finalization quite a bit, cause now finalizing a header would need to perform the transfers. We could possibly do it lazily in `on_initialize` or require the users to enact transfers manually if the claim gets finalized, but that adds extra complexity and UX complications - the users now how to submit two transactions instead of one (could obviously be batched by some 3rd party though).
Given all these arguments, I feel that for convenient usage of option 2, using some 3rd party would be required anyway, and that doesn't really make that much of a difference compared to option 1 (like in theory, with 2-transaction approach, the 3rd party can't really censor the transaction completely, the enactment would just be delayed and would require an extra manual intervention).

I initially thought about option (3) as well, but without CHT - just storing `Map<Number, Hash>` forever and allow users to submit headers as part of the proof. Knowing the entire hash-chain also has other benefits - for instance we could easily do manual (but verifiable) rollback in case of bugs (simply submit enough pruned headers and reset last finalized block). CHT is probably another optimization on top of that, but I don't know how complicated it would be - maybe the implementation cost is not worth the gains (mid-term; cause long-term obviously minimizing the storage would be desirable).

So overall, let's start with Option 1, and have both Options 2 & 3 logged as possible "nice to have" improvements.

HCastano:
I assume that the discussions about fixing interaction between the `exchange` and `ethereum` pruning choices can be done in another PR, so I'm going to go ahead and merge this.

svyatonik:
Yeah, thanks. I'll open PR and submit enhancement-issue once finality-cache PR is in.

TODO for myself - I think option (1) should have been described and implemented slightly differently. Instead of counting ~24hrs (20000 blocks) from best finalized PoA block, known to Substrate, we should leave (unpruned) every finalized PoA header in Substrate for at least ~24hrs. Example of why it is required:
1) on PoA chain: authorities fail to finalize blocks `[10_000; 50_000]` - they need 1 additional signature to finalize all these blocks;
2) on PoA chain: fund locks tx is at the block `10_000`;
3) on PoA chain: last required authority signs block `50_001`;
4) on Substrate chain: block `50_001` is imported and all headers in `[10_000; ~50_000]` range are marked for pruning (and in current impl header `10_000` is instantly pruned).

So the claimant has 0 possibility to present tx proof. And guarantees are not working.

tomusdrw:
> The biggest problem is that we don't want to leave headers for 24hrs when we're still syncing. So probably rely on PoA timestamp (TBD)?

Let's introduce an "admin" method to enabled/disable claims - that will allow us to prevent even a situation like this.
When the bridge matures we can simply set it to an empty account via governance decision.

PR #97, opened 2020-05-18
Add missing error handling

svyatonik:


tomusdrw:
Is there a way to get a unit or e2e test to prevent something like this in the future?

svyatonik:
It should be doable after #80 , but now there's no way

PR #99, opened 2020-05-18
Remove unused files

svyatonik:
(seems like artifacts of merging several repos into single one?)


HCastano:
Haha, yeah that was my bad 😅

PR #100, opened 2020-05-18
Bump rust-web3 to latest v0.11

HCastano:


relays/ethereum/src/substrate_types.rs
```diff
@@ -88,7 +88,7 @@ pub fn into_substrate_ethereum_header(header: &EthereumHeader) -> SubstrateEther
 		extra_data: header.extra_data.0.clone(),
 		state_root: header.state_root,
 		receipts_root: header.receipts_root,
-		log_bloom: header.logs_bloom.data().into(),
+		log_bloom: header.logs_bloom.unwrap_or_default().data().into(),
```
svyatonik, line 91: Probably worth checking this when header is received (`header_by_number` and `header_by_hash` in `ethereum_client.rs`). There's similar check for hash and number. This isn't critical, though

relays/ethereum/Cargo.toml
```diff
@@ -28,7 +28,7 @@ serde = { version = "1.0.110", features = ["derive"] }
 serde_json = "1.0.53"
 sp-bridge-eth-poa = { path = "../../primitives/ethereum-poa" }
 time = "0.2"
-web3 = { git = "https://github.com/svyatonik/rust-web3.git", branch = "add-receipt-root" }
+web3 = { git = "https://github.com/tomusdrw/rust-web3" }
```
tomusdrw, line 31: Just released `0.11` on crates.io, would you mind updating to this version?

relays/ethereum/src/substrate_types.rs
HCastano: Are you talking about adding a check [here](https://github.com/paritytech/parity-bridges-common/blob/master/relays/ethereum/src/ethereum_client.rs#L156)?

If so, I'm not sure I understand. Not having a `log_bloom` shouldn't be reason to mark a header an incomplete, right?

svyatonik: I guess 'incomplete' word now has too many meanings in relay :/ In this check, incomplete means that Ethereum node haven't returned all the fields we use. Since we use lob bloom (we are passing it to the runtime module), then yes - it should fail with `Error::IncompleteHeader` as well.

That differs from 'incomplete' that has been introduced recently, which is about some state of the header at the receiving side.

PR #101, opened 2020-05-18
Use `CallRequest` type from rust-web3 crate

HCastano:
I think we should be moving towards using "standardized" types where possible, this is a small step towards that.

Bit of a side note - right now it's a little ugly to have four `None` fields added to each call, but `CallRequest` doesn't implement `Default` so I can't change it to be something like this:

```rust
CallRequest {
	to: contract_address,
	data: Some(encoded_call.into()),
        ..Default::default()
}
```

Something for the future, but not a big deal.

### DIFF
--- a/relays/ethereum/src/ethereum_client.rs
+++ b/relays/ethereum/src/ethereum_client.rs
@@ -14,7 +14,9 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
 
-use crate::ethereum_types::{Address, Bytes, EthereumHeaderId, Header, Receipt, TransactionHash, H256, U256, U64};
+use crate::ethereum_types::{
+	Address, Bytes, CallRequest, EthereumHeaderId, Header, Receipt, TransactionHash, H256, U256, U64,
+};
 use crate::substrate_types::{GrandpaJustification, Hash as SubstrateHash, QueuedSubstrateHeader, SubstrateHeaderId};
 use crate::sync_types::{HeaderId, MaybeConnectionError};
 use crate::{bail_on_arg_error, bail_on_error};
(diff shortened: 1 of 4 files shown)


relays/ethereum/src/ethereum_client.rs
```diff
@@ -396,8 +401,12 @@ async fn submit_ethereum_transaction(
 		estimate_gas(
 			client,
 			CallRequest {
-				to: contract_address,
+				to: contract_address.unwrap_or_default(),
```
HCastano, line 404: When creating a contract, an address of `None` was used. I'm not sure how to represent that now since the `to` field isn't an `Option` anymore.

@svyatonik, what do you think we should have here?

relays/ethereum/src/ethereum_client.rs
```diff
@@ -223,8 +216,12 @@ pub async fn best_substrate_block(
 	let (encoded_call, call_decoder) = bridge_contract::functions::best_known_header::call();
 	let call_request = bail_on_arg_error!(
 		to_value(CallRequest {
-			to: Some(contract_address),
+			to: contract_address,
 			data: Some(encoded_call.into()),
+			from: None,
```
tomusdrw, line 221: We should consider adding https://github.com/colin-kiegel/rust-derive-builder to `CallRequest` in web3 crate.

relays/ethereum/src/ethereum_client.rs
tomusdrw: This is not really correct. Deploying a contract must not provide `to` field at all, with current code it will test a call to `0x0` contract. Maybe it's an issue with `CallRequest` type from `web3` crate?

svyatonik: In eth client it is `Option` - so probably yes, it could be changed to `Option` in rust-web3 as well. There's also `TransactionRequest`, which has almost same fields, but is used in different methods (`eth_sendTransaction`).

HCastano:
@tomusdrw FYI, this doesn't use a tagged release of `rust-web3` anymore, but should once one is available. Since the repo doesn't move that fast, I don't think you should rush to make a release just for this change.

Issue #59, opened 2020-04-07
Add more configuration options to runtime module


svyatonik:
follow up of #45 

 - [x] [Aura configuration](https://github.com/paritytech/parity-bridges-common/blob/74803274f0eef23282357992f756254c671205cf/modules/ethereum/src/lib.rs#L624);
- [x] [Kovan validators configuration](https://github.com/paritytech/parity-bridges-common/blob/74803274f0eef23282357992f756254c671205cf/modules/ethereum/src/lib.rs#L638);
- [ ] [Transaction pool configuration](https://github.com/paritytech/parity-bridges-common/blob/74803274f0eef23282357992f756254c671205cf/modules/ethereum/src/lib.rs#L751);
 - [ ] [Priority of unsigned transactions](https://github.com/paritytech/parity-bridges-common/pull/45#discussion_r402960072);
 - [ ] [Longevity of unsigned transactions](https://github.com/paritytech/parity-bridges-common/pull/45#discussion_r402960366);
 - [ ] [Prefix for unsigned transaction tags](https://github.com/paritytech/parity-bridges-common/pull/45#discussion_r402971379);

tomusdrw:
Obsolete by #1210

PR #102, opened 2020-05-20
Remove in memory storage

### LINKED ISSUE #46: Test BridgeStorage trait
There are now 2 `trait Storage` implementors in PoA module: [`InMemoryStorage`](https://github.com/paritytech/parity-bridges-common/blob/b4877dbf87da88c843be57771c6a05d4cdd0426d/modules/ethereum/src/lib.rs#L765) and [`BridgeStorage`](https://github.com/paritytech/parity-bridges-common/blob/b4877dbf87da88c843be57771c6a05d4cdd0426d/modules/ethereum/src/lib.rs#L383). When the module was created, I've used in-memory storage to embed everything in parity-ethereum. #45 adds more logic to some storage methods && there are no actual tests for it.

So we either need to: (1) add tests for `BridgeStorage`, or (2) change all existing tests to use `BridgeStorage`.

svyatonik:
closes #46
part of #59 

on top of #91 

I had to do some tests for #62 && decided to remove `InMemoryStorage` first, because otherwise I was writing tests for test code. When removing, I have also moved everything Kovan-related to the runtime level, so PoA -> Substrate bridge module now can be easily configured to support any PoA chain.


svyatonik:
I'll prepare another PR on top of master branch

PR #103, opened 2020-05-20
Use runtime storage in PoA -> Substrate module tests

### LINKED ISSUE #46: Test BridgeStorage trait

svyatonik:
closes #46
part of #59

This PR removes `InMemoryStorage` + moves all Kovan references from runtime module to runtime level => now we can easily reconfgure it to work with any PoA chain.


modules/ethereum/src/finality.rs
```diff
@@ -214,106 +214,110 @@ pub(crate) fn ancestry<'a, S: Storage>(
...
+					&Header::default(),
+					0,
+				),
+				Err(Error::NotValidator),
+			);
+		});
 	}
 
 	#[test]
 	fn prepares_votes() {
```
HCastano, line 239: I'd like to see all the stuff within `custom_test_ext()` be cleaned up with some functions, so that the test looks something like this:

```rust
	#[test]
	fn validators_finalize_block_when_threshold_is_met() {
		...

		let header1 = get_header(...);
		let finalized_blocks = get_finalized_blocks(header1, ...);
		assert_eq!(finalized_blocks, Ok(Vec::new()));

		let header2 = get_header(...);
		let finalized_blocks = get_finalized_blocks(header2, ...);
		assert_eq!(finalized_blocks, Ok(Vec::new()));

		let header3 = get_header(...);
		let finalized_blocks = get_finalized_blocks(header3, ...);
		assert_eq!(finalized_blocks, Ok(vec![(1, header1.hash, None)]);
	}

```

I found it hard to follow the meat of what the test was checking with all the setup involved.

modules/ethereum/src/import.rs
```diff
@@ -166,214 +166,221 @@ pub fn header_import_requires_receipts<S: Storage>(
...
+			let mut storage = BridgeStorage::<TestRuntime>::new();
+			storage.finalize_headers(Some((100, Default::default())), None);
+			assert_eq!(
+				import_header(
+					&mut storage,
+					&test_aura_config(),
+					&test_validators_config(),
+					PRUNE_DEPTH,
+					None,
+					Default::default(),
```
HCastano, line 189: Does this test work because the default block number would be 0? If so, it would be nice to make that more explicit

modules/ethereum/src/mock.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+/// Build default i-th block, using data from runtime storage.
+pub fn block_i(number: u64, validators: &[KeyPair]) -> Header {
+	custom_block_i(number, validators, |_| {})
+}
+
+/// Build custom i-th block, using data from runtime storage.
+pub fn custom_block_i(number: u64, validators: &[KeyPair], customize: impl FnOnce(&mut Header)) -> Header {
+	let validator_index: u8 = (number % (validators.len() as u64)) as _;
```
HCastano, line 116: Rust Q: What's `as _` meant to do?

modules/ethereum/src/mock.rs
```diff
@@ -0,0 +1,168 @@
...
+	let mut header = Header {
+		number,
+		parent_hash: HeadersByNumber::get(number - 1).unwrap()[0].clone(),
+		gas_limit: 0x2000.into(),
+		author: validator(validator_index).address().to_fixed_bytes().into(),
```
HCastano, line 121: I've seen this a few times, could probably be put behind a function so we can just do `author: author(i)`

modules/ethereum/src/mock.rs
```diff
@@ -0,0 +1,168 @@
...
+		seal: vec![vec![number as u8 + 42].into(), vec![].into()],
+		difficulty: number.into(),
+		..Default::default()
+	};
+	customize(&mut header);
+	signed_header(validators, header, number + 42)
```
HCastano, line 127: Does this 42 have anything to do with the seal, or something else?

modules/ethereum/src/mock.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+/// Return key pair of given test validator.
+pub fn validator(index: u8) -> KeyPair {
+	KeyPair::from_secret(Secret::from([index + 1; 32])).unwrap()
+}
+
+/// Return key pairs of all test validators.
+pub fn validators(count: u8) -> Vec<KeyPair> {
+	(0..count as usize).map(|index| validator(index as u8)).collect()
```
HCastano, line 148: Why's the `as usize` cast necessary here?

modules/ethereum/src/mock.rs
```diff
@@ -0,0 +1,168 @@
...
+use primitives::{Address, Header, H256, U256};
+use sp_runtime::{
+	testing::Header as SubstrateHeader,
+	traits::{BlakeTwo256, IdentityLookup},
+	Perbill,
+};
+
+pub type AccountId = u64;
+
+#[derive(Clone, Eq, PartialEq)]
```
HCastano, line 32:
```suggestion
#[derive(Clone, Eq, PartialEq, Debug)]
```

modules/ethereum/src/finality.rs
```diff
@@ -214,106 +214,110 @@ pub(crate) fn ancestry<'a, S: Storage>(
...
-				0,
-			),
-			Err(Error::NotValidator),
-		);
+		custom_test_ext(genesis(), validators_addresses(5)).execute_with(|| {
+			assert_eq!(
+				finalize_blocks(
+					&BridgeStorage::<TestRuntime>::new(),
+					&Default::default(),
+					(&Default::default(), &[]),
```
HCastano, line 227: I'm confused as why this test returns `Error::NotValidator`, but I suspect is has to do with this. I think we should have a variable like this

```rust
// A valid author would be x, y, or z
let invalid_author: ... =  (&Default::default(), &[]);
```

to make it clear.

modules/ethereum/src/import.rs
```diff
@@ -166,214 +166,221 @@ pub fn header_import_requires_receipts<S: Storage>(
...
+				import_header(
+					&mut storage,
+					&test_aura_config(),
+					&test_validators_config(),
+					PRUNE_DEPTH,
+					None,
+					block.clone(),
+					None,
+				)
+				.map(|_| ()),
```
HCastano, line 213: Instead of doing this, could we not do `assert!(import_header(...).is_ok())`

modules/ethereum/src/import.rs
```diff
@@ -166,214 +166,221 @@ pub fn header_import_requires_receipts<S: Storage>(
...
+				Err(Error::KnownHeader),
+			);
+		});
 	}
 
 	#[test]
 	fn import_header_works() {
```
HCastano, line 233: I think this test is either doing too much, or the name needs to be updated.

If we're just checking that importing headers work, I don't think we need that `Multi` authority set thing that's going on. That could be part of a different test that checks validator set changes.

bin/node/runtime/src/kovan.rs
```diff
@@ -0,0 +1,191 @@
...
+
+/// Validators configuration for Kovan chain.
+pub fn kovan_validators_configuration() -> ValidatorsConfiguration {
+	ValidatorsConfiguration::Multi(vec![
+		(0, ValidatorsSource::List(kovan_genesis_validators())),
+		(
+			10960440,
+			ValidatorsSource::List(vec![
+				[
+					0x00, 0xD6, 0xCc, 0x1B, 0xA9, 0xcf, 0x89, 0xBD, 0x2e, 0x58, 0x00, 0x97, 0x41, 0xf4, 0xF7, 0x32,
```
HCastano, line 43: How are these generated?

modules/ethereum/src/verification.rs
```diff
@@ -362,41 +367,88 @@ mod tests {
...
-		)
+		custom_test_ext(genesis(), validators_addresses(3)).execute_with(|| {
+			let validators = vec![validator(0), validator(1), validator(2)];
+			let mut storage = BridgeStorage::<TestRuntime>::new();
+			let block1 = block_i(1, &validators);
+			storage.insert_header(
+				storage
+					.import_context(None, &block1.parent_hash)
+					.unwrap()
+					.into_import_header(true, block1.hash(), block1, 0.into(), None, None),
```
HCastano, line 391: I think this can be done in a function. It would make the test flow in a similar way as before

modules/ethereum/src/verification.rs
```diff
@@ -362,41 +367,88 @@ mod tests {
...
+				&test_aura_config(),
+				&validators_config,
+				&pool_configuration(),
+				&header,
+				receipts.as_ref(),
+			)
+		})
+	}
+
+	fn change_validators_set_at(number: u64, finalized_set: Vec<Address>, signalled_set: Option<Vec<Address>>) {
```
HCastano, line 425:
```suggestion
	fn change_validators_set_at(block_number: u64, finalized_set: Vec<Address>, signalled_set: Option<Vec<Address>>) {
```

HCastano:
I realize that I may have made more comments to do with tests than the changes introduced in this PR. I wouldn't blame you if you don't want to fix them here (since they're not super relevant), but I think the tests should be cleaned up a bit at some point.

With that said, I actually do dig the changes, makes things a lot nicer to work with :)

modules/ethereum/src/mock.rs
tomusdrw, line 148:
```suggestion
	(0..count).map(validator).collect()
```

modules/ethereum/src/mock.rs
```diff
@@ -0,0 +1,168 @@
...
+	KeyPair::from_secret(Secret::from([index + 1; 32])).unwrap()
+}
+
+/// Return key pairs of all test validators.
+pub fn validators(count: u8) -> Vec<KeyPair> {
+	(0..count as usize).map(|index| validator(index as u8)).collect()
+}
+
+/// Return addresses of all test validators.
+pub fn validators_addresses(count: u8) -> Vec<Address> {
```
tomusdrw, line 152:
```suggestion
pub fn validator_addresses(count: u8) -> Vec<Address> {
```

double plural sounds okay for me (:slavic:), but afair it's not how you do it in english though :man_shrugging:

modules/ethereum/src/mock.rs
```diff
@@ -0,0 +1,168 @@
...
+	(0..count as usize)
+		.map(|i| validator(i as u8).address().as_fixed_bytes().into())
```
tomusdrw, line 154: ditto

modules/ethereum/src/mock.rs
svyatonik: IIRC it is some random offset to convert from block number to step number in tests

HCastano: We should make that something like `const BLOCK_TO_STEP_OFFSET = 42` to make it clear

modules/ethereum/src/import.rs
HCastano: Logged #111

modules/ethereum/src/finality.rs
HCastano: Logged #111

modules/ethereum/src/import.rs
HCastano: Logged #111

modules/ethereum/src/import.rs
HCastano: Logged #112

HCastano:
There are still some outstanding comments I have. One those are fixed/disputed I think this can be merged

modules/ethereum/src/mock.rs
svyatonik: This is unrelated to this PR, because this fn has only been moved to that file. Can be addressed in #111

modules/ethereum/src/finality.rs
svyatonik: Unrelated to this PR. Can be addressed in #111

tomusdrw:
@HCastano can you take a look if all things were addressed?

Issue #78, opened 2020-04-28
Prepare actual weights for runtime transactions


svyatonik:
In #69 I had to switch to Substrate master to be able to use `GrandpaJustification::decode_and_verify_finalizes()` from builtin. It turned out that the main change is that now all runtime calls (transactions) require annotated weights. So before considering PoA -> Substrate headers sync to be finished, we need to prepare actual weights for all calls. I've started to do that, but it's a quite complicated process + as some runtime changes are still planned (I'm mostly talking about #62 and maybe #38 ), it isn't the right moment to do that now.

The main problem is that we can't compute weight of header import call by looking at the header itself. The main source of uncertainty is finality computations - we may need to read arbitrary number of headers (and write some updates after that) if validators haven't been finalized headers for too long. At the beginning I had some ideas that we'll cache computation results in the storage, so we don't need to recompute it on every block, but iirc I've stuck with something - so most probably this is impossible, and we'll need storage reads anyway :/ But may worth looking at it again.

That said, most probably, we need to introduce some amortized weight for import calls. This may also affect some params/default constants in relay, because weight may be too big to fit into 'normal' block.

So I'm proposing to mark all methods with `#[weight=0]` (or whatever you suggest) and change that afterwards.

tomusdrw:
> So I'm proposing to mark all methods with #[weight=0] (or whatever you suggest) and change that afterwards.

Sounds good. Let's add a `TODO` and this issue number to make sure we address that.

> The main problem is that we can't compute weight of header import call by looking at the header itself. The main source of uncertainty is finality computations - we may need to read arbitrary number of headers (and write some updates after that) if validators haven't been finalized headers for too long. 

I'd be down for defining some upper bounds and possibly (at least have a) fall-back to a slower mechanism, where we have to process the finality in smaller chunks to make sure we don't go over that upper bound.

Number of headers that are expected to be read can also be passed with the call itself to avoid storage read - then we check if that's actually the case when the call is being processed. That will allow having a dynamic weight without incurring extra state read on computation.

tomusdrw:
Also worth noting:
![image](https://user-images.githubusercontent.com/581548/80488658-fb0bb100-895e-11ea-80ce-9ad799dd1149.png)

HCastano:
We should probably get some help from those doing the Substrate benchmarking when it comes time to figure out our weight values

svyatonik:
[...]

PR #105, opened 2020-05-21
Limit number of headers that are pruned within single import Call

### LINKED ISSUE #62: Prune old headers with separate transaction

svyatonik:
closes #62 
on top of #103 

This is alternative approach of solving #62 - I've decided not to introduce additional call that'll just prune blocks. The main reasons are:
1) it is not clear when someone (relay in our case) should call this method. What't the criteria - when there are too much unpruned headers, or when we are not importing too much blocks? In first case we might slow down import by replacling import-txs with prune-txs in blocks. In second case we might end up with too large state trie, where most of entries are not used - that's what I was trying to avoid by introducing pruning;
2) I was not sure if the 'pruner' should be rewarded? Our schema was to either reward for headers, or reward for messages (that are requiring headers). But pruning is slightly different - why would anyone pay for transactions that bring no benefits?
3) the whole schema was too complex, because it was affecting both module and relay.

So the idea of this PR is to introduce 'amortized pruning' - i.e. we're limiting max number of headers that might be pruned in single import call (8 for now). In future (as a part of #78 ?) the plan is to 'unconst' this number - i.e. if we have imported 100 headers and finalized 1000 headers in single call, then probably we do not want to perform pruning at all (in this block). But if we have imported just 4 headers and finalized another 4, then we might want to import e.g. 32 headers (if we still have anything to prune).


modules/ethereum/src/lib.rs
```diff
@@ -480,10 +499,94 @@ impl<T: Trait> frame_support::unsigned::ValidateUnsigned for Module<T> {
...
+
+		// update oldest block we want to keep
+		if let Some(prune_end) = prune_end {
+			new_pruning_range.oldest_block_to_keep = prune_end;
+		}
+
+		// start pruning blocks
+		let begin = new_pruning_range.oldest_unpruned_block;
+		let end = new_pruning_range.oldest_block_to_keep;
+		for number in begin..end {
```
tomusdrw, line 521: This may panic if `begin > end` - we don't seem to check the `prune_end` before applying it here.

modules/ethereum/src/lib.rs
```diff
@@ -480,10 +499,96 @@ impl<T: Trait> frame_support::unsigned::ValidateUnsigned for Module<T> {
...
+			new_pruning_range.oldest_unpruned_block = number + 1;
+		}
+
+		// update pruning range in storage
+		if pruning_range != new_pruning_range {
+			BlocksToPrune::put(new_pruning_range);
+		}
+	}
+
+	/// Prune od blocks with given hashes.
```
HCastano, line 556:
```suggestion
	/// Prune old blocks with given hashes.
```

Issue #118, opened 2020-06-02
Problems with new async-std


svyatonik:
#107 bumps `async-std` to `1.6.0`. This breaks `PoA` <-> `Substrate` relay - it stops waking up => new headers are not submitted.

svyatonik:
So the reason seems to be migration of `async-std` to [smol](https://github.com/async-rs/async-std/pull/757) that has happened recently and the fact that current smol has some [by-design problems](https://github.com/stjepang/smol/issues/153) with how we use timers.

The minimal (hanging) test would be:
```rust
let mut local_pool = futures::executor::LocalPool::new();
local_pool.run_until(async move {
    async_std::task::sleep(std::time::Duration::from_secs(1)).await;
});
```

So my suggestion is to wait until `async-std` will support our way of `sleep`ing.

PR #119, opened 2020-06-02
Downgrade async-std to 1.5.0

svyatonik:
it is impossible to test sync-using-relay now. Long-term fix is logged in #118


tomusdrw:
Any chance we can get some basic test to prevent this from happening again?

svyatonik:
I do not yet know what causes this - once #118 is fixed, I can answer that question. I assume that for detecting this particular issue we need some kind of integration tests. But having basic tests for sync loop would be a benefit (it is already traitified now, so it should be possible).

PR #122, opened 2020-06-05
Add diagrams from draw.io

HCastano:
I'm trying to figure out how best to share these in such a way that we
can all contribute to them easily. While integration between draw.io
and GitHub exists, it requires read/write permissions to all ParityTech
repos, which doesn't comfort me. I think having the .drawio file and
the corresponding pictures available in a folder is a good start.

I imagine the workflow would be something like this: Someone wants to
update a diagram, so they go to draw.io and upload the .drawio file
from the repo. They make their changes, download the new .drawio file
and corresponding SVG, and commit their changes to the repo.

I plan to make a Wiki page for showing these diagrams off and possibly
going into more in-depth explanations about certain things.


tomusdrw:
The workflow looks okay to me for start.

> I plan to make a Wiki page for showing these diagrams off and possibly
going into more in-depth explanations about certain things.

We can just have a md file in the repo that embeds the SVG. Or set up the wiki so that it's generated from files in the repo - I'd prefer to have it version controlled.

Comments to the diagrams:
- `bridge-relay.svg`
    - I feel like "Substrate Client" and "Ethereum Client" should be two separate blocks, now it kinds of suggest you can have one client that works for both chains.
    - The blocks within `sync loop` are not verify descriptive, but I guess more elaborate documentation will be included in a text form.
    - The diagram does not have a background set (but that's probably alright)
- `cross-chain-fund-transfer.svg`:
    - Nor the diagram neither the boxes have any background. I think it would be good to for the boxes to have white background.
    - My impression was that relay is not responsible for submitting transaction proof, and we make this users responsibility. The events are also not necessary, since we can just look at the transactions itself, not the events. If that's the case then I think we should prepare a tx-proof-submitter relay component for users, so that it can be configured to relay particular kind of transactions. I think initially we can relay the transactions for them (but we have to be careful with tx costs - first relaying might be on us, but subsequent transfers might require topping up our account first (like giving a tip prior to relaying)).
- `ethereum-pallet.svg`
    - "Punish Bad Summitter" - typo
- Some other SVGs are missing boxes backgrounds as well.

Looks pretty good overall, I think adding text descriptions will clear more things up.

HCastano:
So I made the background for each diagram transparent (you need to tick a box when you export them from draw.io), and I filled in all the actual diagram boxes to white. I've also dumped the diagrams in an ARCHITECTURE.md file. I'm not super keen on making write ups for them right now, but if you guys think that it should be required as part of this PR I can do it.

Also, regarding this:

> My impression was that relay is not responsible for submitting transaction proof, and we make this users responsibility.

I think this is a larger discussion outside the scope of this PR, but it's good that these visualizations can help us get an idea of what we each understand about the project so we can work towards a common consensus :)

Issue #51, opened 2020-04-03
Check header timestamp against timestamp module


tomusdrw:
https://github.com/paritytech/parity-bridges-common/pull/45/files#diff-78128fbffca4ac7effa57faf86586991R205

We could perform optional checks against `frame_timestamp` module if available. This should be implemented via associated type (don't reference `frame_timestamp` directly) and should be optional.

svyatonik:
Another benefit of having this fixed. Right now, the logic of relayer is simple:
1) it submits bunch of headers to target node;
2) it waits until either `best_block` is updated, or timeout is elapsed;
2.1) if `best_block` is updated, relayer goes to (1);
2.2) if timeout is elapsed, relayer 'restarts' sync process (i.e. it forgets all headers) and goes to (1).

The bad thing about PoA -> Substrate sync is that if we don't make timestamps checks, then some authority may submit header with step too far from the future => this header will be marked as best. This header will stay best until we'll finalize some ancestor of this block (given #137 is fixed). This may cause pauses in headers relay.

PR #128, opened 2020-06-10
Move pruning strategy to runtime level

svyatonik:
This PR implements option (1) from this discussion:
https://github.com/paritytech/parity-bridges-common/pull/91#issuecomment-638854485
https://github.com/paritytech/parity-bridges-common/pull/91#issuecomment-638923992

New runtime gives ~24 hours (which is ~20000 blocks on current Kovan chain) after 'mining' block `B` with funds-lock transaction to:
1) PoA validators to finalize block `B`;
2) relay to synchronize this header of `B` and finality proof (i.e. some headers built on top of `B`);
3) claimant to prepare and submit proof-of-funds-lock transaction;
4) Substrate validators to 'mine' block with proof-of-funds-lock transaction.

===

This, however doesn't solve the problem discussed in:
https://github.com/paritytech/parity-bridges-common/pull/91#issuecomment-639339730

Why? Mostly because of technical difficulties. I've moved pruning strategy to the runtime level, because imo different bridges may have different strategies. So to support pruning that is timestamp-dependent, we need to: (1) read hashes of all headers at given depth (2) read all headers and compute `max(timestamp)` (or at least read timestamp of canonical header) (3) go to depth + 1 if we have decided to prune headers at `depth` (4) continue until we'll met header that we do not want to prune, or unfinalized header. It seems too heavy for me - I'd better implement option (2) from https://github.com/paritytech/parity-bridges-common/pull/91#issuecomment-638854485 in near future. At least it is heavy in more convenient (mostly for claimant) way.

Another option would be to maintain some additional map like `[ at 12:00 => prune headers with depth < 1000, at 13:00 => prune headers with depth < 1100, ... ]`, but this map also needs to be updated on every header insertion/canonicalization and it needs to reside at runtime level, because other pruning strategies might not need that info. And I have found no easy way to organize map on runtime level (other than introducing some module inside runime and declaring map there).

That said, I'm ready to implement something like that ^^^ if you find that current `KovanPruningStrategy` doesn't match our requirements.

===

I'm also not sure if we 100% need to introduce what is described in this comment: https://github.com/paritytech/parity-bridges-common/pull/91#issuecomment-639560578 . First, I've realized that noone is actually going to sync PoA from the scratch - so we'll most probably will start with near-best PoA block. Then, if we'll disable claims, the claimant will still be able to lock his funds on PoA chain and then just won't have an option to submit proof-of-lock.

### DIFF
--- a/modules/ethereum/src/import.rs
+++ b/modules/ethereum/src/import.rs
@@ -33,6 +33,7 @@ use sp_std::{collections::btree_map::BTreeMap, prelude::*};
 /// imported in this case.
 pub fn import_headers<S: Storage, PS: PruningStrategy>(
 	storage: &mut S,
+	pruning_strategy: &mut PS,
 	aura_config: &AuraConfiguration,
 	validators_config: &ValidatorsConfiguration,
 	submitter: Option<S::Submitter>,
@@ -42,8 +43,9 @@ pub fn import_headers<S: Storage, PS: PruningStrategy>(
 	let mut useful = 0;
 	let mut useless = 0;
 	for (header, receipts) in headers {
-		let import_result = import_header::<_, PS>(
+		let import_result = import_header(
 			storage,
+			pruning_strategy,
 			aura_config,
 			validators_config,
 			submitter.clone(),
@@ -76,6 +78,7 @@ pub fn import_headers<S: Storage, PS: PruningStrategy>(
 /// Returns imported block id and list of all finalized headers.
 pub fn import_header<S: Storage, PS: PruningStrategy>(
 	storage: &mut S,
+	pruning_strategy: &mut PS,
 	aura_config: &AuraConfiguration,
 	validators_config: &ValidatorsConfiguration,
 	submitter: Option<S::Submitter>,
@@ -130,7 +133,7 @@ pub fn import_header<S: Storage, PS: PruningStrategy>(
 	// compute upper border of updated pruning range
 	let new_best_block_id = if is_best { header_id } else { best_id };
 	let new_best_finalized_block_id = finalized_blocks.finalized_headers.last().map(|(id, _)| *id);
-	let pruning_upper_bound = PS::pruning_upper_bound(
+	let pruning_upper_bound = pruning_strategy.pruning_upper_bound(
 		new_best_block_id.number,
 		new_best_finalized_block_id
 			.map(|id| id.number)
@@ -160,7 +163,7 @@ mod tests {
 	use super::*;
 	use crate::mock::{
 		block_i, custom_block_i, custom_test_ext, genesis, signed_header, test_aura_config, test_validators_config,
-		validator, validators, validators_addresses, Keep10HeadersBehindBest, TestRuntime,
+		validator, validators, validators_addresses, KeepSomeHeadersBehindBest, TestRuntime,
 	};
 	use crate::validators::ValidatorsSource;
 	use crate::{BlocksToPrune, BridgeStorage, Headers, PruningRange};
@@ -178,8 +181,9 @@ mod tests {
 				0,
 			);
 			assert_eq!(
-				import_header::<_, Keep10HeadersBehindBest>(
+				import_header(
 					&mut storage,
+					&mut KeepSomeHeadersBehindBest::default(),
 					&test_aura_config(),
 					&test_validators_config(),
 					None,
@@ -198,8 +202,9 @@ mod tests {
 			let mut storage = BridgeStorage::<TestRuntime>::new();
 			let block = block_i(1, &validators);
 			assert_eq!(
-				import_header::<_, Keep10HeadersBehindBest>(
+				import_header(
 					&mut storage,
+					&mut KeepSomeHeadersBehindBest::default(),
 					&test_aura_config(),
 					&test_validators_config(),
 					None,
@@ -210,8 +215,9 @@ mod tests {
 				Ok(()),
 			);
 			assert_eq!(
-				import_header::<_, Keep10HeadersBehindBest>(
+				import_header(
 					&mut storage,
+					&mut KeepSomeHeadersBehindBest::default(),
 					&test_aura_config(),
 					&test_validators_config(),
 					None,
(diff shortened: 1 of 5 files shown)


modules/ethereum/src/lib.rs
```diff
@@ -286,15 +286,31 @@ pub trait Storage {
 	fn scheduled_change(&self, hash: &H256) -> Option<ScheduledChange>;
 	/// Insert imported header.
 	fn insert_header(&mut self, header: HeaderToImport<Self::Submitter>);
-	/// Finalize given block and prune all headers with number < prune_end.
+	/// Finalize given block and prune schedules pruning of all headers
```
HCastano, line 289:
```suggestion
	/// Finalize given block and schedules pruning of all headers
```

modules/ethereum/src/lib.rs
```diff
@@ -770,7 +786,7 @@ impl<T: Trait> Storage for BridgeStorage<T> {
 		);
 	}
 
-	fn finalize_headers(&mut self, finalized: Option<HeaderId>, prune_end: Option<u64>) {
+	fn finalize_headers(&mut self, finalized: Option<HeaderId>, prune_end: u64) {
```
HCastano, line 789: nit: We should consider naming this `finalize_and_prune_headers`

modules/ethereum/src/import.rs
```diff
@@ -140,15 +127,19 @@ pub fn import_header<S: Storage>(
...
 
-	// now mark finalized headers && prune old headers
-	storage.finalize_headers(
-		finalized_blocks.finalized_headers.last().map(|(id, _)| *id),
-		match is_best {
-			true => header_number.checked_sub(prune_depth),
-			false => None,
-		},
+	// compute upper border of updated pruning range
+	let new_best_block_id = if is_best { header_id } else { best_id };
```
tomusdrw, line 131: if the imported header is not new best (i.e. `is_best == false`) then we don't really need to run the whole code at all (pruning & finalizing), do we?

We can assume that it was already run when `best_id` was imported.

modules/ethereum/src/lib.rs
```diff
@@ -286,15 +286,31 @@ pub trait Storage {
...
+	/// Return upper bound (exclusive) of headers pruning range.
+	///
+	/// Every value that is returned from this function, must be greater or equal to the
+	/// previous value. Otherwise it will be ignored (we can't revert pruning).
+	///
+	/// Module may prune both finalized and unfinalized blocks. But it can't give any
+	/// guarantees on when it will happen. Example: if some unfinalized block at height N
+	/// has scheduled validators set change, then the module won't prune any blocks with
+	/// number >= N even if strategy allows that.
+	fn pruning_upper_bound(best_number: u64, best_finalized_number: u64) -> u64;
```
tomusdrw, line 310: I like how `Storage` traits require an instance and allows to pass different values if needed. I feel like we should do the same here (we can require `Default` bound to simplify creation of that instance). That would probably make the test code a bit nicer, since passing the type explicitly would not be required.

```suggestion
	fn pruning_upper_bound(&mut self, best_number: u64, best_finalized_number: u64) -> u64;
```

modules/ethereum/src/mock.rs
```diff
@@ -182,3 +183,12 @@ pub fn insert_header<S: Storage>(storage: &mut S, header: Header) {
 		finality_votes: FinalityVotes::default(),
 	});
 }
+
+/// Pruning strategy that keeps 10 headers behind best block.
+pub struct Keep10HeadersBehindBest;
```
tomusdrw, line 188: By making `PruningStrategy` work with `&mut self` we could make parametrisation of this type easier (i.e. just pass `10` as parameter). `Kovan` one would be just a wrapper in such case.

svyatonik: Kovan keeps some blocks behind best finalized && this keeps some blocks behind best known block.

modules/ethereum/src/import.rs
svyatonik: That's a very good point! But while I hope you're right, I'm unsure about this - will need to check it twice. My concern is whether situation like that is possible:
1) we need 2-of-3 headers for finality;
2) unfinalized headerchain is `H10, H11, H12, H13` - all headers are signed by signer1;
3) signer2 broadcasts the block `H12'` (build on top of `H11`) for some old step.

This header is not the best, because iiuc difficulty is determined by total number of steps in the chain. But at the same time it finalizes headers `H10` and `H11`. I'm not sure if current module impl would reject, or allow import of `H12'` and how it must behave actually.

tomusdrw: Hmm, interesting. Worth making a test case out of your example to make sure it behaves as we intend.
I assume that if we don't finalize & prune nothing really bad would happen, eventually a chain built on top of `H12'` should take over the `H13` one and then we would finalize & prune `H10` & `H11`
To be honest I think I'd prefer that behavior as more predictable for the users, even though technically `H10`&`H11` can be considered finalized earlier. The worst thing that may happen is that submitting proofs is delayed because of that, but I don't think that extra pruning we have to do when `H13'` gets imported is opening an attack vector.

svyatonik: OK, so imo we shouldn't implement this ^^^ behavior. Reasons:
1) if header H has signaled validators change and this block is finalized at block H+N, then starting from block H+N+1 we start verifying headers using new validators set. So we can't miss the moment when block B has been finalized even for non-best chain (i.e. we should track finality on non-best forks);
2) since we don't have any timestamp-related checks now (adding #51 here for reference), malicious authority can maintain competing fork by continuously generating headers => piling up steps => making this fork best => we never finalize anything.

I'll add a test for this case though.

tomusdrw: I'm convinced :) Thanks!

svyatonik: Test is added. Another round of thinking about your comment has resulted in #137 (will try to fix asap) and https://github.com/paritytech/parity-bridges-common/issues/51#issuecomment-644744314 .

PR #129, opened 2020-06-11
Update Substrate to latest master (rc3)

### LINKED ISSUE #98: Eth2Sub transactions are removed from txpool
Seems like after last Substrate update, there's a problem with eth2sub syncing. Some transactions are removed from tx pool with `ExhaustsResources` error. Relay currently do not watch for submitted transactions => because nothing is updated in Substrate state, it waits 5 minutes before 'restarting' itself and resubmitting same headers again.

I'm not sure, if that's the problem with:
1) our configuration - maybe it works that way with `#[weight = 0]`? something is missed from the runtime;
2) some already-fixed Substrate issue - probably ref update will solve this;
3) not-yet-fixed Substrate issue.

In any case, dropping transactions from pool with that error seems strange - the pool is empty (there's at most 4 signed transactions there during eth2sub), so which resources are exhausted?

Required for #91 

[log](https://gist.github.com/svyatonik/6cf28a1e0238986f72799ff994655dfc)

svyatonik:
closes #98 

There's some black magic behind this PR - there was dependencies conflict between `sc-rpc-api` and `sc-network` - they were depending on different version of `quote` (1.0.6/7 vs 1.0.3). I have manually changed versions in Cargo.lock to 1.0.7 for now. The reason why I want this PR in, is that I'm starting to write weight-related benchmarks and wanted to have up-to-date runtime for that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -171,7 +171,7 @@ version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0d0864d84b8e07b145449be9a8537db86bf9de5ce03b913214694643b4743502"
 dependencies = [
- "quote 1.0.3",
+ "quote 1.0.7",
  "syn 1.0.18",
 ]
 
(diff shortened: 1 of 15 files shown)


svyatonik:
(the second commit also brings updated `jsonrpsee` with 100% CPU fix)

PR #131, opened 2020-06-12
Switch to new RPC interface

### LINKED ISSUE #72: Use shared client for RPC calls instead of passing an owned version

HCastano:
This PR hooks up the changes introduced in #80 with the rest of the codebase. The motivation behind this PR is mainly to simplify the existing code, as well as making a more clear separation between the clients and the RPC.

I've split the RPCs into what I consider "base" and "high" level calls. The "base" calls are captured in the `SubstrateRpc`/`EthereumRpc` traits, and the "high" level calls in `SubstrateEthereumHeaders` and `EthereumHighLevelRpc` (creative, I know) traits. The base RPCs try to do the base minimum on top of the APIs defined in `rpc_api!`, while the high level RPCs are made up of a combination of base calls, or perform extra logic (like interacting with a smart contract).

The implementations of the RPCs introduced in #80 were moved from the `rpc` module into the respective `client` modules. I left the traits however, since I figured that they should be close to the API definitions.

Some other things to note:

Right now this fails to sync the bridge. Every now and then the relay complains about the Substrate chain missing the `EthereumHeadersApi-*` runtime APIs, and I'm not sure what's going on (the APIs exist, `master` syncs just fine).

This uses a fork of `jsonrpsee`. This is because at the moment the `jsonrpsee::rpc_api!` macro doesn't support the use of a `Client` (only `RawClient`'s). By using `Client` over `RawClient` we're able to use regular immutable references when working with the RPC, removing the headache that comes with shared mutable references + async/await.

Closes #72.

### DIFF
--- a/relays/ethereum/src/ethereum_deploy_contract.rs
+++ b/relays/ethereum/src/ethereum_deploy_contract.rs
@@ -61,13 +61,13 @@ pub fn run(params: EthereumDeployContractParams) {
 	let mut local_pool = futures::executor::LocalPool::new();
 
 	let result = local_pool.run_until(async move {
-		let mut eth_client = EthereumRpcClient::new(params.eth); // ethereum_client::client(params.eth);
-		let mut sub_client = SubstrateRpcClient::new(params.sub); // substrate_client::client(params.sub);
+		let eth_client = EthereumRpcClient::new(params.eth);
+		let sub_client = SubstrateRpcClient::new(params.sub);
 
-		let (initial_header_hash, initial_header) = prepare_initial_header(&mut sub_client, params.sub_initial_header).await?;
+		let (initial_header_hash, initial_header) = prepare_initial_header(&sub_client, params.sub_initial_header).await?;
 		let initial_set_id = params.sub_initial_authorities_set_id.unwrap_or(0);
 		let initial_set = prepare_initial_authorities_set(
-			&mut sub_client,
+			&sub_client,
 			initial_header_hash,
 			params.sub_initial_authorities_set,
 		).await?;
@@ -97,7 +97,7 @@ pub fn run(params: EthereumDeployContractParams) {
 
 /// Prepare initial header.
 async fn prepare_initial_header(
-	sub_client: &mut SubstrateRpcClient,
+	sub_client: &SubstrateRpcClient,
 	sub_initial_header: Option<Vec<u8>>,
 ) -> Result<(SubstrateHash, Vec<u8>), String> {
 	match sub_initial_header {
@@ -116,7 +116,7 @@ async fn prepare_initial_header(
 
 /// Prepare initial GRANDPA authorities set.
 async fn prepare_initial_authorities_set(
-	sub_client: &mut SubstrateRpcClient,
+	sub_client: &SubstrateRpcClient,
 	sub_initial_header_hash: SubstrateHash,
 	sub_initial_authorities_set: Option<Vec<u8>>,
 ) -> Result<Vec<u8>, String> {
--- a/relays/ethereum/src/ethereum_deploy_contract.rs
+++ b/relays/ethereum/src/ethereum_deploy_contract.rs
@@ -16,7 +16,7 @@
 
 use crate::ethereum_client::{DeployContract, EthereumConnectionParams, EthereumRpcClient, EthereumSigningParams};
 use crate::rpc::SubstrateRpc;
-use crate::substrate_client::{self, SubstrateConnectionParams, SubstrateRpcClient};
+use crate::substrate_client::{SubstrateConnectionParams, SubstrateRpcClient};
 use crate::substrate_types::{Hash as SubstrateHash, Header as SubstrateHeader};
 
 use codec::{Decode, Encode};
@@ -81,7 +81,6 @@ pub fn run(params: EthereumDeployContractParams) {
 			hex::encode(&initial_set),
 		);
 
-		// ethereum_client::deploy_bridge_contract(
 		eth_client.deploy_bridge_contract(
 			&params.eth_sign,
 			params.eth_contract_code,
--- a/relays/ethereum/src/ethereum_deploy_contract.rs
+++ b/relays/ethereum/src/ethereum_deploy_contract.rs
@@ -64,12 +64,7 @@ pub fn run(params: EthereumDeployContractParams) {
 
 	let result = local_pool.run_until(async move {
 		let eth_client = EthereumRpcClient::new(params.eth);
-		let sub_client = SubstrateRpcClient::new(params.sub).await;
-
-		let sub_client = match sub_client {
-			Ok(s) => s,
-			Err(e) => return Err(e.into()),
-		};
+		let sub_client = SubstrateRpcClient::new(params.sub).await?;
 
 		let (initial_header_hash, initial_header) = prepare_initial_header(&sub_client, params.sub_initial_header).await?;
 		let initial_set_id = params.sub_initial_authorities_set_id.unwrap_or(0);
--- a/relays/ethereum/src/ethereum_deploy_contract.rs
+++ b/relays/ethereum/src/ethereum_deploy_contract.rs
@@ -15,7 +15,7 @@
 // along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
 
 use crate::ethereum_client::{
-	EthereumConnectionParams, EthereumHighLevelRpc, EthereumRpcClient, EthereumSigningParams,
+	bridge_contract, EthereumConnectionParams, EthereumHighLevelRpc, EthereumRpcClient, EthereumSigningParams,
 };
 use crate::rpc::SubstrateRpc;
 use crate::substrate_client::{SubstrateConnectionParams, SubstrateRpcClient};
@@ -23,9 +23,6 @@ use crate::substrate_types::{Hash as SubstrateHash, Header as SubstrateHeader};
 
 use codec::{Decode, Encode};
 use num_traits::Zero;
-
-// to encode/decode contract calls
-ethabi_contract::use_contract!(bridge_contract, "res/substrate-bridge-abi.json");
 
 /// Ethereum synchronization parameters.
 #[derive(Debug)]
--- a/relays/ethereum/src/ethereum_deploy_contract.rs
+++ b/relays/ethereum/src/ethereum_deploy_contract.rs
@@ -14,13 +14,18 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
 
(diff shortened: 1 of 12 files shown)


relays/ethereum/src/substrate_client.rs
```diff
@@ -71,301 +81,187 @@ impl Default for SubstrateSigningParams {
...
+		headers: Vec<QueuedEthereumHeader>,
+	) -> Result<Vec<EthereumHeaderId>> {
+		let ids = headers.iter().map(|header| header.id()).collect();
+
+		let genesis_hash = match self.genesis_hash {
+			Some(genesis_hash) => genesis_hash,
+			None => {
+				let genesis_hash = self.block_hash_by_number(Zero::zero()).await?;
+				// TODO: Fix, need &mut
+				// self.genesis_hash = Some(genesis_hash);
```
HCastano, line 239: For this, I really don't want to pass a `&mut` into this function (it bubbles up like 3 levels), so should I just use a `RefCell` or something to do this?

relays/ethereum/src/rpc_errors.rs
```diff
@@ -58,12 +55,28 @@ impl From<SubstrateNodeError> for RpcError {
...
+impl From<ethabi::Error> for RpcError {
+	fn from(err: ethabi::Error) -> Self {
+		Self::Ethereum(EthereumNodeError::ResponseParseFailed(format!("{}", err)))
+	}
+}
+
+impl MaybeConnectionError for RpcError {
+	fn is_connection_error(&self) -> bool {
+		match *self {
+			// Q: Is blanket handling `Request` too broad?
```
tomusdrw, line 73: @svyatonik wdyt?

relays/ethereum/src/substrate_client.rs
tomusdrw: Since this is just cache, it's totally fine to use `RefCell` if Send/Sync is not required for this struct.
If it is maybe use something like `once` initialization (or `lazy_static`). Even a piece of unsafe code here would be fine - we really don't care that much about consistency in the cache.

relays/ethereum/src/sync_loop.rs
```diff
@@ -488,55 +461,50 @@ pub fn run<P: HeadersSyncPipeline>(
...
+		async_std::task::sleep(timeout).await;
 		Some(((), ()))
 	})
 }
 
 /// Process result of the future that may have been caused by connection failure.
-fn process_future_result<TClient, TResult, TError, TGoOfflineFuture>(
-	maybe_client: &mut Option<TClient>,
-	client: TClient,
+fn process_future_result<TResult, TError, TGoOfflineFuture>(
```
tomusdrw, line 480: Might be worth updating the docs to indicate what the returned `bool` is.

relays/ethereum/src/substrate_client.rs
svyatonik: Let's move this to constructor maybe?

svyatonik: (I mean read genesis hash from `SubstrateRpcClient::new()`)

relays/ethereum/src/rpc_errors.rs
svyatonik: I'm not sure, but imo we should only return true if we see `Request(RequestError::TransportError(_))` here. The idea is that if there are some problems with **connection**, we sleep some time and then retry.

relays/ethereum/src/sync_loop.rs
```diff
@@ -131,7 +126,7 @@ pub fn run<P: HeadersSyncPipeline>(
 		let mut stall_countdown = None;
 		let mut last_update_time = Instant::now();
 
-		let mut source_maybe_client = None;
+		let mut source_client_is_online = false;
```
svyatonik, line 129: I think that concept of we-can-execute-at-most-1-request-at-a-time is lost here. Before this PR, if you have e.g. asked Ethereum node about best header, you was unable to ask it for some header until best header has been received. In your concept, this flag has changed its meaning - it is `false` only when transport error has been reported by the underlying client (OK - and initially : ) ). I think that's the reason behind errors like this:
```
Substrate: [2020-06-19 15:26:59] 2020-06-19 15:26:59.941 http.worker30 DEBUG txpool  [0x31d7836d46b14977db3012267046897bbc13307fa78f8ce4d41c98ac61626b6b] Error importing: TooLowPriority { old: 10020844, new: 10020844 }

Substrate: [2020-06-19 15:26:59] 2020-06-19 15:26:59.941 http.worker30 DEBUG rpc  Response: {"jsonrpc":"2.0","error":{"code":1014,"message":"Priority is too low: (10020844 vs 10020844)","data":"The transaction has too low priority to replace another transaction already in the pool."},"id":330}.
```

I.e. we're spawning future1 that submits `headers[1..16]` to substrate node and then immediately spawning future2 that submits `headers[17..32]`. And there's a race, because both futures are asking Substrate node about account' nonce => two transactions are receiving the same nonce => submit fails.

Let's revert to previous behavior here (1 future at a time). There may be some other problems with this as well - i.e. when future that asks for best block is polled/completes after future that asks for best block, ...

svyatonik:
I'm also getting errors like this when starting sub2eth synchronization:
```
Bridge: [2020-06-19 15:50:40] 2020-06-19 15:50:40 +0300 DEBUG bridge Received best known header from Ethereum: HeaderId(0, 0x0000000000000000000000000000000000000000000000000000000000000000)

Bridge: [2020-06-19 15:50:40] 2020-06-19 15:50:40 +0300 DEBUG bridge Substrate header HeaderId(0, 0x0000000000000000000000000000000000000000000000000000000000000000) is now Synced

Bridge: [2020-06-19 15:50:40] 2020-06-19 15:50:40 +0300 DEBUG bridge Asking Ethereum about incomplete headers

Bridge: [2020-06-19 15:50:40] 2020-06-19 15:50:40 +0300 DEBUG bridge Going to download new header from Substrate node: 1

Bridge: [2020-06-19 15:50:40] 2020-06-19 15:50:40 +0300 DEBUG bridge Going to download new header from Substrate node: 1

Bridge: [2020-06-19 15:50:40] 2020-06-19 15:50:40 +0300 DEBUG bridge Queueing new Substrate header: HeaderId(1, 0x8b46ac7f41ed642f51f0f68f885b5a0a7992348a8eba9cefa8d8a6c9e64dc0a8). Queue: MaybeOrphan.

Bridge: [2020-06-19 15:50:40] 2020-06-19 15:50:40 +0300 DEBUG bridge Asking Ethereum node for existence of: HeaderId(0, 0xe91918eabe75c066a6c5e8455b4cd1f6e9bdb6aeef0c16d22d18f3ba41d94238)

Bridge: [2020-06-19 15:50:40] 2020-06-19 15:50:40 +0300 DEBUG bridge Going to download new header from Substrate node: 2

Bridge: [2020-06-19 15:50:40] 2020-06-19 15:50:40 +0300 DEBUG bridge Substrate header HeaderId(1, 0x8b46ac7f41ed642f51f0f68f885b5a0a7992348a8eba9cefa8d8a6c9e64dc0a8) is now Orphan

Bridge: [2020-06-19 15:50:40] 2020-06-19 15:50:40 +0300 ERROR bridge Misconfiguration. Genesis Substrate header is considered orphan by Ethereum node

```

As you could see, ^^^ PoA bridge contract is thinking that the genesis header of Substrate is `HeaderId(0, 0x0000000000000000000000000000000000000000000000000000000000000000)`. So there's some issue either with deploying contract, or calling methods. I'm trying to debug now

svyatonik:
OK, my bad - I have built PoA client from invalid branch, so disregard my last comment about failing sub2eth :)

relays/ethereum/src/ethereum_deploy_contract.rs
```diff
@@ -14,12 +14,19 @@
...
+};
+use crate::rpc::SubstrateRpc;
+use crate::substrate_client::{SubstrateConnectionParams, SubstrateRpcClient};
 use crate::substrate_types::{Hash as SubstrateHash, Header as SubstrateHeader};
+
 use codec::{Decode, Encode};
 use num_traits::Zero;
 
+// to encode/decode contract calls
+ethabi_contract::use_contract!(bridge_contract, "res/substrate-bridge-abi.json");
```
svyatonik, line 28: It'll generate 2x more code (with incompatible types?). Probably only include once?

relays/ethereum/src/substrate_client.rs
HCastano: So I wasn't able to use a `RefCell` because `Send/Sync` is required. I wasn't sure how to use `lazy_static` for this, so I didn't go that way either. I couldn't quite do what Slava suggested, since to get the genesis hash I would need an instance of `SubstrateRpcClient`, which I couldn't have in the constructor.

I added a helper instead, `SubstrateRpcClient::set_genesis_hash()`, which we call after we've created a client and used it to get the genesis hash. I do have some doubts with this approach, since I'm not sure how to handle the case where we don't get a genesis hash back (maybe the RPC connection's been interrupted).

One thing we can do is continue, but then every time we call `SubmitEthereumHeader::submit_signed_ethereum_headers()` we'll be doing an extra call to get the genesis hash, which isn't ideal. Another thing (what's in the code now) is we can just fail right there, but this also doesn't seem nice. What do you guys think?

tomusdrw: > I would need an instance of SubstrateRpcClient, which I couldn't have in the constructor.

What do you mean by that? It's not Java, `SubstrateRpcClient::new` is simply a builder function so it ends with you returning the instance.
I think this should definitely go to `SubstrateRpcClient::new`, but of course it is now required to be `async` - now the logic is duplicated in multiple places.
I think we should also return a `Result<>` and let the caller handle the error instead of panicking. The node might be down for instance, so the caller might choose to retry instead.

HCastano: What I mean is that in order to get the genesis block, I need to call `SubstrateRpcClient::block_hash_by_number()`, but since `new()` doesn't take `&self` I'm not able to do that

relays/ethereum/src/ethereum_deploy_contract.rs
```diff
@@ -59,18 +63,21 @@ pub fn run(params: EthereumDeployContractParams) {
 	let mut local_pool = futures::executor::LocalPool::new();
 
 	let result = local_pool.run_until(async move {
-		let eth_client = ethereum_client::client(params.eth);
-		let sub_client = substrate_client::client(params.sub);
+		let eth_client = EthereumRpcClient::new(params.eth);
+		let sub_client = SubstrateRpcClient::new(params.sub).await;
+
+		let sub_client = match sub_client {
```
tomusdrw, line 69: Doesn't `?` work?

HCastano: Derp, this match was the whole reason I went to the trouble of implementing `String` related traits for `RpcError`

Issue #111, opened 2020-05-27
Improve Test Readability


HCastano:
We should strive to make tests more readable by abstracting common setup/teardown stuff into helper functions. We should also try and be explicit about what is being checked in a test with the use of intermediate variables. There's some room for improvement in places like `modules/ethereum/src/{finality, import}.rs`, but we should be checking the entire codebase.

Clean tests are not limited to the things I've suggested here, but it's just what I've noticed we could work on with the existing tests.

tomusdrw:
Closing, since it has been here for a while and since the PoA stuff is pretty low priority now, I don't expect this being worked on any time soon.

PR #136, opened 2020-06-15
Runtime benchmarks: start

svyatonik:
part of #78 

So the plan for computing final weights is:
1) plug-in 'benchmarking framework' - that's what this PR does (+also introduce simplest possible benchmark);
2) add more benchmarks in follow-up PR(s);
3) execute benchmarks on reference machine and propose final weights in separate PR;
4) tune (or make configurable) dependent constants in relay (and probably in runtime modules) according to final weights.

WIP, because I need to clean up + use single set of functions for both tests and benchmarks.

### DIFF
--- a/modules/ethereum/src/lib.rs
+++ b/modules/ethereum/src/lib.rs
@@ -43,6 +43,9 @@ mod benchmarking;
 #[cfg(test)]
 mod mock;
 
+#[cfg(any(feature = "runtime-benchmarks", test))]
+mod test_utils;
+
 /// Maximal number of blocks we're pruning in single import call.
 const MAX_BLOCKS_TO_PRUNE_IN_SINGLE_IMPORT: u64 = 8;
 
@@ -844,19 +847,25 @@ pub(crate) fn initialize_storage<T: Trait>(
 		oldest_block_to_keep: initial_header.number,
 	});
 	HeadersByNumber::insert(initial_header.number, vec![initial_hash]);
-	Headers::<T>::insert(initial_hash, StoredHeader {
-		submitter: None,
-		header: initial_header.clone(),
-		total_difficulty: initial_difficulty,
-		next_validators_set_id: 0,
-		last_signal_block: None,
-	});
+	Headers::<T>::insert(
+		initial_hash,
+		StoredHeader {
+			submitter: None,
+			header: initial_header.clone(),
+			total_difficulty: initial_difficulty,
+			next_validators_set_id: 0,
+			last_signal_block: None,
+		},
+	);
 	NextValidatorsSetId::put(1);
-	ValidatorsSets::insert(0, ValidatorsSet {
-		validators: initial_validators.to_vec(),
-		signal_block: None,
-		enact_block: initial_id,
-	});
+	ValidatorsSets::insert(
+		0,
+		ValidatorsSet {
+			validators: initial_validators.to_vec(),
+			signal_block: None,
+			enact_block: initial_id,
+		},
+	);
 	ValidatorsSetsRc::insert(0, 1);
 }
 
@@ -924,7 +933,8 @@ pub(crate) mod tests {
 	use super::*;
 	use crate::finality::FinalityAncestor;
 	use crate::mock::{
-		block_i, custom_block_i, custom_test_ext, genesis, insert_header, validators, validators_addresses, TestRuntime,
+		genesis, insert_header, run_test, run_test_with_genesis, validators_addresses, HeaderBuilder, TestRuntime,
+		GAS_LIMIT,
 	};
 	use primitives::compute_merkle_root;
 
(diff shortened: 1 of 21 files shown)


svyatonik:
[Last commit](https://github.com/paritytech/parity-bridges-common/pull/136/commits/88f5b757641de9206ba6d058daa616834c62123e) brings many changes and makes this PR hard to review. It should have been a separate PR, but sorry - just wanted to finish this asap. So the idea is that tests and benchmarking will have many in common => I had to refactor tests infrastructure a bit before using in in benchmarks. Most significant changes:
1) code that is used both by benchmarks and tests is in `test_utils.rs`;
2) tests should not reference this file directly - I'm reexporting required stuff from `mock.rs`;
3) benchmarks are executing in wasm && we want to generate headers => we can't use `parity-crpypto` for signing => switch to `libsecp256k1` that is now used both by tests and benchmarks;
4) headers may be (and should be, if possible) built by `HeaderBuilder`;
5) I've changed `custom_test_ext()` to `run_test(TestContext { validators })` - this would allow us to remove infinite calls to `validators()` and `validators_addresses()` in tests. This is a step towards #111 imo.

svyatonik:
One last thing - to run the only available benchmark, you can use script like that:
```
#!/bin/bash
set -e

cargo build --manifest-path=../parity-bridges-common/bin/node/node/Cargo.toml --target-dir=../parity-bridges-common/target --release --features runtime-benchmarks
cp ../parity-bridges-common/target/release/bridge-node .

RUST_LOG=runtime=trace ./bridge-node benchmark --pallet bridge-eth-poa --extrinsic import_unsigned_header_best_case
```

modules/ethereum/src/lib.rs
```diff
@@ -1225,7 +1261,7 @@ pub(crate) mod tests {
 
 	#[test]
 	fn verify_transaction_finalized_works_for_best_finalized_header_ancestor() {
-		custom_test_ext(genesis(), validators_addresses(3)).execute_with(|| {
+		run_test(3, |_| {
```
tomusdrw, line 1264: The number passed to `run_test` is a bit cryptic. Maybe we could extract it to a constant to give it some more meaning, also given that it repeats for mulitple tests.

modules/ethereum/src/verification.rs
```diff
@@ -282,7 +289,7 @@ fn validator_checks(
 
 /// Returns expected number of seal fields in the header.
 fn expected_header_seal_fields(config: &AuraConfiguration, header: &Header) -> usize {
-	if header.number >= config.empty_steps_transition {
+	if header.number != u64::max_value() && header.number >= config.empty_steps_transition {
```
tomusdrw, line 292: is `u64::max_value()` used a special marker or sth?

HCastano:
This isn't running on "standard" hardware, but I got numbers!

```
Pallet: "bridge-eth-poa", Extrinsic: "import_unsigned_header_best_case", Lowest values: [], Highest values: [], Steps: [], Repeat: 1
Median Slopes Analysis
========

Model:
Time ~=    205.6
    + n        0
              µs

Min Squares Analysis
========

Data points distribution:
    n   mean µs  sigma µs       %
    1     233.1         0    0.0%
  100     206.5         0    0.0%
  199     200.6         0    0.0%
  298     203.5         0    0.0%
  397     200.4         0    0.0%
  496     200.6         0    0.0%
  595       201         0    0.0%
  694     201.4         0    0.0%
  793     197.3         0    0.0%
  892     199.5         0    0.0%
  991     200.9         0    0.0%

Quality and confidence:
param     error
n         0.007

Model:
Time ~=    213.3
    + n        0
              µs
```

primitives/ethereum-poa/src/lib.rs
```diff
@@ -303,6 +324,7 @@ impl SealedEmptyStep {
 	}
 
 	/// Returns rlp for the vector of empty steps (we only do encoding in tests).
+	// TODO: Do we really need an entire feature flag for one method?
```
HCastano, line 327: I added a few more things under this feature flag, so it might be sticking around

tomusdrw, line 327:
```suggestion
```

modules/ethereum/src/test_utils.rs
```diff
@@ -0,0 +1,223 @@
...
+
+use primitives::{rlp_encode, secret_to_address, sign, Address, Bloom, Header, SealedEmptyStep, H256, U256};
+use secp256k1::SecretKey;
+use sp_std::prelude::*;
+
+/// Gas limit valid in test environment.
+pub const GAS_LIMIT: u64 = 0x2000;
+
+/// Test header builder.
+// TODO: I think this needs to be behind a testing + bench feature flag
```
tomusdrw, line 27: Please address before merging.

primitives/ethereum-poa/src/lib.rs
```diff
@@ -457,6 +498,13 @@ pub fn compute_merkle_root<T: AsRef<[u8]>>(items: impl Iterator<Item = T>) -> H2
 	triehash::ordered_trie_root::<Keccak256Hasher, _>(items)
 }
 
+// TODO: Move elsewhere? I move it here since I want the flow of imports to be from primitives ->
```
tomusdrw, line 501:
```suggestion
```

It's fine here.

primitives/ethereum-poa/src/lib.rs
```diff
@@ -436,6 +458,25 @@ pub fn public_to_address(public: &[u8; 64]) -> Address {
 	result
 }
 
+/// Return author's signature over given message.
+#[cfg(feature = "test-helpers")]
```
tomusdrw, line 462: Instead of adding a bunch of feature-gated methods to `src/lib.rs` you can move everything to a separate module and feature-gate it.

That will require decoupling `sign_by` and `sign_by_set` from `Header`, but it's okay imho. If really needed we can extend `Header` by adding an extension trait.

modules/ethereum/src/verification.rs
HCastano: ¯\\_(ツ)_/¯

PR #140, opened 2020-06-18
Update web3 dependency.

tomusdrw:


HCastano:
I'd prefer if we used a released version of `rust-web3` instead of `master`

tomusdrw:
Published & updated :)

PR #149, opened 2020-06-25
Benchmark Ethereum Pallet

HCastano:
The goal of this PR is to add benchmarks to the Ethereum bridge pallet. There are only two extrinsics we care about here: `import_unsigned_headers()` and `import_signed_headers()`. For each of these extrinsics we need to make sure that we cover the best and worst case run times.

There are four main operations whose performance we should check each extrinsic:
- Finalization
- Pruning
- Receipts (through authority set changes I think)
- Chain Re-Orgs

I think we should benchmark each of these individually, and combine them later if needed to find the worst case performance.

If I've understood the benchmarking correctly we want to see how the run time is affected by the complexity parameter. For finalization I think we should use the number of blocks being finalized as the complexity parameter. For pruning it _should_ be the number of blocks, but in practice it doesn't matter because we're limited to the number of blocks we can prune per import anyways (eight in our case). For receipts I'm not sure how the complexity parameter will relate to the bench. For re-orgs it would be the length of the chain being re-org'd.


svyatonik:
For receipts I think the complexity could be `sizeof(receipts)` - nothing else comes to my mind. Nothing irregular happens during reorg => it shouldn't affect complexity at all (because we don't have concept of 'canonical unfinalized chain' and reorgs could only happen in unfinalized part). One other thing that may affect performance a bit (not as much as `len(unfinalized-chain)`, though) could be number of validators in the set.

Dump of my thoughts:
1) I haven't finished reading benchmarking docs yet, but I believe that at the end, we should have some 'worst case' test (with our guesses for worst parameters values) and 'average case' test (which will be used to determine default final weights) right?
2) we may use benchmarking results (in future?) to adjust actual weight using [this suggestion](https://github.com/paritytech/parity-bridges-common/issues/78#issuecomment-620583714);
3) for finality worst case - we should guess some meaningful value that we hope won't actually happen on the PoA chain? I think that maybe we could use [this guess](https://github.com/paritytech/parity-bridges-common/blob/master/bin/node/runtime/src/kovan.rs#L23)? It is used in similar context. Wdyt?

modules/ethereum/src/benchmarking.rs
```diff
@@ -55,6 +60,163 @@ benchmarks! {
...
+		let storage = BridgeStorage::<T>::new();
+		assert_eq!(storage.best_block().0.number, 1);
+		assert_eq!(storage.finalized_block().number, 0);
+	}
+
+	// Our goal with this bench is to try and see the effect that finalizing difference ranges of
+	// blocks has on our import time. As such we need to make sure that we keep the number of
+	// validators fixed while changing the number blocks finalized (the complixity parameter) by
+	// importing the last header.
+	import_unsigned_finality {
```
svyatonik, line 72: One last thing I forgot to add to my top-level comment - imo `weight(import_signed_headers)` is actually `weight(import_unsigned_header) * len(headers) + some_constant`, where `some_constant` stands for recovering signature + iterating headers + invoking callbacks (which are noop in our runtime - not sure how to measure that in general case?). So we may only introduce benchmarks for unsigned version and use results for signed version.

modules/ethereum/src/benchmarking.rs
```diff
@@ -55,6 +60,163 @@ benchmarks! {
...
+			&initial_header,
+			initial_difficulty,
+			&initial_validators,
+		);
+
+		let mut headers = Vec::new();
+		let mut ancestry: Vec<FinalityAncestor<Option<Address>>> = Vec::new(); // Is this needed?
+		let mut parent = initial_header.clone();
+
+		// Q: Is this a sketchy way of "delaying" finality, or should I be manually editing the
```
svyatonik, line 100: Regular import of this header should fail because of 'wrong step', but since you're inserting header without actual verification, it is ok.

svyatonik:
So I read benchmarking doc and [it suggests](https://www.shawntabrizi.com/substrate-graph-benchmarks/docs/#/runtime?id=overview) to use worst case estimate in weights computation. Worst case means many unfinalized blocks for us && it may be too much for regular header import. So my suggestion is to implement point 2 from above list (i.e. use [this suggestion](https://github.com/paritytech/parity-bridges-common/issues/78#issuecomment-620583714)). The final weight would be: `regular-block-import-weight + isset(receipts.is_some()) * receipts-degrade-factor + number-of-unfinalized-headers * finalization-degrade-factor + number-of-pruned-headers * pruning-degrade-factor`.

So we need benchmarks that Hernando has originally suggested to get these factors, then everything else should be quite easy to implement.

HCastano:
I think this is ready for feedback now. One thing I want to point out is that I'm not sure if my usage of the complexity parameter is correct. For example, in the pruning and receipt benches it's not used at all. I don't see a way to use it meaningfully since both of these are fixed regardless of the number of blocks imported.

I also did a first pass for the benchmarks using a `release` build of the branch and `--steps 50 --repeat 20`. The table below summarizes the results of the "Min Squares Analysis" provided by the benchmark.

| Benchmark      | Time (µs)     |
| :------------- | :---------- |
| import_unsigned_best_case | 195.1 + 0.002
| import_unsigned_finality   | 198.7 + 12.34 |
| import_unsigned_finality_with_cache  | 251.9 + 0.243 |
| import_unsigned_pruning   | 317.7 + 0 |
| import_unsigned_with_receipts   | 206.3 + 0.054 |

I haven't included any CI changes in this PR to `build`/`test` the benchmarks. I think we should wait until some of the new CI changes are merged to master before doing that.

modules/ethereum/src/benchmarking.rs
```diff
@@ -55,6 +51,219 @@ benchmarks! {
...
+	// will contain a validator set change. Verifying the receipt root is an expensive operation to
+	// do, which is why we're interested in benchmarking it.
+	import_unsigned_with_receipts {
+		let n in 1..100;
+
+		let mut storage = BridgeStorage::<T>::new();
+
+		let num_validators = 1;
+		let initial_header = initialize_bench::<T>(num_validators as usize);
+		let receipts = vec![validators_change_receipt(Default::default())];
```
svyatonik, line 194: You could add more (say `n`) receipts here to measure how receipts size affects import time.

modules/ethereum/src/benchmarking.rs
```diff
@@ -55,6 +51,219 @@ benchmarks! {
...
+		assert_eq!(storage.finalized_block().number, num_blocks as u64);
+	}
+
+	// A block import may trigger a pruning event, which adds extra work to the import progress.
+	// In this bench we trigger a pruning event in order to see how much extra time is spent by the
+	// runtime dealing with it. In the Ethereum Pallet, we're limited pruning to eight blocks in a
+	// single import, as dictated by MAX_BLOCKS_TO_PRUNE_IN_SINGLE_IMPORT.
+	import_unsigned_pruning {
+		let n in 1..100;
+		let num_headers = 10;
```
svyatonik, line 151: Imo it would be better to make number of headers that have to be pruned variable. Then we'll actually may get single header prune 'weight'.

modules/ethereum/src/benchmarking.rs
HCastano: Ah, yeah good point!

modules/ethereum/src/benchmarking.rs
HCastano: Yeah, I guess we can have an equivalent `MAX_BLOCKS_TO_PRUNE_IN_SINGLE_IMPORT` that's only used during benches for this

svyatonik: Not exactly what I mean :) I mean that you could have something like:
```
let n in 1..MAX_BLOCKS_TO_PRUNE_IN_SINGLE_IMPORT;
...
for i in 1..n+1 {
 insert_initial_header(n);
}
mark_headers_for_prune(0..n);
...
// run extrinsic that imports next header and will prune headers [0; n]
```

So we don't need to change `MAX_BLOCKS_TO_PRUNE_IN_SINGLE_IMPORT`, but instead we only prune given number of blocks (1..MAX_BLOCKS_TO_PRUNE_IN_SINGLE_IMPORT in our case).

modules/ethereum/src/benchmarking.rs
```diff
@@ -55,6 +51,223 @@ benchmarks! {
...
+		assert_eq!(storage.finalized_block().number, num_blocks as u64);
+	}
+
+	// Basically the exact same as `import_unsigned_finality` but with a different range for the
+	// complexity parameter. In this bench we use a larger range of blocks to see how performance
+	// changes when the finality cache kicks in (>16 blocks).
+	import_unsigned_finality_with_cache {
+		// Our complexity parameter, n, will represent the number of blocks imported before
+		// finalization.
+		let n in 7..100;
```
tomusdrw, line 110: ditto

modules/ethereum/src/benchmarking.rs
```diff
@@ -55,6 +51,223 @@ benchmarks! {
...
+		assert_eq!(storage.best_block().0.number, (num_blocks + 1) as u64);
+		assert_eq!(storage.finalized_block().number, num_blocks as u64);
+	}
+
+	// A block import may trigger a pruning event, which adds extra work to the import progress.
+	// In this bench we trigger a pruning event in order to see how much extra time is spent by the
+	// runtime dealing with it. In the Ethereum Pallet, we're limited pruning to eight blocks in a
+	// single import, as dictated by MAX_BLOCKS_TO_PRUNE_IN_SINGLE_IMPORT.
+	import_unsigned_pruning {
+		let n in 1..MAX_BLOCKS_TO_PRUNE_IN_SINGLE_IMPORT as u32;
```
tomusdrw, line 150:
```suggestion
		let blocks_to_prune in 1..MAX_BLOCKS_TO_PRUNE_IN_SINGLE_IMPORT as u32;
```

modules/ethereum/src/benchmarking.rs
```diff
@@ -33,15 +36,8 @@ benchmarks! {
 	import_unsigned_header_best_case {
 		let n in 1..1000;
```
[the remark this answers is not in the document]
svyatonik: It has to be one-letter variable iirc :) That's `benchmarks` macro requirements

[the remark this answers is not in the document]
HCastano: Yeah, it's mentioned [here](https://github.com/paritytech/substrate/blob/master/frame/benchmarking/src/lib.rs#L38).

bin/node/runtime/src/lib.rs
```diff
@@ -215,6 +215,28 @@ impl pallet_aura::Trait for Runtime {
 	type AuthorityId = AuraId;
 }
 
+// We want to use a different validator configuration for benchmarking than what's used in Kovan,
+// but we can't configure a new validator set on the fly which means we need to wire the runtime
+// together like this
```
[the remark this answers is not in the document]
HCastano: Yeah, I can look into it

HCastano:
I've generated some plots. I first got the benchmarking datapoints as so:

```bash
./target/release/bridge-node benchmark --pallet bridge-eth-poa --extrinsic $extrinsic --steps 50 --repeat 20 --raw > $extrinsic.txt
```

Then then I pasted the contents of the file into [here](https://www.shawntabrizi.com/substrate-graph-benchmarks/).

### Import Unsigned Header

import_unsigned_header_best_case

### Import Unsigned Finality
import_unsigned_finality

### Import Unsigned Finality with Cache
import_unsigned_finality_with_cache

### Import Unsigned Pruning
import_unsigned_pruning

### Import Unsigned Receipts
import_unsigned_with_receipts


The finality cache plot is my personal favourite, lol.

tomusdrw:
The plots looks good! So it seems that either things are constant (or at least amortized) or linear, which is really good. Can we plot our formulas on top of this to make sure we have an upper bound? Then let's get this merged and deployed :)

bin/node/runtime/src/lib.rs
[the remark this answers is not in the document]
HCastano: Looks like `ValidatorsConfiguration` has to be configured from the runtime, so I don't think we can address this right now

PR #157, opened 2020-06-30
Create accounts from DepositInto::deposit_into()

svyatonik:
actually resolves [this comment](https://github.com/paritytech/parity-bridges-common/pull/91/files#r430005788) for our test runtime

### DIFF
--- a/modules/currency-exchange/src/lib.rs
+++ b/modules/currency-exchange/src/lib.rs
@@ -83,6 +83,8 @@ decl_error! {
 		FailedToConvertCurrency,
 		/// Deposit has failed.
 		DepositFailed,
+		/// Deposit has partially failed (changes to recipient acount were made).
+		DepositPartiallyFailed,
 		/// Transaction is not finalized.
 		UnfinalizedTransaction,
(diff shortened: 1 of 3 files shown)


modules/currency-exchange/src/lib.rs
```diff
@@ -83,6 +83,8 @@ decl_error! {
 		FailedToConvertCurrency,
 		/// Deposit has failed.
 		DepositFailed,
+		/// Deposit has partially failed (changes to recipient acount were made).
```
tomusdrw, line 86:
```suggestion
		/// Deposit has partially failed (changes to recipient account were made).
```

primitives/currency-exchange/src/lib.rs
```diff
@@ -35,6 +35,8 @@ pub enum Error {
 	FailedToConvertCurrency,
 	/// Deposit has failed.
 	DepositFailed,
+	/// Deposit has partially failed (changes to recipient acount were made).
```
tomusdrw, line 38:
```suggestion
	/// Deposit has partially failed (changes to recipient account were made).
```

PR #158, opened 2020-07-01
Exchange pallet benchmarks

svyatonik:
WIP: should be rebased on master once #155 is reviewed + merged
part of #78

The main difficulty with benchmarking this pallet is that it is too generic && I was unable to write any runtime specific code there. So there's extended `Trait` and `fn Trait::make_proof()`, which asks runtime to prepare environment+proof for executing import extrinsic.

There are now 5 benchmarks there:
1) `import_peer_transaction_best_case` - imports minimal proof (transaction itself), transaction has minimal size (see below) and recipient account exists;
2) `import_peer_transaction_when_recipient_does_not_exists` - the same as (1), but recipient account is created during import;
3) `import_peer_transaction_when_transaction_size_increases` - test runtime only accepts transactions of minimal size (where data field is 32-bytes pub key of Substrate' recipient), so this is the same as (1);
4) `import_peer_transaction_when_proof_size_increases` - the same as (1), but proof size changes from 1 to 1000. Proof size here = number of transactions in PoA block. So worst case is 1000 txs/block. If you think we should have another value for that - please share your thoughts;
5) `import_peer_transaction_worst_case` - recipient account does not exists, proof size is 1000 transactions, every transaction has maximal size (though in test runtime it it the same minimal size).


svyatonik:
(rebased on master. will leave it open until monday)

modules/ethereum/src/test_utils.rs
```diff
@@ -206,6 +223,20 @@ where
 	custom_header.sign_by(author)
 }
 
+/// Insert unverified header into storage.
+pub fn insert_header<S: Storage>(storage: &mut S, header: Header) {
```
HCastano, line 227: I made the same move in my PR, lol

primitives/ethereum-poa/src/lib.rs
```diff
@@ -117,9 +117,9 @@ pub struct UnsignedTransaction {
 	pub gas: U256,
 	/// Transaction destination address. None if it is contract creation transaction.
 	pub to: Option<Address>,
-	/// Transaction value.
+	/// Value.
```
HCastano, line 120: This doesn't really add much value in terms of describing what "value" is. I'd beef up the description a bit

primitives/ethereum-poa/src/lib.rs
```diff
@@ -117,9 +117,9 @@ pub struct UnsignedTransaction {
...
 	pub value: U256,
-	/// Transaction payload.
+	/// Associated data.
```
HCastano, line 122: Same here

bin/node/runtime/src/lib.rs
```diff
@@ -615,7 +615,47 @@ impl_runtime_apis! {
...
+						<Runtime as pallet_bridge_currency_exchange::Trait>::DepositInto::deposit_into(
+							proof_params.recipient.clone(),
+							ExistentialDeposit::get(),
+						).unwrap();
+					}
+
+					let transaction = crate::exchange::prepare_ethereum_transaction(
+						&proof_params.recipient,
+						|tx| {
+							// our runtime only supports transactions where data is exactly 32 bytes long
```
HCastano, line 640: In the PR description you elaborate that this is because of Substrate pubkeys, I think it's worth mentioning that in this comment as well

bin/node/runtime/src/lib.rs
```diff
@@ -615,7 +615,47 @@ impl_runtime_apis! {
...
+					crate::exchange::EthereumTransactionInclusionProof {
+						block: block_hash,
+						index: 0,
+						proof: transactions,
+					}
+				}
+			}
+
 			add_benchmark!(params, batches, b"bridge-eth-poa", BridgeEthPoA);
+			add_benchmark!(params, batches, b"bridge-currency-exchange", BridgeCurrencyExchangeBench::<Runtime>);
```
HCastano, line 658: Not relevant for this PR, but there was recently a change introduced in Substrate that changed the arguments to this macro. We'll need to update that next time we bump Substrate

modules/currency-exchange/src/benchmarking.rs
```diff
@@ -0,0 +1,134 @@
...
+
+use super::{Blockchain, Call, Module as CurrencyExchangeModule, Trait as CurrencyExchangeTrait};
+use sp_std::prelude::*;
+
+use frame_benchmarking::{account, benchmarks};
+use frame_system::RawOrigin;
+
+const SEED: u32 = 0;
+const WORST_TX_SIZE_FACTOR: u32 = 1000;
+const WORST_PROOF_SIZE_FACTOR: u32 = 1000;
```
HCastano, line 29: I think 1000 is fine, we just need to see how the time to run the extrinsic scales as this increases anyways

HCastano:
There do seem to be some problems with the CI (it doesn't look like they're caused by this PR). Should we wait until those are addressed to merge this?

modules/currency-exchange/src/benchmarking.rs
```diff
@@ -0,0 +1,134 @@
...
+			proof_size_factor: n,
+		});
+	}: import_peer_transaction(RawOrigin::Signed(recipient), proof)
+
+	// Benchmark `import_peer_transaction` extrinsic with the worst possible conditions:
+	// * Proof is large.
+	// * Transaction has large size.
+	// * Recipient account does not exists.
+	import_peer_transaction_worst_case {
+		let i in 1..100;
```
tomusdrw, line 121: Can we name the one-letter variables to something more descriptive? The code now is pretty small, but if it grows over time it might get harder to read.

svyatonik: Responded [here](https://github.com/paritytech/parity-bridges-common/pull/149#discussion_r450844218)

Issue #162, opened 2020-07-02
Rework Relay CLI


tomusdrw:
Currently the CLI is build with `clap` and the way we handle params has imho two disadvantages:
1. There is a bunch of back-and-forth and stringifyied parameters between `cli.yml` and `main.rs`
2. We start with `default` values of all parameters and only override them if they are  provided in the CLI. Some parameters however should be mandatory (like `eth-contract` address). The defaults are configured for `--dev` chain, but to be fair I'd prefer to have them required instead of learning one-by-one which are misconfigured.

I'd suggest to rewrite the CLI to `structopt`.

svyatonik:
(I'm currently adding another (auto) mode for tx proofs relay)

svyatonik:
OK - it seems that I've misclicked. I'm actually working on #164 for last couple of days :/

tomusdrw:
Closing this, cause it's only the ethereum relay now using `clap`, which is much lower priority.

Issue #154, opened 2020-06-30
Relay healthcheck endpoint


tomusdrw:
Would be good if relay was able to report it's health. This is useful for dockerized deployments or simply for monitoring.

Related: #141

tomusdrw:
Changing milestone to "Some Day Maybe" since the prometheus endpoint should be sufficient for now.

tomusdrw:
I don't think we need anything more, it seems to work just fine with prometheus endpoint.

PR #163, opened 2020-07-02
TestPoa <> Bridge Node test network setup

tomusdrw:
Still working on the Sub->PoA relay.


relays/ethereum/src/cli.yml
```diff
@@ -112,13 +126,16 @@ subcommands:
...
                 value_name: SUB_AUTHORITIES_SET
                 help: Encoded initial GRANDPA authorities set.
                 takes_value: true
+                required: true
             - sub-initial-header:
                 long: sub-initial-header
                 value_name: SUB_INITIAL_HEADER
                 help: Encoded initial Substrate header.
                 takes_value: true
+                required: true
```
svyatonik, line 141: I don't think it should be required - deploy is able to read it from the substrate node.

relays/ethereum/src/cli.yml
```diff
@@ -112,13 +126,16 @@ subcommands:
                 value_name: SUB_AUTHORITIES_SET_ID
                 help: ID of initial GRANDPA authorities set.
                 takes_value: true
+                required: true
             - sub-authorities-set:
                 long: sub-authorities-set
                 value_name: SUB_AUTHORITIES_SET
                 help: Encoded initial GRANDPA authorities set.
                 takes_value: true
+                required: true
```
svyatonik, line 135: I don't think it should be required - deploy is able to read it from the substrate node.

relays/ethereum/src/cli.yml
```diff
@@ -112,13 +126,16 @@ subcommands:
                 value_name: SUB_AUTHORITIES_SET_ID
                 help: ID of initial GRANDPA authorities set.
                 takes_value: true
+                required: true
```
svyatonik, line 129: I don't think it should be required - it will be zero fro most of deployments.

relays/ethereum/src/cli.yml
```diff
@@ -61,10 +62,17 @@ subcommands:
                 value_name: ETH_CONTRACT
                 help: Address of deployed bridge contract.
                 takes_value: true
+                required: true
+            - eth-chain-id:
+                long: eth-chain-id
+                value_name: ETH_CHAIN_ID
+                help: Chain ID to use for signing.
+                required: true
```
svyatonik, line 70: Making `eth-chain-id` required without making e.g. `eth-signer` required seems strange to me. Let's either mark both `required` or make both work with OE instant seal chain by default. I would prefer to leave everything with default values for now so that we'll be able to work with our Substrate runtime and OE instant-seal network by default. We may revisit cli args in #162 later - this PR makes cli args a bit inconsistent.

tomusdrw:
@svyatonik Sorry about the CLI parameters - that's a good catch. I was marking them as required, cause I didn't check that the defaults are configured for `InstantSeal` engine. It wasn't working with a custom chain, so I assumed the default are simply `0x0` and that they are rather required every time anyway.

deployments/README.md
```diff
@@ -0,0 +1,75 @@
...
+To kill all left over containers and start the network from scratch next time:
+```bash
+docker ps -a --format "{{.ID}}" | xargs docker rm # This removes all containers!
+```
+
+To configure the frontend
+
+## Kovan -> Westend
+
+???
```
HCastano, line 70:
```suggestion
👷 Under Construction 👷‍♀️
```

deployments/README.md
```diff
@@ -0,0 +1,75 @@
...
+### Polkadot.js UI
+
+To teach the UI decode our custom types used in the pallet, go to: `Settings -> Developer`
+and import the [`./types.json`](./types.json)
+
+## TestPoa
+
+`TestPoa` is a test bridge deployment between a test Ethereum PoA network and
+test Substrate network.
+It's main purpose is to make sure that basic PoA<>Substrate bridge operation works.
```
HCastano, line 35:
```suggestion
Its main purpose is to make sure that basic PoA<>Substrate bridge operation works.
```

deployments/README.md
```diff
@@ -0,0 +1,75 @@
...
+The network is being reset every now and then without a warning.
+
+### UI
+
+Use [wss://testpoa.bridges.test-installations.parity.io/](https://polkadot.js.org/apps/)
+as a custom endpoint for [https://polkadot.js.org/apps/](https://polkadot.js.org/apps/).
+
+### How to run locally?
```
HCastano, line 43: I think this should come before the UI stuff, probably right after the `Requirements` section

deployments/testpoa/OpenEthereum.Dockerfile
```diff
@@ -0,0 +1,84 @@
+FROM ubuntu:xenial AS builder
+
+# show backtraces
+ENV RUST_BACKTRACE 1
+
+ENV LAST_DEPS_UPDATE 2020-06-19
```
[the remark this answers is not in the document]
tomusdrw: This is left as a convenient way to re-build the image from scratch. Changing `ENV` invalidates docker cache. In my previous projects I sometimes had issues with image failing to build on fresh installations, but working fine on CI, etc. So having some way clear the cache easily was needed, and just adding the date of last "forced" update was decided to be the best way to handle that.

deployments/testpoa/OpenEthereum.Dockerfile
```diff
@@ -0,0 +1,84 @@
...
+COPY --chown=openethereum:openethereum --from=builder /openethereum/target/release/openethereum ./
+# Solve issues with custom --keys-path
+RUN mkdir -p ~/.local/share/io.parity.ethereum/keys/
+# check if executable works in this container
+RUN ./openethereum --version
+
+EXPOSE 8545 8546 30303/tcp 30303/udp
+
+HEALTHCHECK --interval=2m --timeout=5s \
+  CMD curl -f http://localhost:8545/api/health || exit 1
```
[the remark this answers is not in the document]
tomusdrw: `HEALTHCHECK` is better done in the `Dockerifle`, but the top-level one is being used for both `bridge-node` and `ethereum-poa-relay`, and since there is no good way to add `HEALTHCHECK` conditionally (depending on the project being build) I went with adding the health check on the compose file level.
When #154 is resolved we can conditionally change the port and move healthcheck to the `Dockerfile` then.

deployments/testpoa/docker-compose.yml
```diff
@@ -0,0 +1,147 @@
...
+      - "10044:9944"
+
+  bridge-node-charlie:
+    <<: *bridge-node
+    environment:
+      <<: *bridge-env
+      VIRTUAL_HOST: testpoa.bridges.test-installations.parity.io
+      VIRTUAL_PORT: 9944
+      LETSENCRYPT_HOST: testpoa.bridges.test-installations.parity.io
+      LETSENCRYPT_EMAIL: admin@parity.io
```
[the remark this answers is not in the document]
tomusdrw: We only want to expose one node, not all of them. In theory `nginx` can load-balance between different nodes, but I've made each node responsible only for one thing, to make it easier to check the logs:
- Alice is for `eth2sub`
- Bob for `sub2eth`
- Charlie for external RPC.

deployments/testpoa/docker-compose.yml
```diff
@@ -0,0 +1,147 @@
...
+  #   volumes:
+  #     - ./bridge-config:/config
+  #   environment:
+  #     RUST_LOG: rpc=trace,bridge=trace
+  #   depends_on: *all-nodes
+
+  bridge-node-alice: &bridge-node
+    build:
+      context: .
+      dockerfile: ../../Dockerfile
```
HCastano, line 85: This path didn't work out of the box for me. I kept getting the following error

```
Building bridge-node-alice
ERROR: Forbidden path outside the build context: ../../Dockerfile ()
```

It was only after changing it to

```yaml
context: ../../
dockerfile: ./Dockerfile
```

that I managed to get these built.

deployments/testpoa/docker-compose.yml
```diff
@@ -0,0 +1,147 @@
...
+      - --engine-signer=0x004e7a39907f090e19b0b80a277e77b72b22e269
+    ports:
+      - "8745:8545"
+      - "8746:8546"
+      - "32303:30303"
+
+  relay-eth2sub:
+    build:
+      context: .
+      dockerfile: ../../Dockerfile
```
HCastano, line 51: Same problem here

deployments/testpoa/docker-compose.yml
tomusdrw: I rather duplicated the file, having the entire repo in a context for that specific deployment is a big wasteful. The top-level `Dockerfile` is now using local files instead of checking out the repo from GH.

HCastano: I think that's reasonable!

PR #156, opened 2020-06-30
Add gitlab build config.

tomusdrw:
Marking as draft, waiting for the devops team to review & setup gitlab sync.

PR #166, opened 2020-07-02
Add cargo deny

General-Beck:


General-Beck:
it works. But I think it would be much more convenient if we had a file with deny.toml rules. In this case, we would get a more demonstrative report.

HCastano:
> lgtm, seems the build fails though due to incompatible compiler version & cache.
> We can add the `deny.toml` file later on.

Why don't we add the `deny.toml` file that was part of #156 in this PR?

General-Beck:
Added, [checked](https://github.com/paritytech/parity-bridges-common/actions/runs/156758328)
The first option is to create a log file, the second is actions from Embark (I like the second option)

tomusdrw:
I think it'd be sensible to whitelist `substrate` as a git dependency for now and also ignore the entire substrate sub-tree for warnings and security notifications. Main reasoning being that substrate is not stable enough to rely on it yet. And I really don't want the bridges PRs being blocked by some upstream fixes in substrate.

I can take a look at implementing this later today.

tomusdrw:
Thanks @General-Beck

PR #167, opened 2020-07-02
Add CI cargo build

General-Beck:


.github/workflows/build.yml
```diff
@@ -0,0 +1,126 @@
+name:                  Build Suite
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+      - stable
```
HCastano, line 8: We don't have a `stable` branch

.github/workflows/build.yml
```diff
@@ -0,0 +1,126 @@
...
+    tags:
+      - v*
+    paths-ignore:
```
HCastano, line 11: I don't think we should run if anything inside `diagrams/` is changed either

.github/workflows/build.yml
```diff
@@ -0,0 +1,126 @@
...
+        run:  |
+          echo "::set-env name=RUSTC_WRAPPER::sccache"
+          echo "::set-env name=CMAKE_C_COMPILER_LAUNCHER::sccache"
+          echo "::set-env name=CMAKE_CXX_COMPILER_LAUNCHER::sccache"
+      - name:          Building bridge node ${{ matrix.platform }} rust-${{ matrix.toolchain }}
+        uses:          actions-rs/cargo@master
+        with:
+          command:     build
+          toolchain:   ${{ matrix.toolchain }}
+          args:        --release --verbose -p bridge-node
```
HCastano, line 112: I don't think we should be building this in release mode, it's just going to make build times longer

.github/workflows/build.yml
```diff
@@ -0,0 +1,126 @@
...
+      - name:          Building relayer ${{ matrix.platform }} rust-${{ matrix.toolchain }}
+        uses:          actions-rs/cargo@master
+        with:
+          command:     build
+          toolchain:   ${{ matrix.toolchain }}
+          args:        --release --verbose -p bridge-poa-relay
```
HCastano, line 118: Same here

.github/workflows/build.yml
```diff
@@ -0,0 +1,126 @@
...
+      - name:          Stop sccache
+        if:            always()
+        run:           sccache --stop-server
+      - name:          Upload artifact's
```
HCastano, line 122: Oh, is this why we're building in release mode? I think we should only build in release mode and post artifacts if we trigger a release, but not for everyday development

HCastano:
I'd prefer to have the CI build using `cargo build --all` than specifying individual packages

.github/workflows/build.yml
HCastano: This should be `ethereum-poa-relay`, but I don't think we should be building individual packages

.github/workflows/build.yml
tomusdrw: But the whole point of this step is to prepare ready-to-use binaries. I don't think `debug` binaries would be any useful. I'm voting for `--release` here.

.github/workflows/build.yml
tomusdrw: Yeah, that build only runs on `master` and `v*` tags,  so it's for releases and maybe `nightly` builds.

HCastano:
Looks like everything is passing now, great!

About Clippy, we're getting a ton of warnings. Is there any way we can indicate that the Clippy step failed without failing the whole build? It would be nice to have _some_ indication that we have warnings so we can address them

.github/workflows/rust.yml
```diff
@@ -1,121 +1,88 @@
...
+        if:                        always()
+        run:                       sccache --show-stats
+      - name:                      Prepare artifact's
+        run:                       |
+          mkdir -p ./artifacts
+          mv -v $CARGO_TARGET_DIR/release/bridge-node ./artifacts/
+          mv -v $CARGO_TARGET_DIR/release/ethereum-poa-relay ./artifacts/
+          mv -v $CARGO_TARGET_DIR/release/substrate-bridge ./artifacts/
+        shell:                     bash
+      - name:                      Upload artifact's
```
HCastano, line 84:
```suggestion
      - name:                      Upload artifacts
```

.github/workflows/rust.yml
```diff
@@ -1,121 +1,88 @@
...
-          args:        --all -- --check
-      - name:          Testing ${{ matrix.platform }} rust-${{ matrix.toolchain }}
-        uses:          actions-rs/cargo@master
+          command:                 build
+          toolchain:               ${{ matrix.toolchain }}
+          args:                    --all --release --verbose
+      - name:                      Stats sccache
+        if:                        always()
+        run:                       sccache --show-stats
+      - name:                      Prepare artifact's
```
HCastano, line 77:
```suggestion
      - name:                      Prepare artifacts
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,49 @@
+name:                              Check style
+
+on:
+# Uncomment when needed
```
HCastano, line 4: So we wouldn't run the RustFmt and Clippy stages unless this is uncommented, right?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,49 @@
...
+#    paths-ignore:
+#      - 'README.md'
+#      - diagrams/
+  schedule:                        #Nightly build
+    - cron:                        '0 0 * * *'
+jobs:
+
+## Check stage
+  check-style:
+    name:                          Check format and clippy
```
HCastano, line 20:
```suggestion
    name:                          Check RustFmt and Clippy
```

.github/workflows/rust.yml
```diff
@@ -1,121 +1,88 @@
-name:                  Test Suite
+name:                              Check, test and build Suite
```
HCastano, line 1:
```suggestion
name:                              Compilation and Testing Suite
```

This one's a bit more of a nitpick since there's a stage called "Check, test and build" and as an Actions newbie I want to differentiate between them

.github/workflows/lint.yml
General-Beck: Yes, only the night build, but it will be possible to configure it further

tomusdrw: Can we split `rust-fmt` and `clippy` and run `fmt` on every push & pr?

`clippy` is not part of our workflow yet, but `fmt` is crucial to be checked on every PR.

.github/workflows/rust.yml
```diff
@@ -1,121 +1,88 @@
...
   push:
     branches:
       - master
-      - stable
     tags:
       - v*
     paths-ignore:
       - 'README.md'
+      - diagrams/
+  schedule:                        #Nightly build
```
tomusdrw, line 13: I think it would be better to have weekly builds instead (even tagged and published on Github), we don't move fast enough for daily builds.

PR #172, opened 2020-07-03
Fix sub2eth synchronization

svyatonik:
- [x] - [this commit](https://github.com/paritytech/parity-bridges-common/commit/f7069ef2db6314a45f28303ef65a534217fec86d) makes it easy to compile openethereum with builtin;
- [x] - [this commit](https://github.com/paritytech/parity-bridges-common/commit/7bb68e28e62488c4630a2dac821f4c90762bc20a) is fixing completion notifications receival. Otherwise synchronization stucks at first header that requires finality proof;
- [x] - [this commit](https://github.com/paritytech/parity-bridges-common/pull/172/commits/17cff7d5bee3c5ed4ef8c917675fd000914460ff) fixes situation when we keep submitting PoA TX with header H when txpool has transaction that imports incomplete H-1. Now we ask target node if headers is incomplete **before submitting it**. This means that we won't submit transactions that fail with `Revert` anymore. This is accompanied with [this](https://github.com/svyatonik/substrate-bridge-sol/commit/1d0fa475a2ba3a70a47ed2dd870568c42ec16c8c) solidity contract update
- [x] - [this commit](https://github.com/paritytech/parity-bridges-common/pull/172/commits/fe8adfe2ac74d0d8e9b01528a32ee426c5c94b45) fixes situation when: (1) best block on target node that we know of is N (2) we have recently submitted headers N+1,N+2,N+3,N+4 (3) headers queue is empty (4) relay keeps asking source node about block N until best target node' block is updated.


relays/ethereum/src/ethereum_client.rs
```diff
@@ -246,25 +246,35 @@ impl EthereumHighLevelRpc for EthereumRpcClient {
...
+					submitted: Vec::new(),
+					incomplete: Vec::new(),
+					rejected: headers.iter().rev().map(|header| header.id()).collect(),
+					fatal_error: Some(error),
+				}
+			}
+		};
 
-		Ok(ids)
+		// submit headers. Note that w're cloning self here. It is ok, because
```
tomusdrw, line 264:
```suggestion
		// submit headers. Note that we're cloning self here. It is ok, because
```

relays/ethereum/src/ethereum_client.rs
```diff
@@ -363,3 +373,201 @@ impl EthereumHighLevelRpc for EthereumRpcClient {
...
+
+/// Submit multiple Substrate headers.
+async fn submit_substrate_headers(
+	mut header_submitter: impl HeadersSubmitter,
+	headers: Vec<QueuedSubstrateHeader>,
+) -> SubmittedHeaders<SubstrateHeaderId, RpcError> {
+	let mut ids = headers.iter().map(|header| header.id()).collect::<VecDeque<_>>();
+	let mut submitted_headers = SubmittedHeaders::default();
+	for header in headers {
+		let id = ids.pop_front().expect("both collections have same size; qed");
```
tomusdrw, line 444: You can `.zip` `headers` and `ids` instead.

relays/ethereum/src/ethereum_client.rs
```diff
@@ -363,3 +373,201 @@ impl EthereumHighLevelRpc for EthereumRpcClient {
...
+		submit_substrate_header(&mut header_submitter, &mut submitted_headers, id, header).await;
+
+		if submitted_headers.fatal_error.is_some() {
```
tomusdrw, line 447: Why not return an `Option<Error>` from `submit_substrate_header` instead of relying on the function to mutate passed arguments?

relays/ethereum/src/substrate_client.rs
```diff
@@ -228,29 +229,55 @@ impl SubmitEthereumHeaders for SubstrateRpcClient {
...
 	async fn submit_unsigned_ethereum_headers(
 		&self,
 		headers: Vec<QueuedEthereumHeader>,
-	) -> Result<Vec<EthereumHeaderId>> {
-		let ids = headers.iter().map(|header| header.id()).collect();
+	) -> SubmittedHeaders<EthereumHeaderId, RpcError> {
+		let mut ids = headers.iter().map(|header| header.id()).collect::<VecDeque<_>>();
+		let mut submitted_headers = SubmittedHeaders::default();
 		for header in headers {
+			let id = ids.pop_front().expect("both collections have same size; qed");
```
tomusdrw, line 267: Same here, `.zip` would look nicer.

relays/ethereum/src/sync_loop.rs
```diff
@@ -288,14 +290,24 @@ pub fn run<P: HeadersSyncPipeline>(
...
-				target_submit_header_result = target_submit_header_future => {
-					target_client_is_online = process_future_result(
-						target_submit_header_result,
-						|submitted_headers| sync.headers_mut().headers_submitted(submitted_headers),
-						&mut target_go_offline_future,
-						|| async_std::task::sleep(CONNECTION_ERROR_DELAY),
-						|| format!("Error submitting headers to {} node", P::TARGET_NAME),
-					);
+				submitted_headers = target_submit_header_future => {
+					// this closure helps Rust understand the type of `submitted_headers` :/
```
tomusdrw, line 294: Can't you do:
```rust
let submitted_headers: SubmittedHeaders<HeaderId<P::Hash, P::Number>, TC::Error> = submitted_headers;
```
instead of introducing a closure?

relays/ethereum/src/ethereum_client.rs
svyatonik: Nope - this consumes `Vec<SubstrateHeaderId>`, but I need remaining ids when error occurs :(

relays/ethereum/src/sync_loop.rs
svyatonik: yeah, thanks - this also helps :)

PR #174, opened 2020-07-06
Remove less maintenance friendly deny.

tomusdrw:


.github/workflows/deny.yml
```diff
@@ -13,47 +13,10 @@ on:
...
-          name:            deny.log
-          path:            deny.log
-
   cargo-deny:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
     - uses: EmbarkStudios/cargo-deny-action@v1
+      with:
+        command: "check --hide-inclusion-graph"
```
HCastano, line 22: Why do we want to hide this?

tomusdrw: It generates super long output and makes it hard to find actual errors, especially due to Github Actions interface loading file lazily in chunks (which makes scrolling difficult). The file in artifacts contains full inclusion graph, so it's still available.

PR #175, opened 2020-07-06
Update Substrate to rc4

HCastano:
Since all crates, with the exception of `node-primitives`, are on crates.io I've removed the `rev` and `git` fields from the imports.

### DIFF
--- a/scripts/update_rc.sh
+++ b/scripts/update_rc.sh
@@ -2,6 +2,7 @@
 
 # One-liner to update between Substrate `rc` releases
 # Usage: ./update_rc.sh rc4 rc5
+set -xeu
 
 OLD_RC_VERSION=$1
 NEW_RC_VERSION=$2
--- a/scripts/update_rc.sh
+++ b/scripts/update_rc.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+# One-liner to update between Substrate `rc` releases
+# Usage: ./update_rc.sh rc4 rc5
+
+OLD_RC_VERSION=$1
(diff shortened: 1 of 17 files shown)


tomusdrw:
> Iiuc everything with -rc* suffix may break semver requirements => may break our build when (unintentionally?) updated
I'm fine with crates.io versions, but we can make require the exact version via `~` I believe.

Also please include a bash script to update the versions, I don't want anyone to do that manually even once again - it has to be an automated process.

HCastano:
> Also please include a bash script to update the versions, I don't want anyone to do that manually even once again - it has to be an automated process.

I'm not really sure a committed script is needed. It was a one liner to update the versions:

```bash
find . -type f -name 'Cargo.toml' -exec sed -i '' -e 's/rc3/rc4/g' {} \;
```

And if we remove the `git`/`rev` tags that one liner is all we'd need anyways. Is this too simplistic a way of looking at the update process?

tomusdrw:
> It was a one liner to update the versions:

Well it would probably take me a couple of minutes to come up with that one-liner I think it's worth adding anyway. We already have `scripts` folder, so we can collect such utilities there.

scripts/update_rc.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+# One-liner to update between Substrate `rc` releases
+# Usage: ./update_rc.sh rc4 rc5
+
```
tomusdrw, line 5:
```suggestion
set -xeu

```

PR #177, opened 2020-07-07
Relay PoA lock-funds transactions proofs

svyatonik:
Brief overview:
1) I've added another option to or relayer, even though initially I wanted to build separate utility for this. It seems faster option to me;
2) the utility is able to relay **single transaction**. So there's no 'listening mode' - you have to start it manually after submitting PoA transaction. I think it's enough for demonstration purposes;
3) you can start utility just after receiving transaction hash. Even before you'll submit it to PoA node. The utility would wait for tx to be mined, for corresponding header to be imported and finalized by the bridge module and then the proof-transaction would be submitted.

How to:
```
RUST_LOG=bridge=trace ./ethereum-poa-relay eth-exchange-sub --eth-tx-hash=7669b7a77c103ed330b3416843fd05dd0c294f61e3b3de27c22e21b33fc9f668
```

Related `rust-web3` PR: https://github.com/tomusdrw/rust-web3/pull/366


relays/ethereum/src/ethereum_exchange.rs
```diff
@@ -0,0 +1,211 @@
...
+
+		Ok(Some((eth_header_id, eth_tx)))
+	}
+
+	async fn transaction_proof(
+		&self,
+		eth_header_id: &EthereumHeaderId,
+		eth_tx: EthereumTransaction,
+	) -> Result<EthereumTransactionInclusionProof, Self::Error> {
+		let eth_header = self.client.header_by_hash(eth_header_id.1).await?;
```
tomusdrw, line 101: IMHO we should get both with transactions in one RPC query. It would also simplify the code a tiny bit.

relays/ethereum/src/ethereum_exchange.rs
```diff
@@ -0,0 +1,211 @@
...
+					Some(eth_raw_tx) => eth_raw_tx.0,
+					None => return Err(EthereumNodeError::MissingTransaction(eth_tx_hash).into()),
+				});
+			} else {
+				eth_relay_tx_index = Some(index as u64);
+				transaction_proof.push(match eth_relay_tx.take().and_then(|eth_tx| eth_tx.raw) {
+					Some(eth_raw_relay_tx) => eth_raw_relay_tx.0,
+					None => {
+						return Err(
+							EthereumNodeError::DuplicateBlockTransaction(*eth_header_id, eth_relay_tx_hash).into(),
```
tomusdrw, line 119: I don't think it's worth an error, let's just skip such case.

relays/ethereum/src/ethereum_exchange.rs
```diff
@@ -0,0 +1,211 @@
...
+		let eth_relay_tx_hash = eth_tx.hash;
+		let mut eth_relay_tx = Some(eth_tx);
+		let mut eth_relay_tx_index = None;
+		let mut transaction_proof = Vec::with_capacity(eth_header.transactions.len());
+		for (index, eth_tx_hash) in eth_header.transactions.into_iter().enumerate() {
+			if eth_tx_hash != eth_relay_tx_hash {
+				let eth_tx = self.client.transaction_by_hash(eth_tx_hash).await?;
+				transaction_proof.push(match eth_tx.and_then(|eth_tx| eth_tx.raw) {
+					Some(eth_raw_tx) => eth_raw_tx.0,
+					None => return Err(EthereumNodeError::MissingTransaction(eth_tx_hash).into()),
```
tomusdrw, line 111: This error would no longer be required if we fetch block with transactions in one go.

relays/ethereum/src/main.rs
```diff
@@ -224,3 +235,18 @@ fn ethereum_deploy_contract_params(
...
 }
+
+fn ethereum_exchange_params(matches: &clap::ArgMatches) -> Result<ethereum_exchange::EthereumExchangeParams, String> {
+	let mut params = ethereum_exchange::EthereumExchangeParams::default();
+	params.eth = ethereum_connection_params(matches)?;
+	params.sub = substrate_connection_params(matches)?;
+	params.sub_sign = substrate_signing_params(matches)?;
+
+	if let Some(eth_tx_hash) = matches.value_of("eth-tx-hash") {
+		params.eth_tx_hash = eth_tx_hash
```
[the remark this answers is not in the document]
svyatonik: It's already marked as required in `cli.yml`, so it's actually would always me `Some()`. I'll change to `.expect()` here

relays/ethereum/src/ethereum_exchange.rs
svyatonik: Done, but the bad thing is that now we have this:
```
		#[rpc(method = "eth_getBlockByHash", positional_params)]
		fn get_block_by_hash(hash: H256, full_tx_objs: bool) -> EthereumHeader;
		#[rpc(method = "eth_getBlockByHash", positional_params)]
		fn get_block_by_hash_with_transactions(hash: H256, full_tx_objs: bool) -> EthereumHeaderWithTransactions;
```

in rpc.rs. So basically if wrong flag is passed, we'll fail to decode results. I do not know any other how to fix that, but to introduce custom deserialization (which I wanted to avoid).

tomusdrw: Yeah, I think that's fine. The ethereum API is not really well-typed. AFAIR in ParityEthereum we were returning an `enum` and depending on the flag different variant was returned. I'm also fine to return `(EthereumHeader, Option<Vec<Transactions>>>)` to avoid introducing another type, but your solution looks good too.

primitives/ethereum-poa/src/lib.rs
```diff
@@ -508,6 +508,9 @@ sp_api::decl_runtime_apis! {
 		/// (or leads to making) other header the best one.
 		fn best_block() -> (u64, H256);
 
+		/// Returns number abd hash of the best finalized block known to the bridge module.
```
HCastano, line 511:
```suggestion
		/// Returns number and hash of the best finalized block known to the bridge module.
```

relays/ethereum/src/ethereum_exchange.rs
```diff
@@ -0,0 +1,212 @@
...
+/// PoA exchange transaction relay params.
+#[derive(Debug, Default)]
+pub struct EthereumExchangeParams {
+	/// Ethereum connection params.
+	pub eth: EthereumConnectionParams,
+	/// Hash of the Ethereum transaction to relay.
+	pub eth_tx_hash: EthereumTransactionHash,
+	/// Substrate connection params.
+	pub sub: SubstrateConnectionParams,
+	/// Substrate connection params.
```
HCastano, line 47:
```suggestion
	/// Substrate signing params.
```

relays/ethereum/src/cli.yml
```diff
@@ -6,22 +6,22 @@ subcommands:
     - eth-to-sub:
         about: Synchronize headers from Ethereum node to Substrate node.
         args:
-            - eth-host:
+            - eth-host: &eth-host
```
HCastano, line 9: What's the point of adding `&arg`?

relays/ethereum/src/rpc_errors.rs
```diff
@@ -101,20 +102,34 @@ pub enum EthereumNodeError {
...
+			Self::IncompleteHeader => "Incomplete Ethereum Header Received".to_string(),
+			Self::IncompleteReceipt => "Incomplete Ethereum Receipt Recieved".to_string(),
+			Self::IncompleteTransaction => "Incomplete Ethereum Transaction".to_string(),
+			Self::InvalidSubstrateBlockNumber => "Received an invalid Substrate block from Ethereum Node".to_string(),
+			Self::DuplicateBlockTransaction(header_id, tx_hash) => format!(
+				"Ethereum block {}/{} includes Ethereum transaction {} more than once",
+				header_id.0, header_id.1, tx_hash,
+			),
+			Self::BlockMissingTransaction(header_id, tx_hash) => format!(
+				"Ethereum block {}/{} is missing Ethereum transaction {} we believe is a part of this block",
```
HCastano, line 129:
```suggestion
				"Ethereum block {}/{} is missing Ethereum transaction {} which we believe is a part of this block",
```

relays/ethereum/src/substrate_client.rs
```diff
@@ -281,14 +293,50 @@ impl SubmitEthereumHeaders for SubstrateRpcClient {
...
+		params: SubstrateSigningParams,
+		proof: bridge_node_runtime::exchange::EthereumTransactionInclusionProof,
+	) -> Result<()> {
+		let account_id = params.signer.public().as_array_ref().clone().into();
+		let nonce = self.next_account_index(account_id).await?;
+
+		let transaction = create_signed_transaction(
+			bridge_node_runtime::Call::BridgeCurrencyExchange(
+				bridge_node_runtime::BridgeCurrencyExchangeCall::import_peer_transaction(proof),
+			),
```
HCastano, line 322: I'm a little confused at the wrapped `Call`s here. So the inner call is to ask the Substrate node to import a "lock funds" transaction, what's the outer call for?

relays/ethereum/src/ethereum_exchange.rs
```diff
@@ -0,0 +1,212 @@
...
+
+	async fn is_header_finalized(&self, id: &EthereumHeaderId) -> Result<bool, Self::Error> {
+		// we check if header is finalized by simple comparison of the header' number and
+		// number of best finalized PoA' header known to Substrate node.
+		//
+		// this may lead to failure in tx proof import if PoA reorganization has happened
+		// after we have checked that our tx has been included into given block
+		//
+		// the fix is easy, but since this code is mostly developed for demonstration purposes,
+		// I'm leaving this KISS-based design here
```
HCastano, line 166: In 10 years some poor soul gonna be shaking their fist at this "demonstration" code ;)

relays/ethereum/src/ethereum_exchange.rs
```diff
@@ -0,0 +1,212 @@
...
+	) -> Result<Option<(EthereumHeaderId, EthereumTransaction)>, Self::Error> {
+		let eth_tx = match self.client.transaction_by_hash(*hash).await? {
+			Some(eth_tx) => eth_tx,
+			None => return Ok(None),
+		};
+
+		// we need transaction to be mined => check if it is included in the block
+		let eth_header_id = match (eth_tx.block_number, eth_tx.block_hash) {
+			(Some(block_number), Some(block_hash)) => HeaderId(block_number.as_u64(), block_hash),
+			_ => return Ok(None),
```
HCastano, line 90: How can we get to this point? If a transaction hasn't been included in a block wouldn't the call to `transaction_by_hash()` return `None`?

HCastano:
It looks like the meat of this PR is really in`ethereum_exchange.rs` and `exchange.rs`. How I'm understanding it now is that `ethereum_exchange.rs` is kinda like `{ethereum, substrate}_sync_loop.rs`, and `exchange.rs` is like `sync_loop.rs`. That's to say, `ethereum_exchange.rs` is has a lot of the support code (e.g creating the Tx proofs), while `exchange.rs` is responsible for the high level sync logic and leverages the support code.

I think the code looks fine in general, although I'm not going to approve it just yet because there's a couple of things I want to take a closer look at tomorrow.

relays/ethereum/src/cli.yml
svyatonik: That's what Tomek has suggested [here](https://github.com/paritytech/parity-bridges-common/pull/177#discussion_r451415868). TLDR: to avoid duplicate args declarations.

relays/ethereum/src/ethereum_exchange.rs
svyatonik: `None` is returned when transaction is completely unknown to the node. But if it is in the tx pool, it would rather return tx with these fields set to `None`.

relays/ethereum/src/substrate_client.rs
svyatonik: There are two `Call` enums (as well as `Event`, `Digest`, ...) on different levels. First one is at pallet level - it just enumerates all calls that pallet provides. The second one is on runtime level - it includes calls from all runtime' pallets. So something like that:
```
enum BridgeCall { ImportHeader };
enum ExchangeCall { ImportTransaction };
enum RuntimeCall {
    Bridge(BridgeCall),
    Exchange(ExchangeCall),
}
```

relays/ethereum/src/cli.yml
HCastano: Ah, I missed that, thanks!

relays/ethereum/src/exchange.rs
```diff
@@ -0,0 +1,493 @@
...
+	async fn tick(&self);
+	/// Returns `Ok(true)` if header is known to the target node.
+	async fn is_header_known(&self, id: &HeaderId<P>) -> Result<bool, Self::Error>;
+	/// Returns `Ok(true)` if header is finalized by the target node.
+	async fn is_header_finalized(&self, id: &HeaderId<P>) -> Result<bool, Self::Error>;
+	/// Submits transaction proof to the target node.
+	async fn submit_transaction_proof(&self, proof: P::TransactionProof) -> Result<(), Self::Error>;
+}
+
+/// Relay single transaction' proof.
```
HCastano, line 82:
```suggestion
/// Relay single transaction proof.
```

relays/ethereum/src/exchange.rs
```diff
@@ -0,0 +1,493 @@
...
+		let is_header_known = target_client.is_header_known(&source_header_id).await.map_err(|err| {
+			format!(
+				"Failed to check existence of header {}/{} on {} node: {:?}",
+				source_header_id.0,
+				source_header_id.1,
+				P::TARGET_NAME,
+				err,
+			)
+		})?;
+		match is_header_known {
```
HCastano, line 174:
```suggestion
		if is_header_known {
```

relays/ethereum/src/exchange.rs
```diff
@@ -0,0 +1,493 @@
...
+			.map_err(|err| {
+				format!(
+					"Failed to check finality of header {}/{} on {} node: {:?}",
+					source_header_id.0,
+					source_header_id.1,
+					P::TARGET_NAME,
+					err,
+				)
+			})?;
+		match is_header_finalized {
```
HCastano, line 220:
```suggestion
		if is_header_finalized {
```

relays/ethereum/src/exchange.rs
```diff
@@ -0,0 +1,493 @@
...
+		}
+	}
+}
+
+/// Wait until target node finalizes required header.
+async fn wait_header_finalized<P: TransactionProofPipeline>(
+	target_client: &impl TargetClient<P>,
+	source_header_id: &HeaderId<P>,
+) -> Result<(), String> {
+	// now wait until target node finalizes required header
```
HCastano, line 206: This is a duplicate of the doc comment, I don't think we need it

relays/ethereum/src/exchange.rs
```diff
@@ -0,0 +1,493 @@
...
+
+	fn ensure_success(source: TestTransactionsSource, target: TestTransactionsTarget) {
+		assert_eq!(
+			async_std::task::block_on(relay_single_transaction_proof(
+				&source,
+				&target,
+				test_transaction_hash(),
+			)),
+			Ok(()),
+		);
```
HCastano, line 391:
```suggestion
		assert!(
			async_std::task::block_on(relay_single_transaction_proof(
				&source,
				&target,
				test_transaction_hash(),
			)).is_ok()
		);
```

relays/ethereum/src/exchange.rs
```diff
@@ -0,0 +1,493 @@
...
+				);
+
+				target_client.tick().await;
+			}
+		}
+	}
+}
+
+#[cfg(test)]
+mod tests {
```
HCastano, line 248: I'm a fan of how the tests are written, they're nice and readable!

relays/ethereum/src/ethereum_exchange.rs
```diff
@@ -0,0 +1,212 @@
...
+		async_std::task::sleep(SUBSTRATE_TICK_INTERVAL).await;
+	}
+
+	async fn is_header_known(&self, id: &EthereumHeaderId) -> Result<bool, Self::Error> {
+		self.client.ethereum_header_known(*id).await
+	}
+
+	async fn is_header_finalized(&self, id: &EthereumHeaderId) -> Result<bool, Self::Error> {
+		// we check if header is finalized by simple comparison of the header' number and
+		// number of best finalized PoA' header known to Substrate node.
```
HCastano, line 160:
```suggestion
		// number of best finalized PoA header known to Substrate node.
```

HCastano:
There are a few small things I would like to see:
- Module level comments for `ethereum_exchange.rs` and `exchange.rs`
- An updated README that describes the new utility, and its usage
- Optional for this PR, but a flow diagram would also be helpful

svyatonik:
Updated readme + module-level docs. Diagrams are still ok imo, since, iiuc, we're going to extend tx-proofs-relay functionality (i.e. listen to contract on PoA network).

relays/ethereum/Cargo.toml
```diff
@@ -29,7 +29,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.56"
 sp-bridge-eth-poa = { path = "../../primitives/ethereum-poa" }
 time = "0.2"
-web3 = { version = "0.12.0", default-features = false }
+web3 = { git = "https://github.com/tomusdrw/rust-web3", rev = "f8ed7137fefb1b1ba41e17b86078a6eddf3ca52b", default-features = false }
```
tomusdrw, line 32:
```suggestion
web3 = { version = "0.13.0", default-features = false }
```

Could you update to crates.io version? Just published it.

relays/ethereum/src/ethereum_client.rs
```diff
@@ -120,13 +121,35 @@ impl EthereumRpc for EthereumRpcClient {
...
 		}
 	}
 
+	async fn header_by_hash_with_transactions(&self, hash: H256) -> Result<HeaderWithTransactions> {
+		let get_full_tx_objects = true;
+		let header = Ethereum::get_block_by_hash_with_transactions(&self.client, hash, get_full_tx_objects).await?;
+
+		let is_complete_header = header.number.is_some() && header.hash.is_some() && header.logs_bloom.is_some();
+		if !is_complete_header {
+			return Err(RpcError::Ethereum(EthereumNodeError::IncompleteHeader));
```
tomusdrw, line 138: This happens if the header/block is still pending?

relays/ethereum/src/ethereum_client.rs
```diff
@@ -120,13 +121,35 @@ impl EthereumRpc for EthereumRpcClient {
...
+		}
+
+		let is_complete_transactions = header.transactions.iter().all(|tx| tx.raw.is_some());
+		if !is_complete_transactions {
+			return Err(RpcError::Ethereum(EthereumNodeError::IncompleteTransaction));
```
tomusdrw, line 143: How could that happen?

relays/ethereum/src/rpc_errors.rs
```diff
@@ -101,20 +102,34 @@ pub enum EthereumNodeError {
...
 	fn to_string(&self) -> String {
 		match self {
-			Self::ResponseParseFailed(e) => e,
-			Self::IncompleteHeader => "Incomplete Ethereum Header Received",
-			Self::IncompleteReceipt => "Incomplete Ethereum Receipt Recieved",
-			Self::InvalidSubstrateBlockNumber => "Received an invalid Substrate block from Ethereum Node",
+			Self::ResponseParseFailed(e) => e.to_string(),
+			Self::IncompleteHeader => "Incomplete Ethereum Header Received".to_string(),
+			Self::IncompleteReceipt => "Incomplete Ethereum Receipt Recieved".to_string(),
+			Self::IncompleteTransaction => "Incomplete Ethereum Transaction".to_string(),
```
tomusdrw, line 122: Can you add the explanation about `raw` field here?

relays/ethereum/src/rpc_errors.rs
```diff
@@ -101,20 +102,34 @@ pub enum EthereumNodeError {
...
 
 impl ToString for EthereumNodeError {
 	fn to_string(&self) -> String {
 		match self {
-			Self::ResponseParseFailed(e) => e,
-			Self::IncompleteHeader => "Incomplete Ethereum Header Received",
-			Self::IncompleteReceipt => "Incomplete Ethereum Receipt Recieved",
-			Self::InvalidSubstrateBlockNumber => "Received an invalid Substrate block from Ethereum Node",
+			Self::ResponseParseFailed(e) => e.to_string(),
+			Self::IncompleteHeader => "Incomplete Ethereum Header Received".to_string(),
```
tomusdrw, line 120: Could you be more explicit about what's missing in the header? This would help the users debug errors.

relays/ethereum/src/ethereum_client.rs
svyatonik: I haven't found this information in [web3](https://docs.rs/web3/0.13.0/web3/types/struct.Block.html) documentation. But yeah - I think this could happen when block is pending.

relays/ethereum/src/ethereum_client.rs
svyatonik: When we are connected to the node which doesn't return `raw` field,

PR #179, opened 2020-07-09
Fix clippy suggestions.

tomusdrw:


.github/workflows/lint.yml
```diff
@@ -36,3 +36,41 @@ jobs:
         with:
           command:      fmt
           args:         --all -- --check
+  check-clippy:
+    name:                          Check Clippy
+    runs-on:                       self-hosted
+    container:
+      image:                       paritytech/ci-linux:production
```
svyatonik, line 43: Not related to this PR - just found that we use different images here vs the cargo check step. Is this OK? Also seems that `cargo check` isn't using the cache, right (iiuc it requires compilation)?

tomusdrw: You mean `cargo fmt` and `cargo deny`? They run on github-hosted runners, cause they are pretty fast anyway. AFAICT the `build.yml` file is running on self-hosted with the same image. Can you link to the exact line you mean?

svyatonik: Sorry, I just wrongly interpreted `check-fmt` step :/ It is actually `cargo fmt -- --check`, not `cargo check`. My bad - sorry again :)

tomusdrw: No worries! :)

PR #183, opened 2020-07-10
Fix benchmarks compilation

### LINKED ISSUE #142: Check benchmarks compilation on CI
#136 brings first benchmark and there are some others planned. We need to run `cargo check` on CI to test if benchmarks are still compilable - they're too fragile, especially when we touch test code.

svyatonik:
closes #142


tomusdrw:
I've made similar changes here: https://github.com/paritytech/parity-bridges-common/pull/163/files#diff-cf310bd8aed83600b1472d2ea1ec4290R96

PR #187, opened 2020-07-13
Precompiled cache

General-Beck:
If cache is missing, copy it from `master`

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -49,6 +49,7 @@ jobs:
           echo "::set-env name=CARGO_HOME::/cache/${{ github.head_ref }}/${{ matrix.toolchain }}"
           echo "::set-env name=CARGO_TARGET_DIR::/cache/${{ github.head_ref }}/${{ matrix.toolchain }}"
       - name:                      Checking for the cache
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -48,6 +48,15 @@ jobs:
         run:                       |
           echo "::set-env name=CARGO_HOME::/cache/${{ github.head_ref }}/${{ matrix.toolchain }}"
           echo "::set-env name=CARGO_TARGET_DIR::/cache/${{ github.head_ref }}/${{ matrix.toolchain }}"
+      - name:                      Checking for the cache
+        run:                       |
+          if ! [ -d $CARGO_TARGET_DIR ]; then
+            echo "Copy cache from master";
+            mkdir -p $CARGO_TARGET_DIR;
+            apt update&&apt install rsync -y;
+            rsync -aq /cache/${{ matrix.toolchain }}/ $CARGO_TARGET_DIR/;
```
tomusdrw, line 57: Why `rsync` and not simply `cp`? The directory does not exist anyway, so it has to copy everything.

General-Beck: If there is an active build in the master branch, some of the files in the source may change. rsync makes the copy better.

PR #190, opened 2020-07-14
[WIP] Fix cleanup cache

General-Beck:


tomusdrw:
@General-Beck how is that going?

General-Beck:
@tomusdrw today I plan to finish

PR #221, opened 2020-07-17
Add Prometheus and Grafana to Docker Compose

HCastano:
Adds Prometheus monitoring and a Grafana dashboard as part of the Rialto
test network's Docker Compose setup as a follow up to #191.

Currently having issues getting Prometheus to feed data to Grafana, but
both web UIs are running fine.


HCastano:
I think Grafana isn't getting any data because Prometheus isn't getting any data from the bridge relay. I know Grafana and Prometheus are up since I can access both of their UIs, but when I try and connect to the bridge node with `curl http://localhost:9616/metrics` I get an empty response from the server.

Headers are being synced, so it is up and running fine.  This is the port config for the relay from`docker ps`

```
PORTS
0.0.0.0:9616->9616/tcp
```

Am I doing something dumb with the networking in the `docker-compose`?

svyatonik:
I think the reason is that by default relay is exposing its metrics only to local clients - i.e. via `127.0.0.1` adapter. So you may want to pass `--prometheus-host=0.0.0.0` cli option to relay.

tomusdrw:
In docker compose the best way to have two containers connect to each other is using their service names as hostnames. So for instance to connect to PoA node arthur we do this:
https://github.com/paritytech/parity-bridges-common/blob/a5b3df9b4e0f975c22d518503e1fead83993233b/deployments/rialto/bridge-config/relay-eth2sub-entrypoint.sh#L5

Afaict it means you will need to update `prometheus.yml` file to have `relay-eth2sub:9616` as target here:
https://github.com/paritytech/parity-bridges-common/blob/a5b3df9b4e0f975c22d518503e1fead83993233b/relays/ethereum/dashboard/prometheus.yml#L9

Also as @svyatonik mentioned, you need to configure the relay to expose prometheus interface outside of the container, which can be achieved by listening on all interfaces as Slava suggested.

deployments/rialto/docker-compose.yml
```diff
@@ -145,3 +147,22 @@ services:
...
+    volumes:
+      - ../../relays/ethereum/dashboard/prometheus.yml:/etc/prometheus/prometheus.yml
+    ports:
+      - "9090:9090"
+    depends_on: *all-nodes
+
+  grafana-dashboard:
+    image: grafana/grafana
+    volumes:
+      - ../../relays/ethereum/dashboard/grafana-datasource.yaml:/etc/grafana/provisioning/datasources/grafana-datasource.yaml
```
tomusdrw, line 162: I'd consider moving the entire `dashboard` directory somewhere closer to `deployments`. But I'm not super strong on this, cause having it nearby the `relay` code also is reasonable - so your call :)

deployments/dashboard/prometheus/prometheus.yml
```diff
@@ -6,4 +6,4 @@ scrape_configs:
     scrape_interval: 5s
 
     static_configs:
-      - targets: ['127.0.0.1:9616']
+      - targets: ['relay-eth2sub:9616']
```
tomusdrw, line 9: Can we include both `127.0.0.1` and `relay-eth2sub` here? This will allow re-using the config outside of docker-compose as well.

Issue #153, opened 2020-06-29
Support multiple instances of Bridge pallet


tomusdrw:
It should be possible to have multiple instances of bridge pallet in the runtime. For instance to support Kovan + Custom PoA testnet.

Features:
- [ ] Runtime flags to enable / disable particular instance of the bridge ("admin" rights).
- [ ] Relay to specify which instance should it be connected to
- [ ] (?) A `BridgeHub` pallet to list available bridges and their metadata (to use in UIs for instance).

Alternatively we can have a `BridgeHub` pallet being a proxy to individual deployments (so the access control is done in `BridgeHub` for instance)

May involve tweaking Relay to speak to the right instance.

tomusdrw:
Closed by #226

Issue #209, opened 2020-07-17
Relay Substrate Headers


tomusdrw:
Extend relayer to support `Substrate` as a source chain. Make sure it works with the generic API (#204).

svyatonik:
Let's close this - we're able to relay Substrate headers && rutime relay API may be built on top of `HeaderChain` later, if we need that.

Issue #205, opened 2020-07-17
Migrate `CurrencyExchange` to use `HeaderChain`


tomusdrw:
After #204 is complete we should migrate `CurrencyExchange` pallet to be compatible with anything that implements `HeaderChain` trait.

PR #226, opened 2020-07-21
Allow Multiple Bridge Pallet Instances

HCastano:
This starts tackling #153. The idea behind this PR is to allow multiple PoA bridge pallets to be used in the same runtime. To do this we use the [instantiable pallet pattern](https://substrate.dev/recipes/3-entrees/instantiable.html). Ultimately this would allow us to run a Substrate chain which supports both Rialto and Kovan simultaneously.

It's still too early to tell if I'll end up including changes to the bridge relay node as part of this PR, or if that'll be done in a follow up.


bin/node/runtime/src/lib.rs
```diff
@@ -448,7 +458,8 @@ construct_runtime!(
 		TransactionPayment: pallet_transaction_payment::{Module, Storage},
 		Sudo: pallet_sudo::{Module, Call, Config<T>, Storage, Event<T>},
 		Session: pallet_session::{Module, Call, Storage, Event, Config<T>},
-		BridgeEthPoA: pallet_bridge_eth_poa::{Module, Call, Config, Storage, ValidateUnsigned},
+		Rialto: pallet_bridge_eth_poa::<Instance1>::{Module, Call, Config, Storage, ValidateUnsigned},
```
tomusdrw, line 461:
```suggestion
		BridgeRialto: pallet_bridge_eth_poa::<Instance1>::{Module, Call, Config, Storage, ValidateUnsigned},
```

bin/node/runtime/src/lib.rs
```diff
@@ -448,7 +458,8 @@ construct_runtime!(
...
+		Kovan: pallet_bridge_eth_poa::<Instance2>::{Module, Call, Config, Storage, ValidateUnsigned},
```
tomusdrw, line 462:
```suggestion
		BridgeKovan: pallet_bridge_eth_poa::<Instance2>::{Module, Call, Config, Storage, ValidateUnsigned},
```
`Bridge` prefix will add a bit more context to the pallet name and also will help with grouping since I think that the names are sorted alphabetically in polkadotjs combo box.

Also let's put all bridge stuff at the top, wdyt?

bin/node/runtime/src/lib.rs
```diff
@@ -537,24 +548,44 @@ impl_runtime_apis! {
 
 	impl sp_bridge_eth_poa::EthereumHeadersApi<Block> for Runtime {
```
tomusdrw, line 549: We might need separate runtime API for each bridge or we might pass bridge instance name as a parameter. That aligns with the `BridgeHub` pallet idea, which could be a unified interface for all bridge instances.

HCastano:
So I ran into a few problems today, and I'd like to hear your thoughts on them.

1. Conditional Compilation in Construct Runtime Macro

I asked around, and apparently this is not supported as part of the macro. I'm not sure if we'll get support for it soon, but I feel like it should be added to Substrate eventually. For now my workaround during development is to simply include all instances of a pallet in the macro and have the default Rust feature include all PoA pallets.

2. Multiple PoA instances with Currency Exchange

As I wired up the runtime with multiple instances of the PoA pallet I realized that I would need to update the currency exchange pallet to support multiple PoA pallet instances. This turned out to be a bit of a problem since the implementation of the `Blockchain` (now `PeerBlockchain`) trait used by the currency exchange pallet was calling [`BridgeEthPoA`](https://github.com/paritytech/parity-bridges-common/blob/master/bin/node/runtime/src/exchange.rs#L77). Now we had to choose between `BridgeKovan` and `BridgeRialto`.

The first solution that came to mind was to extend the currency exchange pallet to support multiple chains by referring to a `peer_id`. Using this `peer_id` you'd be able to know which transaction and proof format to use, and also keep track of what transactions you've accepted from a particular chain. However, this path didn't look totally straightforward so I abandoned it.

Instead what I've done is made the currency exchange pallet instantiable. Now each instance of the currency exchange pallet can be paired with an instance of a bridge. While this isn't the most elegant solution, I think it's nice because basically nothing changes in the currency exchange pallet. I also don't think it comes with that much storage overhead since we'd have to keep track of transactions for each chain anyways with a unified model.

svyatonik:
2. imo what you have done is the only correct solution :)

tomusdrw:
> Conditional Compilation in Construct Runtime Macro

I don't think it's needed. We always want to construct a full-featured runtime. We might add a `disabled` flag to the pallet itself, so that it's not operational if not needed, but it would be a storage entry, so it's easy to turn it on and off via governance.

> Multiple PoA instances with Currency Exchange

Agree with Slava, that's was the way to go. In the future it might be possible for the CE pallet to connect to `BridgeHub` instead of individual bridges, but then it would need to deal with a completely opaque data. That said, I think I prefer having separate instances, since it allows more flexibility (for instance CurrencyExchange for PoA, but more sophisticated message delivery for Substrate).

HCastano:
> > Conditional Compilation in Construct Runtime Macro
> 
> I don't think it's needed. We always want to construct a full-featured runtime. We might add a `disabled` flag to the pallet itself, so that it's not operational if not needed, but it would be a storage entry, so it's easy to turn it on and off via governance.
> 

I'm not sure if I agree with this. I think it would be fine to have the default be full-featured, but I don't think we should stop people from being able to easily build a runtime which only supports their PoA network. Having to disable Kovan through governance seems annoying for someone that only wants to support one network from the get-go.

tomusdrw:
> I don't think we should stop people from being able to easily build a runtime which only supports their PoA network. Having to disable Kovan through governance seems annoying for someone that only wants to support one network from the get-go.

The `bridge-node` is just an example (similar to `substrate-node` or even `node-template`). Anyone who would like to get a bridge into their network, would simply just include one of the pallets to their own runtime. Running a production blockchain with `bridge-node` is kind of pointless beyond testing purposes.

HCastano:
> > I don't think we should stop people from being able to easily build a runtime which only supports their PoA network. Having to disable Kovan through governance seems annoying for someone that only wants to support one network from the get-go.
> 
> The `bridge-node` is just an example (similar to `substrate-node` or even `node-template`). Anyone who would like to get a bridge into their network, would simply just include one of the pallets to their own runtime. Running a production blockchain with `bridge-node` is kind of pointless beyond testing purposes.

Hmm, alright you've convinced me

HCastano:
I'm gonna do the bridge relay changes in another PR to keep things focused. What it means for this PR is that usage of Rialto is hardcoded into `substrate_client`. I don't think this should be a big deal though.

.github/workflows/rust.yml
```diff
@@ -55,6 +55,9 @@ jobs:
             # TODO add rsync in docker
             rsync -aq /cache/${{ matrix.toolchain }}/ $CARGO_TARGET_DIR/ || true;
         shell:                     bash
+      - name:                      Sccache stats
```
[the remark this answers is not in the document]
HCastano: General Beck made this change for debugging purposes iirc

bin/node/node/src/chain_spec.rs
```diff
@@ -152,7 +152,8 @@ fn testnet_genesis(
 		pallet_aura: Some(AuraConfig {
 			authorities: Vec::new(),
 		}),
-		pallet_bridge_eth_poa: load_bridge_config(),
+		pallet_bridge_eth_poa_Instance1: load_rialto_bridge_config(),
```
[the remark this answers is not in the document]
HCastano: Nope, these are the names generated by the Substrate macros :(

bin/node/runtime/src/kovan.rs
```diff
@@ -131,6 +134,25 @@ impl BridgePruningStrategy for PruningStrategy {
 	}
 }
 
+/// The Kovan Blockchain as seen by the runtime.
+pub struct KovanBlockchain;
+
+impl PeerBlockchain for KovanBlockchain {
```
tomusdrw, line 140: This implementation should not be necessary, and hopefully will be addressed by #205

bin/node/runtime/src/lib.rs
```diff
@@ -684,8 +724,8 @@ impl_runtime_apis! {
 				}
 			}
 
-			add_benchmark!(params, batches, b"bridge-eth-poa", BridgeEthPoA);
-			add_benchmark!(params, batches, b"bridge-currency-exchange", BridgeCurrencyExchangeBench::<Runtime>);
+			add_benchmark!(params, batches, b"bridge-eth-poa", BridgeKovan);
+			add_benchmark!(params, batches, b"bridge-currency-exchange", BridgeCurrencyExchangeBench::<Runtime, KovanCurrencyExchange>);
```
tomusdrw, line 728: shall we have benchmarks for Rialto as well?

relays/ethereum/src/substrate_client.rs
```diff
@@ -317,7 +317,8 @@ impl SubmitEthereumExchangeTransactionProof for SubstrateRpcClient {
 		let nonce = self.next_account_index(account_id).await?;
 
 		let transaction = create_signed_transaction(
-			bridge_node_runtime::Call::BridgeCurrencyExchange(
+			// TODO: Change so that that it's dynamic
```
tomusdrw, line 320: Can you add a specific issue number? I believe it's going to be #209

bin/node/runtime/src/lib.rs
HCastano: I thought about it, but given that we previously only benchmarked Kovan I wanted to keep it like that for now. We can look into adding Rialto once we update the benches to use the automated approach

relays/ethereum/src/substrate_client.rs
HCastano: Yeah, it's related. I can add it

Issue #197, opened 2020-07-17
Set-up Monitoring Alerts


tomusdrw:
Relay monitoring of Rialto (both persistent and ephemeral) should be pushed to some central server where we can configure alerts in case the network is not running correctly.

PR #228, opened 2020-07-22
Relay healthcheck

tomusdrw:
Fix healthcheck of bridge node and the relay. Related: #154


tomusdrw:
Is it possible to auto-restart on healtcheck error? My understanding was that this is something that external (host) applications can inspect and act upon, but it's not automated like that.
I think it would be cool to get a notification in case healthcheck is failing indeed - will research what we can do about that.

svyatonik:
I've found [this](https://stackoverflow.com/questions/47088261/restarting-an-unhealthy-docker-container-based-on-healthcheck) but still not sure whether we want this or not.

tomusdrw:
I'm not convinced we should do it, I think I'd rather see an alert being sent instead of auto-restarting. Will research that further as part of #197

PR #229, opened 2020-07-22
Submit exchange transactions to PoA node

### LINKED ISSUE #198: PoA Transactions generator
Write a tool to automatically send some transactions on the PoA network that will be relayed. The tool should be running for both ephemeral and persitent Rialto (perhaps with different settings though).

svyatonik:
closes #198

This PR:
1) removes `test-helpers` feature from `sp-bridge-eth-poa`, because tx signatures are now required by all crates that use it && it isn't that big overhead;
2) adds `relay-eth-submit-exchange-tx` command to relay. This subcommand submits exchange tx to PoA node;
3) adds script that submits exchange transactions to docker-compose.


deployments/rialto/docker-compose.yml
```diff
@@ -65,6 +65,20 @@ services:
       - bridge-node-bob
       - bridge-node-charlie
 
+  relay-eth-submit-exchange-tx:
```
tomusdrw, line 68: I'd rather call this `poa-exchange-tx-generator`, I know it uses `relay` binary to actually submit them, but I feel the `generator` part emphasizes that this thing is producing (test) transactions.

deployments/rialto/bridge-config/relay-eth-submit-exchange-tx-entrypoint.sh
```diff
@@ -0,0 +1,60 @@
...
+ETH_SIGNERS=(
+	# Harcoded account on OE dev chain
+	"4d5db4107d237df6a3d58ee5f70ae63d73d7658d4026f2eefd2f204c81682cb7"
+)
+# Minimal exchange amount (in finney)
+MIN_EXCHANGE_AMOUNT_FINNEY=1 # 0.1 ETH
+# Maximal exchange amount (in finney)
+MAX_EXCHANGE_AMOUNT_FINNEY=100000 # 100 ETH
+# Max delay before submitting transactions (s)
+MAX_SUBMIT_DELAY_S=60
```
tomusdrw, line 29: Depending on the deployment it might be too often or too rare. Can we maybe make it configurable via ENV variable passed from docker-compose? (see another comment there)

deployments/rialto/docker-compose.yml
```diff
@@ -65,6 +65,20 @@ services:
...
+      context: .
+      dockerfile: ./Bridge.Dockerfile
+      args:
+        BRIDGE_HASH: master
+        PROJECT: ethereum-poa-relay
+    entrypoint: /config/relay-eth-submit-exchange-tx-entrypoint.sh
+    volumes:
+      - ./bridge-config:/config
+    environment:
+      RUST_LOG: rpc=trace,bridge=trace
```
tomusdrw, line 79:
```suggestion
      RUST_LOG: rpc=trace,bridge=trace
      GENERATOR_MAX_DELAY: ${GENERATOR_MAX_DELAY:-60}
```

Adding this env variable will make it 60s by default, but if ENV variable is present when running `docker-compose up` (or a `.env` file is present) it can be overwritten.
I think it would be cool to have, because for tests we can easily set it to 60, but in production we can send less/more frequently.

deployments/rialto/bridge-config/relay-eth-submit-exchange-tx-entrypoint.sh
```diff
@@ -0,0 +1,60 @@
...
+	# select signer
+	ETH_SIGNERS_MAX_INDEX=$((${#ETH_SIGNERS[@]} - 1))
+	ETH_SIGNERS_INDEX=`shuf -i 0-$ETH_SIGNERS_MAX_INDEX -n 1`
+	ETH_SIGNER=${ETH_SIGNERS[$ETH_SIGNER_INDEX]}
+	echo $ETH_SIGNER
+
+	# select amount
+	EXCHANGE_AMOUNT_FINNEY=`shuf -i $MIN_EXCHANGE_AMOUNT_FINNEY-$MAX_EXCHANGE_AMOUNT_FINNEY -n 1`
+	EXCHANGE_AMOUNT_ETH=`printf "%s000" $EXCHANGE_AMOUNT_FINNEY`
+	echo $EXCHANGE_AMOUNT_ETH
```
tomusdrw, line 53: Wouldn't it make more sense to write this loop in Rust as part of the relay? With some CLI parameters tuning the default behaviour could be to send only one transaction (as it is right now).

relays/ethereum/src/ethereum_exchange_submit.rs
```diff
@@ -0,0 +1,114 @@
...
+/// Submit single Ethereum -> Substrate exchange transaction.
+pub fn run(params: EthereumExchangeSubmitParams) {
+	let mut local_pool = futures::executor::LocalPool::new();
+
+	let result: Result<_, String> = local_pool.run_until(async move {
+		let eth_client = EthereumRpcClient::new(params.eth);
+
+		let eth_signer_address = params.eth_sign.signer.address();
+		let sub_recipient_encoded = params.sub_recipient;
+		let nonce = eth_client
```
tomusdrw, line 63: This might become tricky if the CLI is called too frequently (i.e. there is a pending transaction in the pool), since this API doesn't take that into account. We might need to tune the minimal delay to make sure that pending transaciton is always mined before we submit a new one. However, I think it would be cool to also sometimes test with more than one transaction in a block, so maybe better try some other fix here.
Moving the generation logic here, would allow you to track nonces internally instead of relying on the RPC.

relays/ethereum/src/main.rs
```diff
@@ -266,6 +311,8 @@ fn ethereum_exchange_params(matches: &clap::ArgMatches) -> Result<ethereum_excha
 		.parse()
 		.map_err(|e| format!("Failed to parse eth-tx-hash: {}", e))?;
 
+	log::debug!(target: "bridge", "Etehereum exchange params: {:?}", params);
```
tomusdrw, line 314:
```suggestion
	log::debug!(target: "bridge", "Ethereum exchange params: {:?}", params);
```

deployments/rialto/bridge-config/relay-eth-submit-exchange-tx-entrypoint.sh
svyatonik: I don't see much point in this, with the single exception - handling nonces. It is much easier to modify script than to modify relay code. We may e.g. want to replace this loop with some simulation modeling software && the only thing that will change (it'll be dropped) is that small script. If we'll have rust loop, removing it will affect its author morale much harder (and it's not only the loop that will be dropped - we'll also need to add some config file parsing). So imo this loop is something that should be easy to change + easy to drop. So script language is imo the best choice here.

Re nonces - that's the good point. I think we may just print/return nonce from the relay submit tx call && maintain nonces map in the script. Will try to do that

svyatonik:
(sorry that I'm committing after approval, but I've found a typo + changed script a bit so it could be used from within other test environments - now it is configured for our docker environment by default, but could be used for other configurations as well)

HCastano:
I'll take a look at this later today :)

deployments/rialto/bridge-config/poa-exchange-tx-generator-entrypoint.sh
```diff
@@ -0,0 +1,89 @@
...
+MIN_EXCHANGE_AMOUNT_FINNEY=${EXCHANGE_GEN_MIN_AMOUNT_FINNEY:-1} # 0.1 ETH
+# Maximal exchange amount (in finney)
+MAX_EXCHANGE_AMOUNT_FINNEY=${EXCHANGE_GEN_MAX_AMOUNT_FINNEY:-100000} # 100 ETH
+# Max delay before submitting transactions (s)
+MAX_SUBMIT_DELAY_S=${EXCHANGE_GEN_MAX_SUBMIT_DELAY_S:-60}
+
+while true
+do
+	# sleep some time
+	SUBMIT_DELAY_S=`shuf -i 0-$MAX_SUBMIT_DELAY_S -n 1`
```
HCastano, line 43: TIL about `shuf`

deployments/rialto/bridge-config/poa-exchange-tx-generator-entrypoint.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
```
HCastano, line 1: Can you add some documentation as to what the goal of the script is? Some things I think are worth noting are that it's not meant from production use, and it more for lightly "fuzzing" a test network

bin/node/runtime/src/exchange.rs
```diff
@@ -37,7 +37,7 @@ use sp_currency_exchange::{
 use sp_std::vec::Vec;
 
 /// Ethereum address where locked PoA funds must be sent to.
-const LOCK_FUNDS_ADDRESS: [u8; 20] = hex!("DEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEF");
+pub const LOCK_FUNDS_ADDRESS: [u8; 20] = hex!("DEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEF");
```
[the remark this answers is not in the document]
svyatonik: The idea is that relay points works with concrete runtime and primitives may be used by many runtimes + many relays. Even in our runtime we may have 1 address for Kovan and another for TestPoA. So I'd prefer to leave it here

relays/ethereum/src/ethereum_exchange_submit.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+impl Default for EthereumExchangeSubmitParams {
+	fn default() -> Self {
+		EthereumExchangeSubmitParams {
+			eth: Default::default(),
+			eth_sign: Default::default(),
+			eth_nonce: None,
+			eth_amount: 1_000_000_000_000_000_000_u64.into(), // 1 ETH
+			sub_recipient: hex!("1cbd2d43530a44705ad088af313e18f80b53ef16b36177cd4b77b846f2a5f07c"), // ferdie
```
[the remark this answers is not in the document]
svyatonik: Some pre-history - since the very beginning, all relay CLI options had default values that were pointing to: (1) some well-known addresses in OE development chain at PoA side (i.e. to 0x00a329c0648769a73afac7f9381e08fb43dbea72) and (2) some well-known addresses in our Substrate runtime (i.e. Alice, Bob, ...). So now we have default values for everything, including contract code, contract address (which depends on eth-signer and its nonce), sub-signers, ... The idea was that you could execute some relay subcommand  with both OE and Substrate nodes active and everything will work by default. 

I agree that these times it isn't actual anymore, because we (even myself) no longer use OE dev chain. But I'd prefer to leave all cli options consistent, than to make some work in has-defaults way and some in no-defaults way. I suggest to address this altogether in #162 by requiring values for all parameters + introducing configuration files (OE-like) as replacement for default values.

Feel free to insist on your change though - I'll add `required` to `cli.yml`. But I feel it is wrong way to fix it now.

svyatonik:
Last commit changes exchange transactions signer from Arthur to Bertha because Arthur will generate contract deployment + Substrate headers sync transactions.

relays/ethereum/src/ethereum_exchange_submit.rs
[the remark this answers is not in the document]
HCastano: I do agree that adding `required` is mainly just a bandaid fix, but I think in this case we should favour safety over consistency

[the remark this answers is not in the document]
tomusdrw: IMHO this should be reworked as part of #162 I'd rather keep it consistent for now, maybe we will come back to `--dev` testing :)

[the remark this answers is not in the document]
svyatonik: ok, then I'm merging

PR #230, opened 2020-07-23
Faster re-build of docker images.

tomusdrw:
This PR adds two-step compilatiion of each project:
1. We first compile against `master` version of `parity-bridges-common` repo
2. Then we switch to `BRIDGE_HASH` and compile again.

The first step should (hopefully) be retrieved from docker cache, so the second step will only incrementally compile changes between `master..BRIDGE_HASH`, which should result in much faster compilation.

Also updates `update.sh` script used on our test deployment, to force rebuild on every pull.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 **/target/
+**/.env
 
 **/*.rs.bk
 
--- a/deployments/rialto/Bridge.Dockerfile
+++ b/deployments/rialto/Bridge.Dockerfile
@@ -34,14 +34,19 @@ RUN rustc -vV && \
 WORKDIR /parity-bridges-common
 
 ### Build from the repo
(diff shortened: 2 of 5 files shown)


svyatonik:
Dumb q: won't we get into the the error described [here](https://www.ctl.io/developers/blog/post/more-docker-image-cache-tips/)? I.e. "pathspec '???' did not match any file(s) known to git" - because `git clone` would be cached and repo won't be updated and then when we'll get to the `git checkout $BRIDGE_HASH` step (with say some branch that has been pushed recently), the build will fail?

tomusdrw:
Ugh, yeah, you are definitely right. Doing `git fetch && git checkout $BRIDGE_HASH` should work though, right?

PR #232, opened 2020-07-23
Expose Grafana & Set-up first alert

tomusdrw:
Related #197


deployments/rialto/dashboard/grafana/provisioning/dashboards/grafana-dashboard.json
```diff
@@ -18,66 +18,143 @@
...
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
       "targets": [
         {
-          "expr": "best_block_numbers",
-          "format": "time_series",
-          "instant": true,
+          "expr": "max(best_block_numbers{node=\"source\"}) - max(best_block_numbers{node=\"target\"})",
```
svyatonik, line 102: Technically we may not receive alerts even if sync is not working, because both metrics are coming from relay => relay may be offline && report the same (<5) difference again and again. But we have nothing better now.

Except, maybe additional alerts if 'max(best_block_numbers{node=\"source\"})' or `max(best_block_numbers{node=\"target\"})` aren't changing for too long. But here comes additional relay feature - it doesn't ask source node (PoA in eth2sub mode) about best block if the difference is too big. Because we don't actually need this information until we have synchronized almost everything. Example: we start Kovan -> Rialto sync. Kovan has almost 20_000_000 headers. Let's say it'll take 4 days to sync this chain (this just an example - not an estimation). So for all this time, `max(best_block_numbers{node=\"source\"})` will be 20_000_000 even though another 1_000_000 may have been generated on Kovan. This is just a small optimization (because we do not need this information for headers synchronization) and can be eliminated if required.

deployments/rialto/dashboard/grafana-matrix/config.yml
```diff
@@ -0,0 +1,52 @@
...
+matrix:
+- name: matrix-parity-io
+  url: https://matrix.parity.io
+  # Create a user - log that user in using a post request
+  # curl -XPOST -d '{"type": "m.login.password",
+  #                  "user":"grafana",
+  #                  "password":"2m4ny53cr3t5"}'
+  #      "https://my-matrix-server/_matrix/client/r0/login"
+  # Fill that access token in here
+  access_token: "MDAxZWxvY2F0aW9uIG1hdHJpeC5wYXJpdHkuaW8KMDAxM2lkZW50aWZpZXIga2V5CjAwMTBjaWQgZ2VuID0gMQowMDMwY2lkIHVzZXJfaWQgPSBAYnJpZGdlcy1ib3Q6bWF0cml4LnBhcml0eS5pbwowMDE2Y2lkIHR5cGUgPSBhY2Nlc3MKMDAyMWNpZCBub25jZSA9IF5oTWZDc29xZytnI3RKcywKMDAyZnNpZ25hdHVyZSD8viAa8CCGowlWi1qOHLhXAjznDKdeqDRGZ4I6SJgQNwo"
```
HCastano, line 17: Umm, should this be part of version control?

deployments/rialto/docker-compose.yml
```diff
@@ -148,16 +148,34 @@ services:
...
+      GF_SERVER_ROOT_URL: ${GRAFANA_SERVER_ROOT_URL}
     volumes:
-      - ../dashboard/grafana/provisioning/:/etc/grafana/provisioning/
+      - ./dashboard/grafana/provisioning/:/etc/grafana/provisioning/
     ports:
       - "3000:3000"
     depends_on:
       - prometheus-metrics
+
+  grafana-matrix-notifier:
```
HCastano, line 172: Is there some way to make this an optional step? If I want to test a network locally I don't really want to have this

deployments/rialto/dashboard/grafana/provisioning/dashboards/grafana-dashboard.json
tomusdrw: IMHO Initial sync is a special case anyway, we should probably temporarily turn off alerts during that time anyway. I like the idea of adding an alert if there is no progress in PoA blocks. Will do.

deployments/rialto/docker-compose.yml
tomusdrw: The container will just fail to start or will simply not send notifications. AFAIK there is no way to have conditional services in compose file.

PR #233, opened 2020-07-25
Fix Local Dockerfile Entrypoint

HCastano:
The old `ENTRYPOINT` in the bridge Dockerfiles weren't actually resolving the `$PROJECT` variable. This means that when you did `docker run ...` the container wouldn't spin up. This is because we were using `exec` form of ENTRYPOINT. From the [Docker docs](https://docs.docker.com/engine/reference/builder/#entrypoint):

> Unlike the shell form, the exec form does not invoke a command shell. This means that normal shell processing does not happen. For example, ENTRYPOINT [ "echo", "$HOME" ] will not do variable substitution on $HOME.

I've changed it to use `shell` form by introducing an `entrypoint` script. I first tried to keep the contents of the script in-line in the Dockerfile, but it didn't work.


deployments/scripts/bridge-node-dev-entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
tomusdrw, line 1: IMHO the file should be `bridge-entrypoint.sh`, since we can run either bridge-node or the relay using it.

PR #237, opened 2020-07-27
Auto exchange transactions relay metrics

svyatonik:
Dashboard will appear in one of follow-up PRs, because it'll only be useful after #229 is in + I've had problems deploying rialto locally today. So I'd prefer to make changes to our docker configuration + test it once in several PRs, when everything required is merged.

Exposed metrics:
1) standard 'system' metrics (memory, CPU, workload);
2) number of block that we have processed (`best_block_numbers::processed`) and that we need to process (`best_block_numbers::known`);
3) total number of processed blocks (reset after every restart);
4) total number of processed transactions - total, relayed and failed (reset after every restart).


relays/ethereum/src/exchange_loop.rs
```diff
@@ -92,6 +95,19 @@ pub fn run<P: TransactionProofPipeline>(
 		let mut state = storage.state();
 		let mut current_finalized_block = None;
 
+		let mut metrics_global = GlobalMetrics::new();
```
tomusdrw, line 98: I feel like this kind of code + `expose_metrics` is duplicated in every relay loop. Is there a way we could wrap it into a nicer API? Ideally it would be 2-3 lines at most to add new metrics to another `*_loop.rs` file imho (init + get_state + update).

tomusdrw:
Needs rebase, but lgtm.

Issue #75, opened 2020-04-24
Avoid hardcoded magic values in relayer - move them to constants.


tomusdrw:
Some of the constants should be placed in it's own module, cause they are hardcoded, but potentially configurable.

Other constants (like RPC method names) can be safely placed in the top of the module.

See also: #60

tomusdrw:
I guess it has been addressed over time. The issue is too generic now.

Issue #61, opened 2020-04-07
Make relay (easy) configurable for other chains


svyatonik:
original comment: https://github.com/paritytech/parity-bridges-common/pull/58#discussion_r404764786

The problem is that we can't (at least I do not know how atm) make relay generic over runtime/chain. That's why if we're going to point it to other chain, it would need to be reconfigured and rebuilt before any transaction will be submitted.

The only way I could think of, is to extract everything that is runtime/chain specific (including crypto type) to separate file, like [this one](https://github.com/paritytech/secret-store/blob/master/bin/substrate/src/runtime.rs#L17). This also must be noted in crate' readme.

tomusdrw:
Related #153

tomusdrw:
Closed via #849

PR #238, opened 2020-07-28
Support Pallet Instances in Relay

HCastano:
This is a first step in making the relay a bit more flexible in terms of what networks it is able to sync. It brings the relay up to speed with the changes introduced in #226. Right now we're making a few simplifying assumptions:

1. We know the runtime of the Substrate node at compile time
2. We know the pallets which support instances at compile time

With these assumption we are able to support syncing headers from an Ethereum network to a specific instance on the Substrate node. For example, we can have Kovan headers synced to the Kovan instance on the Substrate node. A different run of the relay can sync Rialto headers to the Rialto instance on the _same_ Substrate node.

I've tested this PR with Rialto and headers were synced fine. I haven't tested it with Kovan, but since it's basically the same code I think that it'll also be fine.


relays/ethereum/src/ethereum_sync_loop.rs
```diff
@@ -64,6 +64,8 @@ pub struct EthereumSyncParams {
 	pub sync_params: HeadersSyncParams,
 	/// Metrics parameters.
 	pub metrics_params: Option<MetricsParams>,
+	/// Substrate bridge pallet instance to syncronize headers with
```
tomusdrw, line 67:
```suggestion
	/// Substrate bridge pallet instance to synchronize headers with.
```

relays/ethereum/src/substrate_client.rs
```diff
@@ -83,12 +83,20 @@ impl Default for SubstrateSigningParams {
 	}
 }
 
+// TODO: Remove
+enum BridgeInstance {
```
tomusdrw, line 87: I think long-term I'd prefer if this was a generic parameter instead of an enum, i.e.:
```rust
trait BridgeInstance {
 fn whatever_parameters_you_need(&self);
}
struct Kovan;
impl BridgeInstance for Kovan {
 fn ...
}
struct Rialto;
impl BridgeInstance of Rialto {
  fn..
}
```
The reason for this is that it will allow using `relay` code as a library and easily implement support for custom chains. I.e. IMHO what we should target with the relay is an architecture with crates like that:
```
- relay-cli // a BINARY - CLI runner for the relay, might list "known" networks
- relay // a LIBRARY - that allows building your own relay, should most likely be generic over some parameters
```
Related: #162 #75 #61 #60

svyatonik: Just sharing my thoughts here. I was thinking that we may actually have several relay binaries - one for every pair of chains. We may extract almost everything from relay to shared crate. What should stay in chain-specific binaries are files like: `cli.yml`, `ethereum_client.rs`, `ethreum_deploy_contract.rs`, `substrate_types.rs`, ... So we'll have separate binary for Rialto <-> Kovan, Rialto <-> TestPoA, ... We may achieve same functionality with conditional compilation, though, but it'll be too much for single binary crate imo.

tomusdrw: I'm not a fan of feature gating either. I like the multi-binary approach as well, pretty much it's another take on the `CLI + Library` split, where we basically have multiple CLI (binaries) ourselves. Probably an advantage of that approach would be that the CLI creation process would be generic/abstract as well, since we would like to keep the code DRY.

HCastano: I didn't really consider the multi-binary approach. The vision I had for this was one where you would configure the relay through the CLI:

```bash
$ relay-cli relay --from ETH --to SUB --instance Kovan
$ relay-cli relay --from KSM --to DOT
$ relay-cli exchange --from KSM --to DOT
```

Although a multi-binary approach might be necessary to support different runtimes. From a user point of view that seems annoying though. 

I like the library + binary split, and maybe that'll ease of the UX around having different binaries for pairs of chains. I'll think about it some more.

relays/ethereum/src/instances.rs
```diff
@@ -20,13 +20,51 @@ use crate::substrate_types::{into_substrate_ethereum_header, into_substrate_ethe
 use bridge_node_runtime::exchange::EthereumTransactionInclusionProof as Proof;
 use bridge_node_runtime::Call;
 
+use std::fmt::Debug;
+
 pub trait BridgeInstance {
```
tomusdrw, line 25: Maybe add `Send + Sync` bounds here to avoid repetition?

relays/ethereum/src/ethereum_deploy_contract.rs
```diff
@@ -42,10 +43,12 @@ pub struct EthereumDeployContractParams {
...
 	/// Initial header.
 	pub sub_initial_header: Option<Vec<u8>>,
+	/// Instance of the bridge pallet being synchronized.
+	pub instance: Box<dyn BridgeInstance>,
 }
 
-impl Default for EthereumDeployContractParams {
-	fn default() -> Self {
+impl EthereumDeployContractParams {
+	pub fn with_instance(instance: Box<dyn BridgeInstance>) -> Self {
```
tomusdrw, line 51: Docs?
```suggestion
	pub fn new(instance: Box<dyn BridgeInstance>) -> Self {
```
To be fair I'd prefer simply `new` instead of `with_instance`. I feel like `with_*` constructors/builder functions are useful if you want to parametrize some extra stuff that is set to default values in `new`. I.e. they allow customization over default `new` constructor. Here however we only have one way to create the params and I don't feel like `with_instance` adds any extra value. Looking at the std library we have:
https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.new
https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.with_capacity
but there is no `with_value` for things like `Box`, `Cell`, etc:
https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html#method.new

relays/ethereum/src/instances.rs
```diff
@@ -0,0 +1,129 @@
...
+	fn build_signed_header_call(&self, headers: Vec<QueuedEthereumHeader>) -> Call;
+	/// Used to build a `Call` for importing an unsigned header to a Substrate runtime.
+	fn build_unsigned_header_call(&self, header: QueuedEthereumHeader) -> Call;
+	/// Used to build a `Call` for importing peer transactions to a Substrate runtime.
+	fn build_currency_exchange_call(&self, proof: Proof) -> Call;
+}
+
+/// Corresponds to the Rialto instance used in the bridge runtime.
+#[derive(Default, Clone, Debug)]
+pub struct RialtoInstance;
```
tomusdrw, line 47: I feel like this belongs to a separate file that is more chain-specific, but it seem we don't have one yet. If we split binary & library, the trait should live in the library, but the instances in binary(ies).

```suggestion
pub struct Rialto;
```

Since currently the struct is within `instances` module, I feel that adding `Instance` prefix is redundant, would rather prefer the type to be used like that: `instances::Rialto`.

relays/ethereum/src/instances.rs
```diff
@@ -0,0 +1,129 @@
...
+
+impl RialtoInstance {
+	/// Create a new RialtoInstance.
+	pub fn new() -> Self {
```
tomusdrw, line 51: IMHO that's redundant, let's use `instances::Rialto::default()`

relays/ethereum/src/instances.rs
```diff
@@ -0,0 +1,129 @@
...
+		Self
+	}
+}
+
+impl BridgeInstance for RialtoInstance {
+	fn build_signed_header_call(&self, headers: Vec<QueuedEthereumHeader>) -> Call {
+		let pallet_call = bridge_node_runtime::BridgeEthPoACall::import_signed_headers(
```
tomusdrw, line 58: We should start extracting all things referencing `bridge_node_runtime` to a single module that can be easily replaced. Now it's impossible to use relay with a different target runtime (e.g. Westend). See #61

relays/ethereum/src/main.rs
```diff
@@ -41,12 +42,14 @@ mod utils;
...
 use ethereum_client::{EthereumConnectionParams, EthereumSigningParams};
 use ethereum_sync_loop::EthereumSyncParams;
+use instances::{BridgeInstance, KovanInstance, RialtoInstance};
 use parity_crypto::publickey::{KeyPair, Secret};
 use sp_core::crypto::Pair;
-use std::io::Write;
 use substrate_client::{SubstrateConnectionParams, SubstrateSigningParams};
 use substrate_sync_loop::SubstrateSyncParams;
 
+use std::io::Write;
```
tomusdrw, line 51: I prefer a single (but sorted) group of imports. IMHO splitting them arbitrarily never works long-term in a bigger project.

relays/ethereum/src/main.rs
```diff
@@ -210,50 +213,55 @@ fn substrate_signing_params(matches: &clap::ArgMatches) -> Result<SubstrateSigni
 }
 
 fn ethereum_sync_params(matches: &clap::ArgMatches) -> Result<EthereumSyncParams, String> {
-	let mut eth_sync_params = EthereumSyncParams::default();
-	eth_sync_params.eth = ethereum_connection_params(matches)?;
-	eth_sync_params.sub = substrate_connection_params(matches)?;
-	eth_sync_params.sub_sign = substrate_signing_params(matches)?;
-	eth_sync_params.metrics_params = metrics_params(matches)?;
+	let instance = instance_params(matches)?;
+	let mut params = EthereumSyncParams::with_instance(instance);
```
tomusdrw, line 217: Why not create the `EthereumSyncParams` object directly here and get rid of the constructor completely? You wouldn't need defaults at all and it would be easier to ensure that all parameters are set to non-default values.

relays/ethereum/src/main.rs
```diff
@@ -210,50 +213,55 @@ fn substrate_signing_params(matches: &clap::ArgMatches) -> Result<SubstrateSigni
...
 }
 
 fn substrate_sync_params(matches: &clap::ArgMatches) -> Result<SubstrateSyncParams, String> {
-	let mut sub_sync_params = SubstrateSyncParams::default();
-	sub_sync_params.eth = ethereum_connection_params(matches)?;
-	sub_sync_params.eth_sign = ethereum_signing_params(matches)?;
-	sub_sync_params.sub = substrate_connection_params(matches)?;
-	sub_sync_params.metrics_params = metrics_params(matches)?;
+	let instance = instance_params(matches)?;
+	let mut params = SubstrateSyncParams::with_instance(instance);
```
tomusdrw, line 244: Same here.

relays/ethereum/src/main.rs
```diff
@@ -352,3 +362,17 @@ fn metrics_params(matches: &clap::ArgMatches) -> Result<Option<metrics::MetricsP
 
 	Ok(Some(metrics_params))
 }
+
+fn instance_params(matches: &clap::ArgMatches) -> Result<Box<dyn BridgeInstance>, String> {
+	let instance: Box<dyn BridgeInstance> = if let Some(instance) = matches.value_of("sub-pallet-instance") {
+		match instance.to_lowercase().as_str() {
```
tomusdrw, line 368: Is `to_lowercase` still required if parameter is marked as `case_insensitive` in the yaml file?

relays/ethereum/src/cli.yml
```diff
@@ -43,6 +44,16 @@ subcommands:
                 long: sub-signer-password
                 value_name: SUB_SIGNER_PASSWORD
                 help: The password for the SURI of secret key to use when transactions are submitted to the Substrate node.
+            - sub-pallet-instance: &sub-pallet-instance
+                long: instance
+                short: i
+                value_name: PALLET_INSTANCE
+                help: Which instance of the bridge pallet should the relay follow?
```
tomusdrw, line 51: Should probably mention that it defaults to `Rialto`.

relays/ethereum/src/main.rs
HCastano: I've been splitting imports like this:

1. Crate imports
2. Crates.io imports
3. Std imports

I'd be okay merging 2 and 3, but I'd like to keep the crate imports separate since it gives me a way to tell what we're putting in from the rest of the project at a glance

relays/ethereum/src/main.rs
HCastano: The problem lies with `sync_params`. For the two different sync params (Ethereum and Substrate) we require a different "default" for `HeaderSyncParams`. To build this default it requires us to pull in ~5 different constants from the corresponding module. It seemed nicer to keep that as part of the corresponding modules, but I can look into changing it

relays/ethereum/src/main.rs
HCastano: I'm not sure actually, I'll have to look into it

HCastano: So it looks like `case_insensitive` is just a gatekeeper for what comes through the CLI, but it doesn't actually do anything to the input - like make it lowercase

relays/ethereum/src/instances.rs
HCastano: I agree, but I don't think that refactor should be done in this PR

relays/ethereum/src/main.rs
tomusdrw: I've noticed that it's quite common in other files already, so I'm okay with that. The issue I have with `crate` stuff being in separate section is that I've seen in all of our other "big" projects that at some point some stuff is moved to a separate crate, and the imports are usually changed by search & replace (then format to sort) and over time if we don't pay enough attention we might end up with two sections, but the original idea will be lost. Hope it's not going to happen here though :)

relays/ethereum/src/main.rs
```diff
@@ -307,7 +326,8 @@ fn ethereum_exchange_submit_params(
 }
 
 fn ethereum_exchange_params(matches: &clap::ArgMatches) -> Result<ethereum_exchange::EthereumExchangeParams, String> {
-	let mut params = ethereum_exchange::EthereumExchangeParams::default();
+	let instance = instance_params(matches)?;
+	let mut params = ethereum_exchange::EthereumExchangeParams::new(instance);
```
tomusdrw, line 330: No defaults here as well?

relays/ethereum/src/main.rs
```diff
@@ -210,50 +214,64 @@ fn substrate_signing_params(matches: &clap::ArgMatches) -> Result<SubstrateSigni
...
-	Ok(sub_sync_params)
+	Ok(params)
 }
 
 fn ethereum_deploy_contract_params(
 	matches: &clap::ArgMatches,
 ) -> Result<ethereum_deploy_contract::EthereumDeployContractParams, String> {
-	let mut eth_deploy_params = ethereum_deploy_contract::EthereumDeployContractParams::default();
+	let instance = instance_params(matches)?;
+	let mut eth_deploy_params = ethereum_deploy_contract::EthereumDeployContractParams::new(instance);
```
tomusdrw, line 274: No defaults here as well?

relays/ethereum/src/main.rs
HCastano: My bad, I missed these 🤦

relays/ethereum/src/main.rs
```diff
@@ -210,110 +218,134 @@ fn substrate_signing_params(matches: &clap::ArgMatches) -> Result<SubstrateSigni
...
+		hex::decode(include_str!("../res/substrate-bridge-bytecode.hex")).expect("code is hardcoded, thus valid; qed")
+	};
 
-	log::debug!(target: "bridge", "Deploy params: {:?}", eth_deploy_params);
+	let params = EthereumDeployContractParams {
+		eth_params: ethereum_connection_params(matches)?,
+		eth_sign: ethereum_signing_params(matches)?,
+		sub_params: substrate_connection_params(matches)?,
+		instance: instance_params(matches)?,
+		sub_initial_authorities_set_id: None,
```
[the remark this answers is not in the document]
HCastano: Done: #266.

PR #240, opened 2020-07-28
Fix script copy in Bridge.Dockerfile

svyatonik:
follow up for this commit https://github.com/paritytech/parity-bridges-common/pull/233/commits/e58891b7f8cbc3530df690e52a7bc44560cd5c45

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,7 +61,7 @@ WORKDIR /home/user
 ARG PROJECT=ethereum-poa-relay
 
 COPY --chown=user:user --from=builder /parity-bridges-common/target/release/${PROJECT} ./
-COPY ./deployments/scripts/bridge-entrypoint.sh ./
+COPY --chown=user:user --from=builder /parity-bridges-common/deployments/scripts/bridge-entrypoint.sh ./
 
 # check if executable works in this container
 RUN ./${PROJECT} --version
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,7 +61,7 @@ WORKDIR /home/user
 ARG PROJECT=ethereum-poa-relay
 
 COPY --chown=user:user --from=builder /parity-bridges-common/target/release/${PROJECT} ./
-COPY ./deployments/scripts/bridge-node-dev-entrypoint.sh ./
+COPY ./deployments/scripts/bridge-entrypoint.sh ./
 
 # check if executable works in this container
 RUN ./${PROJECT} --version
--- a/deployments/rialto/Bridge.Dockerfile
+++ b/deployments/rialto/Bridge.Dockerfile
@@ -72,7 +72,7 @@ ARG PROJECT=ethereum-poa-relay
 ARG HEALTH=http://localhost:9616/metrics
 
 COPY --chown=user:user --from=builder /parity-bridges-common/target/release/${PROJECT} ./
-COPY ./deployments/scripts/bridge-node-dev-entrypoint.sh ./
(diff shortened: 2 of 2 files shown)


svyatonik:
@HCastano @tomusdrw I need some help here, please :) The fix itself is ok afaui, but now running `update.sh` fails:
1) when running `./update.sh` with current folder = `/deployments/rialto`, it fails because there's no folder `./deployments/scripts/` (it is two levels above);
2) when running `./update.sh` with current folder = root of the repo, it fails because there's following command in in update.sh:
```bash
sed -i "s/access_token.*/access_token: \"$MATRIX_ACCESS_TOKEN\"/" ./dashboard/grafana-matrix/config.yml
```

So we either need to fix `update.sh` to work from root (that'd need Tomek auto-deployment script fix as well, iiuc), or fix it in some other way. Can you please give me hint here, guys?

Dockerfile
```diff
@@ -61,7 +61,7 @@ WORKDIR /home/user
 ARG PROJECT=ethereum-poa-relay
 
 COPY --chown=user:user --from=builder /parity-bridges-common/target/release/${PROJECT} ./
-COPY ./deployments/scripts/bridge-node-dev-entrypoint.sh ./
+COPY ./deployments/scripts/bridge-entrypoint.sh ./
```
tomusdrw, line 64:
```suggestion
COPY --chown=user:user --from=builder /parity-bridges-common/deployments/scripts/bridge-entrypoint.sh ./
```

deployments/rialto/Bridge.Dockerfile
```diff
@@ -72,7 +72,7 @@ ARG PROJECT=ethereum-poa-relay
 ARG HEALTH=http://localhost:9616/metrics
 
 COPY --chown=user:user --from=builder /parity-bridges-common/target/release/${PROJECT} ./
-COPY ./deployments/scripts/bridge-node-dev-entrypoint.sh ./
+COPY ./deployments/scripts/bridge-entrypoint.sh ./
```
tomusdrw, line 75:
```suggestion
COPY --chown=user:user --from=builder /parity-bridges-common/deployments/scripts/bridge-entrypoint.sh ./
```

svyatonik:
My first build attempt with your changes has ended up with the same error as before (file is not found). When building without using docker cache, I've stuck with #242 . I'm a docker looser indeed. Will try tomorrow with `--locked` or with #242 resolved.

HCastano:
So it looks like there might be an issue unrelated to `update.sh`. If I do `docker-compose build` on the `master` branch I get the following `COPY` error:

```
ERROR: Service 'bridge-node-alice' failed to build: COPY failed: stat /var/lib/docker/tmp/docker-builder192103447/deployments/scripts/bridge-node-dev-entrypoint.sh: no such file or directory
```

Is this the same error you're running into?

I can get around this by changing the `COPY` command to

```bash
COPY --chown=user:user --from=builder /parity-bridges-common/deployments/scripts/ ./
```

If I add the `bridge-endpoint.sh` at the end there to copy just the single file it doesn't find the file, not sure why.

svyatonik:
@HCastano That's the issue I've started with :) There's actually nothing wrong with `update.sh`. In https://github.com/paritytech/parity-bridges-common/commit/e58891b7f8cbc3530df690e52a7bc44560cd5c45 you've renamed `bridge-node-dev-entrypoint.sh` tot `bridge-entrypoint.sh`, but not everywhere.

PR #241, opened 2020-07-28
Exchange transactions relay dashboard

svyatonik:
This PR adds prefix for metrics that relay exposes - now in `eth-to-sub` mode it'll be exposing metrics that are starting with "Ethereum_to_Substrate_Sync_" prefix. In `eth-exchange-sub` mode, prefix would be "Ethereum_to_Substrate_Exchange_". There's now 2nd dashboard for `eth-exchange-sub` relay.

I was unable to test it in docker env (mostly because of #240), so it is only tested using retrograde [bash scripts](https://github.com/svyatonik/parity-bridges-common.test) :)

### DIFF
--- a/deployments/rialto/dashboard/prometheus/prometheus.yml
+++ b/deployments/rialto/dashboard/prometheus/prometheus.yml
@@ -11,7 +11,7 @@ scrape_configs:
   - job_name: 'eth_exchange_sub_relay_node'
 
     # Override the global default and scrape targets from this job every 5 seconds.
-    scrape_interval: 5s
+    scrape_interval: 15s
 
     static_configs:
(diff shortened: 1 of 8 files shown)


deployments/rialto/dashboard/prometheus/prometheus.yml
```diff
@@ -11,7 +11,7 @@ scrape_configs:
   - job_name: 'eth_exchange_sub_relay_node'
 
     # Override the global default and scrape targets from this job every 5 seconds.
```
tomusdrw, line 13:
```suggestion
    # Override the global default and scrape targets from this job every 15 seconds.
```

svyatonik:
It looks like that: Dashboard

PR #243, opened 2020-07-29
Bump Substrate to RC5

HCastano:
I figured I'd end my day with something easy, but turns out that the changes in `service.rs` were a bit hairy (introduced in paritytech/substrate#6557). I ended up copying the updated `node-template` instead of digging through everything.

Keeping it as a draft because I need to fix the RPC extensions in the service, and I want to double check I haven't accidentally screwed the node up.


bin/node/node/src/service.rs
```diff
@@ -16,13 +16,14 @@
 
 //! Service and ServiceFactory implementation. Specialized wrapper over substrate service.
 
-use bridge_node_runtime::{self, opaque::Block, RuntimeApi};
-use grandpa::{self, FinalityProofProvider as GrandpaFinalityProofProvider, SharedVoterState, StorageAndProofProvider};
-use sc_client_api::ExecutorProvider;
-use sc_consensus::LongestChain;
+use bridge_node_runtime::{self, Block, RuntimeApi};
```
tomusdrw, line 19: Is it necessary for us to have that file at all? Isn't it possible to reference the service builder from substrate somehow?

HCastano: No, I don't think so

HCastano:
Ah, I forgot about the benchmark changes. I'll fix those tomorrow-ish

bin/node/node/src/service.rs
```diff
@@ -36,137 +37,175 @@ native_executor_instance!(
...
+		std::sync::Arc::new(pool_api),
+		config.prometheus_registry(),
+		task_manager.spawn_handle(),
+		client.clone(),
+	);
+
+	let (grandpa_block_import, grandpa_link) =
+		sc_finality_grandpa::block_import(client.clone(), &(client.clone() as Arc<_>), select_chain.clone())?;
+
+	let aura_block_import =
```
tomusdrw, line 83: I didn't notice that earlier, but we run with `AuRa + Grandpa`? That's pretty cool and hopefully works :)

PR #247, opened 2020-07-29
Remove substrate-bridge artifact.

tomusdrw:
Follow up on #244


tomusdrw:
Clippy failure unrelated, merging :)

Issue #194, opened 2020-07-17
Setup Substrate → PoA part of the bridge


tomusdrw:
There is a commented-out code in docker-compose to run a second part of the bridge, it requires some extra work to get running.

PR #248, opened 2020-07-29
Prepare for sub to eth relay

svyatonik:
related #194

I've been trying to start `sub-to-eth` relay locally today and there were some issues. PR can be reviewed commit-by-commit:
1) https://github.com/paritytech/parity-bridges-common/commit/00752061342da2474d31dec7d3763f3e9a6bafcd: fixes retrieval of GRANDPA authorities set, required to deploy bridge contract on PoA side;
2) https://github.com/paritytech/parity-bridges-common/commit/6b35575b57175e6036c77dabd36dec65d8cd2521: if relay has been trying to submit some headers and all headers were rejected by the contract (e.g. if they parent were unknown), relay retried the same operation without any delay. This commit adds usual retry backoff in this situation;
3) https://github.com/paritytech/parity-bridges-common/commit/a2bcdb03df56aa0c6465a91f0d65c16b23b5384c: gives funds to both Bertha and Carlos. Bertha would be signing auto-generated exchange transactions and we could use Carlos for manual tests;
4) https://github.com/paritytech/parity-bridges-common/commit/7fea754f14c0569d381d8bf18d7c0071ffa9885f: I've realized that when there's `tx1(nonce=1, data=submit(SubstrateHeader#1))` in PoA node txpool and I'm trying to submit `tx2(nonce=2, data=submit(SubstrateHeader#2))`, it'll fail because we're asking if SubstrateHeader#2 is incomplete && the call fails, because it is executed on top of best block && SubstrateHeader#1 is unknown there (and we require it for the call). So I've limited number of active PoA transactions to 1 - this means that we'll be unable to sync headers if Substrate generates headers with higher rate than PoA. The fix would be to add method to the PoA contract which will allow to import multiple headers at once. I've wanted to avoid this, because it'll complicate solidity code, but iiuc there's no other way :( Will do that in follow-up PR. P.S.: this worked well on instant seal engine because transaction was mined instantly;
5) https://github.com/paritytech/parity-bridges-common/commit/89d01deeae6b82949de2db5196f065bfe582d1a0: changes deploy-contract so that it shows id of initial Substrate header.


svyatonik:
Re 4: apparently I've tried to add the same `function importHeaders(bytes[] memory rawHeaders)` before, but passing `bytes[]` requires using `ABIEncoderv2` - it is experimental + iiuc I've had some problems using it (probably with `ethabi` crate?). So I can't add this function without introducing complex encoding + decoding stuff in solidity/builtin. So I suggest to add `function importHeaders(bytes memory rawHeader1, bytes memory rawHeader2, bytes memory rawHeader3, bytes memory rawHeader4)` - this way we could synchronize even if Substrate generate blocks 4x times faster than PoA. What do you think?

tomusdrw:
> Re 4: apparently I've tried to add the same `function importHeaders(bytes[] memory rawHeaders)` before, but passing `bytes[]` requires using `ABIEncoderv2` - it is experimental + iiuc I've had some problems using it (probably with `ethabi` crate?). So I can't add this function without introducing complex encoding + decoding stuff in solidity/builtin. So I suggest to add `function importHeaders(bytes memory rawHeader1, bytes memory rawHeader2, bytes memory rawHeader3, bytes memory rawHeader4)` - this way we could synchronize even if Substrate generate blocks 4x times faster than PoA. What do you think?

Can we use a hack and pass `bytes32[] data + uint16[] offsets`? `data` would contain all the headers concatenated and chunked into 256bits parts, while `offsets` would contain boundaries of headers? I hope it's possible to transmute `bytes32[]` into `bytes` and then chunk it with the `offsets`to get individual headers?
Also is `memory` required? I think it would be cheaper to read from transaction data directly, no?

If above is not possible, I'm fine with fixed number of headers at once.

>  even if Substrate generate blocks 4x times faster than PoA

Even with your approach I think it's fine to submit multiple transactions in one block, right? So it'd only require some extra logic in the relayer and also high enough block gas limit.

relays/ethereum/src/substrate_client.rs
```diff
@@ -187,9 +187,9 @@ impl SubstrateRpc for SubstrateRpcClient {
 
 	async fn grandpa_authorities_set(&self, block: Hash) -> Result<GrandpaAuthorityList> {
 		let call = SUB_API_GRANDPA_AUTHORITIES.to_string();
-		let data = Bytes(block.as_bytes().to_vec());
+		let data = Bytes(Vec::new());
 
-		let encoded_response = Substrate::state_call(&self.client, call, data, None).await?;
+		let encoded_response = Substrate::state_call(&self.client, call, data, Some(block)).await?;
```
tomusdrw, line 192: Just for posterity, why is `data` empty and the 4th parameter is set?

relays/ethereum/src/sync_types.rs
```diff
@@ -192,3 +188,17 @@ impl<Id, Error> Default for SubmittedHeaders<Id, Error> {
...
+
+impl<Id: std::fmt::Debug, Error> std::fmt::Display for SubmittedHeaders<Id, Error> {
+	fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+		let submitted = format_ids(self.submitted.iter());
+		let incomplete = format_ids(self.incomplete.iter());
+		let rejected = format_ids(self.rejected.iter());
+
+		write!(
+			f,
+			"Submitted: {} Incomplete: {} Rejected: {}",
```
tomusdrw, line 200:
```suggestion
			"Submitted: {}, Incomplete: {}, Rejected: {}",
```
?

relays/ethereum/src/utils.rs
```diff
@@ -51,3 +95,22 @@ pub fn retry_backoff() -> ExponentialBackoff {
...
+		0 => "<nothing>".into(),
+		1 => format!("{:?}", ids.next().expect(NTH_PROOF)),
+		2 => {
+			let id0 = ids.next().expect(NTH_PROOF);
+			let id1 = ids.next().expect(NTH_PROOF);
+			format!("[{:?}, {:?}]", id0, id1)
+		}
+		len => {
+			let id0 = ids.next().expect(NTH_PROOF);
+			let id_last = ids.nth(len - 2).expect(NTH_PROOF);
```
tomusdrw, line 112:
```suggestion
			let id_last = ids.last().expect(NTH_PROOF);
```
?

relays/ethereum/src/substrate_client.rs
svyatonik: Because block hash is the 4th argument of RPC method, not the runtime call - i.e. we call method `call` with arguments `data` at the state-od-block `Some(block)`.

svyatonik:
> Can we use a hack and pass bytes32[] data + uint16[] offsets? data would contain all the headers concatenated and chunked into 256bits parts, while offsets would contain boundaries of headers? I hope it's possible to transmute bytes32[] into bytes and then chunk it with the offsetsto get individual headers?

TBH I don't have enough experience with solidity + evm to tell you if that works right now - I mean transmute. This may worth checking, but since I've already implemented 4-headers endpoint in parallel PR, let it be possible optimization for the future. I guess migrating to updated bridge contract should be quite easy (we just need to select proper moment). 

> Also is memory required? I think it would be cheaper to read from transaction data directly, no?

We need to pass address of this header to builtin. Iiuc calldata is another address space, right? I'm not sure, though

> Even with your approach I think it's fine to submit multiple transactions in one block, right? So it'd only require some extra logic in the relayer and also high enough block gas limit.

No - we can't submit multiple transactions in one block. Ok - we actually may, but it'll be waste of gas. The reason is that when I'm going to submit `tx2`, I'm checking if contract accepts headers from this tx. If there's another `tx1` that submits ancestors, then the contract will reject `tx2` just because parents of headers are unknown to the contract. I've added this check because PoA contract stops accepting headers when it believes parent of this header needs justification. So there may be a situation when we submit `TX1(H1, H2, H3, H4)` and then immediately `TX2(H5, H6, H7, H8)` and after importing `H2` contract decides that it'll require justification for `H4` => all headers from `TX2` will be rejected => waste of gas.

tomusdrw:
>  let it be possible optimization for the future.

fair enough!

> We need to pass address of this header to builtin. Iiuc calldata is another address space, right? I'm not sure, though

Right, I think calling builtins would require copying to memory anyway.

> No - we can't submit multiple transactions in one block. 

Thanks for explanations, makes sense!

PR #250, opened 2020-07-29
Pin nightly to fix clippy.

tomusdrw:


HCastano:
I think it looks fine, but I don't know why the CI is complaining that a nightly compiler isn't installed - so maybe your change is not fine? lol

PR #252, opened 2020-07-29
Expose metrics port for exchange-tx relay.

tomusdrw:


deployments/rialto/docker-compose.yml
```diff
@@ -80,6 +80,8 @@ services:
       - ./bridge-config:/config
     environment:
       RUST_LOG: rpc=trace,bridge=trace
+    ports:
+      - "9616:9617"
```
svyatonik, line 84: Shouldn't it be "9617:9616" instead? According to [documentation](https://docs.docker.com/compose/compose-file/) first number is the port at host machine and it is already occupied by `relay-eth2sub` container (it has "9616:9616" mapping).

deployments/rialto/dashboard/prometheus/prometheus.yml
```diff
@@ -14,4 +14,4 @@ scrape_configs:
     scrape_interval: 15s
 
     static_configs:
-      - targets: ['relay-eth-exchange-sub:9616']
+      - targets: ['relay-eth-exchange-sub:9617']
```
svyatonik, line 17: But relay node in `relay-eth-exchange-sub` node still uses port 9616 to expose its prometheus metrics. So iiuc prometheus container won't be able to fetch metrics from that relay.

svyatonik:
I think I understand what you've been trying to fix here - imo instead of this ^^^ we should add `--prometheus-host=0.0.0.0` to `relay-eth-exchange-sub-entrypoint.sh`

tomusdrw:
You're right, I didn't think it through :)

PR #253, opened 2020-07-29
Prepare for sub to eth relay - part2

svyatonik:
related #194
on top of #248 (yes, it uses stuff from that PR)

Significant commits of this PR (I omit compilation fix + fmt commits):
1) https://github.com/paritytech/parity-bridges-common/commit/de1a3426e61786d76dacfcf153b9e74656637e78: I've updated solidity contract to support 4 headers import in single transaction (instead of 1 as before). Corresponding contract commit is: https://github.com/svyatonik/substrate-bridge-sol/commit/591023388322ff57f77d1810601610d4d46c392e . This would allow us to follow Substrate chain even if it generates blocks 4x times faster than PoA chain;
2) https://github.com/paritytech/parity-bridges-common/commit/7c781821a57cb373bec63ce4aa6e695f6394b1c2: improves headers sync behavior when contract rejects all headers we were going to submit (this is true when we try to submit 2nd PoA transaction while previous is not yet mined). #248 introduces backoff in this case, and this PR just pauses headers sync until that active transaction is mined (or until sync has stalled).

This resolves all the problems I've found with `sub-to-eth` synchronization => the only thing that is left is replicating last changes from https://github.com/svyatonik/parity-bridges-common.test/blob/master/start.sh to https://github.com/paritytech/parity-bridges-common/blob/master/deployments/rialto/docker-compose.yml


svyatonik:
Last commit brings [this commit](https://github.com/svyatonik/substrate-bridge-sol/commit/6456d3e016c95cd5e6d5e817c23e9e69e739aa78) which fixes some typos in contract code.

relays/ethereum/src/ethereum_client.rs
```diff
@@ -418,15 +420,27 @@ impl EthereumHighLevelRpc for EthereumRpcClient {
...
-	/// Submit given header to Ethereum node.
-	async fn submit_header(&mut self, header: QueuedSubstrateHeader) -> Result<()>;
+	/// Returns Err(()) if contract has rejected headers. This means that the contract is
+	/// unable to import first header (e.g. it may already be imported).
+	async fn is_headers_incomplete(
+		&self,
+		header1: &QueuedSubstrateHeader,
+		header2: Option<&QueuedSubstrateHeader>,
+		header3: Option<&QueuedSubstrateHeader>,
+		header4: Option<&QueuedSubstrateHeader>,
```
HCastano, line 433: I'm not a fan of this. Couldn't we package this into a collection and process up to four items at a time? We could even make the number of items processed a global constant to make it more flexible if we decide to change the number of headers processed in the future

relays/ethereum/src/sync.rs
```diff
@@ -59,6 +59,8 @@ pub struct HeadersSync<P: HeadersSyncPipeline> {
 	target_best_header: Option<HeaderIdOf<P>>,
 	/// Headers queue.
 	headers: QueuedHeaders<P>,
+	/// Pause headers submit.
```
HCastano, line 62:
```suggestion
	/// Pause headers submission.
```

relays/ethereum/src/ethereum_client.rs
svyatonik: Ok, I've tried && found it worse than passing exactly 4 headers - simply because contract api also accepts headers as 4 separate arguments. So I had to actually pack headers first in `(&QueuedSubstrateHeader, [Option<&QueuedSubstrateHeader>; 3])` then unpack them again to call the contract. With additional indexing it seems worse to me than the current approach.

tomusdrw: The thing is that it matches the Solidity contract API, IMHO it's better to keep it consistent - pretending on the API level that we can support arbitrary number of headers, while we in fact it's 1-4 will not end up well.
For convenience I'd probably implement a custom type like this:
```
Headers {
 pub header1: QueuedSubstrateHeader,
 pub header2: Option<QueuedSubstrateHeader>,
 pub header3: Option<QueuedSubstrateHeader>,
 pub header4: Option<QueuedSubstrateHeader>,
}
```
and implement some builders or `From` implementations, but it's used in only one place, so I don't feel like it's a strict requirement.

relays/ethereum/src/ethereum_client.rs
```diff
@@ -439,30 +453,54 @@ struct EthereumHeadersSubmitter {
...
 		let call_request = CallRequest {
 			to: Some(self.contract_address),
 			data: Some(encoded_call.into()),
 			..Default::default()
 		};
 
 		let call_result = self.client.eth_call(call_request).await?;
-		let is_incomplete = call_decoder.decode(&call_result.0)?;
+		let incomplete_index: U256 = call_decoder.decode(&call_result.0)?;
+		if incomplete_index > 4.into() {
```
tomusdrw, line 477:
```suggestion
		if incomplete_index > MAX_SUPPORTED_HEADERS.into() {
```
Please move to a constant.

relays/ethereum/src/ethereum_client.rs
HCastano: > So I had to actually pack headers first in (&QueuedSubstrateHeader, [Option<&QueuedSubstrateHeader>; 3])

Could we not just make everything an `Option`, even though the first header is mandatory? Then we'd be able to have an `[Option<&QueuedSubstrateHeader; MAX_SOLIDITY_HEADERS]` which we can use

HCastano: > For convenience I'd probably implement a custom type like this:
> ```
> Headers {
>  pub header1: QueuedSubstrateHeader,
>  pub header2: Option<QueuedSubstrateHeader>,
>  pub header3: Option<QueuedSubstrateHeader>,
>  pub header4: Option<QueuedSubstrateHeader>,
> }
> ```


This is slightly better, and I'd be okay with this approach as longs as the builders/`From` implementations make it somewhat ergonomic to use

relays/ethereum/src/ethereum_client.rs
```diff
@@ -410,18 +412,97 @@ impl EthereumHighLevelRpc for EthereumRpcClient {
 	}
 }
 
+pub const HEADERS_BATCH: usize = 4;
```
HCastano, line 415: Docs

relays/ethereum/src/ethereum_client.rs
```diff
@@ -410,18 +412,97 @@ impl EthereumHighLevelRpc for EthereumRpcClient {
...
+
+#[derive(Debug)]
+#[cfg_attr(test, derive(Clone))]
+pub struct Headers {
```
HCastano, line 419: Docs

I think they should mention that the Solidity contract since that's driving the design, and the reason behind that

relays/ethereum/src/ethereum_client.rs
```diff
@@ -410,18 +412,97 @@ impl EthereumHighLevelRpc for EthereumRpcClient {
...
+		[
+			encode(&self.header1),
+			self.header2.as_ref().map(encode).unwrap_or_default(),
+			self.header3.as_ref().map(encode).unwrap_or_default(),
+			self.header4.as_ref().map(encode).unwrap_or_default(),
+		]
+	}
+	/// Returns number of contained headers.
+	pub fn len(&self) -> usize {
+		let isset = |h: &Option<QueuedSubstrateHeader>| if h.is_some() { 1 } else { 0 };
```
HCastano, line 473: nitpick:

```suggestion
		let is_set = |h: &Option<QueuedSubstrateHeader>| if h.is_some() { 1 } else { 0 };
```

relays/ethereum/src/ethereum_client.rs
```diff
@@ -600,13 +707,92 @@ mod tests {
...
+				header(11).id(),
+				header(12).id(),
+				header(13).id()
+			]
+		);
 		assert!(submitted_headers.fatal_error.is_some());
 	}
+
+	#[test]
+	fn headers_methods() {
```
HCastano, line 743: I would split this into four smaller tests - one for each method

relays/ethereum/src/ethereum_client.rs
```diff
@@ -490,28 +581,31 @@ async fn submit_substrate_headers(
 	submitted_headers
 }
 
-/// Submit single Substrate header.
-async fn submit_substrate_header(
+/// Submit 4 Substrate headers in single PoA transaction.
+async fn submit_4_substrate_headers(
```
HCastano, line 585: Should this be changed to be something like `submit_batch_substrate_headers`, referring to `HEADERS_BATCH`?

relays/ethereum/src/ethereum_client.rs
```diff
@@ -32,12 +32,12 @@ use jsonrpsee::transport::http::HttpTransportClient;
...
 use parity_crypto::publickey::KeyPair;
 
-use std::collections::{HashSet, VecDeque};
+use std::collections::HashSet;
 
 // to encode/decode contract calls
 ethabi_contract::use_contract!(bridge_contract, "res/substrate-bridge-abi.json");
 
-type Result<T> = std::result::Result<T, RpcError>;
+type RpcResult<T> = std::result::Result<T, RpcError>;
```
HCastano, line 40: If we're gonna change this in the Ethereum client we should make the same change in the Substrate client for consistency

relays/ethereum/src/ethereum_client.rs
```diff
@@ -600,13 +707,92 @@ mod tests {
...
+			vec![
+				header(5),
+				header(6),
+				header(7),
+				header(8),
+				header(9),
+				header(10),
+				header(11),
+				header(12),
+				header(13),
```
HCastano, line 721: I don't think we need these to get the idea of the test across

PR #254, opened 2020-07-29
Substrate -> PoA relay in docker environment

### LINKED ISSUE #194: Setup Substrate → PoA part of the bridge

svyatonik:
closes #194
requires #248 and #253 to actually work

We'll need dashboards for this sync later - will add in follow-up PR

### DIFF
--- a/deployments/rialto/docker-compose.yml
+++ b/deployments/rialto/docker-compose.yml
@@ -119,20 +119,6 @@ services:
       - "9618:9616"
     depends_on: *all-nodes
 
-  # relay-sub2eth:
-  #   build:
-  #     context: .
-  #     dockerfile: ./Bridge.Dockerfile
-  #     args:
-  #       BRIDGE_HASH: master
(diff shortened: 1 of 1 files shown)


HCastano:
Why do we need Dave and Eve?

svyatonik:
They're [initial validators](https://github.com/paritytech/parity-bridges-common/blob/ea325c2df4f5594f9ccd0b9c103647f7e70da664/bin/node/node/src/chain_spec.rs#L100) => without Dave or Eve we won't finalize any blocks in first two sessions => will be unable to sync sub2eth. There are 5 initial validators so that we may select 3 validators starting from session 2 [here](https://github.com/paritytech/parity-bridges-common/blob/ea325c2df4f5594f9ccd0b9c103647f7e70da664/bin/node/runtime/src/lib.rs#L403).

PR #255, opened 2020-07-29
Sub2eth relay dashboard

svyatonik:
related #194

That's a copy of eth2sub dashboard with changed metrics references + labels

### DIFF
--- a/deployments/rialto/dashboard/grafana/provisioning/dashboards/relay-sub2eth-sync-dashboard.json
+++ b/deployments/rialto/dashboard/grafana/provisioning/dashboards/relay-sub2eth-sync-dashboard.json
@@ -50,7 +50,7 @@
         "frequency": "5m",
         "handler": 1,
         "message": "",
-        "name": "Rialto Block Difference > 5 (eth2sub)",
+        "name": "Rialto Block Difference > 5 (sub2eth)",
         "noDataState": "no_data",
         "notifications": []
       },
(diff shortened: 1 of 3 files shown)


deployments/rialto/dashboard/grafana/provisioning/dashboards/relay-sub2eth-sync-dashboard.json
```diff
@@ -0,0 +1,691 @@
...
+              "type": "diff_abs"
+            },
+            "type": "query"
+          }
+        ],
+        "executionErrorState": "alerting",
+        "for": "3m",
+        "frequency": "5m",
+        "handler": 1,
+        "name": "No New Blocks (sub2eth)",
```
tomusdrw, line 559: Do we need both alerts? If Substrate does not produce blocks we will see it in other metrics as well, no? I feel like 3 alerts: No PoA blocks, Block Difference (sub2eth), Block Difference (eth2sub) cover all cases and extra alerts might mean more false positives.

svyatonik: `No New Blocks` is more about relay health, not PoA/Substrate node health. Like Substrate/PoA may still continue to generate blocks, but relay has lost connection => we probably need to restart this specific relay instance (we now have 3 relays running).

svyatonik:
(I've updated "No New Blocks" alert with what we have discussed externally)

PR #256, opened 2020-07-30
Fix env issues.

tomusdrw:
Bash `.` import expects the variables to be quoted if they contain special characters (i.e. `SOME_VAR="asdf:"`), while `docker-compose` doesn't like the quotes at all.
Since we only need `MATRIX_ACCESS_TOKEN` in the script, I've added a grep here to extract it from `.env` to a temporary file and import that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 **/target/
 **/.env
+**/.env2
 
 **/*.rs.bk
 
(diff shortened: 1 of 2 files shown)


tomusdrw:
> What about $WITH_PROXY - it is used in the script as well?

Oh, good catch. I forgot about this one.

PR #258, opened 2020-07-30
Add traces to PoA builtin

svyatonik:
They're here to replace debug printlns [here](https://github.com/svyatonik/parity/blob/substrate-builtins-stubs/ethcore/builtin/src/lib.rs#L794) and below. I'll remove them once this PR is merged

### DIFF
--- a/modules/ethereum-contract/builtin/src/lib.rs
+++ b/modules/ethereum-contract/builtin/src/lib.rs
@@ -114,7 +114,11 @@ pub fn parse_substrate_header(raw_header: &[u8]) -> Result<Header, Error> {
 	log::debug!(
 		target: "bridge-builtin",
 		"Parsed Substrate header {}: {:?}",
-		hex::encode(raw_header),
+		if substrate_header.is_ok() {
+			format!("<{}-bytes-blob>", raw_header.len())
+		} else {
+			hex::encode(raw_header)
+		},
 		substrate_header,
 	);
 
@@ -136,7 +140,11 @@ pub fn verify_substrate_finality_proof(
 	log::debug!(
 		target: "bridge-builtin",
 		"Parsed Substrate authorities set {}: {:?}",
-		hex::encode(raw_best_set),
+		if best_set.is_ok() {
+			format!("<{}-bytes-blob>", raw_best_set.len())
+		} else {
+			hex::encode(raw_best_set)
+		},
 		best_set,
 	);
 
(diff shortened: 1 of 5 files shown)


modules/ethereum-contract/builtin/src/lib.rs
```diff
@@ -113,16 +131,35 @@ pub fn verify_substrate_finality_proof(
...
 	)
 	.map_err(Box::new)
 	.map_err(Error::JustificationVerify)
-	.map(|_| ())
+	.map(|_| ());
+
+	log::debug!(
+		target: "bridge-builtin",
+		"Verified Substrate finality proof {}: {:?}",
+		hex::encode(raw_finality_proof),
```
tomusdrw, line 158: Maybe full hex value should only be displayed if we are `tracing`? I supposed the output would be cleaner if we just do `0x123...beef` and display the decoded values and to see the actual hex we enable `trace`, wdyt?

modules/ethereum-contract/builtin/src/lib.rs
```diff
@@ -113,16 +131,35 @@ pub fn verify_substrate_finality_proof(
 ) -> Result<(), Error> {
 	let best_set = AuthorityList::decode(&mut &raw_best_set[..])
 		.map_err(Error::BestSetDecode)
-		.and_then(|authorities| VoterSet::new(authorities.into_iter()).ok_or(Error::InvalidBestSet))?;
-	sc_finality_grandpa::GrandpaJustification::<Block>::decode_and_verify_finalizes(
+		.and_then(|authorities| VoterSet::new(authorities.into_iter()).ok_or(Error::InvalidBestSet));
+
+	log::trace!(
```
tomusdrw, line 136: Why `trace`?

svyatonik: I was thinking that most of times we'll end up in next trace, but actually you're right

modules/ethereum-contract/builtin/src/lib.rs
svyatonik: Ok, now it is only printed when trace is enabled

svyatonik:
(I've also updated RUST_LOG in `docker-compose.yml` for PoA nodes)

modules/ethereum-contract/builtin/src/lib.rs
```diff
@@ -113,16 +131,39 @@ pub fn verify_substrate_finality_proof(
...
 	)
 	.map_err(Box::new)
 	.map_err(Error::JustificationVerify)
-	.map(|_| ())
+	.map(|_| ());
+
+	log::debug!(
+		target: "bridge-builtin",
+		"Verified Substrate finality proof {}: {:?}",
+		if log::log_enabled!(log::Level::Trace) {
```
tomusdrw, line 158: I wonder if this works correctly, since we override `target` here, looking at the code of `log_enabled` it passes `module_path!()` to the check.

Also I think we might do something similar for authorities as well.

svyatonik: OK, it doesn't work - will try to find another way :/ I'd prefer to leave header - it has helped me a lot, so imo it is important to see it even in `debug!()` mode. Will do the same for authorities

svyatonik: Ok - I'll change header as well, because it is parsing result that has helped me

svyatonik:
So the final solution is to display hex blob only if operation has failed. In all other cases we'll show something like `<64-bytes-blob>` instead. Parsed data is still shown, though, but I'd prefer to keep it

PR #260, opened 2020-07-30
Fix generator in case tty is not available.

tomusdrw:
It seems it still doesn't work because of insufficient funds though:
```
2020-07-30 12:56:52 +0000 ERROR bridge Error submitting exchange transaction to Ethereum node: error submitting transaction: Request(Request(Error { code: ServerError(-32010), message: "Insufficient funds. The account you tried to send transaction from does not have enough funds. Required 185410166640640 and got: 0.", data: None }))
```

### DIFF
--- a/deployments/rialto/bridge-config/poa-exchange-tx-generator-entrypoint.sh
+++ b/deployments/rialto/bridge-config/poa-exchange-tx-generator-entrypoint.sh
@@ -6,7 +6,7 @@
 # exchange transaction from hardcoded PoA senders (assuming they have
 # enough funds) to hardcoded Substrate recipients.
 
-set -e
+set -eu
 
 # Path to relay binary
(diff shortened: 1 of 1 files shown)


svyatonik:
It'll exit a bit later anyway here:
```
echo "Missing nonce in relay response: $SUBMIT_OUTPUT"
exit 1
```

svyatonik:
(it needs #248 to actually work - I've gave Bertha enough funds there)

tomusdrw:
> It'll exit a bit later anyway here:
> 
> ```
> echo "Missing nonce in relay response: $SUBMIT_OUTPUT"
> exit 1
> ```

How so? I think `SUBMIT_OUTPUT` will conain the correct value and the ` | tee /dev/tty` was only required to display it to the logs as well (which is now done with `set -x`), no?

PR #261, opened 2020-07-30
Fix gas_used fields in receipts

svyatonik:
After introducing Substrate -> PoA sync, most exchange transactions are not relayed to Susbtrate. I've found that this is because of receipts root mismatch (and this only happens when there are >1 tx in PoA block => that's why after sub2eth) :/ The reason is that I've been copying `Receipt::gas_used` from `EthReceipt::gas_used` field. But actually it should have been copied from `EthReceipt::cumulative_gas_used` field. I should reread yellow paper :)

I've also added more tracing to runtime - it has helped me to localize this problem.

### DIFF
--- a/modules/ethereum/src/lib.rs
+++ b/modules/ethereum/src/lib.rs
@@ -893,17 +893,40 @@ pub fn verify_transaction_finalized<S: Storage>(
 	proof: &[(RawTransaction, RawTransactionReceipt)],
 ) -> bool {
 	if tx_index >= proof.len() as _ {
+		frame_support::debug::trace!(
+			target: "runtime",
+			"Tx finality check failed: transaction index ({}) is larger than number of transactions ({})",
+			tx_index,
+			proof.len(),
+		);
+
 		return false;
 	}
 
 	let header = match storage.header(&block) {
 		Some((header, _)) => header,
-		None => return false,
+		None => {
+			frame_support::debug::trace!(
+				target: "runtime",
+				"Tx finality check failed: can't find header in the storage: {}",
+				block,
+			);
+
+			return false;
+		}
 	};
 	let finalized = storage.finalized_block();
 
 	// if header is not yet finalized => return
 	if header.number > finalized.number {
+		frame_support::debug::trace!(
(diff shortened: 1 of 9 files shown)


modules/ethereum/src/lib.rs
```diff
@@ -893,17 +893,40 @@ pub fn verify_transaction_finalized<S: Storage>(
...
+			return false;
+		}
 	};
 	let finalized = storage.finalized_block();
 
 	// if header is not yet finalized => return
 	if header.number > finalized.number {
+		frame_support::debug::trace!(
+			target: "runtime",
+			"Tx finality check failed: header {} is not finalized: {} > {}",
```
HCastano, line 924:
```suggestion
			"Tx finality check failed: header {} (#{}) - best finalized: {}",
```

Or something like this. I don't think it would be clear from the logs what `{} > {}` means

modules/ethereum/src/lib.rs
```diff
@@ -915,24 +938,60 @@ pub fn verify_transaction_finalized<S: Storage>(
...
+		);
+
 		return false;
 	}
 
 	// verify that transaction is included in the block
 	if !header.verify_transactions_root(proof.iter().map(|(tx, _)| tx)) {
+		frame_support::debug::trace!(
+			target: "runtime",
+			"Tx finality check failed: transactions root is invalid. Expected: {}",
```
HCastano, line 955: Should we also print what we got?

modules/ethereum/src/lib.rs
```diff
@@ -915,24 +938,60 @@ pub fn verify_transaction_finalized<S: Storage>(
...
+		);
+
 		return false;
 	}
 
 	// verify that transaction receipt is included in the block
 	if !header.verify_raw_receipts_root(proof.iter().map(|(_, r)| r)) {
+		frame_support::debug::trace!(
+			target: "runtime",
+			"Tx finality check failed: receipts root is invalid. Expected: {}",
```
HCastano, line 966: Same question here

PR #264, opened 2020-08-04
Fix vars substitution on update.

tomusdrw:


deployments/rialto/update.sh
```diff
@@ -10,8 +10,9 @@ git pull
 sed -i '/BRIDGE_HASH/d' .env || true
 echo "BRIDGE_HASH=$(git rev-parse HEAD)" >> .env
 
-# Update Matrix access token
-(grep -e MATRIX_ACCESS_TOKEN -e WITH_PROXY .env > .env2 && . ./.env2 && rm .env2) || true
+# Update Matrix access token (disable errors in case
```
HCastano, line 13: I don't know what you wanted to say here. We're not disabling errors, we're kinda doing the opposite :P

Issue #204, opened 2020-07-17
HeaderChain trait


tomusdrw:
Define & implement the `HeaderChain` interface to cover current and future bridge pallets. Implement `HeaderChain` for PoA bridge pallet. Proposed interface:
```rust
fn import_header_unsigned(Self::Header, Self::Extra);
fn import_headers_signed((Self::Header, Self::Extra)[]);
fn best_block() -> Self::Header;
fn finalized_block() -> Self::Header;
fn earliest_block() -> Self::Header; // earliest useful for proofs
fn by_number(Self::BlockNumber) -> Self::Header;
fn by_hash(Self::BlockHash) -> Self::Header;
fn verify_proof(Self::Proof) -> bool;
// I imagine Proof to be an enum in case multiple things can be proven
// (Transaction, Event, MultipleBlocks, ValidatorSet, etc)
```

Triggered hooks:
```rust
fn on_valid_headers(
	Option<Submitter>, useful: Self::Header[], useless: Self::Header[]
);

fn on_invalid_headers(Option<Submitter>);
```

`Proof` for each implementation will most likely be an enum containing different things that can be proven (think `Transaction` or `Event`, etc). For that reason it will probably be required to have traits that allow proof creation, like so:
```rust
trait TransactionProof<T> { type ProofData; fn new(T::BlockHash, Self::ProofData) -> Self }

trait EventProof<T> { type ProofData; fn new(T::BlockHash, Self::ProofData) -> Self }
```

So that for instance `CurrencyExchange` (see #205) can specify someting like this:
````rust
/// Blockchain is a trait that is required by CurrencyExchange currently.
impl<T: HeaderChain> Blockchain for T where
 T::Proof: TransactionProof<T>,
{
...
}

```

tomusdrw:
Closed via #287

PR #265, opened 2020-08-05
Lay Groundwork for Header Chain Traits

HCastano:
This PR starts addressing #204. The goal is to define an interface which, if followed, can be used as the common interface for bridge pallets. Additionally, I want to move our existing PoA Ethereum bridge pallet to use this new interface (or parts of it, I'll explain more in a second) in this PR. 

By having a common interface it will make the process of developing bridge pallets for other chains much easier, and will allow for other utility pallets - such as the currency exchange pallet we've developed - to be re-used across bridges.

I've defined two traits: `FinalityHeaderChain` and `FullHeaderChain`. The finality related trait should be the bare minimum a pallet needs to implement to be able to follow an external chain. It wiil be used for importing headers which are allegedly finalized, and verifying their finality.

The full header trait would be a superset of the finality trait. It would be use to track _all_ the headers of a chain, regardless of whether or not they've been finalized. This is useful for chains which use Grandpa finality, where finalized blocks can be hundreds of blocks apart from each other. Applications which would like to build on the bridge but require these intermediate blocks would benefit from having this interface.

Since I said I'd get back to it later, I think some of the methods defined by the finality trait might be hard for the Ethereum pallet to follow in its current state. I'm not sure if a  large refactoring is a worthwhile cause at this point so I'll have to see how it goes as I work through this PR.


modules/header-chain/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Defines traits which represent a common interface for Substrate pallets which want to
+//! incorporate bridge functinality.
+
+use frame_support::dispatch::DispatchResult;
+
+/// A base trait for tracking finalized headers on foreign chains in a Substrate pallet. Should be
+/// used as a primitive for more complicated header tracking logic.
+pub trait FinalityHeaderChain<AccountId> {
```
tomusdrw, line 24: I think the crate should rather be in `primitives` not in `modules`.

modules/header-chain/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+	/// The header used by the chain.
+	type Header;
+	/// Any extra data which helps describe a transaction.
+	type Extra;
+	/// The type of transaction proof used by the chain.
+	type Proof;
```
tomusdrw, line 30: The `Proofs` are going to be different for `FullHeaderChain` and `FinalityHeaderChain`. IMHO it's a bit too early to split these two, we don't really know exactly how it would work, so I'd rather defer this split until we have more knowledge.

modules/header-chain/src/lib.rs
HCastano: I thought about this and was a little conflicted. Ultimately I decided not to put in `primitives` because it's not used by the relay at all - it's only used by pallets.

HCastano:
I've ended up not making `FullHeaderChain` a strict extension of `FinalityHeaderChain`. This is because I don't think the Ethereum pallet will be able to implement `FinalityHeaderChain` in its current state.

One thing I've been struggling with is trying to figure out if `FullHeaderChain`'s interface should support forks, or if the interface should only return blocks which have been confirmed as part of the canonical/finalized chain.

When requesting the earliest known block for example, we can end up in a scenario where we have multiple blocks at the same height which haven't yet been finalized or pruned. Should we return both, or wait until one is finalized or pruned? I'm leaning towards returning only the single block that's part of the finalized chain, but I'm worried about the overhead costs associated which checking for ancestry in every call.

modules/header-chain/src/lib.rs
tomusdrw: I like the split we have in substrate, where `primitives` are things that are cheap to import since they mostly define traits + data types and are often shared. That's my reasoning to put it there :)

modules/header-chain/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "header-chain"
+description = "A common interface for describing what a bridge pallet should be able to do."
+version = "0.1.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+# TODO: I'm pretty sure this should _just_ be GPL3
```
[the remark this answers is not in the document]
HCastano: I still have some doubts about this, but I need to look into it further (at a later point in time)

HCastano:
So I've scaled back this PR a bit, mainly due to ambiguity around what the bridge interfaces should look like. Basically what it's doing now is implementing one trait: `MininalHeaderChain`. This trait is supposed to be a prototype of what a chain which tracks all headers of a foreign chain should be able to do/what kind of data it should store.

I think that it would be better to develop these traits alongside the Substrate <-> Substrate bridge pallet. This would give us a chance to tailor the pallet to what we think the interfaces should look like, and through the implement check our assumptions.

bin/node/runtime/src/lib.rs
```diff
@@ -564,8 +565,8 @@ impl_runtime_apis! {
 		}
 
 		fn finalized_block() -> (u64, sp_bridge_eth_poa::H256) {
-			let finalized_block = BridgeRialto::finalized_block();
-			(finalized_block.number, finalized_block.hash)
+			let finalized_block = BridgeRialto::best_finalized_header();
+			(finalized_block.number, finalized_block.compute_hash())
```
tomusdrw, line 569: `compute_hash` might be expensive, I think the bridge should be able to return that cheaply as well.
Maybe it's not super important in that context (`RuntimeApi`), but the fact that the relay needs this information indicates that it might be worth to extend the `MinimalHeaderChain` to provide that as well.

primitives/header-chain/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+	/// which are at the same height.
+	fn best_headers() -> Vec<Self::Header>;
+
+	/// Get the best finalized block the pallet knows of.
+	fn best_finalized_header() -> Self::Header;
+
+	/// Get a specific block from the pallet given its number.
+	///
+	/// Will return None if this block is not part of the canonical chain tracked by the pallet.
+	fn header_by_number(block_hash: Self::BlockNumber) -> Option<Self::Header>;
```
tomusdrw, line 58: Is this strictly necessary? The implementation is pretty heavy, so maybe best to avoid it?

primitives/header-chain/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+		submitter: AccountId,
+		header: Self::Header,
+		extra_data: Option<Self::Extra>,
+	) -> DispatchResult;
+
+	/// Get the best block the pallet knows of.
+	///
+	/// May return multiple blocks if we have unfinalized blocks on different forks
+	/// which are at the same height.
+	fn best_headers() -> Vec<Self::Header>;
```
tomusdrw, line 50: What would be the use case for this API? How would bridge application use this information?

modules/ethereum/src/lib.rs
```diff
@@ -835,6 +957,26 @@ impl<T: Trait<I>, I: Instance> Storage for BridgeStorage<T, I> {
 		// and now prune headers if we need to
 		self.prune_blocks(MAX_BLOCKS_TO_PRUNE_IN_SINGLE_IMPORT, finalized_number, prune_end);
 	}
+
+	fn blocks_by_number(&self, block_number: u64) -> Vec<Header> {
```
tomusdrw, line 961: If we get rid of `best_blocks()` this should not be needed, since the method is pretty heavy.

primitives/header-chain/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+	type Header;
+	/// Any extra data which helps describe a transaction.
+	type Extra;
+	/// The type of block number used by the chain.
+	type BlockNumber;
+	/// The type of block hash used by the chain.
+	type BlockHash;
+
+	/// Imports a header submitted using an _unsigned_ transaction to the pallet.
+	fn import_header_unsigned(header: Self::Header, extra_data: Option<Self::Extra>) -> DispatchResult;
```
tomusdrw, line 37: Why not have a single method for both signed and unsigned? This interface is higher level, so it really shouldn't care about stuff like "transaction" any more or signatures (especially that they are nowhere present).

If the implementation does not accept headers without submitter it will simply return an `Err`.

```
import_header(
  submitter: Option<AccountId>,
  header: Self::Header,
  extra_data: Option<Self::Extra>
)
```
?

primitives/header-chain/src/lib.rs
HCastano: The use case here would be to see if the block you're about to submit is newer than the best block the bridge knows of. If is it not newer then it would not be a good idea to submit it since it could get flagged as "useless" during the signed header import. We don't do anything with this information at the moment, but further down the line this could be used as a sort of reputation mechanism for peers

HCastano: Actually, at the moment that can be accomplished with `best_block()`, so maybe `best_headers()` is not required

HCastano:
This is heading in an unclear direction. Going to try and rework it with a series of different PRs, starting with #287.

PR #267, opened 2020-08-05
Publish Bridge Node and Relay Docker Images

General-Beck:
Publish Docker image in our technical hub

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -1,7 +1,6 @@
 name:                                               Publish images to Docker hub
 
 on:
-  pull_request:                                     #TODO remove before merge
   push:
     tags:
       - v*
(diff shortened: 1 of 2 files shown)


tomusdrw:
From this one top-level `Dockerfile` we actually want to build two different images by changing `build-args` `PROJECT`:
1. `bridge-node`
2. `ethereum-poa-relay`

Also should probably only build it on `tags` and some rolling `nightly` target (once a day).

General-Beck:
should I make this a separate workflow? for example, publish.yml

tomusdrw:
@General-Beck I think it should be part of the regular workflow, because it makes `Building` job reduntant. Both `Dockerfile` and this job build a binary and it will just do it twice. I think we should optimize this be either:
1. Building the binary on CI and then creating a dockerfile that contains that particular binary simply by copying it (instead of building). This will require matching environment between CI image and the runtime Docker image.
2. Building the binary within Docker and then copying it from resulting docker image to store in artifacts as well.

That said, I'm now having second thoughts that maybe it's too complicated and we can accept the double time, because:
1. Binaries are produced only on `master`
2. Docker images are supposed to be built on `tags` and `nightly`
So it seems there is no overlap between these two.

So finally: Let's add it as a separate workflow, but make sure that:
1. We run only on `tags` and `nighlty`
2. We produce two images as mentioned earlier (`bridge-node`, `ethereum-poa-relay`).

General-Beck:
for the last nightly build, we will use the `latest` _(paritytech/bridge-node:latest)_and the `date` tag _(paritytech/bridge-node:nightly-06-08-2020)_. Will that be enough?

tomusdrw:
> @tomusdrw for the last nightly build, we will use the `latest` _(paritytech/bridge-node:latest)_ and the `date` tag _(paritytech/bridge-node:nightly-06-08-2020)_. Will that be enough?

Yup, perfect!

PR #271, opened 2020-08-06
Rename Primitive Crates

HCastano:
Right now our primitive crates are named using the `sp-` prefix convention. This comes from Substrate, since that's where some of these crates started their life off. However, I think we should move to using a `bp-` prefix to indicate they're from the bridge primitives crates.


relays/ethereum/src/substrate_types.rs
```diff
@@ -18,11 +18,11 @@ use crate::ethereum_types::{
 	Header as EthereumHeader, Receipt as EthereumReceipt, HEADER_ID_PROOF as ETHEREUM_HEADER_ID_PROOF,
 };
 use crate::sync_types::{HeaderId, HeadersSyncPipeline, QueuedHeader, SourceHeader};
-use codec::Encode;
-pub use sp_bridge_eth_poa::{
+pub use bp_eth_poa::{
 	Address, Bloom, Bytes, Header as SubstrateEthereumHeader, LogEntry as SubstrateEthereumLogEntry,
 	Receipt as SubstrateEthereumReceipt, TransactionOutcome as SubstrateEthereumTransactionOutcome, H256, U256,
 };
+use codec::Encode;
```
tomusdrw, line 25: This should go back up, so that `pub` stuff is separate.

PR #276, opened 2020-08-07
Use Images from DockerHub in Rialto Docker-Compose

### LINKED ISSUE #272: Update Docker-Compose to use DockerHub images
Once #267 goes in we should update our `docker-compose` file to pull images from the hub instead of pulling `master` directly. I think this will help speed builds up.

Although, we should make sure there's still some way to pull from GitHub if need-be (maybe for testing a specific branch).

HCastano:
Since we're publishing images to the Docker Hub it makes sense for our "production" deployment pipeline to use those releases instead of building from source every time. I've published the OpenEthereum fork we have under my own Docker Hub account right now mostly out of convenience. If we want to move this under the `paritytech` account that would be fine with me.

I have also added an [override file](https://docs.docker.com/compose/extends/#adding-and-overriding-configuration) for building the compose setup from Git. You can use that config like this:

```bash
docker-compose -f docker-compose.yml -f docker-compose.git.yml build
```

If you want to do a sanity check of the merged config you can also do the following:

```bash
docker-compose -f docker-compose.yml -f docker-compose.git.yml config > docker-compose.merged.yml
```

Note that order is important when it comes to config files, and swapping the order around may result in unexpected configs.

I've also updated the `update.sh` script. By default it'll pull the latest Docker image from the Docker Hub. It now supports a `-g` flag, which if enabled will pull changes from GitHub and build with the Git override config file.

Closes #272.


General-Beck:
if we add openethereum as a submodule, I can add it to the docker hub.

HCastano:
Good point on the README instructions, I'll add those in. As far as working with Git submodules goes, I'd really like to avoid them, especially if it's just to publish Docker images. I'll just manually update the image every now and then. That repo doesn't move very quick anyways

General-Beck:
when we make release, then I think we should use the [openethereum](https://hub.docker.com/r/openethereum/openethereum) images.

HCastano:
> when we make release, then I think we should use the [openethereum](https://hub.docker.com/r/openethereum/openethereum) images.

We can't because we need to use the fork of OpenEthereum with the Grandpa built-in contract for the Sub -> Eth side of the bridge

deployments/rialto/docker-compose.yml
```diff
@@ -144,6 +117,10 @@ services:
       - ./bridge-config:/config
     environment: &bridge-env
       RUST_LOG: runtime=trace,rpc=debug,txpool=trace
+    healthcheck:
```
tomusdrw, line 120: That shouldn't be necessary. The build arg should be passed when image is being built.

deployments/rialto/update.sh
```diff
@@ -1,24 +1,40 @@
...
 # Script used to periodically update the network.
+#
+# If used with no flags then it updates the network using Docker
+# images from the Docker Hub. If used with the `-g` a.k.a the Git
+# flag then it'll build the Docker containers from GitHub.
 
 set -xeu
 
-git pull
+git_flag=false
```
tomusdrw, line 11: Can we stick to upper-cased vars?

deployments/rialto/update.sh
```diff
@@ -1,24 +1,40 @@
...
 }
 
-case "${1:-all}" in
-	all) update_hash "NODE_BRIDGE_HASH"; update_hash "ETH_BRIDGE_HASH"; update_hash "RELAY_BRIDGE_HASH";;
-	node) update_hash "NODE_BRIDGE_HASH";;
-	relay) update_hash "RELAY_BRIDGE_HASH";;
-	eth) update_hash "ETH_BRIDGE_HASH";;
-	*) echo "Invalid parameter: $1 (expected all/node/relay/eth)"; exit 1;;
-esac
+while getopts 'g:' flag; do
```
tomusdrw, line 19: I think we can avoid `getops` and simply add another branch to `case`, no?

tomusdrw: Actually do we really need a flag? Maybe let's control that with `.env` file, like we do with `WITH_PROXY` and other hashes? The reason for that would be that you most likely want the same behaviour on every run (i.e. always build git or from images).

deployments/README.md
```diff
@@ -35,19 +35,42 @@ test Substrate network.
...
 
+### Docker-Compose and Git Deployment
+
+It is also possible to avoid using images from the Docker Hub and instead build
+containers from GitHub. This can be done using an override file for Docker Compose. To
+build the containers you can do the following:
+
+```bash
+cd rialto
+docker-compose -f docker-compose.yml -f docker-compose.git.yml build
```
tomusdrw, line 61: Nice! Thanks for the docs on this.

deployments/rialto/docker-compose.yml
HCastano: This isn't part of the build: https://github.com/paritytech/parity-bridges-common/blob/master/.github/workflows/publish-docker.yml#L50

Between having it (kinda hidden) in a Github Actions workflow and here I'd prefer to have it here, but I don't have a strong opinion on this

deployments/rialto/update.sh
HCastano: Yeah I can give that a try. I think it sounds reasonable

HCastano:
I'm not sure if the changes I made to the Deploy action are correct, so @General-Beck do you mind taking another look? Feel free to push changes to the workflow if it's incorrect.

deployments/rialto/docker-compose.yml
tomusdrw: If it's part of the image everyone can benefit from it since it doesn't require any extra configuration.

.github/workflows/publish-docker.yml
```diff
@@ -18,6 +18,9 @@ jobs:
         project:
           - bridge-node
           - ethereum-poa-relay
+        healthcheck:
```
tomusdrw, line 21: I think this will create all possible combinations. We should rather change `HEALTH` based on `matrix.project`

tomusdrw (approved):

tomusdrw (approved):
lgtm!

PR #281, opened 2020-08-12
Retry Connections on Relay Startup

HCastano:
At the moment if we start the relayer without a Substrate or Ethereum node running it will block the current thread forever. Even if a node is brought up afterwards it isn't able to connect to it since it never retries the connection. This PR aims to implement some sort of re-try mechanism for the relayer so we don't end up stuck forever.

I've been seeing this with our Rialto deployment when running the relayer in various configurations. This is because while `docker-compose` spins up nodes in the "correct" order the relayers may occasionally be brought up too early.

### DIFF
--- a/relays/ethereum/src/instances.rs
+++ b/relays/ethereum/src/instances.rs
@@ -42,7 +42,7 @@ pub trait BridgeInstance: Send + Sync + std::fmt::Debug {
 	fn build_currency_exchange_call(&self, proof: Proof) -> Call;
 }
 
-/// TODO
+/// Different instances supported by the relay.
 #[derive(Debug, Clone)]
 pub enum SupportedInstance {
 	Rialto,
--- a/relays/ethereum/src/instances.rs
+++ b/relays/ethereum/src/instances.rs
@@ -42,6 +42,22 @@ pub trait BridgeInstance: Send + Sync + std::fmt::Debug {
 	fn build_currency_exchange_call(&self, proof: Proof) -> Call;
 }
 
+/// TODO
+#[derive(Debug, Clone)]
+pub enum SupportedInstance {
+	Rialto,
+	Kovan,
+}
+
+impl From<&SupportedInstance> for Box<dyn BridgeInstance> {
+	fn from(i: &SupportedInstance) -> Self {
+		match i {
+			SupportedInstance::Rialto => Box::new(Rialto),
+			SupportedInstance::Kovan => Box::new(Kovan),
+		}
+	}
+}
+
 /// Corresponds to the Rialto instance used in the bridge runtime.
 #[derive(Default, Clone, Debug)]
 pub struct Rialto;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -376,7 +376,10 @@ version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "721c249ab59cbc483ad4294c9ee2671835c1e43e9ffc277e6b4ecfef733cfdc5"
 dependencies = [
+ "async-std",
+ "futures-core",
  "instant",
+ "pin-project",
  "rand 0.7.3",
 ]
 
@@ -7406,7 +7409,7 @@ version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3bfd5b7557925ce778ff9b9ef90e3ade34c524b5ff10e239c69a42d546d2af56"
 dependencies = [
- "rand 0.7.3",
+ "rand 0.3.23",
 ]
 
 [[package]]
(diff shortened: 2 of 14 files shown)


relays/ethereum/src/instances.rs
```diff
@@ -42,6 +42,22 @@ pub trait BridgeInstance: Send + Sync + std::fmt::Debug {
 	fn build_currency_exchange_call(&self, proof: Proof) -> Call;
 }
 
+/// Different instances supported by the relay.
+#[derive(Debug, Clone)]
+pub enum SupportedInstance {
+	Rialto,
+	Kovan,
+}
```
HCastano, line 50: When I introduced the trait objects for the supported instances we talked about not having an `enum`. However, due to ownership issues caused by the closure used for the retry mechanism I needed a way to create a new trait object during each retry, which lead me to introducing this `enum`.

It's not elegant, but I couldn't figure out another way to please the compiler.

HCastano:
> I'd rather crash the relay and then have a supervisor (or entrypoint script) restart it automatically. Since the relay is designed to recover from being temporary offline, we are not really loosing any functionality here by not implementing reconnections within the relay code.

I disagree with this. The main reason is that we're assuming that relays will always be run via Docker Compose. This isn't an assumption that should be made since it affects the ability for our code to recover gracefully from connection problems _in general_.

> we need to try-to-reconnect not only during initialization, but instead on every transport error.

Yes, I agree with this. I didn't want to complicate this PR with that though, as most of the issues I've noticed happened at start up. 

> I'm still not sure whether we'll need that code after the actual issue will be solved

Yeah, this is also unclear. Since I know you're working on a patch for `jsonrpsee` let's get this fixed upstream and then come back to this PR and see how much is still useful.

tomusdrw:
> I disagree with this. The main reason is that we're assuming that relays will always be run via Docker Compose. This isn't an assumption that should be made since it affects the ability for our code to recover gracefully from connection problems in general.

No we don't assume it's run via Docker Compose. There is no mechanism in compose that would auto-restart the node (our entrypoint could do that, sure).
This is a general concept of [Fail-fast](https://en.wikipedia.org/wiki/Fail-fast) approach, where you simply don't even attempt to handle some class of errors, because it's simply not worth it. Similar to how we don't handle OOM, out of disk, serialization/encoding errors, not enough threads etc. Being able to connect to a running & functional & fully synced blockchain client is a strict requirement of the relay, I don't see much point of adding a dozens lines of code to handle this situation with imho zero gains for the system as a whole - as I've mentioned earlier relayer is fine to go offline and resume operation at any point in time, in production environment we might even run multiple instances of the relayer, just for the sake of redundancy.

HCastano:
> No we don't assume it's run via Docker Compose. There is no mechanism in compose that would auto-restart the node (our entrypoint could do that, sure).
> This is a general concept of [Fail-fast](https://en.wikipedia.org/wiki/Fail-fast) approach, where you simply don't even attempt to handle some class of errors, because it's simply not worth it. Similar to how we don't handle OOM, out of disk, serialization/encoding errors, not enough threads etc. Being able to connect to a running & functional & fully synced blockchain client is a strict requirement of the relay, I don't see much point of adding a dozens lines of code to handle this situation with imho zero gains for the system as a whole - as I've mentioned earlier relayer is fine to go offline and resume operation at any point in time, in production environment we might even run multiple instances of the relayer, just for the sake of redundancy.

Yeah, so it makes sense that since the relay is "stateless" it doesn't really matter if it crashes while relaying headers since we can restart it and pick back up without an issue. However, I'd still like to see this handled gracefully not because it's good for the system but because it's good for those that are going to run this infrastructure in the future. Maybe asking people to run an extra supervisor script isn't a big stretch, but from my point-of-view it seems like added complexity if we can handle it from within the relay. 

Maybe at start up it makes sense to crash and restart, but during operation we should at leave have some sort of retry mechanism. If we don't recover after a certain number of attempts then we can crash the node. What do you think about that approach?

tomusdrw:
> because it's good for those that are going to run this infrastructure in the future. Maybe asking people to run an extra supervisor script isn't a big stretch, but from my point-of-view it seems like added complexity if we can handle it from within the relay.

I don't see how that's any good. If you kill the target node I really don't understand why relay should continue working. Same if the host runs out of disk space, is disconnected from the internet, has no RAM or file pointers left - the relay doesn't need to operate under such conditions, I can't see any reason why we would need to add complexity to the code if the solution is simply to restart it. Note that we are not talking about a regular situation here - blockchain node going offline is definitely unusual, since they are designed to be up 24/7.
IMHO taking care of system uptime on the infrastructure level is outside of the scope of this project. This can be simply solved either by:
1. Ensuring proper operation (i.e. restarting the relay in case of networking failure)
2. Implementing a re-connecting jsonrpsee transport layer, which transparently for our app will re-try requests.
Putting so low-level layer logic in high-level relay code, which should be dealing with domain-specific errors is imho wasted effort.

> Maybe at start up it makes sense to crash and restart, but during operation we should at leave have some sort of retry mechanism.

I'd rather accept the opposite: we retry connection initially but if there is any error during operation we crash. The reason for this that if you get an error during operation you can't be sure if the internal state is consistent on either side. Imagine that we start using subscriptions in the relay at some point, re-subscribing after disconnect would require tremendous amount of effort and code which is completely not worth the gains.

That said, I feel we've wasted too much time discussing this already, so I'm not going to object any more if you wish to include some sort of retry mechanism.

HCastano:
I haven't decided how to move forward with this yet. I'll make sure something is done, such as closing this PR and proposing a new one, by the end of the week.

HCastano:
Closing in favour of #318.

PR #287, opened 2020-08-17
Introduce Minimal Header Chain: Proving Interface

HCastano:
Supersedes #265. 

This takes another approach at the minimal headerchain trait. The current trait is just a proving interface for transaction proofs (previously known as `PeerBlockchain`), but it will grow in the future as we implement the Substrate bridge pallet. Maybe `MinimalHeaderChain` isn't the right name for the trait at the moment, but I'm being a bit forward thinking here.

### DIFF
--- a/primitives/header-chain/src/lib.rs
+++ b/primitives/header-chain/src/lib.rs
@@ -33,5 +33,3 @@ pub trait MinimalHeaderChain {
 	/// Returns Some(transaction) if proof is valid and None otherwise.
 	fn verify_transaction_inclusion_proof(proof: &Self::TransactionInclusionProof) -> Option<Self::Transaction>;
 }
-
-
--- a/primitives/header-chain/src/lib.rs
+++ b/primitives/header-chain/src/lib.rs
@@ -19,10 +19,19 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
-use frame_support::Parameter;
+use core::clone::Clone;
+use core::cmp::Eq;
+use core::fmt::Debug;
+use parity_scale_codec::{Codec, EncodeLike};
+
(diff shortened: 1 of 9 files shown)


primitives/header-chain/src/lib.rs
```diff
@@ -0,0 +1,35 @@
...
+
+//! Defines traits which represent a common interface for Substrate pallets which want to
+//! incorporate bridge functionality.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use frame_support::Parameter;
+
+/// A base trait for pallets which want to keep track of a full set of headers from a bridged chain.
+pub trait MinimalHeaderChain {
```
svyatonik, line 25: `BaseHeaderChain`, `HeaderChainBase`, `HeaderChainEssence`? :)

primitives/header-chain/src/lib.rs
```diff
@@ -0,0 +1,35 @@
...
+	/// Transaction type.
+	type Transaction: Parameter;
```
tomusdrw, line 27: Are these bounds needed? Can we get away with no `frame-support` dependency?

HCastano: Yeah they are needed. The Substrate macros complain if we don't have them

tomusdrw: Well it probably compliants about a bunch of traits, but not this specific one :) We can simply import `Codec + EncodeLike + Clone + Eq + core::fmt::Debug` and it should have the same effects without extra `frame_support` dependency, which surprisingly is pretty heavy.

HCastano: Lol, kinda gross but sure I can change that

PR #288, opened 2020-08-18
Update jsonrpsee reference

svyatonik:
This PR brings updated `jsonrpsee` with fixed 100% CPU usage + adds transport errors propagation. All changes are in [this PR](https://github.com/svyatonik/jsonrpsee/pull/1).

I've also updated our CPU-related widgets a bit - they were showing average CPU usage over last 5m (iiuc). If everything would have worked perfectly, I'd prefer them to show CPU usage at this moment (which is just `"instant": true`). But this leads to issues with Grafana UI - sometimes it stops showing widget with error like ".toFixed() is not defined". So the solution is to show average CPU load for last 1m (instant=true + range=1m). For ranges less than 10s it sometimes shows "No Data" => I've left 1m

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2558,7 +2558,7 @@ dependencies = [
 [[package]]
 name = "jsonrpsee"
 version = "1.0.0"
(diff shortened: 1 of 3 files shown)


tomusdrw:
LGTM! Let's try it out!

PR #290, opened 2020-08-18
Update open ethereum.

tomusdrw:
Fixes `serde` resolution issues and allows overriding the hash via `.env`.


HCastano:
We'll need to publish some new Docker images from this repo, but that can be done later.

PR #291, opened 2020-08-18
Reduce PoA block time

tomusdrw:


tomusdrw:
Closing, most likely unnecessary.

PR #296, opened 2020-08-20
Finality Header Chain Pallet

HCastano:
This PR introduces the skeleton of a finality header chain pallet. The idea behind this pallet is that it's a "dumb" pallet which only cares about being able to do two things:

1. That a header is valid.
2. That a header has been finalized.

This pallet doesn't, and shouldn't, care about the specifics of what the means. In theory if I want to have a blockchain where blocks are valid as long as they have a block number of 42, and are finalized if they have a seal of `b"Nando"` then that should be fine.

The storage layout is very simple here as well as it consists of just the finalized blocks which form the finalized header chain.

I'm hoping that this can be a basic building block for more complex pallets. This can either be done by refactoring and extending this pallet as time goes or, or just using its primitives (the header and finality verifiers) as part of a larger crate. This larger crate could track all the headers and forks of a given chain, and delegate all the finality tracking to this pallet.


primitives/header-chain/src/lib.rs
```diff
@@ -42,3 +42,22 @@ pub trait BaseHeaderChain {
...
+	fn validate_header(header: &Self::Header, extra_data: &Option<Self::Extra>) -> bool;
+}
+
+/// A trait for verifying that a given header has been finalized.
+pub trait FinalityVerifier {
+	type Header: Parameter;
+	type Proof: Parameter;
+
+	/// Verify that the given header has been finalized and is part of the canonical chain.
+	fn verify_finality(header: &Self::Header, proof: &Self::Proof) -> bool;
```
tomusdrw, line 62: Any non-trivial finality verification will require storage access to:
1. Proof ancestry (we build on the previously finalized chain)
2. Check current validator set.

I've made some small proposal how we could handle that in the issue (just expanded it):
```rust
// contains instructions to the upper module what should happen with the headers.
struct ImportOutcome<HeaderId> {
 finalize: Vec<HeaderId>,
 enact: Vec<HeaderId>,
 retract: Vec<HeaderId>,
}

/// T type will give us access to underlying header storage (might be a different trait)
trait ChainVerifier<T: HeaderChain> {
/// Header type
type Header;
/// Extra data for header import (may contain finality proof, receitps, etc)
type HeaderImportExtraData;
/// A way to access custom storage items that are needed by this particular implementation
type CustomStorage;
/// Supported proof kinds
type Proof;
/// A proof error type
type ProofError: core::fmt::Debug;

/// basic (cheap) verification of the header
fn verify_header(Self::Header, Option<Self::HeaderImportExtraData>, &Self::CustomStorage) -> bool;
/// import header against latest state - this should verify finality internally already.
fn import_header(Self::Header, Option<Self::HeaderImportExtraData> &mut Self::CustomStorage) -> ImportOutcome;
/// Check proof correctness (custom storage mutable to allow for caches)
fn verify_proof(Self::Proof, &mut Self::CustomStorage) -> Result<(), Self::ProofError>;
}


/// And the implementation could look like this:
mod impl {
  trait SubstrateExtraStorage<C: ChainVerifier> {
    fn insert_authority_set(&mut self, ...);
  }

  enum Proof {
    TransactionProof(Vec<u8>),
  }

  struct SubstrateLikeChain;
  impl<T: HeaderChain> ChainVerifier<T> for SubstrateLikeChain {
    type Header = sp_core::generic::Header<...>;
    type HeaderImportExtraData = GrandpaFinalityProof;
    type CustomStorage = Box<dyn SubstrateExtraStorage<Self>>;
    type Proof = Proof;
    type ProofError = String;
  
    fn verify_header(h: Self::Header, Option<Self::HeaderImportExtraData>, &Self::CustomStorage) -> bool {
      // Check that it's latest (although that could done externally in the generic bridge pallet, but it's header to show how to access the storage)
      if T::last_finalized_number() > h.number() {
         false
      } else {
         true
      }
    }
    fn import_header(Self::Header, Option<Self::HeaderImportExtraData>, s: &mut Self::CustomStorage) -> ImportOutcome {
         // TODO: verify header, finality proof (optional), check the authority set, enact the authority set if it changes,
         /// mark some more blocks as finalized, etc.
         s.insert_authority_set(...);
         ImportOutcome {
            ...
         }
    }
    fn verify_proof(proof: Self::Proof, p: &mut Self::CustomStorage) -> Result<(), Self::ProofError> {
          let { header_hash, transaction_proof } = extract_tx_proof(proof)?;
          let header = T::header_by_hash(header_hash);
          if verify_merkle_proof(header.extrinsics_root, transaction_proof) { 
             Ok(())
          } else {
             Err("Invalid Proof")
          }
    }
  }
}


```
I don't think that splitting `HeaderVerifier` and `FinalityVerifier` is reasonable, especially in it's current form, since we do both of them anyway.

HCastano:
I've made some progress on the pallet, but I have some questions:

1. Has there been any progress in a `no_std` way to verify Grandpa proofs? 

I don't recall seeing anything in the last couple of months, but maybe I've missed something. I thought our `built-in` crate 
looked promising, but it's using `sc-finality-grandpa` crate to verify proofs, which unfortunately isn't `no_std`.

2. Does the `BridgeStorage` trait generally make sense?

I know it's a kinda wimpy wrapper around our storage right now, but I'm traitifying it with future use cases in mind. Does the API _generally_ make sense?

3. Is the finality check flow in `Verifier::verify_finality` alright?

Basically is Finality -> Ancestry -> Authorities the correct way to going through the finality check pipeline. Are there any other steps that are missing?

4. Am I missing any basic checks in `Verifier::import_header`?

5. How should I bootstrap the initial authority set?

I'm thinking that I have a hardcoded set of authorities which are configured by the runtime, kinda like what the Ethereum pallet does with the Aura authorities. Asking mainly to see if you guys have any different ideas.

Also, I know that the code is a bit messy at the moment, so don't focus too much on the code quality aspects just yet. I'm looking for more high level feedback, lol.

@svyatonik @tomusdrw

modules/substrate/src/verifier.rs
```diff
@@ -0,0 +1,472 @@
...
+		if parent_header.is_none() {
+			return Err(ImportError::MissingParent);
+		}
+
+		// A block at this height should come with a justification and signal a new
+		// authority set. We'll want to make sure it is valid
+		//
+		// This defaults to 0, should it maybe be an Option?
+		let scheduled_change_height = storage.scheduled_set_change().height;
+		if *header.number() == scheduled_change_height && finality_proof.is_some() {
```
svyatonik, line 89: Looks like you do not support forks. Imagine following simple chain:
```
        / -> B2'
B1 -   -> B2
```

Let's say both `B2` and `B2'`are signalling authorities set change. But `B2` signals it in `10` blocks, while `B2'` signals it in `100`blocks. Seems like your storage only supports one signal. But instead imo it is easier to implement linked list on top of headers hierarchy. Something like `prevSignalHeaderHash` and `prevSignalTargetNumber` [here](https://github.com/svyatonik/substrate-bridge-sol/blob/6456d3e016c95cd5e6d5e817c23e9e69e739aa78/substrate-bridge.sol#L73). This way you could check if justification is required without additional storage reads.

svyatonik: Also I'm not sure about condition. Probably it's just unfinished code, but please remember that you can't finalize block `C` that is descendant of block `B` if `B` doesn't have justification but requires it (i.e. it enacts new authority set). Seems like here you just may insert `B` without justification. And you don't have API/Call to finalize existing block.

modules/substrate/src/lib.rs
```diff
@@ -13,3 +13,186 @@
...
+decl_module! {
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+		type Error = Error<T>;
+
+		/// Import a signed Substrate header into the runtime.
+		///
+		/// Will check for finality proofs, and if available finalize any already imported
+		/// headers which are finalized by the newly imported header.
+		#[weight = 0]
+		pub fn import_signed_header(
```
svyatonik, line 81: You have both `import()` and `finalize()` API in single call. This means that during period `P` when the header is generated, but justification is not (yet), your pallet won't be able to import blocks. So if this call will have some large weight, your pallet may be out of sync for a while, because it'll: (1) need to wait till the end of `P` and submit justification (2) sync all descendants that have appeared during `P`

svyatonik: What I mean that it should be meaningful decision, because it may became bottleneck in the future

modules/substrate/src/verifier.rs
```diff
@@ -0,0 +1,472 @@
...
+			// Maybe pass the scheduled_change in here so we don't have to query storage later
+			let finalized_headers = Self::verify_finality(
+				storage,
+				header,
+				finality_proof.expect("Checked for `finality_proof` before entering if-block"),
+			)?;
+
+			// TODO: Would need to prune blocks from the non-canonical chain at some point
+			for header in finalized_headers.iter() {
```
svyatonik, line 98: `verify_finality` only returns ancestors? Probably should also write current header too?

svyatonik:
2. Imo it doesn't actually matters which API it has, if it works and decouples client code from runtime storage. So imo ok.
3 + 4. I've left some comments - hope they're helpful :)
5. In Aura they're hardcoded only when they're hardcoded in the chain configuration. If there's contract, only initial set needs to be hardcoded/provided. In your case I suggest designing it in some way to allow pallet to be started from any header (not necessary genesis). I.e. config of your pallet should have `initial_header` + `initial_authorities_set` + `initial_set_id`. Also, since authorities change signals are scheduled in advance, then config should also have `next_change: Option<AuthoritiesChangeSignal>` (probably another field for change+2? not sure - need to think :) ). But it could be left for another PRs - we may want to get something that is able to sync from genesis first :)

tomusdrw:
> 1. Has there been any progress in a no_std way to verify Grandpa proofs?

I thought that `finality-grandpa` provides everything that's needed for this? Is there anything missing? Also to verify signatures you should rather use the substrate-specific way of calling the host functions instead of re-implementing crypto in the runtime.

> 2. Does the BridgeStorage trait generally make sense?

I think the current version has some logic built into it. IMHO it should be completely minimal and just serve as a layer of indirection between substrate storage and our "high level" code.

> 5. How should I bootstrap the initial authority set?

Should be part of `GenesisConfig` I think, so that we can put it to the chainspec.

modules/substrate/src/verifier.rs
```diff
@@ -0,0 +1,472 @@
...
+			return Ok(());
+		}
+
+		// Don't like having to take ownership of header...
+		storage.write_header(&ImportedHeader::new(header.clone(), false));
+
+		Ok(())
+	}
+
+	fn verify_finality(storage: &mut S, header: &H, proof: FinalityProof) -> Result<Vec<H>, ImportError> {
```
tomusdrw, line 111: I'd prefer to not mix storage reads & writes here. Since the `import_header` function is already doing some writes, I think we should keep a really strict boundary between verification (state reads), preparation of update data (also only state reads), error handling and actual state writes. Otherwise it's really easy to miss `?` which can leave the state inconsistent.

svyatonik:
> I thought that finality-grandpa provides everything that's needed for this? Is there anything missing? Also to verify signatures you should rather use the substrate-specific way of calling the host functions instead of re-implementing crypto in the runtime.

@tomusdrw `finality-grandpa` provides most of types/functionality. The only exception is that finality proof is `Justification` and it is declared in `sc-finality-grandpa`. So we either need to move this type to smth like `sp-finality-grandpa` or manually parse it in runtime.

modules/substrate/src/verifier.rs
HCastano: Yeah, I know this pallet doesn't support forks at the moment. I figured I'd start with the "happy path" case first, where all the headers imported are from the same chain, and later on I'd build in fork awareness.

> remember that you can't finalize block C that is descendant of block B if B doesn't have justification but requires it

So I think I have that covered the when checking the ancestry proof. When importing C we check that the last block with a justification, B, and the new block, C, are related. The code is [here](https://github.com/paritytech/parity-bridges-common/pull/296/files#diff-16898582e891075aa30c1b1330962d86R119).

modules/substrate/src/lib.rs
HCastano: No, the `Verifier::import_header` function is still supposed to import blocks without justifications if it they don't provide a finality proof. They're marked as unfinalized, but the pallet has made _some_ progress.

It's true that all the blocks imported during `P` won't be finalized until we get a justification at the end of `P`, but that's how Grandpa works anyways, so I don't think it's an issue.

modules/substrate/src/verifier.rs
HCastano: I didn't want to have any header writes in `verify_finality`. I wanted `import_header` to be the only thing doing as a way to minimize side-effects (or at least have one clear location where writes are taking place)

modules/substrate/src/verifier.rs
HCastano: I was trying to keep it kinda split. Like I mentioned to Slava earlier, there are no header writes in `verify_finality`, that only happens in `import_headers`. The only things that are written during the finality verification process are the authority set changes. It might be a little strange to have the authority set changes updated in `import_header`, but  I can look into doing all the writes there.

modules/substrate/src/verifier.rs
svyatonik: > Yeah, I know this pallet doesn't support forks at the moment. I figured I'd start with the "happy path" case first, where all the headers imported are from the same chain, and later on I'd build in fork awareness.

Ok then :) (you were asking if "Am I missing any basic checks in Verifier::import_header" : )

> So I think I have that covered the when checking the ancestry proof. When importing C we check that the last block with a justification, B, and the new block, C, are related. The code is here.

I don't see that :/ Looks like you're just building ancestry from `best_finalized` to new header there.

modules/substrate/src/lib.rs
svyatonik: > No, the Verifier::import_header function is still supposed to import blocks without justifications if it they don't provide a finality proof. They're marked as unfinalized, but the pallet has made some progress.

You do not have separate `finalize()` call - that's what I mean.

> It's true that all the blocks imported during P won't be finalized until we get a justification at the end of P, but that's how Grandpa works anyways, so I don't think it's an issue.

Without providing separate `finalize()` call for already-imported headers AND allowing to import headers on top of header that requires justification, you'll stuck.

modules/substrate/src/verifier.rs
svyatonik: Are you answering my question here? :) What I said is that `finalized_headers` is missing `header` (i.e. the header which you're importing now).

modules/substrate/src/lib.rs
HCastano: We mark old headers as finalized when a block with a justification is built on top of them. You're right in that you can't build a chain on top of a header that's supposed to have a justification until you get that header. So in that sense the chain is stuck.

I'm confused as to how this would work otherwise. Let's say that we have previously imported scheduled change that's supposed to take effect at block N. When we get to block N we expect that this block will have a justification, otherwise it's not valid and nobody should be building on top of it. This is because they'll eventually be on a fork when the header with justification at N comes in.

Am I misunderstanding this?

modules/substrate/src/verifier.rs
HCastano: > Ok then :) (you were asking if "Am I missing any basic checks in Verifier::import_header" : )

Ah, I didn't realize you were answering that, lol. I thought it was more of a general observation, since I think fork support goes a little deeper than just `Verifier::import_header` (requires changes in pallet storage, pruning, etc).

> I don't see that :/ Looks like you're just building ancestry from best_finalized to new header there.

In the way the pallet is written now, `best_finalized` will always be the best block we know of that was imported with a justification. This means it's fine if we build from `(best_finalized, being_imported]`, assuming that `best_finalized` and `being_imported` are related.

Although I do realize now that I never update the `BestFinalized` storage item, so maybe that's adding to the confusion.

modules/substrate/src/verifier.rs
HCastano: > Are you answering my question here? 

I thought you were suggesting that `Verifier::verify_finality` should be writing to storage in general.

> What I said is that finalized_headers is missing header (i.e. the header which you're importing now)

I don't think so. If you take a look at the test `correctly_verifies_and_finalizes_chain_of_headers()`, which calls  `Verifier::import_header` with a `header`, you can see that `header` gets written to storage correctly. Maybe the name `finalized_headers` is a little misleading here?

modules/substrate/src/lib.rs
svyatonik: > I'm confused as to how this would work otherwise. Let's say that we have previously imported scheduled change that's supposed to take effect at block N. When we get to block N we expect that this block will have a justification, otherwise it's not valid and nobody should be building on top of it. This is because they'll eventually be on a fork when the header with justification at N comes in.

Blocks generation and finalization are completely separate in Substrate, so the fact that some blocks are finalized and some are not **is not affecting** block production. So let's only concentrate on finalization now. Let's also omit forks for now.

At some block `B` you see the `ScheduledChange(N, S2)` digest that signals change to `S2` in `N` headers. That'd mean 2 things: (1) starting from block `B+N+1` blocks will be finalized with new set `S2` and (2) current set (`S1`) must emit justification for block `B+N`. The justification is produced asynchronously (remember that block production + finalization are separate) - i.e. block production may broadcast **valid** blocks `B+N`, `B+N+1`, `B+N+2`, ... before set `S1` will generate justification for `B+N`. The most important thing is that you can't accept any justifications from `S2` (i.e. from blocks >= `B+N+1`) until you'll accept justification for `B+N` (generated by `S1`).

That's why if there's no API to finalize already-imported block (`B+N`) + you keep accepting descendants of this block, then finalization would stuck - you're unable to import justification for `B+N` and you can't accept justifications for any of its descendants.

modules/substrate/src/verifier.rs
svyatonik: Ok, I think you're right - sorry for confusion :)

modules/substrate/src/lib.rs
HCastano: @svyatonik I ended up splitting the header import and finality. Can you check that part again to see if it's alright?

svyatonik: You still accept all justifications - i.e. in the example above you'll accept justification for block `B+N+100` from `S2` without waiting for justification for `B+N` from `S1`.

Why this is important, especially for light client (at least that's my understanding - probably there are some other concerns): let's say single malicious block author makes a block `X` where he signals authorities set change to some `MaliciousSet` on next block `X+1`. Then block `X+1` and some of its descendants are created by the same block author and then the `MaliciousSet` just generates justification for, say, `X+3`. This makes chain `X->X+1->X+2->X+3` final and irrevertible. At the same time, all others block authors and old GRANDPA authorities are keeping to produce valid headers + justifications. So there's now competing finalized forks (keep in mind that light client only looks at the headers => it can't ask runtime which header is actually valid from runtime PoV).

So justification of `B+N` (or `X+1` in latter example) by previous set `S1` is treated as a confirmation that the set `S1` agrees to transfer its 'power-to-finalize-blocks' to the next set `S2`. Without this confirmation, you can't trust `S2` => can't accept any justifications for blocks > `B+N`.

Hope it makes sense :)

svyatonik: Actually, I've found that you've added `requires_justification` field to the header, but you never check it. Probably some commits are missing?

HCastano: I was unsure about the strategy to use when checking `requires_justification`. I've followed your advice and am iterating through the ancestors of the header that we're trying to finalize and checking if there's any ancestor which `requires_justification`. If they do then we don't have the "best" justification so we need to reject it.

I've added a test `does_not_import_future_justification()` which I think capture the correct logic here. Let me know if it looks better now.

primitives/substrate/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+
+// Parity Bridges Common is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Primitives for the Substrate light client (a.k.a bridge) pallet.
```
svyatonik, line 17: Imo if it is primitives of the pallet, then it should have types that are used in pallet APIs. Here you have e.g. `struct ImportedHeader`, which is the struct describing how pallet is storing the header. Why to expose that? The same about other types - imo this file should be moved to smth like `pallet/storage.rs`. No pressure about that, though

svyatonik (approved):
If the scope of this PR is to create light-client for happy-path, then it is mostly OK imo. The priority of design-fixes imo should be:
1) forcing pallet to work with bridged chain headers - it isn't that important to continue coding pallet/relay/apps (just because all chains that we're going to bridge right now are using the same header struct), but it is quite important to see a strict border between chains;
2) everything that affects storage - i.e. forks, authorities set changes, storage-based optimizations. If we want to bridge persistent chains, then we should minimize any changes to the storage. I hope we'll be able to bridge more-or-less persistent chains before bridging actual chains.

I haven't left any performance-related comments in the code - it'll make too much noise && I believe they'll be addressed in follow-up PRs anyway. But I'll leave some general comments here:
1) there are too much calls to the `Header::hash()` function - most of times you already know the header hash (e.g. you've just read the header by the hash from storage) && may reuse result of previous computation;
2) there should be some other way on how to implement `fn are_ancestors()` - probably skip list?
3) I'm not sure we need to explicitly mark every finalized header as finalized - iirc there was large gap of ~30_000 unfinalized headers at the beginning of some Substrate-based chain. I guess now GRANDPA is much more stable, but iiuc offline/misbehaving authorities may lead to large gaps again. Probably use the same skip-list-like struct from (2)?

modules/substrate/src/verifier.rs
```diff
@@ -0,0 +1,667 @@
...
+		for header in finalized_headers.iter_mut() {
+			// TODO: Maybe mutate() storage?
+			header.is_finalized = true;
+			header.requires_justification = false;
+			storage.write_header(header);
+		}
+
+		let best_finalized = finalized_headers
+			.last()
+			.expect("We just iterated through these headers, therefore the last header must exist");
```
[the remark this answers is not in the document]
HCastano: Yeah good point

modules/substrate/src/lib.rs
```diff
@@ -14,5 +14,238 @@
...
+decl_storage! {
+	trait Store for Module<T: Trait> as SubstrateBridge {
+		/// Hash of the best finalized header.
+		BestFinalized: T::Hash;
+		/// Headers which have been imported into the pallet.
+		ImportedHeaders: map hasher(identity) T::Hash => Option<ImportedHeader<T::Header>>;
+		/// The current Grandpa Authority set.
+		CurrentAuthoritySet: AuthoritySet;
+		/// The next scheduled authority set change.
+		NextScheduledChange: ScheduledChange<Number<T::Header>>;
```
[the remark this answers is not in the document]
HCastano: If you take a look at the GenesisConfig in `decl_storage` I make sure to always write `NextScheduledChange`

modules/substrate/src/lib.rs
```diff
@@ -14,5 +14,238 @@
...
+type Number<T> = <T as HeaderT>::Number;
+
+pub trait Trait: frame_system::Trait {}
+
+decl_storage! {
+	trait Store for Module<T: Trait> as SubstrateBridge {
+		/// Hash of the best finalized header.
+		BestFinalized: T::Hash;
+		/// Headers which have been imported into the pallet.
+		ImportedHeaders: map hasher(identity) T::Hash => Option<ImportedHeader<T::Header>>;
```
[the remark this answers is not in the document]
HCastano: My assumption here is that we're bridging a chain that is just like our chain. But you're right, maybe that's not a good assumption since even between Substrate chains we can have differences

modules/substrate/src/verifier.rs
```diff
@@ -0,0 +1,667 @@
...
+		storage.write_header(&ImportedHeader::new(header, requires_justification, is_finalized));
+
+		Ok(())
+	}
+
+	fn verify_finality(storage: &mut S, hash: H::Hash, proof: FinalityProof) -> Result<(), ImportError> {
+		// Make sure that we've previously imported this header
+		let header = storage.get_header_by_hash(hash).ok_or(ImportError::UnknownHeader)?;
+
+		let current_authority_set = storage.current_authority_set();
```
tomusdrw, line 106: It seems weird we don't verify that the `header` actually should be finalized by `current_authority_set`. i.e. It seems this code assumes that `next_validator_set_enactment_block_number > header.number() > last_finalized_header_number`, but I don't think it's clear where it's enforced (and why not here).

modules/substrate/src/lib.rs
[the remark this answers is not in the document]
svyatonik: What I mean is not about coding, but about GRANDPA itself. Are you sure that at any point we'll have something scheduled? I'm not sure - that's why I'm asking

modules/substrate/src/verifier.rs
```diff
@@ -0,0 +1,667 @@
...
+		let is_finalized = prove_finality(&header, &current_authority_set, proof);
+		if !is_finalized {
+			return Err(ImportError::UnfinalizedHeader);
+		}
+
+		let last_finalized = storage.best_finalized_header();
+		let mut finalized_headers = if let Some(ancestors) = are_ancestors(storage, last_finalized, header.clone()) {
```
tomusdrw, line 113: This method only works if `last_finalized.block_number() < header.block_number()` but we don't check it anywhere. Without the check the `are_ancestors` function may traverse the entire blockchain up to genesis before it fails.

modules/substrate/src/verifier.rs
```diff
@@ -0,0 +1,667 @@
...
+where
+	S: BridgeStorage<Header = H>,
+	H: HeaderT,
+{
+	let mut ancestors = vec![];
+	let mut current_header = child;
+
+	while ancestor.hash() != current_header.hash() {
+		// We've gotten to the same height and we're not related
+		if ancestor.number() == current_header.number() {
```
tomusdrw, line 162:
```suggestion
		if ancestor.number() >= current_header.number() {
```
Early exit if `ancestor.number() > child.number()`

modules/substrate/src/verifier.rs
tomusdrw: Proposed a change to `are_ancestors` to prevent that.

modules/substrate/src/verifier.rs
```diff
@@ -0,0 +1,667 @@
...
+			.last()
+			.expect("We just iterated through these headers, therefore the last header must exist");
+		storage.update_best_finalized((*best_finalized).hash());
+
+		Ok(())
+	}
+}
+
+// Returns the lineage of headers from (ancestor, child]
+fn are_ancestors<S, H>(
```
tomusdrw, line 148: `is/are` functions usually just return `bool`, why not call it `headers_between(storage, a, b)` or something like that? 
The doc should also mention the lineage is actually in reverse order.

primitives/substrate/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+///
+/// Keeps track of two important fields aside from the header. First,
+/// if the header requires a Grandpa justification. This is required
+/// for headers which signal new authority set changes.
+///
+/// Secondly, whether or not this header has been finalized. A header
+/// does not need to be finalized explictly, but instead may be finalized
+/// implicitly when one of its children gets finalized.
+#[derive(Default, Encode, Decode, Clone, RuntimeDebug, PartialEq)]
+pub struct ImportedHeader<H: HeaderT> {
```
tomusdrw, line 73:
```suggestion
pub struct ImportedHeader<H> {
```

primitives/substrate/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+	pub fn new(header: H, requires_justification: bool, is_finalized: bool) -> Self {
+		Self {
+			header,
+			requires_justification,
+			is_finalized,
+		}
+	}
+
+	/// Get the hash of this header.
+	pub fn hash(&self) -> H::Hash {
```
tomusdrw, line 90: Maybe just `Deref` to regular header or add `fn header(&self) -> &H` function?

primitives/substrate/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+/// of N, assuming we get a valid justification for the header at N.
+#[derive(Default, Encode, Decode, RuntimeDebug, PartialEq, Clone)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct ScheduledChange<N> {
+	pub authority_set: AuthoritySet,
+	pub height: N,
+}
+
+impl<N> ScheduledChange<N> {
+	pub fn new(authority_set: AuthoritySet, height: N) -> Self {
```
tomusdrw, line 58: Is it really needed? Both fields are public and it's just more readable if you initialize them explicitly.

primitives/substrate/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+#[derive(Default, Encode, Decode, Clone, RuntimeDebug, PartialEq)]
+pub struct ImportedHeader<H: HeaderT> {
+	pub header: H,
+	pub requires_justification: bool,
+	pub is_finalized: bool,
+}
+
+impl<H: HeaderT> ImportedHeader<H> {
+	/// Create a new ImportedHeader.
+	pub fn new(header: H, requires_justification: bool, is_finalized: bool) -> Self {
```
tomusdrw, line 81: Same here, instead of creating a `let is_finalized = false` variable in one of the functions you could just initialize `ImportedHeader {}` without calling `new` to make it clear what's being set - epscially given this function takes two non-obvious `bool`s.

modules/substrate/src/justification.rs
```diff
@@ -20,12 +20,13 @@
 // TODO: remove on actual use
 #![allow(dead_code)]
```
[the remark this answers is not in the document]
HCastano: No. I still haven't hooked this up to the `Verifier`. I was planning on doing that in a different PR so I could add tests and make sure it was working properly

modules/substrate/src/verifier.rs
```diff
@@ -0,0 +1,667 @@
...
+	/// This will typically have to do with missing authority set change signals.
+	MissingConsensusDigest,
+	/// Trying to prematurely import a justification
+	PrematureJustification,
+	/// The height of the next authority set change overflowed.
+	ScheduledHeightOverflow,
+}
+
+/// A trait for verifying whether a header is valid for a particular blockchain.
+pub trait ChainVerifier<S, H: HeaderT> {
```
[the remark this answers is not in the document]
HCastano: This trait was from when I initially had envisioned a more "generic" trait for the pallet. It didn't end up panning out, so yeah makes sense to remove it now

primitives/substrate/src/lib.rs
HCastano: I was getting annoyed of the full struct creation syntax when I added this. I agree that it does hinder readability a bit though, so I'll change it

modules/substrate/src/lib.rs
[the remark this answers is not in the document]
HCastano: Oh, that's a good point. I'm not sure actually, I'll have to look into it

modules/substrate/src/verifier.rs
```diff
@@ -0,0 +1,667 @@
...
+
+		if storage.header_exists(header.hash()) {
+			return Err(ImportError::HeaderAlreadyExists);
+		}
+
+		let parent_header = storage.get_header_by_hash(*header.parent_hash());
+		if parent_header.is_none() {
+			return Err(ImportError::MissingParent);
+		}
+
```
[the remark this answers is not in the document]
HCastano: Good call! Some of my tests broke when I added this. I had modified my test headers to container the correct parent hash, but their number was off-by-one

modules/substrate/src/verifier.rs
HCastano: > Without the check the `are_ancestors` function may traverse the entire blockchain up to genesis before it fails.

Yes, good catch!

HCastano:
I've spent the day basically just fixing Tomek's comments. I haven't gotten around to fixing your comments yet Slava 😅

modules/substrate/src/lib.rs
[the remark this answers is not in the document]
svyatonik: Let's probably change hasher/block number type that is used in Millau test-chain? E.g. `Hasher` would return `H512` instead of `H256` (we can just concat two `H256` returned by `BlakeTwo256`) and block number e.g. `u128`. This would also help in testing other sub2sub-related pallets as well (`call-dispatch`, `message-lane`) - we'll be sure that our pallets are ready to work with "incompatible" chains? Wdyt?

(not a part of this PR of course)

modules/substrate/src/lib.rs
```diff
@@ -14,5 +14,238 @@
...
 mod storage_proof;
+mod verifier;
+
+#[cfg(test)]
+mod mock;
+
+type Hash<T> = <T as HeaderT>::Hash;
+type Number<T> = <T as HeaderT>::Number;
+
+pub trait Trait: frame_system::Trait {}
```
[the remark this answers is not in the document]
HCastano: Logged in #360

modules/substrate/src/lib.rs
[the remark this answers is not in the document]
HCastano: It turns out that most of the time Grandpa will be sitting around without any pending changes, so yes this should be an `Option`

HCastano:
Can you guys take another look at this? I've fixed a lot (but not all) of the grumbles, and notably I've reworked how the authority set handoffs are done. It's not quite perfect, and I have a note about that in the code, but I think it's more correct than it was before. I've left one test related to it marked as `#[ignore]` for now. I think I'll remove it and add it back in a different PR, but right now I think it'll be good for you guys to check if my logic there makes sense.

For the comments I haven't addressed yet I think it'll be best to address them in a series of follow-up PRs, since this one is getting a bit hairy. If there any comment you think needs to be addressed in this PR let me know.

bin/rialto-runtime/Cargo.toml
```diff
@@ -150,6 +155,12 @@ tag = 'v2.0.0-rc6'
 default-features = false
 git = "https://github.com/paritytech/substrate/"
 
+[dependencies.sp-finality-grandpa]
```
svyatonik, line 158: Are you sure it is required?

bin/rialto-runtime/Cargo.toml
```diff
@@ -55,6 +55,11 @@ version = "0.1.0"
 default-features = false
 path = "../../modules/currency-exchange"
 
+[dependencies.pallet-substrate-bridge]
+version = "0.1.0"
+default-features = false
```
svyatonik, line 60: More a question - can there be some negative effects of not-enabling "std" feature when runtime is compiled with std? Probably something related to panic-handling or disabled logs/debug info, ...?

modules/substrate/src/lib.rs
```diff
@@ -14,5 +14,254 @@
...
+					is_finalized: true,
+				},
+			);
+
+			let authority_set =
+				AuthoritySet::new(config.initial_authority_list.clone(), config.initial_set_id);
+			CurrentAuthoritySet::put(authority_set);
+
+			let change = config.first_scheduled_change.clone();
+			if change.is_some() {
```
svyatonik, line 97: By using `if let Some(ref change) = config.first_scheduled_change { ... }` you could remove that expect :)

modules/substrate/src/verifier.rs
```diff
@@ -0,0 +1,764 @@
...
+		let parent_number = *parent_header.number();
+		if parent_number + One::one() != *header.number() {
+			return Err(ImportError::InvalidChildNumber);
+		}
+
+		// This header requires a justification since it enacts an authority set change. We don't
+		// need to act on it right away (we'll update the set once this header gets finalized), but
+		// we need to make a note of it.
+		//
+		// Note: This assumes that we can only have one authority set change pending at a time.
```
svyatonik, line 108: Just asking - we're going to fix it in next PRs, right? I'd replace 'Note' with 'TODO' here, even without submitting an issue (because pallet is in-progress anyway).

modules/substrate/src/lib.rs
```diff
@@ -14,5 +14,254 @@
...
+	fn current_authority_set(&self) -> AuthoritySet {
+		CurrentAuthoritySet::get()
+	}
+
+	fn update_current_authority_set(&self, new_set: AuthoritySet) {
+		CurrentAuthoritySet::put(new_set)
+	}
+
+	fn enact_authority_set(&mut self) {
+		let new_set = <NextScheduledChange<T>>::take().expect("TODO").authority_set;
```
svyatonik, line 256: .expect("TODO")

modules/substrate/src/verifier.rs
```diff
@@ -0,0 +1,764 @@
...
+		for header in finalized_headers.iter_mut() {
+			header.is_finalized = true;
+			header.requires_justification = false;
+			self.storage.write_header(header);
+		}
+
+		let best_finalized = finalized_headers
+			.first()
+			.expect("Asserted that `ancestors` (now `last_finalized`) had at least one element.");
+		self.storage.update_best_finalized((*best_finalized).hash());
```
svyatonik, line 212:
```suggestion
		self.storage.update_best_finalized(hash);
```

Iiuc `best_finalized` is always the `header`, right?

svyatonik (approved):

modules/substrate/Cargo.toml
```diff
@@ -8,17 +8,32 @@ license = "GPL-3.0-or-later WITH Classpath-exception-2.0"
...
-codec = { package = "parity-scale-codec", version = "1.3.4", default-features = false }
+bp-header-chain = { path = "../../primitives/header-chain", default-features = false }
+bp-substrate = { path = "../../primitives/substrate", default-features = false }
 finality-grandpa = { version = "0.12.3", default-features = false }
 hash-db = { version = "0.15.2", default-features = false }
+serde = { version = "1.0", optional = true }
 
 # Substrate Based Dependencies
+
+[dependencies.codec]
```
tomusdrw, line 19: Why is it in substrate based section? It's just a regular crate on crates.io

modules/substrate/src/lib.rs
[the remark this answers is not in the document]
tomusdrw: Just for the record, I realised that my previous comment might have been intepreted as endorsing usage of `T::Header` stuff. It's not. I only meant that we can focus on just `sp_runtime_traits::Header`-based headers for now.
We for sure need to decouple source and target types.

modules/substrate/src/lib.rs
```diff
@@ -14,5 +14,254 @@
...
+		) -> dispatch::DispatchResult {
+			let _ = ensure_signed(origin)?;
+			frame_support::debug::trace!(target: "sub-bridge", "Got header {:?}", header);
+
+			let mut verifier = verifier::Verifier {
+				storage: PalletStorage::<T>::new(),
+			};
+
+			let _ =
+				verifier.import_header(header).map_err(|_| <Error<T>>::InvalidHeader)?;
```
tomusdrw, line 138:
```suggestion
			let _ = verifier.import_header(header).map_err(|_| <Error<T>>::InvalidHeader)?;
```

modules/substrate/src/lib.rs
```diff
@@ -14,5 +14,238 @@
...
+#[derive(Default)]
+pub struct PalletStorage<T>(PhantomData<T>);
+
+impl<T> PalletStorage<T> {
+	fn new() -> Self {
+		Self(PhantomData::<T>::default())
+	}
+}
+
+impl<T: Trait> BridgeStorage for PalletStorage<T> {
```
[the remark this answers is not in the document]
tomusdrw: Still applies, please at least mark it as a todo and have some checklist for the follow up PR or split all things into it's own issues.

modules/substrate/src/lib.rs
```diff
@@ -14,5 +14,254 @@
...
+		self.header_by_hash(hash)
+			.expect("A finalized header was added at genesis, therefore this must always exist")
+	}
+
+	fn update_best_finalized(&self, hash: Hash<T::Header>) {
+		<BestFinalized<T>>::put(hash)
+	}
+
+	fn header_exists(&self, hash: Hash<T::Header>) -> bool {
+		self.header_by_hash(hash).is_some()
```
tomusdrw, line 240: It was already reported earlier by Slava that this should rather use `contains_key` to avoid decoding the header.
This could be left as a default implementation on the trait level instead.

modules/substrate/src/verifier.rs
```diff
@@ -0,0 +1,667 @@
...
+use sp_finality_grandpa::{ConsensusLog, SetId, GRANDPA_ENGINE_ID};
+use sp_runtime::generic::OpaqueDigestItemId;
+use sp_runtime::traits::{CheckedAdd, Header as HeaderT};
+use sp_std::{prelude::Vec, vec};
+
+/// The finality proof used by the pallet.
+///
+/// For a Substrate based chain using Grandpa this will
+/// be an encoded Grandpa Justification.
+pub type FinalityProof<'a> = &'a [u8];
```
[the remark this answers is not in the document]
tomusdrw: Not addressed yet :)

modules/substrate/src/verifier.rs
```diff
@@ -0,0 +1,764 @@
...
+			return Err(FinalizationError::UnfinalizedHeader);
+		}
+
+		let mut finalized_headers =
+			if let Some(ancestors) = headers_between(&self.storage, last_finalized, header.clone()) {
+				// Since we only try and finalize headers with a height strictly greater
+				// than `best_finalized` if `headers_between` returns Some we must have
+				// at least one element. If we don't something's gone wrong, so best
+				// to die before we write to storage.
+				assert_eq!(ancestors.is_empty(), false);
```
tomusdrw, line 178: Maybe add a description here?
```suggestion
				assert_eq!(ancestors.is_empty(), false, "Emtpy ancestry list returned from headers_between");
```

modules/substrate/src/verifier.rs
```diff
@@ -0,0 +1,764 @@
...
+
+				// Check if any of our ancestors `requires_justification` a.k.a schedule authority
+				// set changes. If they're still waiting to be finalized we must reject this
+				// justification.
+				//
+				// We do this because it is important to to import justifications _in order_,
+				// otherwise we risk finalizing headers on competing chains.
+				let requires_justification = ancestors.iter().skip(1).find(|h| h.requires_justification);
```
tomusdrw, line 186: Why do we `skip(1)` here? Maybe add to the docs?

primitives/substrate/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+/// Keeps track of two important fields aside from the header. First,
+/// if the header requires a Grandpa justification. This is required
+/// for headers which signal new authority set changes.
+///
+/// Secondly, whether or not this header has been finalized. A header
+/// does not need to be finalized explictly, but instead may be finalized
+/// implicitly when one of its children gets finalized.
+#[derive(Default, Encode, Decode, Clone, RuntimeDebug, PartialEq)]
+pub struct ImportedHeader<H> {
+	pub header: H,
```
tomusdrw, line 67: missing docs. Please add `#![warn(missing_docs)]` to th ecrate.

tomusdrw (commented):
lgtm! Few things are still remaining unaddressed though

bin/rialto-runtime/Cargo.toml
HCastano: The stuff remaining in the Rialto runtime crate were missed during my initial runtime cleanup, thanks for catching it

modules/substrate/src/verifier.rs
HCastano: > It seems this code assumes that next_validator_set_enactment_block_number > header.number() > last_finalized_header_number, but I don't think it's clear where it's enforced (and why not here).

I think there is some work left to be done around the authority set changes (and its assumptions), but I'm going to address that in a follow up PR if that's alright.

> It seems weird we don't verify that the header actually should be finalized by current_authority_set

I was under the assumption that this would at least partly be handled by the function which checks the finality proofs but maybe that's not the correct thing to do. I'll need to revisit that once I start integrating it into the pallet.

HCastano:
Based off your review comments, here is a list of things left for other PRs:

- Move things from `bp-primitives` closer to the pallet [[ref](https://github.com/paritytech/parity-bridges-common/pull/296#discussion_r490128662)]
- More realistic assumptions around authority set changes [[ref](https://github.com/paritytech/parity-bridges-common/pull/296#discussion_r490092276), [ref](https://github.com/paritytech/parity-bridges-common/pull/296#discussion_r490295849)]
- Create a deeper `BridgeStorage` abstraction [[ref](https://github.com/paritytech/parity-bridges-common/pull/296#discussion_r490202616)]
- Accommodate different Substrate chain characteristics (Hash, BlockNumber, etc) [[ref](https://github.com/paritytech/parity-bridges-common/pull/296#discussion_r490083775)]
- Reduce calls to `Header::hash()` [[ref](https://github.com/paritytech/parity-bridges-common/pull/296#pullrequestreview-490374230)]
- More efficient ancestry checker implementation [[ref](https://github.com/paritytech/parity-bridges-common/pull/296#pullrequestreview-490374230)]
- More efficient method for marking headers as finalized [[ref](https://github.com/paritytech/parity-bridges-common/pull/296#pullrequestreview-490374230)]

This is on top of the changes I mentioned I wanted to address in the PR description.

tomusdrw (approved):
Looks good! Given we have a list of things that are yet to be done, I'm down to merge as-is and parallelize efforts on this.
@HCastano please create a set of issues that we (Slava and I) can assign to and create a bunch of smaller PRs.

bin/rialto-runtime/Cargo.toml
HCastano: I'm not entirely sure tbh. But I assume that having `no_std` code running in a `std` environment is more forgiving than the other way around

HCastano:
@tomusdrw I've created some issues which I think can be picked up independently from each other. I didn't create issues for everything since some of these things are either too small, or too abstract to be nicely put into an issue.

Issue #215, opened 2020-07-17
Message Lanes Delivery Protocol


tomusdrw:
Write a pallet to support a more involved than #212  delivery protocol: Message Lanes. The spec can be found in Jim's bridge design doc.

tomusdrw:
Pretty much covered with our current Message Lanes pallet.

PR #299, opened 2020-08-21
Initial version of DummyOrdered pallet

### LINKED ISSUE #212: `DummyOrdered` delivery protocol
Write a pallet that allows queueing outbound messages and is able to receive inbound messages. The messages should have a simple format defined below and the dispatcher should be pluggable.
```
struct Message {
  lane_id: [u8; 3],
  nonce: u64,
  payload: Vec<u8>,
}
```
- Messages from each lane should be submitted separately (initially it's fine to mix them up though)
- Initially we will support only one lane.

To receive messages the pallet needs to keep track of:
`Map<LaneId, Nonce>` to make sure messages are not replayed and processed in order.

svyatonik:
closes #212

There's no default lane in the pallet - it is up to message creators + relay - which lanes to use.

The lane may have 4 endpoints (again - it is up to clients+relay to decide which endpoints are actually used and served) - `OutboundLane` at `Chain1` to send messages to `Chain2`. `InboundLane` at `Chain1` to receive messages from `Chain2`. And corresponding `OutboundLane` and `InboundLane` at `Chain2`.

The workflow for sending and receiving message `M` is as follows:
1) user submits transaction `MessageLane::send_message(LaneId, Payload)` to the `Chain1`. Pallet assigns 1-based nonce to the message and saves it to the runtime storage;
2) relay that is responsible for serving this lane, uses `OutboundLaneApi::latest_generated_nonce()` on source chain and `InboundLaneApi::latest_received_nonce()` to check if there are new messages are to be relayed;
3) once relay sees that there's new message`M`, it prepares proof-of-message using some RPC method (to be done in one of follow-up PRs);
4) the proof-of-message is sent to the target node. Please keep in mind that the entry point for this transaction is outside of scope of this pallet, because it requires proof, that uses source chain headers. So it should be part of the pallet that bridges two chains together (let it be `BridgePallet`). Once `BridgePallet` receives this proof, it recovers message(s) from the proof and calls `MessageLane::receive_messages()`;
5) there are two options on how to process incoming messages:
5.1) we can process messages as they come - in this case message processor could process it and return `MessageAction::Drop`. But some messages may be too heavy to process in the same block => ...;
5.2) we may enqueue message and process it later. For this, message processor should return `MessageAction::Queue` and call `MessageLane::process_lane_messages()` later;
6) once message is received by the target chain, we may relay proof-of-receival back to source chain. It must be sent if `InboundLaneApi::latest_received_nonce()` is larger than `OutboundLaneApi::latest_received_nonce()`. The proof is generated by RPC call, then submitted to `BridgePallet` of source chain && finally transforms into `MessageLane::confirm_receival()` call. The pruning isn't happening currently, because I'm not sure again - when we should do that. Probably in `on_finalize`? But then we need some separate pruning-track map. Probably when next message is sent? This is to be decided yet. **UPD**: messages are pruned whenever outbound lane is used.


modules/message-lane/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+	///
+	/// Trusted here means that the function itself doesn't check whether the bridged chain has
+	/// actually received these messages.
+	///
+	/// The caller may break the channel by providing `latest_received_nonce` that is larger
+	/// than actual one. Not-yet-sent messages may be pruned in this case.
+	pub fn confirm_receival(lane_id: &LaneId, latest_received_nonce: MessageNonce) {
+		let mut lane = outbound_lane::<T, I>(*lane_id);
+		lane.confirm_receival(latest_received_nonce);
+		lane.prune_messages(T::MaxHeadersToPruneAtOnce::get());
```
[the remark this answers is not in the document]
svyatonik: I've added 3 events + another entry point - `fn confirm_processing(nonce)`. So we may now relay confirm that the message has been processed.

modules/message-lane/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+		#[weight = 0] // TODO: update me (https://github.com/paritytech/parity-bridges-common/issues/78)
+		pub fn send_message(
+			origin,
+			lane_id: LaneId,
+			payload: T::Payload,
+		) {
+			let _ = ensure_signed(origin)?;
+			let mut lane = outbound_lane::<T, I>(lane_id);
+			lane.send_message(payload);
+			lane.prune_messages(T::MaxHeadersToPruneAtOnce::get());
```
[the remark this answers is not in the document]
tomusdrw: Why do we even wait with pruning till the message is relayed? My idea is that we remove all the messages in `on_finalize` (or maybe rather `on_initialize` of the next block) i.e. they occupy state only to allow inspection, similar to have you access `frame_system::Events`.

IMHO confirmation is not strictly necessary just yet. If one of the messages is not relayed, others can't be relayed as well (due to nonce issues). Messages to be relayed does not need to be stored on-chain up until they are proven to be relayed - rather I think that it should be the responsibility of relayers to retrieve the messages as they come and queue them up for delivery on the other chain. More sophisticated mechanism, like ACK-based rate control should imho be implemented as part of #215 not the simpler mechanism we have here. To prevent spam we may introduce exponential cost of triggering messages (i.e. if there are already 100 messages in current block, 101 will be much more expensive) or just introduce a fixed-size outbound pool. Scheduling/Queueing could be implemented separately if really needed for the PoC.

Having a `Vec<Message>` imho is going to simplify significantly both the relayer and runtime code, but obviously is less flexible - we assume that all messages are relayed timely. And as mentioned earlier the more sophisticated mechanisms can be implemented as part of #215 

> if messages are not relayed for some (significant) time, we'll need to encode/decode large amount of data on every append;

`StorageValue<Vec<_>>::append` does not require encoding/decoding. That's the whole point of this API, that runtime can append to the storage entry with a constant cost (no matter how much entries is there already).

> the proof-of-message would be large (and duplicate) in similar ^^^ cases (we must prove the whole messages vec from the storage trie).

Yes, however we might want to relay all messages at once anyway, imho sophisticated rate-limitting is not necessary initially (revist in #215). Also the tradeoff between relaying single messages (with multiple smaller proofs) vs batch-relaying all messages in a block (with single, bigger proof) is not obvious and depends on the dispatch cost of the message payload.

modules/message-lane/src/lib.rs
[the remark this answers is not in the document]
tomusdrw: I was actually thinking about notifying pallet that sent a particular message about it's delivery. Imagining a pallet, which serves as some sort of scheduler/queue/ratelimitter which sends highest-priority messages when some previous messages were relayed. However this can potentially be done in `on_intiialize` of the pallet, where it just inspects the storage to see if messages where confirmed.
If you feel the events are useful - go ahead, I haven't put much thought into it.

modules/message-lane/src/lib.rs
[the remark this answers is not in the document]
svyatonik: > Why do we even wait with pruning till the message is relayed? My idea is that we remove all the messages in on_finalize (or maybe rather on_initialize of the next block) i.e. they occupy state only to allow inspection, similar to have you access frame_system::Events.

We never had any requirements for the node. This ^^^ requires node to be archive, because if relay is offline for, say, a day, old states with messages are dropped && the channel is broken. Because we can't relay new messages without relaying previous. I agree that your approach simplifies pruning, but it introduces unnecessary requirement + there'll be problems with submitting messages from several blocks in single tx + I believe even more difficulties. 

> Having a Vec<Message> imho is going to simplify significantly both the relayer and runtime code, but obviously is less flexible - we assume that all messages are relayed timely. And as mentioned earlier the more sophisticated mechanisms can be implemented as part of #215

I don't see how it actually simplifies runtime code :) `OutboundMessages::append()` vs `OutboundMessages::insert()` isn't more simple imo. Re relay - both versions will have the same `prove_messages()` call. The only difference that in your version it'll accept `block_hash` (which you need to save somewhere when relay restarts) and in my case relay is still stateless, because runtime has everything required at every block. This would be even worse if @HCastano choose your approach - restart-on-any-failure in #281 :) Then simple network error may break message delivery if relay has no persistent storage. In my approach there'll be two additional rpc calls which would just return nonces && then I'll call `prove_messages()` with some nonce range.

> StorageValue<Vec<_>>::append does not require encoding/decoding. That's the whole point of this API, that runtime can append to the storage entry with a constant cost (no matter how much entries is there already).

This was written before you've introduced your idea with pruning-at-the-end of block :) Agree that if we only need to append, then my first argument isn't valid.

> Yes, however we might want to relay all messages at once anyway, imho sophisticated rate-limitting is not necessary initially (revist in #215). Also the tradeoff between relaying single messages (with multiple smaller proofs) vs batch-relaying all messages in a block (with single, bigger proof) is not obvious and depends on the dispatch cost of the message payload.

In my case there'll be single proof for multiple messages too - I'm not going to relay messages one-by-one.

===

As a conclusion, I can go your way, if you insist. But I do not see any benefits in it, except for easier pruning. I also do not agree that mine approach is more complex - we'd need to solve problems anyway. Please comment and I'll go your way :)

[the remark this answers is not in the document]
tomusdrw: > This ^^^ requires node to be archive, because if relay is offline for, say, a day, old states with messages are dropped && the channel is broken.

Or it requires relay (or multiple relays) to be on-online. Note we are no longer talking about header-sync, but rather message delivery. The latter will require incentivisation research for relayers anyway, cause they not only need to deliver messages, but also dispatch them (unless we split this two things, which sounds interesting as well). I've noticed that we don't have the incentivisation part in the issue tracker yet, so here it is: https://github.com/paritytech/parity-bridges-common/issues/316
So since I expect message delivery relayers to be incentivisised I'm fine expecting a bit more from them or simply assuming there is a bunch of them (redundancy). That said, I'm convinced that the stateless approach is way better initially.

> I don't see how it actually simplifies runtime code :) 

I was mostly thinking about getting rid of `MessageKey` & pruning, but yeah, maybe it's not worth it.

> In my case there'll be single proof for multiple messages too - I'm not going to relay messages one-by-one.

What if there is too many messages to fit in the block then? We will split the batch, right?

> As a conclusion, I can go your way, if you insist. 

I don't. I feel you're right that the stateless relay is a better option to go with initially even though it might cause a bit more complex runtime code. If storage cost of messages is too high, we might go with Substrate's `IndexingAPI` - the message content would be stored in the offchain-db (that's opt-in, but much less heavy than archive) and the relay would retrieve messages from there, while the state could only contain some references.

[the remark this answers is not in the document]
svyatonik: > What if there is too many messages to fit in the block then? We will split the batch, right?

Sure. I'm going to do the same thing as with headers initially - some constant/option to limit number of **messages** in "submitted" state. Later we may ask target runtime about how many messages it may accept in single tx/block, or some other criteria.

modules/message-lane/src/inbound_lane.rs
```diff
@@ -0,0 +1,236 @@
...
+					lane_id: self.storage.id(),
+					nonce,
+				},
+				payload,
+			};
+
+			let process_result = processor.on_message_received(message);
+			if let MessageResult::NotProcessed(_) = process_result {
+				break;
+			}
```
HCastano, line 116: Would it be possible for us to get stuck with a bad message?

Let's say we have messages A, and B in storage pending processing. We'll process them A->B. Since A is the oldest, it'll get processed first, next time we try and process B but fail for some reason.

Now a new message, C comes in. Next time we try and process messages we'll start at the oldest, B, but we'll continue to get `NotProcessed` and never make our way to C - effectively blocking the lane.

modules/message-lane/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+	/// Stops processing either when all messages are processed, or when processor returns
+	/// MessageResult::NotProcessed.
+	pub fn process_lane_messages(lane_id: &LaneId, processor: &mut impl OnMessageReceived<T::Payload>) {
+		inbound_lane::<T, I>(*lane_id).process_messages(processor);
+	}
+
+	/// Receive TRUSTED proof of message receival.
+	///
+	/// Trusted here means that the function itself doesn't check whether the bridged chain has
+	/// actually received these messages.
```
HCastano, line 134: Would it be the job of the relay to make sure that the messages are in fact in storage before making this call? I imagine this function would (or should) be called immediately after `receive_messages`, right?

modules/message-lane/src/outbound_lane.rs
```diff
@@ -0,0 +1,269 @@
...
+		Some((prev_latest_processed_nonce + 1, latest_processed_nonce))
+	}
+
+	/// Prune at most `max_messages_to_prune` already received messages.
+	pub fn prune_messages(&mut self, max_messages_to_prune: MessageNonce) -> MessageNonce {
+		let mut pruned_messages = 0;
+		let mut anything_changed = false;
+		let mut data = self.storage.data();
+		while pruned_messages < max_messages_to_prune && data.oldest_unpruned_nonce <= data.latest_received_nonce {
+			self.storage.remove_message(&data.oldest_unpruned_nonce);
```
HCastano, line 108: I wonder if there's a way to use a double map or something to more efficiently remove these messages

modules/message-lane/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+		lane_id,
+		_phantom: Default::default(),
+	})
+}
+
+/// Creates new outbound lane object, backed by runtime storage.
+fn outbound_lane<T: Trait<I>, I: Instance>(
+	lane_id: LaneId,
+) -> crate::outbound_lane::OutboundLane<RuntimeOutboundLaneStorage<T, I>> {
+	crate::outbound_lane::OutboundLane::new(RuntimeOutboundLaneStorage {
```
HCastano, line 177: Same thing about the imports here

modules/message-lane/src/inbound_lane.rs
```diff
@@ -0,0 +1,236 @@
...
+	/// Returns saved inbound message payload.
+	fn message(&self, nonce: &MessageNonce) -> Option<Self::Payload>;
+	/// Save inbound message in the storage.
+	fn save_message(&mut self, nonce: MessageNonce, payload: Self::Payload);
+	/// Remove inbound message from the storage.
+	fn remove_message(&mut self, nonce: &MessageNonce);
+}
+
+/// Inbound messages lane.
+pub struct InboundLane<Storage> {
```
HCastano, line 41: This is a bit of a nitpick, but sometimes it can be a bit confusing. I'd like to see this (and other places like `OutboundLane`) follow the standard generic convention of being a single letter, `S`.

modules/message-lane/src/inbound_lane.rs
svyatonik: Sure - bad processor/handler is able to block lane. If handler would be unable to handle message forever, the lane would stuck. The other option is to force it to handle message as it appears, but it can be impossible because of block time/weight. So it is up to handler to decide when and how to process messages.

modules/message-lane/src/lib.rs
svyatonik: > Would it be the job of the relay to make sure that the messages are in fact in storage before making this call?

Nope. The pallet has zero trust to relay. What relay does is very dumb routine - it checks whether any new messages have appeared in target storage. If there are new messages, it prepares a proof-of-storage (proof of inbound lane state) and submits tx to source node. The pallet will verify proof-of-storage (using your pallet :) ) and finally call this function.

Right now the pallet **API** is unfinished. As I said [here](https://github.com/paritytech/parity-bridges-common/pull/299#issuecomment-678325725), initially I thought that it'll be `BridgePallet` who will call exposed mutables, but actually it should be the opposite. So in next PR I'd add some trait like `trait DummyOrderedChain { fn check_message_proof() {} fn check_inbound_lane_state_proof() {} ... }`. Its implementation would be based on top of your Substrate-spceific `HeaderChain` implementation. And then I'll add more calls to this pallet - i.e. all what's currently called 'exposed mutables' would be callable using transactions.

Again - if you insist, I could add this trait to this PR. But would prefer to do this in follow-up. Changes wouldn't be huge.

> I imagine this function would (or should) be called immediately after receive_messages, right?

It is up to relayer to decide when to call this function. But our implementation would call it as soon as it finds that some messages have been received, right.

modules/message-lane/src/outbound_lane.rs
svyatonik: Even if we would have the most natural API for this structure (I mean queue), I doubt the number of writes would be the less than `N`. Unless we'll store it as `Vec<_>` as Tomek has suggested.

HCastano: I didn't realize you and Tomek had already had a similar conversation. I'm okay with whatever outcome comes from that

modules/message-lane/src/inbound_lane.rs
HCastano: I think it's fine for the processor to be able to decide when to handle messages, but I'm still wondering if there's some way to get unstuck. It seems like a pretty simple attack to spam the chain with invalid messages just to block lanes.

Would it be possible to send a different valid message at the same nonce which can take priority?

modules/message-lane/src/outbound_lane.rs
```diff
@@ -0,0 +1,269 @@
...
+
+	#[test]
+	fn confirm_receival_rejects_nonce_lesser_than_latest_received() {
+		run_test(|| {
+			let mut lane = outbound_lane::<TestRuntime, _>(TEST_LANE_ID);
+			lane.send_message(REGULAR_PAYLOAD);
+			lane.send_message(REGULAR_PAYLOAD);
+			lane.send_message(REGULAR_PAYLOAD);
+			assert_eq!(lane.storage.data().latest_generated_nonce, 3);
+			assert_eq!(lane.storage.data().latest_received_nonce, 0);
```
HCastano, line 165: This is part of every test, I think it would be worthwhile to move it into a function

modules/message-lane/src/outbound_lane.rs
```diff
@@ -0,0 +1,269 @@
...
+			// when lane is empty, nothing is pruned
+			assert_eq!(lane.prune_messages(100), 0);
+			assert_eq!(lane.storage.data().oldest_unpruned_nonce, 1);
+			// when nothing is confirmed, nothing is pruned
+			lane.send_message(REGULAR_PAYLOAD);
+			lane.send_message(REGULAR_PAYLOAD);
+			lane.send_message(REGULAR_PAYLOAD);
+			assert_eq!(lane.prune_messages(100), 0);
+			assert_eq!(lane.storage.data().oldest_unpruned_nonce, 1);
+			// after confirmation, some messages are receved
```
HCastano, line 259:
```suggestion
			// after confirmation, some messages are received
```

modules/message-lane/src/outbound_lane.rs
```diff
@@ -0,0 +1,269 @@
...
+
+		let prev_latest_processed_nonce = data.latest_processed_nonce;
+		data.latest_processed_nonce = latest_processed_nonce;
+		self.storage.set_data(data);
+
+		Some((prev_latest_processed_nonce + 1, latest_processed_nonce))
+	}
+
+	/// Prune at most `max_messages_to_prune` already received messages.
+	pub fn prune_messages(&mut self, max_messages_to_prune: MessageNonce) -> MessageNonce {
```
HCastano, line 103: This doesn't return the message nonce, it returns the number of pruned messages. It's subtle, but it tripped me up when reading the pruning test.

modules/message-lane/src/inbound_lane.rs
```diff
@@ -0,0 +1,236 @@
...
+		run_test(|| {
+			let mut lane = inbound_lane::<TestRuntime, _>(TEST_LANE_ID);
+			assert!(lane.receive_message(1, REGULAR_PAYLOAD, &mut TestMessageProcessor));
+			assert!(lane.storage.message(&1).is_none());
+			assert_eq!(lane.storage.data().latest_received_nonce, 1);
+		});
+	}
+
+	#[test]
+	fn process_message_does_noting_when_lane_is_empty() {
```
HCastano, line 183:
```suggestion
	fn process_message_does_nothing_when_lane_is_empty() {
```

modules/message-lane/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+
+// Parity Bridges Common is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Runtime module that allows sending and receiving messages using lane concept.
```
HCastano, line 17: I would expand on this description a bit. Adding information about how the lanes work, and what the expect flow of messages is. I imagine it'll be similar to you explained in the PR description, but maybe a bit less verbose.

modules/message-lane/src/inbound_lane.rs
svyatonik: I the message is invalid, then the handler needs to fail the message. I.e. just ignore + return `Processed`. That's the most straightforward way imo :)

modules/message-lane/src/outbound_lane.rs
svyatonik: Updated doc

modules/message-lane/src/inbound_lane.rs
tomusdrw: Yes. There is an issue with a seemingly valid message that doesn't fit the block size - i.e. the handler (message processor) MUST detect that and fail the message right away. Might be worth adding to the documentation of `OnMessageReceived`.

modules/message-lane/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+//! 4) external component (relay) delivers messages to bridged chain;
+//! 5) messages are processed in order (ordered by assigned nonce);
+//! 6) relay may send proof-of-receiving and proof-of-processing back to this chain.
+//!
+//! Once message is sent, its progress can be tracked by looking at module events.
+//! The assigned nonce is reported using `MessageAccepted` event. When message is
+//! accepted by the bridged chain, `MessagesDelivered` is fired. When message is
+//! processedby the bridged chain, `MessagesProcessed` by the bridged chain.
+
+#![cfg_attr(not(feature = "std"), no_std)]
```
tomusdrw, line 31:
```suggestion
#![cfg_attr(not(feature = "std"), no_std)]
#![warn(missing_docs)]
```

modules/message-lane/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+	type Event: From<Event> + Into<<Self as frame_system::Trait>::Event>;
+	/// Message payload.
+	type Payload: Parameter;
+	/// Maximal number of messages that may be pruned during maintenance. Maintenance occurs
+	/// whenever outbound lane is updated - i.e. when new message is sent, or receival is
+	/// confirmed. The reason is that if you want to use lane, you should be ready to pay
+	/// for it.
+	type MaxMessagesToPruneAtOnce: Get<MessageNonce>;
+	/// Called when message has been received.
+	type OnMessageReceived: Default + OnMessageReceived<Self::Payload>;
```
tomusdrw, line 61: Why do we require require an instance (and `Default` here)? This will make it hard to plug-in another pallet as dispatch. IMHO the callback should rather be a static method.

modules/message-lane/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+		}
+
+		correct_messages
+	}
+
+	/// Process stored lane messages.
+	///
+	/// Stops processing either when all messages are processed, or when processor returns
+	/// MessageResult::NotProcessed.
+	pub fn process_lane_messages(lane_id: &LaneId, processor: &mut impl OnMessageReceived<T::Payload>) {
```
tomusdrw, line 140: How are we expecting this to be called? IMHO it should be part of `on_initialize` to make some progress on message processing. This ensures that we are not competing with other runtime transactions, and some scheduling mechanism can ensure fairness (for instance depending on the number of queued messages you utilize from 10%-90% of the available block weight).

Note we should require the processing fee to be payed upfront and we will probably need weight assessment of the inner-call.  Related to #316, but created a new issue to track it and discuss it https://github.com/paritytech/parity-bridges-common/issues/319

modules/message-lane/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+	}
+
+	/// Receive TRUSTED proof of message processing.
+	///
+	/// Trusted here means that the function itself doesn't check whether the bridged chain has
+	/// actually processed these messages.
+	pub fn confirm_processing(lane_id: &LaneId, latest_processed_nonce: MessageNonce) {
+		let mut lane = outbound_lane::<T, I>(*lane_id);
+		let processed_range = lane.confirm_processing(latest_processed_nonce);
+		lane.prune_messages(T::MaxMessagesToPruneAtOnce::get());
```
tomusdrw, line 168: Why do we prune both here and when sending messages? Wouldn't one of these be enough?

tomusdrw: Actually I wonder if we should rather amortize pruning cost and do it during `on_initialize`. I.e. the relayer should pay a fee (weight) when the message is queued that should include the pruning cost as well. Then the pruning can happen in `on_initialize`.

Otherwise pruning randomly will make `weight` calculations a nightmare and will make the user experience quite unpredictable.

modules/message-lane/src/outbound_lane.rs
tomusdrw: As suggested in some other comment, I'm more inclined to prune in `on_initialize` and have the costs be amortized.

tomusdrw (commented):
Looks good to me, there is a few things to improve, but that could as well be done in a separate PRs.
I'd insist on fixing the pruning though. The costs should be amortized and pruning should be done in `on_initialize` imho to make it easier to calculate weights & fees in the future.

modules/message-lane/src/lib.rs
svyatonik: I was in doubts about where it is more appropriate to do that - see last point in PR description.The final idea is that "if you want to use the lane, then you should pay for its maintenance". So it is called from every place where (outbound) lane state is updated - i.e. whenever message is sent, or we confirm receiving/processing.

I'm not sure I understand your point about nightmare. `prune_messages()` has upper bound - it is: `(1 + T::MaxMessagesToPruneAtOnce::get()`)  writes. Afaiu, we just include that upper bound into default call weight and then return actual weight if less than `MaxMessagesToPruneAtOnce` headers has been pruned.

Re pruning from `on_initialize`/`on_finalize` - the main disadvantage is that we either need parallel structure (map?) to track lanes that actually need pruning, or we need to enumerate all lanes in the storage. Both options may end up with potentially large amount of reads (+writes sometimes) => slow down block production/import time, probably even if nothing needs to be pruned.

In any case - imo "amortized pruning" is exactly what's currently happening - i.e. we amortize cost of pruning between all lane users, not even all blockchain users.

Does it makes sense, or you're still convinced that pruning from on_initialize/on_finalize is better?

modules/message-lane/src/lib.rs
svyatonik: When I've been introducing `PruningStrategy` into ethereum bridge pallet, you've been asking to do `: Default`, because otherwise you have to specify the strategy type explicitly :) And if you have instance of `PruningStrategy`, then we may use rust type deduction mechanism to avoid that. So default isn't actually required here - could remove if desired

modules/message-lane/src/inbound_lane.rs
svyatonik: Updated docs

modules/message-lane/src/lib.rs
tomusdrw: `PruningStrategy` is a bit different, since it's not expected to be another substrate pallet - it was just a tiny piece of rust logic (or a bunch of numbers). Here, my understanding was, that we want to plug in say `Call::decode` dispatch pallet and that's going to be something like `pallet_call_dispatch::Module<Runtime>`, and I don't think it implements `Default`.

modules/message-lane/src/lib.rs
tomusdrw: > I'm not sure I understand your point about nightmare.  `prune_messages()` has upper bound

I've said "nightmare", because I meant exact calculations. Over-estimating and using an upper bound here will make a sub-optimal user experience and block utilisation. Assuming that queueing a message is ~2/3 state writes, and `MaxMessagesToPruneAtOnce` is going to be set to dozens, we may end up with quite high over estimation, and extra code to calculate refunds. I'd rather focus on lowering the cost (and it's variance) for relayers and users.

> Re pruning from on_initialize/on_finalize - the main disadvantage is that we either need parallel structure (map?) to track lanes that actually need pruning, or we need to enumerate all lanes in the storage. 

That's a good point and I really don't have the right answer. The best strategy here actually depends on whether we will have many in-frequently used lanes or rather a small amount of very frequently used ones. My main critique towards `MaxMessagesToPruneAtOnce` and pruning-when-used is that it makes actual costs of some operations a bit unpredicatble (although bonded), so it creates some unnecessary fee dynamics (like back-running a message to prevent paying the pruning costs), which might be significant if the chain fees are high.


That said, I think I'm convinced that it's better to prune-when-used, but I think we should only do it on `send_message`. Here is why. Imagine that we set `MaxMessagesToPruneAtOnce` to some large value. If the relayers are slow (or the other chain is slow), senders will keep adding more and more messages without really pruning anything. Then whenever this situation get's resolved, the relayers will be required to pay a high cost of pruning, when messages are received - without proper incentivisation they may simply stop doing this. There isn't much they loose really.
However senders, should be prevented from sending too many messages that are not properly processed (rate limitting), so the above situation should never really happen. Therefore I'd rather have relayers confirm delivery cheaply and then have senders prune old messages & being rate-limited at the same time. Since, rate-limitting should (or may) cause an (exponentialy) increasing cost of sending a message (which basically creates some upper limit), we may also use up this fee to prune more and more entries, hence the number of messages that gets pruned during send would go up along the undelivered queue size.
At worst we will end up with a bunch of unpruned messages in infrequently used lanes. To cater for this we might consider making the minimal number of messages to prune depenendent on the usage frequency as well. I.e. if the lane is used frequently, `send_message` prunes like at most 1 or 2 entries, if there was no messages for couple of blocks we will prune significantly more old messages on the next `send_message`.

So again, TL;DR I agree it's better to leave as-is for now (maybe without pruning on confirmation though), making `MaxMessagesToPruneAtOnce` dynamic in the future will prevent more sophisticated problems, but pruning in `on_initialize` doesn't really seem to solve any issues either.

modules/message-lane/src/lib.rs
svyatonik: I've remembered another reason when implementing this fix - sorry for not mentioning it earlier - working an another part now :/ We may use different `OnMessageReceived` impls in single pallet. One, default, from `fn receive_messages()`. And second one, custom, from `fn process_lane_messages()`. The reason behind that is that process-existing-messages version may use other scheduling algorithm. I.e. as you suggested elsewhere, we may process calls that are < 10% of block immediately and large calls only when `fn process_lane_messages()` is called. And that custom impl may need to store cumulative weight (or something similar) in-memory => hence instance. Plus, the fact that instance is required in this custom func, imo is a bit clearer than requiring a generic param.

That said, I have commit that removes instances [here](https://github.com/paritytech/parity-bridges-common/commit/50038e3af5e573ed5d8eb13fc8d009d03bf738d2). Can cherry-pick on demand. But imo better to leave as is.

modules/message-lane/src/lib.rs
svyatonik: Ok, removed pruning from confirmation calls

modules/message-lane/src/lib.rs
svyatonik: I've added this, but now it shows warning about undocumented struct. I believe it comes somewhere from decl_module, or decl_storage macros. So I've reverted this for now

svyatonik: Have checked Substrate - we do not have this attr in pallets there either :)

modules/message-lane/src/lib.rs
tomusdrw: Let's see how it will look like when we plug in actual "production" modules and revisit then.

tomusdrw (approved):

HCastano:
@svyatonik You should probably rebase https://github.com/paritytech/parity-bridges-common/pull/317 on top of this before it gets squash merged. It'll help keep the Git history cleaner.

Issue #211, opened 2020-07-17
`Call::decode` dispatcher


tomusdrw:
Write a pallet that takes a blob and tries to decode a runtime `Call` out of it and then dispatch with bridge-specific origin. Make sure to check `spec_version` to avoid dispatching stale stuff.

PR #300, opened 2020-08-21
Initial version of `Call::decode` dispatch.

tomusdrw:
Closes: #211 

Yet to do:
- [x] Plug into runtime
- [ ] ~~Change `MessageDispatch::dispatch` to actually return a `DispatchResult` with meaningful errors~~
- [x] Generate events on successful/failed dispatch
- [x] More tests
- [x] Weights (how much we plan the dispatch to take should be part of the message, so that the scheduler can make decisions if it wants to dispatch or not. Secondly the message delivery might need that info as well to calculate it's own weight).


svyatonik:
I've changed interface a bit:
1) instead of bool (and `DispatchResult` as you've suggested), `MessageDispatch::dispatch()` now returns nothing. I'm not sure how to use this `bool`/`DispatchResult` from outside of dispatch (i.e. in scheduler). So now `MessageDispatch` is like a black box - you're just throwing messages there and don't care about result. It is scheduler duty to select messages from the queue and feed them to dispatch. If you have another vision, please share;
2) I've added events as you've suggested. My vision is that these are mostly useful for message submitters to know **how** message has been processed. They already know that message has been delivered/processed from message delivery mechanism. So this is just to know result of dispatch;
3) message submitter now must include weight of `Call` into his message. We only care about static weights, not the post-dispatch weight. Scheduler will determine which messages he's going to process in this block using that information.

I'm leaving this PR in draft state, because I want to reuse `InstanceId` introduced in #317 . 

In future PRs:
1) implement scheduler (component that connects delivery + dispatch). This most probably would be combination of `bp_message_lane::OnMessageReceived` implementation that will try to dispatch everything immediately, but if block is already full, will use message-lane storage to remember lanes-with-queued-messages and will process something in the next block' on_initialize;
2) deal with fees + rewards + weights

modules/call-dispatch/src/lib.rs
```diff
@@ -0,0 +1,348 @@
...
+		let expected_version = <T as frame_system::Trait>::Version::get().spec_version;
+		if spec_version != expected_version {
+			frame_support::debug::trace!(
+				"Message {:?}/{:?}: spec_version mismatch. Expected {:?}, got {:?}",
+				origin,
+				id,
+				expected_version,
+				spec_version,
+			);
+			Self::deposit_event(Event::<T>::MessageVersionSpecMismatch(origin, id));
```
tomusdrw, line 107: Shall we pass the expected / got in the event to make it easier for debugging?

modules/call-dispatch/src/lib.rs
```diff
@@ -0,0 +1,348 @@
...
+		let expected_weight = call.get_dispatch_info().weight;
+		if weight < expected_weight {
+			frame_support::debug::trace!(
+				"Message {:?}/{:?}: passed weight is too low. Expected at least {:?}, got {:?}",
+				origin,
+				id,
+				expected_weight,
+				weight,
+			);
+			Self::deposit_event(Event::<T>::MessageWeightMismatch(origin, id));
```
tomusdrw, line 123: Same as above.

modules/call-dispatch/src/lib.rs
```diff
@@ -0,0 +1,348 @@
...
+			return;
+		}
+
+		// finally dispatch message
+		let origin_account = Self::bridge_account_id(&origin);
+		let dispatch_result = call.dispatch(frame_system::RawOrigin::Signed(origin_account).into());
```
tomusdrw, line 129: We will most likely need to keep track of consumed `weight` at some level. AFAIU The stack looks like this:
1. Delivery (external call)
2. Scheduler
3. Dispatch (internal `call.dispatch`)

Substrate requires us to provide `weight` upper bound at (1). The simplest calculation would be just a `Min(block_weight_left, SUM(weight_of_each_message))`, a better solution would delegate to scheduler, which could restrict this even further. Then during the actual dispatch, we should keep track of actually consumed weight to: a) let the scheduler process more calls if possible and needed, b) let the relayer pay smaller fees for (1), and c) better utilize available block weight.

modules/call-dispatch/src/lib.rs
```diff
@@ -0,0 +1,348 @@
...
+		Self::deposit_event(Event::<T>::MessageDispatched(
+			origin,
+			id,
+			dispatch_result.map(drop).map_err(|e| e.error),
+		));
+	}
+}
+
+impl<T: Trait> Module<T> {
+	fn bridge_account_id(origin: &T::MessageOrigin) -> T::AccountId {
```
tomusdrw, line 146: Just realised that it might be useful in other pallets, so should probably be part of some `bp-` crate (maybe same place as `InstanceId`).

tomusdrw (commented):
Looks good! I think we should somehow pass the actually consumed weight up in the stack.

modules/call-dispatch/src/lib.rs
svyatonik: Changed return type to `Weight`. I had to 'fix' `Dispatchable::PostInfo` to be `frame_support::dispatch::PostDispatchInfo`. Otherwise it seems impossible to get actual weight.

Some extended context: I expect scheduler to only rely to remaining weight of current block (i.e. `frame_system::BlockWeight`) and maximal weight of current block. So imo this change isn't required - it is dispatcher who needs to update `frame_system::BlockWeight` (it is missing from current PR, but most probably will be changed in follow-up PRs).

modules/call-dispatch/src/lib.rs
svyatonik: Extracted. Also removed recently introduced `MessageOrigin` - message origin is always `InstanceId` now.

tomusdrw (commented):
LGTM! (can't approve since it's my own PR :rofl:)

PR #317, opened 2020-08-24
Custom RPC methods for DummyOrdered pallet

svyatonik:
The most arguable decision here is: I think that it is better to start with custom RPCs for our pallet just because we hide storage details behind RPC => relay or whoever won't need to deal with storage keys. So you could operate with high-level concepts like lane + nonces && leave it up to runtime to deal with storage. But probably you, guys, have another opinion?

TBD:
- [x] - do we want to use `InstanceId` or several RPC instances?
- [x] - tests;
- [x] - `SerializableStorageProof` should be just `Bytes`, which is scale-encoded `StorageProof`.


modules/message-lane/rpc/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+}
+
+/// Provides RPC methods for interacting with message-lane pallet.
+#[rpc]
+pub trait MessageLaneApi<BlockHash> {
+	/// Returns proof-of-message(s) in given inclusive range.
+	#[rpc(name = "prove_messages")]
+	fn prove_messages(
+		&self,
+		instance: InstanceId,
```
svyatonik, line 69: I do not like this argument (`instance`), but the alternative is to introduce several instances of `MessageLaneRpcHandler`, which is even worse imo. Maybe you have different opinion?

tomusdrw: I think it would make sense to have some kind of pallet-instances namespacing. I.e. we would have a bunch of pallets under say `eth-pow`  ID (bridge, message lanes, message dispatch, etc). Then the id would be common for these pallets.

svyatonik: Ok, then I'm leaving `InstanceId` arg. Do you think it worth to change it to `String` then? Also - leaving typedef here, because we do not have any appropriate crate now (that is accessible from both node && this crate). Probably worth adding new `bp-runtime-primitives`?

tomusdrw: >  Do you think it worth to change it to String then?
No, I think `InstanceId` or some other couple-of-bytes typedef is fine.

> Probably worth adding new bp-runtime-primitives?
:+1:

modules/message-lane/rpc/src/error.rs
```diff
@@ -0,0 +1,59 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Possible errors and results of message-lane RPC calls.
+
+/// Future Result type.
+pub type FutureResult<T> = Box<dyn jsonrpc_core::futures::Future<Item = T, Error = jsonrpc_core::Error> + Send>;
```
tomusdrw, line 20: That's `jsonrpc_core::BoxFuture` actually.

tomusdrw (approved):

primitives/runtime/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+
+//! Primitives that may be used at (brides) runtime level.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+/// Id of deployed module instance. We have a bunch of pallets that may be used in
+/// different bridges. E.g. message-lane pallet may be deployed twice in the same
+/// runtime to bridge ThisChain with Chain1 and Chain2. Sometimes we need to be able
+/// to identify deployed instance dynamically. This type is used for that.
+pub type InstanceId = [u8; 4];
```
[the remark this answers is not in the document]
svyatonik: Not at this moment

svyatonik:
> So if I understand this PR correctly, it adds a set of RPCs so we can query the chain storage to see if there message have been sent/received.

Right

> But at no point does it actually interact with the message-lane pallet via Runtime APIs - it's all done through the "general" chain storage, right?

Yes, it only generates proofs. And proofs generation is the same for every storage cell across all runtime. The only point where we 'touch' message-lane pallet would be `Runtime` trait implementation. Since we'll have single instance of RPCs for (potentially) multiple instances of message-lane pallet, it'll need some way to get `StorageKey` inside this particular pallet (that is referenced by `InstanceId`).

> Is it correct to say that the main consumer of the RPC will be the bridge relay which is managing the sending/receiving of messages?

Yeah - I'm not sure if these methods are usable anywhere outside of our relay.

svyatonik:
> > So if I understand this PR correctly, it adds a set of RPCs so we can query the chain storage to see if there message have been sent/received.

> Right

Actually, no - for querying we have separate runtime API that is accessible using regular Substrate APIs. This PR introduces APIs that are proving storage values. So before proving relay would need to call into runtime to check if messages "have been sent/received" and then call these RPCs to prove that to the bridged chain.

HCastano (approved):

PR #318, opened 2020-08-25
Relay DummyOrdered messages

### LINKED ISSUE #213: Extend Relayer to support `DummyOrdered` delivery protocol
Extend the relayer to be able to query substrate chain for outbound messages and deliver them to the foreign chain (wrap in a transaction).

svyatonik:
closes #213

I'm opening this early to be able to discuss/convince @HCastano not to invest much time in #281 :)

This PR is in active development phase => please do not look at the code now. Only ideas matters.

So the idea is:
1) for now it is the `dead_code`, because we do not have any exposed modules in runtime;
2) this 'loop' is more complex than exchange transactions relay loop, but much easier than headers sync loop. Anyway - there's a lot of similar/duplicated code right now. The only solution I see is to use macros => probably I'll be able to introduce something in this PR, or we may extract similar code later. But it needs to be done. **UPD**: I've spent some time trying to make message-loop-macros, but itsn't trivial => I suggest deferring this for now;
3) @HCastano I'm suggesting to implement similar error handling in sync loop too. I mean - run loop until any transport error is met. After error, exit from loop, wait for some time && restart. Since we may use same set of macro(s) for both loops, probably not worth continuing #281 now? It is up to you, of course;
4) why `futures::select` instead of easy-to-read + straightforward await-s (i.e. why this loop is more like headers sync loop than like exchange loop)? In headers loop the main idea is that we're filling in-memory buffer (`Headers`) with data from source node, while still talking to target node. This allows us batching headers + not blocking on data waits from one of nodes. Here, the idea is the same - since e.g. finalization isn't instant, we may batch several messages in single delivery transaction, without blocking on finality.await;
5) there are 3 sub-lanes within single one-way lane: (1) first is used to deliver messages, (2) 2nd is used to confirm receiving and (3) 3rd is used to confirm processing. I've called them races :) Right now there's only one unfinished race implementation - message delivery race. ~I think~ I'll add other two in follow-up PRs.


relays/ethereum/src/message_race_loop.rs
```diff
@@ -0,0 +1,337 @@
...
+
+// Parity Bridges Common is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+//! Message confirm (either receiving or processing) race within one-way message lane.
+
+// Until there'll be actual message-lane in the runtime.
+#![allow(dead_code)]
```
tomusdrw, line 17: same. I think we should get rid of this now.

relays/ethereum/src/message_race_loop.rs
```diff
@@ -0,0 +1,337 @@
...
+	fn is_empty(&self) -> bool;
+	/// Called when latest nonce is updated at source node of the race.
+	fn source_nonce_updated(&mut self, at_block: SourceHeaderId, nonce: MessageNonce);
+	/// Called when latest nonce is updated at target node of the race.
+	fn target_nonce_updated(
+		&mut self,
+		nonce: MessageNonce,
+		race_state: &mut RaceState<SourceHeaderId, TargetHeaderId, MessageNonce, Proof>,
+	);
+	/// Should return `Some(_)` with if we need to deliver proof of given nonces from
```
tomusdrw, line 108: I think the docs needs an update here, it returns `Some` with the range of nonces that require proof, right?

relays/ethereum/src/message_race_loop.rs
```diff
@@ -0,0 +1,337 @@
...
+
+						strategy.target_nonce_updated(latest_nonce, &mut race_state);
+					},
+					&mut target_go_offline_future,
+					|delay| async_std::task::sleep(delay),
+					|| format!("Error retrieving latest nonce from {}", P::target_name()),
+				).fail_if_connection_error(FailedClient::Target)?;
+			},
+
+			// proof generation and submittal
```
tomusdrw, line 229:
```suggestion
			// proof generation and submission
```

relays/ethereum/src/message_race_loop.rs
```diff
@@ -0,0 +1,337 @@
...
+			source_client_is_online = false;
+
+			if let Some(nonces_range) = strategy.select_nonces_to_deliver(&race_state) {
+				log::debug!(
+					target: "bridge",
+					"Asking {} to prove nonces in range {:?}",
+					P::source_name(),
+					nonces_range,
+				);
+				let at_block = race_state.source_state.as_ref().expect("TODO").best_self.clone();
```
tomusdrw, line 288: TODO?

relays/ethereum/src/message_race_loop.rs
```diff
@@ -0,0 +1,337 @@
...
+				source_generate_proof.set(race_source.generate_proof(at_block, nonces_range).fuse());
+			} else if source_latest_nonce_required {
+				log::debug!(target: "bridge", "Asking {} about latest generated message nonce", P::source_name());
+				let at_block = race_state.source_state.as_ref().expect("TODO").best_self.clone();
```
tomusdrw, line 292: ditto

relays/ethereum/src/utils.rs
```diff
@@ -114,3 +122,102 @@ pub fn format_ids<Id: std::fmt::Debug>(mut ids: impl ExactSizeIterator<Item = Id
...
+			ProcessFutureResult::Failed => Ok(false),
+			ProcessFutureResult::ConnectionFailed => Err(failed_client),
+		}
+	}
+}
+
+/// Process result of the future from a client.
+///
+/// Returns bool which means whether or not the client we're interacting with
+/// is online. In this context what online means is that the client is currently not
```
tomusdrw, line 177: I feel we should think of a better name instead of reusing `online/offline` to indicate that - you immediately think about the underlying connection when you hear this. Why not say that the client is `ready/notready` or `idle/busy`?

relays/ethereum/src/utils.rs
```diff
@@ -114,3 +122,102 @@ pub fn format_ids<Id: std::fmt::Debug>(mut ids: impl ExactSizeIterator<Item = Id
...
+	futures::stream::unfold((), move |_| async move {
+		async_std::task::sleep(timeout).await;
+		Some(((), ()))
+	})
+}
+
+/// Which client has caused error.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum FailedClient {
+	Source,
```
tomusdrw, line 137: Nit: docs.

relays/ethereum/src/message_lane_loop.rs
```diff
@@ -0,0 +1,601 @@
...
+//! Single relay instance delivers messages of single lane in single direction.
+//! To serve two-way lane, you would need two instances of relay.
+//! To serve N two-way lanes, you would need N*2 instances of relay.
+//!
+//! Please keep in mind that the best header in this file is actually best
+//! finalized header. I.e. when talking about headers in lane context, we
+//! only care about finalized headers.
+
+// Until there'll be actual message-lane in the runtime.
+#![allow(dead_code)]
```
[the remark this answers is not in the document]
svyatonik: No, we don't have anything yet

relays/ethereum/src/message_lane_loop.rs
```diff
@@ -0,0 +1,601 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Message delivery loop. Designed to work with message-lane pallet.
+//!
+//! Single relay instance delivers messages of single lane in single direction.
+//! To serve two-way lane, you would need two instances of relay.
+//! To serve N two-way lanes, you would need N*2 instances of relay.
```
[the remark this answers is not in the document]
svyatonik: I don't like that idea. Imo it is better to run one instance of relay per one application (i.e. one-direction lane).

In any case - this could be achieved just by spawning several instances of message-lane-loop on start. Since spawning is not a part of this PR (just because there are no chains, and therefore implementations of `SourceClient` and `TargetClient`), I suggest to discuss/solve it later.

relays/ethereum/src/message_race_delivery.rs
```diff
@@ -0,0 +1,384 @@
...
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Parity Bridges Common is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+use crate::message_lane::{MessageLane, SourceHeaderIdOf, TargetHeaderIdOf};
```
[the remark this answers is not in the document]
svyatonik: This concept is introduced in `message_lane.rs`. I've added some module-level docs to `message_race_*.rs`

relays/ethereum/src/message_race_loop.rs
svyatonik: :) Sorry- should have checked for TODOs left

relays/ethereum/src/utils.rs
svyatonik: I've actually replaced `bool` with enum here => should be self-descriptive

svyatonik: (=> removed obsolete docs)

relays/ethereum/src/main.rs
```diff
@@ -27,6 +27,10 @@ mod exchange_loop;
 mod exchange_loop_metrics;
 mod headers;
 mod instances;
+mod message_lane;
```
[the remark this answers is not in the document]
svyatonik: I've extracted this to separate issue - #344. Otherwise this PR will blow up

relays/ethereum/src/message_lane_loop.rs
```diff
@@ -0,0 +1,591 @@
...
+async fn run_until_connection_lost<P: MessageLane, SC: SourceClient<P>, TC: TargetClient<P>>(
+	source_client: SC,
+	source_tick: Duration,
+	target_client: TC,
+	target_tick: Duration,
+	stall_timeout: Duration,
+	exit_signal: impl Future<Output = ()>,
+) -> Result<(), FailedClient> {
+	let mut source_retry_backoff = retry_backoff();
+	let mut source_client_is_online = false;
```
tomusdrw, line 173: I'm still not sold on the `online/offline` concept, but since it's prevalent in other sync loops already, I can accept it.

tomusdrw (approved):
The abstraction looks good to me. Happy to merge as is and then see improvements when this is actually being integrated.

relays/ethereum/src/message_lane_loop.rs
```diff
@@ -0,0 +1,591 @@
...
+	let mut target_client_is_online = false;
+	let mut target_state_required = true;
+	let target_state = target_client.state().fuse();
+	let target_go_offline_future = futures::future::Fuse::terminated();
+	let target_tick_stream = interval(target_tick).fuse();
+
+	let (
+		(delivery_source_state_sender, delivery_source_state_receiver),
+		(delivery_target_state_sender, delivery_target_state_receiver),
+	) = (unbounded(), unbounded());
```
HCastano, line 189: Do we not risk memory leaks with `unbounded` channels?

relays/ethereum/src/message_race_delivery.rs
```diff
@@ -0,0 +1,385 @@
...
+
+use async_trait::async_trait;
+use futures::stream::FusedStream;
+use std::{collections::VecDeque, marker::PhantomData, ops::RangeInclusive, time::Duration};
+
+/// Maximal number of messages to relay in single transaction.
+const MAX_MESSAGES_TO_RELAY_IN_SINGLE_TX: u32 = 4;
+
+/// Run message delivery race.
+pub async fn run<P: MessageLane>(
```
HCastano, line 32: This is basically just a pass-through to `message_race_loop`, so why not call it from `message_lane_loop` directly?

relays/ethereum/src/message_lane_loop.rs
```diff
@@ -0,0 +1,591 @@
...
+				let mut data = self.data.lock();
+				(self.tick)(&mut *data);
+				data.is_source_reconnected = true;
+			}
+			self
+		}
+
+		async fn state(&self) -> Result<SourceClientState<TestMessageLane>, Self::Error> {
+			let mut data = self.data.lock();
+			(self.tick)(&mut *data);
```
HCastano, line 379: On first read this was pretty confusing

HCastano (commented):
Honestly, I don't quite follow all the code here. Since Tomek's approved it I'm okay with this getting merged without my approval.

relays/ethereum/src/message_race_loop.rs
```diff
@@ -0,0 +1,359 @@
...
+}
+
+/// State of race source client.
+type SourceClientState<P> = ClientState<<P as MessageRace>::SourceHeaderId, <P as MessageRace>::TargetHeaderId>;
+
+/// State of race target client.
+type TargetClientState<P> = ClientState<<P as MessageRace>::TargetHeaderId, <P as MessageRace>::SourceHeaderId>;
+
+/// One of message lane clients, which is source client for the race.
+#[async_trait(?Send)]
```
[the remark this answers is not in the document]
svyatonik: Without it required futures have `Send` requirement, which fails for some of our future(s) (I do not exactly remember which one now). Since we have single-thread executor, this is fine.

relays/ethereum/src/message_lane_loop.rs
svyatonik: How leaks may happen here? You probably mean high memory consumption? The only hypothetical case is when we fail to wake up on non-empty channel. Which should not happen because ready-futures-to-awake are selected in random order.

relays/ethereum/src/message_race_delivery.rs
svyatonik: It is exposing 4 private types vs exposing 1 function. I'm for the latter

PR #321, opened 2020-08-27
Update substrate reference to rc6

svyatonik:
This fixes compilation issue

### DIFF
--- a/bin/node/node/src/service.rs
+++ b/bin/node/node/src/service.rs
@@ -32,9 +32,7 @@ use bridge_node_runtime::{self, opaque::Block, RuntimeApi};
 use sc_client_api::{ExecutorProvider, RemoteBackend};
 use sc_executor::native_executor_instance;
 pub use sc_executor::NativeExecutor;
-use sc_finality_grandpa::{
-	FinalityProofProvider as GrandpaFinalityProofProvider, SharedVoterState,
-};
+use sc_finality_grandpa::{FinalityProofProvider as GrandpaFinalityProofProvider, SharedVoterState};
 use sc_service::{error::Error as ServiceError, Configuration, TaskManager};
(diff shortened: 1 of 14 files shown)


bin/node/node/src/service.rs
```diff
@@ -16,14 +16,24 @@
 
 //! Service and ServiceFactory implementation. Specialized wrapper over substrate service.
 
+// =====================================================================================
+// =====================================================================================
+// =====================================================================================
```
HCastano, line 21: I think the update guide section is useful, but tbh I'm not a fan of these. They add a lot of visual noise when CAPS do the trick here

bin/node/node/src/service.rs
```diff
@@ -16,14 +16,24 @@
...
+// UPDATE GUIDE:
+// 1) replace everything with node-template/src/service.rs contents;
```
HCastano, line 23:
```suggestion
// 1) replace everything with node-template/src/service.rs contents (found in main Substrate repo);
```

HCastano: I don't want a new person coming on-board and thinking that there's a missing`node-template` folder in this repo

HCastano:
Benchmarking is broken. Maybe it has to do with this: https://github.com/paritytech/substrate/pull/6812?

bin/node/node/src/service.rs
svyatonik: The idea is to actually make a visual noice :)))

HCastano: I think sections like this should only be used in places where a critical amount of attention is needed. Here there's not much else happening, so CAPS suffice to draw attention. Worst case we merge to master, and quickly realize our syncing doesn't work because we can't connect to the RPCs

PR #331, opened 2020-08-31
Whitelist extensions in CORS settings for PoA nodes

tomusdrw:
That's to allow Metamask connecting to the node.
CC @Tbaut


svyatonik (approved):
(assuming it'll help :) )

Issue #316, opened 2020-08-24
Relayers Incentivisation


tomusdrw:
Currently the relay code is responsible for the following:
1. Header Sync
2. Message Delivery
3. Message Dispatch

Header Sync is essentially a low-level "utility" of the bridge and it seems it's possible to do it in a way that doesn't impose significant costs on the Relayers - i.e. we can use unsigned (or fee-less) transactions (at least for Substrate<>Substrate bridge, PoA is a different story).

Message Delivery is usually tangled with Message Dispatch when thinking, and that part seems to require some sort of incentivisation mechanism for the relayers, since, especially the dispatch, is going to be quite heavy and user-controlled.
We should think of some kind of fee mechanics to make sure that that when message is dispatched (no matter if successful or not) the party who triggered this dispatch is rewarded.

We may either:
1. Split Delivery & Dispatch, try to make the first free and the second incentivised
2. Or we may introduce fee for a bundled Delivery & Dispatch

The latter approach might cause issues due to in-lane ordering (see #215), similar to how a single low-fee Ethereum transaction can block subsequent (higher-fee) transactions from the same account. Here however, there is no way for the user to substitute transactions after they are submitted, so we can incorporate some sort of  "total fee" metric, where we rather look at average fee/cost of all transactions in a lane, so a parent-pays-for-child-like schemes could work.

tomusdrw:
It would be really nice if the incentivisation scheme was a bit more ambitious than: "the winner takes all" approach, where the fee is distributed directly to the relayer triggering the dispatch. Some sort of:
1. Relay active set elections (per session)
2. Distribution of relay tasks between relayers in the active set (+ fallback relayers)
3. Reward accumulation within session
3. Fair distribution of the reward towards all non-misbehaving relayers in the session

would be nice (basically very similar to how babe/grandpa is done, maybe even with nomination + phragmen).

tomusdrw:
We should make sure to also incentivise submitters of delivery proofs, see more details: https://github.com/paritytech/parity-bridges-common/pull/385#discussion_r500295959

svyatonik:
I think it is closed by #385 - both message-relayer and proof-relayer accounts are reported to the callback. So any complex strategy may be built. There's now instant-payments implementation of the callback that is used both by Rialto and Millau.

PR #332, opened 2020-08-31
Introduce `BridgedHeaderChain` and `LaneMessageVerifier` traits in DummyOrdered pallet

svyatonik:
This PR solves the https://github.com/paritytech/parity-bridges-common/pull/299#issuecomment-678325725 comment + also introduces `LaneMessageVerifier` that we have discussed earlier.

**IMPORTANT THING**: this PR is not about paying for delivery/dispatch, so no any callbacks for taking fees. This should probably be part of #316 ?

ToDo:
- [x] Why events are not deposited in tests?
- [x] More traces in runtime


svyatonik:
I'm closing it, because in context of fees+weights+by-weight-scheduler (future in-progress PR), these traits hanged a bit. So this PR became a part of that future PR.

Issue #319, opened 2020-08-26
Message Lane Processing Scheduler


tomusdrw:
The current form of Message Delivery mechanism (see #299, related #212) splits message delivery and message processing. Independently from the dispatch mechanism (see #214 or #211) we will need some way to make sure that:
- At least some queued messages are always processed (i.e. we make some progress in every block).
- We have some way to assess the complexity of the dispatch (i.e. `Weight`) to prevent going over block weight limit.
- We scale the dispatch in some sensible way, since we compete for block weight with other transactions in the runtime.

I can think of two ways how we can achieve some fairness and scale the amount of dispatch within every block.
1. Relayers are allowed to send a "process messages" transaction, which pays for the execution. The more messages is queued in a lane, the higher fee (see #316) the relayer should get from making progress on that lane. The mechanism should also make sure that the reward from processing the lane is higher than the fee the relayer has to pay to get it's transaction into block in the time of congestion.
2. The processing fee is payed upfront by the relayer when the message is queued and the dispatch happens during `on_intialize` callback of the message lane pallet. With some scaling mechanism, depending on the number of queued messages we could for instance utilize from `10%-90%` of the entire block weight for the dispatch. The advantage of this approach is that processing messages from the bridge does not directly compete with regular chain transactions and we are guaranteed to make some progress (bridge messages can be considered privileged).

Initially I'm in favour of (2), since it seems simpler to implement. In both cases, though, we will need the message payload to be able to tell us upfront the maximal `Weight` it might consume during dispatch.

svyatonik:
Closing, as this is no longer relevant - we now have no scheduled messages concept.

PR #339, opened 2020-09-07
Fees, weights, message delivery and dispatch

svyatonik:
related #316, #319 (I'm not closing issues, because nothing is plugged into runtime yet)

This PR consists of 2 parts: first 5 commits are extracting (already obsolete) `BridgedHeaderChain` and `LaneMessageVerifier` traits. See #332 for some details.

Other commits are about splitting these traits even further and dealing with weights + fees in message delivery and dispatch.

Below is detailed design of the weights+fees. The logic is mostly in `modules/message-lane/src/lib.rs` (`fn send_message`, `fn receive_messages_proof()`, `fn receive_message_receiving_proof()` and `fn receive_message_processing_proof()`).

---

Everything starts when the submitter calls `fn send_message()`. The message is first verified:
1) by `fn TargetHeaderChain::verify_message()` - implementations must at least ensure that the relay will be able to submit transaction, containing this message, to transaction pool of the target chain node. E.g. if target chain is BTC, then the size of the message could not be more than 1MB (minus delivery overhead, if required);
2) by `fn LaneMessageVerifier::verify_message()` - implementations could add e.g. some kind of security over lanes, or reject messages if paid fee is too low (will require oracles).

Then the message fee (called `delivery_and_dispatch_fee` in the code) is withdrawn from the submitter account by `fn MessageDeliveryAndDispatchPayment::pay_delivery_and_dispatch_fee()`. The implementation of `MessageDeliveryAndDispatchPayment` could e.g. move funds from submitter account to some relayers-fund account and then split relayers-fund among all participated relayers.

The sumbitter should select message fee carefully. It should include at least:
1) message-delivery fee - the fee that the relayer will pay for delivery-transaction;
2) message-dispatch fee - the fee that the relayer will pay for message dispatch;
3) message-receiving fee - the fee that the relayer will pay for message receiving confirmation transaction;
4) relayer interest.

After fee is paid, the message is considered accepted and any running relayer may deliver it to the target chain. The altruistic relayer logic is already implemented in #318. Before bridging non-test chains we'll need to make some changes, so that relayer won't deliver messages with low message fee.

When the relayer submits delivery-transaction, he pays for transaction itself (indirectly, using transaction fees) and pays for delivered messages dispatch. Then he has several options on how to control dispatch:
1) if relayer is sure that there are other queued messages blocking the lane AND he do not want to pay additional fees for processing these messages, it sets `max_dispatch_weight` parameter of the delivery-transaction to zero. This guarantees that no messages will be processed during this call && we'll allow other transactions to be included in this block;
2) if relayer is sure that there are no other queued messages in the lane, he may set `max_dispatch_weight` to weight of all bundled messages and they'll be processed immediately (given that there are no other conditions that are blocking messages from dispatching immediately);
3) relayer may choose any value of `max_dispatch_weight` (of course it only makes sense until `MaxExtrinsicWeight` is reached) and affect how fast messages are dispatched. It makes no sense to set `max_dispatch_weight` to values larger than dispatch-weight of all bundled transactions unless relayer is incentivized to do that in some other way;
4) incentivized in some other way relayers may submit delivery-transactions with 0 bundled messages just to be sure that the messages are dispatched.

After message is delivered, it is either processed immediately, or is queued. It is processed immediately iff: (1) there are no other messages queued in the lane and (2) relayer has provided such `max_dispatch_weight`, so that message dispatch fits it. If the message is queued, then it will be processed later, when `fn QueuedMessageDispatch::dispatch()` will be called. And it is called: (1) from block `on_initialize()` and (2) from delivery-transactions if some weight is left after all bundled messages have been processed/queued.

---

Some ideas for future PRs:
I) as you can see, `struct MessageData` now includes message fee field. Since we may also estimate message dispatch weight, it allows us to approximate exchange rate between source chain fee-tokens and target chain weights (which, in turn, may be converted into target chain fee tokens). Probably we may use this information to verify outgoing messages - e.g. rejecting messages which `MessageFee < ConvertToTargetFee(Message.Weight)`. I'm not sure about this yet;
II) on non-test chains, we'll need a way to incentivize relayers. I'm currently leaning towards adding `RelayerSourceChainAccount` parameter to the delivery-transaction + saving set of these accounts in `InboundLaneState`. Old entries from this set are pruned when relayer has submitted nothing for, say, last 1024 blocks. By providing proof of `InboundLaneState` to the source chain, we may split relayers-fund between all participated relayers.

---

TODO:
- [x] Deal with messages with weight > max extrinsic weight (drop)
- [x] Self-review


svyatonik:
I've found some problems trying to plug-in current code into runtime:
1) message payload should be opaque to the message-lane module. Otherwise we may be unable to decode incoming messages => lane would stuck. Messages must only be decoded by the `MessageDispatcher`/`QueuedMessageDispatcher`;
2) there's no **reusable** implementations of `MessageDispatcher`/`QueuedMessageDispatcher` for `ByWeightMessageDispatch`;
3) module that are supporting instances (i.e. message-lane), should also have generic events (depending on `Runtime` and `Instance`).

I don't want this PR to grow any more, so I suggest to fix these issues in follow-up (plugging into runtime) PR. And concentrate on handling fees + weights during delivery and dispatch in this PR.

modules/message-lane/src/lib.rs
```diff
@@ -90,19 +159,237 @@ decl_module! {
...
+					queued_messages_dispatched,
+				);
+			}
+
+			Ok(())
+		}
+
+		/// Receive messages receiving proof from bridged chain.
+		#[weight = 0] // TODO: update me (https://github.com/paritytech/parity-bridges-common/issues/78)
+		pub fn receive_message_receiving_proof(origin, proof: MessagesReceivingProofOf<T, I>) -> DispatchResult {
```
tomusdrw, line 337:
```suggestion
		pub fn receive_message_delivery_proof(origin, proof: MessagesDeliveredProofOf<T, I>) -> DispatchResult {
```
Why is it `MessagesReceiving` and not `MessagesDelivered`?

modules/message-lane/src/lib.rs
```diff
@@ -90,19 +159,237 @@ decl_module! {
...
+
 		/// Send message over lane.
 		#[weight = 0] // TODO: update me (https://github.com/paritytech/parity-bridges-common/issues/78)
 		pub fn send_message(
 			origin,
 			lane_id: LaneId,
 			payload: T::Payload,
-		) {
-			let _ = ensure_signed(origin)?;
+			delivery_and_dispatch_fee: T::MessageFee,
```
tomusdrw, line 184: I feel this should rather be calculated by `LaneMessageVerifier` - it's supposed to verify that so it knows the minimal value anyway. Is it possible to create a market for messages (i.e. higher paying fees go first?). Maybe we should change that to just `tip`, and have the sender pay the fee calculated by `LaneMessageVerifier` plus the tip?

modules/message-lane/src/lib.rs
```diff
@@ -90,19 +159,237 @@ decl_module! {
...
+					"Message to lane {:?} is rejected by lane verifier: {:?}",
+					lane_id,
+					err,
+				);
+
+				Error::<T, I>::MessageRejectedByLaneVerifier
+			})?;
+
+			// let's withdraw delivery and dispatch fee from submitter
+			T::MessageDeliveryAndDispatchPayment::pay_delivery_and_dispatch_fee(
```
tomusdrw, line 218: What is supposed to happen with the fee here? I feel like this is too abstracted from the pallet. IMHO we should import `Currency` trait here and:
1. When someone is sending a message we should `lock` funds for `delivery_and_dispatch_fee`.
2. When the message is delivered (or processed - not sure), we should transfer the fee to the relayer.

Right now it's not clear to me which layer is supposed to be responsible for fees and how to implement this trait. The method name says `pay` so the straightforward way would be to do transfer right away (but to whom?), and I don't see any hooks for actually releasing the `fee`.

modules/message-lane/src/lib.rs
```diff
@@ -90,19 +159,237 @@ decl_module! {
...
+			);
+
 			Self::deposit_event(Event::MessageAccepted(lane_id, nonce));
+
+			Ok(())
+		}
+
+		/// Receive messages proof from bridged chain.
+		#[weight = DELIVERY_BASE_WEIGHT + max_dispatch_weight]
+		pub fn receive_messages_proof(origin, proof: MessagesProofOf<T, I>, max_dispatch_weight: Weight) -> DispatchResult {
```
tomusdrw, line 256: I feel this should rather be called `receive_and_dispatch_source_messages(origin, proof, max_weight)`.

tomusdrw: Why do we let the relayer specify `max_dispatch_weight`? What prevents them from setting `0` to pay the lowest fee and then rely on `on_initialize` to actually do the heavy work?

EDIT: Ah okay, we still charge the relayer for all the messages, but still I wonder if we shouldn't restrict the `minimal_max_dispatch_weight`, to prevent excessive queueing.

modules/message-lane/src/lib.rs
```diff
@@ -90,19 +159,237 @@ decl_module! {
...
+
+			// estimate cumulative weight of messages dispatch
+			let mut dispatcher = T::MessageDispatch::with_allowed_weight(max_dispatch_weight);
+			let messages_dispatch_weight: Weight = messages.iter().map(|m| dispatcher.dispatch_weight(m)).sum();
+
+			// the submitter pays for dispatch in advance, so the whole `messages_dispatch_weight` should
+			// be covered. But it has already paid regular fee for at least `max_dispatch_weight`, so we
+			// only need to withdraw `messages_dispatch_weight - max_dispatch_weight`.
+			let messages_dispatch_extra_weight = messages_dispatch_weight.saturating_sub(max_dispatch_weight);
+			T::MessageDispatchPayment::pay_dispatch_fee(&relayer, messages_dispatch_extra_weight).map_err(|err| {
```
tomusdrw, line 278: IMHO this should use `Currency` trait directly, we shouldn't be abstracting the fees more than necessary.

modules/message-lane/src/lib.rs
```diff
@@ -90,19 +159,237 @@ decl_module! {
...
+				queued_messages,
+				max_dispatch_weight.saturating_sub(weight_left),
+				weight_left,
+			);
+
+			// ok - we have processed or queued all bundled messages. But what if relayer acts aggressivly
+			// (i.e. `max_dispatch_weight` is larger than `messages_dispatch_weight`), or if some messages
+			// have been queued? Then we may have some capacity left from `max_dispatch_weight`.
+			// => let's not waste it and try to process queued messages
+			let mut queued_dispatcher = T::QueuedMessageDispatch::with_allowed_weight(weight_left);
```
tomusdrw, line 321: That's cool, but it means that we actually pay twice for delivering/dispatching these messages. I'd say that here, the relayer who is willing to spend more `max_dispatch_weight` to process extra messages from the queue, should be rewarded for that, by getting back the fee (or some portion of it) payed by the less aggressive relayer, who queued these messages. wdyt?

Currently, the strategy for relayers is to send smallest possible `max_dispatch_weight`, cause they pay small amount upfront and they don't overpay in case the transaction fails (imagine there is a race between multiple relayers). Restricting the minimal value of `max_dispatch_weight` + giving them some incentive to actually process messages in the queue might help with overall bridge efficiency.

modules/message-lane/src/lib.rs
```diff
@@ -90,19 +159,237 @@ decl_module! {
...
+			Ok(())
+		}
+
+		/// Receive messages proof from bridged chain.
+		#[weight = DELIVERY_BASE_WEIGHT + max_dispatch_weight]
+		pub fn receive_messages_proof(origin, proof: MessagesProofOf<T, I>, max_dispatch_weight: Weight) -> DispatchResult {
+			let relayer = ensure_signed(origin)?;
+
+			// verify messages proof && convert proof into messages
+			let messages = T::SourceHeaderChain::verify_messages_proof(proof).map_err(|err| {
```
tomusdrw, line 260: I imagine this will fail if the same messages has already been delivered right? I think in that case, we should still consider using up the remaining `max_dispatch_weight`, which the relayer is willing to pay, to process messages from the queue. In case we reward them with the fee that was payed upfront by the relayer, who actually queued the messages we have a nice incentive to run redundant relayers in case of lane congestion (messages getting queued) and a mechanism that prevents relayers from abusing the queue (i.e. offloading work to `on_initialize`).

modules/message-lane/src/lib.rs
tomusdrw: I imagine we should also charge the relayers who queue messages a tiny bit more, to prevent them from just sending lowest possible `max_dispatch_weight` and abusing  dispatch during `on_initialize`.

modules/message-lane/src/lib.rs
```diff
@@ -90,19 +159,237 @@ decl_module! {
...
+			}
+
+			Ok(())
+		}
+
+		/// Receive messages receiving proof from bridged chain.
+		#[weight = 0] // TODO: update me (https://github.com/paritytech/parity-bridges-common/issues/78)
+		pub fn receive_message_receiving_proof(origin, proof: MessagesReceivingProofOf<T, I>) -> DispatchResult {
+			let _ = ensure_signed(origin)?;
+			let (lane_id, nonce) = T::TargetHeaderChain::verify_messages_receiving_proof(proof).map_err(|err| {
```
tomusdrw, line 339: I think we should extract the relayer address who delivered the messages, so that we can distribute fee (or some portion of it) to him. I imagine that `delivery` will have a fixed reward that the relayer get's, while the rest of the fee should be payed upon processing of the message.

To avoid excessive transfers I think that all the fees should be first transferred to some special "message-lane" account, and relayers are allowed to claim the total whenever they want. IMHO this deserves a separate PR.

modules/message-lane/src/lib.rs
```diff
@@ -90,19 +159,237 @@ decl_module! {
...
+					target: "runtime",
+					"Rejecting invalid proof of messages receiving: {:?}",
+					err,
+				);
+
+				Error::<T, I>::InvalidMessagesReceivingProof
+			})?;
+
+			let mut lane = outbound_lane::<T, I>(lane_id);
+			let received_range = lane.confirm_receival(nonce);
```
tomusdrw, line 350: I've missed this method name in the previous PR, but I think it should rather be called `confirm_delivery`. `Receival` sounds super weird, is that even a word?

modules/message-lane/src/lib.rs
```diff
@@ -90,19 +159,237 @@ decl_module! {
...
+			);
+
+			Ok(())
+		}
+
+		/// Receive messages processing proof from bridged chain.
+		#[weight = 0] // TODO: update me (https://github.com/paritytech/parity-bridges-common/issues/78)
+		pub fn receive_message_processing_proof(origin, proof: MessagesProcessingProofOf<T, I>) -> DispatchResult {
+			let _ = ensure_signed(origin)?;
+			let (lane_id, nonce) = T::TargetHeaderChain::verify_messages_processing_proof(proof).map_err(|err| {
```
tomusdrw, line 369: Same as with delivery confirmation, this should give us the address of relayer who caused the message to be processed, so that we can reward him.
Since it's a `target` chain, we probably either need a map of `target -> source` addresses, or just have the relayers include their "reward" address of source chain, when they deliver / process messages in the target chain.

modules/message-lane/src/lib.rs
```diff
@@ -112,62 +399,17 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
...
-	/// Trusted here means that the function itself doesn't check whether the bridged chain has
-	/// actually processed these messages.
-	pub fn confirm_processing(lane_id: &LaneId, latest_processed_nonce: MessageNonce) {
-		let mut lane = outbound_lane::<T, I>(*lane_id);
-		let processed_range = lane.confirm_processing(latest_processed_nonce);
-
-		if let Some(processed_range) = processed_range {
-			Self::deposit_event(Event::MessagesProcessed(*lane_id, processed_range.0, processed_range.1));
-		}
+	/// Returns empty-lane flag.
```
tomusdrw, line 407:
```suggestion
	/// Returns `true` if there is no more messages in the lane (i.e. the lane is now empty).
```

primitives/message-lane/src/source_chain.rs
```diff
@@ -0,0 +1,103 @@
...
+		proof: Self::MessagesProcessingProof,
+	) -> Result<(LaneId, MessageNonce), Self::Error>;
+}
+
+/// Lane message verifier.
+///
+/// Runtime developer may implement any additional validation logic over message-lane mechanism.
+/// E.g. if lanes should have some security (e.g. you can only accept Lane1 messages from
+/// Submitter1, Lane2 messages for those who has submitted first message to this lane, disable
+/// Lane3 until some block, ...), then it may be built using this verifier.
```
tomusdrw, line 67: I think we should mention that with current design it's important to make sure that the `fee` being passed is sufficient to deliver & dispatch the `payload`. I proposed in some other comment to actually return the minimal required `fee` instead of verifying it.

modules/message-lane/src/lib.rs
tomusdrw: Since the target chain congestion might be dynamic, I imagine we will require some extra way to increase the `tip` for a specific message, to make sure that relayers are either waiting for a less-congested time (stalling the delivery) or overpay for transaction (lose money). I can see that being a completely separate PR, where you just add an extra "unlocking lane reward". I.e. the senders that see their messages (their lane) being stuck due to insufficient fees can just top-up the entire lane.

primitives/message-lane/src/source_chain.rs
```diff
@@ -0,0 +1,103 @@
...
+/// So to be sure that any non-altruist relayer would agree to deliver message, submitter
+/// should set `delivery_and_dispatch_fee` to at least (equialent of): sum of fees from (2)
+/// to (4) above, plus some interest for the relayer.
+pub trait MessageDeliveryAndDispatchPayment<AccountId, Balance> {
+	/// Error type.
+	type Error: Debug + Into<&'static str>;
+
+	/// Withhold/write-off delivery_and_dispatch_fee from submitter account to
+	/// some relayers-fund account.
+	fn pay_delivery_and_dispatch_fee(submitter: &AccountId, fee: &Balance) -> Result<(), Self::Error>;
```
tomusdrw, line 102: As mentioned earlier I think we should keep `fees` less abstracted, and use `Currency` trait directly to lock + distribute rewards. Since this PR is big enough, I'm okay with splitting the fee modelling into a separate one (we can think a bit more on the dynamics there as well and document them in an issue)

primitives/message-lane/src/target_chain.rs
```diff
@@ -0,0 +1,101 @@
...
+
+/// Message dispatch payment. It is called as a part of message-delivery transaction. Delivery
+/// transaction submitter (message relay account) pays for the messages dispatch (probably
+/// in advance).
+pub trait MessageDispatchPayment<AccountId> {
+	/// Error type.
+	type Error: Debug + Into<&'static str>;
+
+	/// Write-off fee for dispatching messages of given cumulative weight.
+	fn pay_dispatch_fee(payer: &AccountId, weight: Weight) -> Result<(), Self::Error>;
```
tomusdrw, line 100: Same with the other trait, this abstraction is imho unnecessary and we should bundle `Currency` trait directly with the message lane.

tomusdrw (commented):
Looks super cool, I'm really exited to see this being close to completion. Have some grumbles about naming and the fee model though.

primitives/message-lane/src/lib.rs
```diff
@@ -23,53 +23,78 @@
...
+	pub payload: Payload,
+	/// Message delivery and dispatch fee, paid by the submitter.
+	pub fee: Fee,
+}
+
 /// Message as it is stored in the storage.
-#[derive(Encode, Decode, Clone)]
-pub struct Message<Payload> {
+#[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug)]
+pub struct Message<Payload, Fee> {
```
HCastano, line 62:
```suggestion
pub struct Message<P, F> {
```

primitives/message-lane/src/lib.rs
```diff
@@ -23,53 +23,78 @@
...
-	/// Message payload.
-	pub payload: Payload,
+	/// Message data.
+	pub data: MessageData<Payload, Fee>,
 }
 
 /// Message processing result.
-pub enum MessageResult<Payload> {
+#[derive(RuntimeDebug, PartialEq)]
+pub enum MessageResult<Payload, Fee> {
```
HCastano, line 71:
```suggestion
pub enum MessageResult<P, F> {
```

primitives/message-lane/src/lib.rs
```diff
@@ -23,53 +23,78 @@
...
 pub struct MessageKey {
 	/// ID of the message lane.
 	pub lane_id: LaneId,
 	/// Message nonce.
 	pub nonce: MessageNonce,
 }
 
+/// Message data as it is stored in the storage.
+#[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug)]
+pub struct MessageData<Payload, Fee> {
```
HCastano, line 53:
```suggestion
pub struct MessageData<P, F> {
```

primitives/message-lane/src/lib.rs
```diff
@@ -23,53 +23,78 @@
...
-	/// `MessageResult::Processed`), or it'll block the lane forever.
-	fn on_message_received(&mut self, message: Message<Payload>) -> MessageResult<Payload>;
+impl<Payload, Fee> MessageResult<Payload, Fee> {
+	/// Return number of processed messages (0 or 1).
+	pub fn messages_processed(&self) -> MessageNonce {
+		match *self {
+			MessageResult::Processed(_) => 1,
+			MessageResult::NotProcessed(_) => 0,
+		}
+	}
```
HCastano, line 85: Is this ever used?

Also, I don't quite get the logic here. Why do we only ever process one message?

primitives/message-lane/src/source_chain.rs
```diff
@@ -0,0 +1,103 @@
...
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Primitives of message lane module, that are used on the source chain.
+
+use crate::{LaneId, MessageNonce};
+
+use frame_support::Parameter;
+use sp_std::fmt::Debug;
+
+/// Target chain API. Used by source chain, to verify target chain proofs.
```
HCastano, line 24:
```suggestion
/// Target chain API. Used by source chain to verify target chain proofs.
```

primitives/message-lane/src/source_chain.rs
```diff
@@ -0,0 +1,103 @@
...
+
+	/// Proof that messages have been received by target chain.
+	type MessagesReceivingProof: Parameter;
+	/// Proof that messages have been processed by target chain.
+	type MessagesProcessingProof: Parameter;
+
+	/// Verify message payload before we accept it.
+	///
+	/// **CAUTION**: this is very important function. Incorrect implementation may lead
+	/// to stucked lanes and/or relayers loses.
```
HCastano, line 41:
```suggestion
	/// to stuck lanes and/or relayers loses.
```

primitives/message-lane/src/source_chain.rs
```diff
@@ -0,0 +1,103 @@
...
+
+use frame_support::Parameter;
+use sp_std::fmt::Debug;
+
+/// Target chain API. Used by source chain, to verify target chain proofs.
+///
+/// All implementations of this trait should only work with finalized data that
+/// can't change. Wrong implementation may lead to invalid lane states (i.e. lane
+/// that's stuck) and/or processing messages without paying fees.
+pub trait TargetHeaderChain<Payload> {
```
HCastano, line 29:
```suggestion
pub trait TargetHeaderChain<P> {
```

I'd avoid using generics with full names unless it was a particularly ambiguous situation like `pub trait Foo<Hash, Header>` where `H` wouldn't quite cut it.

primitives/message-lane/src/target_chain.rs
```diff
@@ -0,0 +1,101 @@
...
+	/// Creates instance of message dispatch, which can spent at most `weight` on
+	/// messages dispatch. Spending more than weight is not allowed.
+	fn with_allowed_weight(weight: Weight) -> Self;
+	/// Return weight that is left at this moment. Initially it is `weight` from
+	/// `with_allowed_weight` call, but it should be decreased after every successful
+	/// `dispatch` call.
+	fn weight_left(&self) -> Weight;
+	/// Estimate dispatch weight.
+	///
+	/// This function must: (1) be instant and (2) return correct, but upper bound
```
HCastano, line 56: What do you mean by it must be "instant"?

modules/message-lane/src/by_weight_dispatch.rs
```diff
@@ -0,0 +1,363 @@
...
+}
+
+/// Message dispatcher that only cares about weights. If message fits within
+/// allowed weight, it is dispatched immediately. Otherwise, message is queued
+/// using message-lane module storage and will be processed later.
+#[derive(RuntimeDebug)]
+pub struct ByWeightMessageDispatch<T, I, WrappedDispatch, Storage> {
+	storage: Storage,
+	instance: InstanceId,
+	weight: Weight,
```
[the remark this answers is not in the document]
HCastano: `weight_remaining` might be better, but no strong opinion here

modules/message-lane/src/lib.rs
```diff
@@ -90,19 +159,237 @@ decl_module! {
...
+					weight_spent_on_queued_dispatch,
+					queued_messages_dispatched,
+				);
+			}
+
+			weight_spent_on_queued_dispatch
+		}
+
 		/// Send message over lane.
 		#[weight = 0] // TODO: update me (https://github.com/paritytech/parity-bridges-common/issues/78)
```
HCastano, line 179: `DELIVERY_BASE_WEIGHT` maybe?

modules/message-lane/src/lib.rs
```diff
@@ -90,19 +159,237 @@ decl_module! {
...
+				"Received messages: total={}, invalid={}, processed={}, queued={}. Weight spent: {}, left: {}",
+				total_messages,
+				invalid_messages,
+				processed_messages,
+				queued_messages,
+				max_dispatch_weight.saturating_sub(weight_left),
+				weight_left,
+			);
+
+			// ok - we have processed or queued all bundled messages. But what if relayer acts aggressivly
```
HCastano, line 317:
```suggestion
			// ok - we have processed or queued all bundled messages. But what if relayer acts aggressively
```

modules/message-lane/src/lib.rs
HCastano: It is a word, lol. But I agree here, delivery would be clearer

modules/message-lane/src/lib.rs
HCastano: I like the idea of the relayers being able to set a coinbase address to send their rewards to

HCastano (commented):
I haven't managed to look through everything just yet. Will do another round tomorrow

modules/message-lane/src/lib.rs
svyatonik: > it's supposed to verify that so it knows the minimal value anyway

Hmm...It **may** verify that. But I don't think that all implementations must do that. I suggested one possible way on how to estimate `source-fees -> target-weight` conversions in the description. But it isn't required and needs to be confirmed yet. 

Imo we can't rely on the fact that **any** source chain knows **current state** of target chain economics (i.e. conversion_rate(source -> target), `TargetWeightToFee(TargetWeight)`). I'm going to hardcode this knowledge for our test nets (simply 1:1), but for other chains (Kusama, Polkadot) we can't rely on that.

My vision is: (1) sender selects `delivery_and_dispatch_fee` on its own, probably we could add some RPCs to relay node (that will has access to both chains + probably market information); (2) relay just ignores messages/lanes with too low `delivery_and_dispatch_fee`.

> I imagine we will require some extra way to increase the tip

I thought we have agreed that this could be solved just by pushing additional message to the lane with some dummy call (probably just empty vec?)? The fee should be enough to cover both this and previous message(s) costs.

modules/message-lane/src/lib.rs
svyatonik: > Right now it's not clear to me which layer is supposed to be responsible for fees and how to implement this trait. The method name says pay so the straightforward way would be to do transfer right away (but to whom?), and I don't see any hooks for actually releasing the fee.

I've left this decision for future PRs - please see section 2 of "description/Some ideas for future PRs".

Re excess abstraction - I don't think that we need to directly use `Currency` (or `LockableCurrency`) anywhere. Why do you think that lock+transfer is the only possible implementation? I see multiple uses for custom logic here. E.g. we may reserve some free slots for relayer transactions instead of paying him directly. Or it may be some increased possibility in next validators elections...Changing pallet API has larger (development) costs than changing implementations, so imo we should leave most generic callbacks where possible.

modules/message-lane/src/lib.rs
svyatonik: Please see description - having some lower-bounds on `max_dispatch_weight` would block other transactions from being included into block. Having upper bounds would block some alternatively-incentivised relayers from submitting "process-my-messages-asap" transactions (i.e. transactions with empty messages vec and non-zero `max_dispatch_weight`).

I agree that excessive queueing could be a problem here. I have no other solutions right now than to completely remove queued messages. Please respond, if you think we should do that.

modules/message-lane/src/lib.rs
svyatonik: > IMHO this should use Currency trait directly, we shouldn't be abstracting the fees more than necessary.

Responded in other comment. Here, imo, relying on `Currency` is even worse, because fee != currency. We may e.g. transfer fees to treasure, burn them, tip validators for processing queued messages, or even combine these actions.

> I imagine we should also charge the relayers who queue messages a tiny bit more, to prevent them from just sending lowest possible max_dispatch_weight and abusing dispatch during on_initialize.

In this case relayers would never deliver messages if there's anything queued in the lane. So this would be semantically equal to having no queued messages. As I said above, we may remove supporting messages queued for now => this should solve the issue. I'm just waiting for your opinion, guys.

modules/message-lane/src/lib.rs
svyatonik: > That's cool, but it means that we actually pay twice for delivering/dispatching these messages

Why? Who's paying twice? Relayer is only paying base delivery tx fee + dispatch fee.

>  I'd say that here, the relayer who is willing to spend more max_dispatch_weight to process extra messages from the queue, should be rewarded for that, by getting back the fee (or some portion of it) payed by the less aggressive relayer, who queued these messages. wdyt?

If there's relayer that is submitting transactions with `max_dispatch_weight` larger than `weights-sum(delivered-messages)`, then either (1) it is alternatively-incentivised to dispatch messages, not to deliver messages - that should be controlled on app-level imo (2) it is trying to fill block with dummy transactions (possibly malicious or dumb) - I don't think we should reward this behavior.

Remember that reserving more weight than necessary decreases transactions throughput. It only works if relayers are paying for such transactions. Rewarding this behavior may result in blocks spammed with dummy transactions.

modules/message-lane/src/lib.rs
svyatonik: > I imagine this will fail if the same messages has already been delivered right?

You're right. I've actually missed the fact that we may deliver duplicate messages => relayer should only pay dispatch fee for actually delivered messages. Will respond to that comment once this is solved.

> I think in that case, we should still consider using up the remaining max_dispatch_weight

It is already done - please see how `ByWeightMessageDispatch` works with passed `max_dispatch_weight`.

>  In case we reward them with the fee that was payed upfront by the relayer, who actually queued the messages we have a nice incentive to run redundant relayers in case of lane congestion (messages getting queued) and a mechanism that prevents relayers from abusing the queue (i.e. offloading work to on_initialize).

As I said in other my comments, we should not reward redundant relayers, or txpool of target chain will be flooded with `deliver_transaction(empty-messages, max-transaction-weight)`.

modules/message-lane/src/lib.rs
svyatonik: Please see (II) of description/Some ideas for future PRs. This isn't a part of this PR. This PR is about weights and fees for message deliver/dispatch. Rewards will be introduced in follow-up PRs.

modules/message-lane/src/lib.rs
svyatonik: > Same as with delivery confirmation, this should give us the address of relayer who caused the message to be processed, so that we can reward him.

Responded in many other comments: (1) we should not reward redundant dispatch-transactions (2) rewards is not a part of this PR.

> Since it's a target chain, we probably either need a map of target -> source addresses, or just have the relayers include their "reward" address of source chain, when they deliver / process messages in the target chain.

Please see (II) of description/Some ideas for future PRs.

primitives/message-lane/src/source_chain.rs
svyatonik: Tips on selecting fee are mentioned in `MessageDeliveryAndDispatchPayment` documentation. As I said in other comment, imo we should not force `LaneMessageVerifier` to deal with `delivery_and_dispatch_fee` - it is like buyer-seller relations between sender and relayer. `LaneMessageVerifier` may just enforce some regulation over these relations - i.e. we know that minimal fee must be at least `N`, or no relayers will ever consider delivering this message. Knowledge of this `N` is out-of-scope of this PR.

primitives/message-lane/src/source_chain.rs
svyatonik: I don't think so. We may provide some default implementations that are based on `Currency` trait, but we should not force everyone to use `Currency` trait.

primitives/message-lane/src/target_chain.rs
svyatonik: I don't think so. We may provide some default implementations that are based on Currency trait, but we should not force everyone to use Currency trait.

primitives/message-lane/src/lib.rs
svyatonik: > Is this ever used?

Not anymore - I'm removing it.

> Also, I don't quite get the logic here. Why do we only ever process one message?

Because `MessageResult` is result of processing single message.

modules/message-lane/src/lib.rs
svyatonik: Not sure what do you mean. `DELIVERY_BASE_WEIGHT` is weight of completely another transaction on different chain

svyatonik:
Looking at the reviews, I think that maybe we should remove support for queued messages now. I thought initially that there may be some additional logic on messages dispatch. E.g. if we have two lanes, one for `Headers` sync and another one for `CurrencyExchange` transactions sync. Obviously, to process `CurrencyExchange` we need `Headers` to be synced first (omitting that incorrect txs may stuck this lane). And dispatch may look at the `CurrencyExchange` and return `NotProcessed` until there's required header for that tx. But it causes a lot of other problems. So we may sacrifice this feature to avoid queued messages at all. 

So basically - delivery-transactions is always dispatching its messages. Weight of delivery tx is always `DELIVERY_BASE_WEIGHT + sum(dispatch-message-weight)` (i.e. relayer no longer controls its weight). We don't need to dispatch queued messages from `on_intialze`. Wdyt?

modules/message-lane/src/lib.rs
tomusdrw: > Hmm...It may verify that.

AFAICT, if it doesn't verify then it's broken. I agree that it's cool to have it extensible, but at the same time too much extensibility will just let the users (or even the future versions of us) shoot themselves in the foot. Added a more detailed comment on this below.

> Imo we can't rely on the fact that any source chain knows current state of target chain economics (i.e. conversion_rate(source -> target), TargetWeightToFee(TargetWeight)). 

Hmm, that's a fair point, but IMHO it needs at least some minimal knowledge, that might be slightly out of date. Obviously we will never be able to synchronize the knowledge fast enough, to respond to the changing market, but imo it's viable to assume we know some long term trend at least.

That said, the fees would be worked on in a separate PR anyway, when we do proper integration with the runtime and I'm happy to pause the discussion here and resume in this other PR. Referencing https://github.com/paritytech/parity-bridges-common/issues/316 here.

modules/message-lane/src/lib.rs
tomusdrw: I feel the fees problem (and also similar `delivery_and_dispatch_fee` discussion earlier) is a tradeoff between:
1. Extensibility of the pallets
2. Auditability / Ease of use.

Basically if we abstract more stuff out of the pallet, we allow more sophisticated schemes as you've mentioned, but it's becoming harder and harder to actually understand the whole thing.
I feel that since fees (and therefore) incentives are so complicated already I would prefer this to be an integral part of the delivery pallet - the pallet without proper incentive scheme is basically useless and it's might be even dangerous to include it in a production runtime.
Auditing the pallet seems also much easier than auditing the pallet + the runtime glue code.

The ease of use argument applies to both:
-  runtime integration (more custom hooks to implement, vs just plugging an existing `Currency` pallet, most likely just `pallet_balances`).
- usage of the external API - i.e. requirement of calling an extra runtime API before submitting a message (i.e. getting `delivery_and_dispatch_fee` is really much worse than simply saying how much you're willing to overpay to get your message faster (including a `tip`)) - this is obviously a bit more tricky, since probably selecting the right `tip` requires knowing the target chain dynamics, but still it seems a bit more ergonomic to me.

primitives/message-lane/src/source_chain.rs
tomusdrw: I disagree. Reading a bunch of one-letter generics is a nightmare, it's like using one-letter variables. I'd flip the rule and say we should only use one-letter generic names if it's clear from the context what they mean. In my past work we used to call generics with `T` prefix, to distinguish from regular types, so it would be `TPayload, THash, TFee`, etc. I'd leave the decision to go with one-letter or full-length up to Slava though.

primitives/message-lane/src/source_chain.rs
tomusdrw: > - it is like buyer-seller relations between sender and relayer.

Well, I perceive this a bit differently. Choosing incorrect `fee` may result in the lane being stuck forever, so it's the lane's responsibility to make sure it attempts to ensure it's existence and usefullness. That said, the discussion should probably be part of a different PR.

modules/message-lane/src/lib.rs
tomusdrw: > Please see description - having some lower-bounds on max_dispatch_weight would block other transactions from being included into block. 

But the point is that "other transactions" and messages should really have the same "priority". I.e. Currently we treat bridge messages in a special way and we allow them to be dispatched with highest possible priority, during `on_initialize` call. IMHO this should be fully tuneable and in most of the blockchains should only be a final necessity to do so. For instance it probably only makes sense for system-dispatches, i.e. where two blockchains communicate with each other, but definitely not for user messages. If we allow arbitrary message to be dispatched, it might be better/cheaper/faster for a `target` blockchain user to send a message over the bridge instead of simply sending a transaction to the `target` chain, cause he/she knows it is more priviledged regarding timely dispatch.

We should definitely make sure that delivering messages doesn't require the entire block weight, but having some sane minimum, and having relayers compete for these resources with other `target` blockchain users should simply be part of the design. The incentive layer should ensure that it's profitable for them.

modules/message-lane/src/lib.rs
tomusdrw: > Why? Who's paying twice?

The first relayer to deliver these queued messages is already charged with `fee` for this delivery&dispatch. Now here the relayer is paying the `fee` again, since it declared high `max_dispatch_weight` to actually process these queued messages, so it's charged a `fee` for that declared weight. Unless I'm missing something.

> it is trying to fill block with dummy transactions (possibly malicious or dumb) - I don't think we should reward this behavior.

Processing messages in the queue is not really malicious. If they pay higher fee than other `target` blockchain users, it's totally legit for them to spend it even just idle-looping, if they actually make progress on the bridge lane, then it's better for us.

Also notice, that I'm not mentioning that we should mint coins out of thing air to reward them, I suggested that the relayer, how decided to queue the messages instead of processing them (because they thought it would be easier for them to get included in a block with lower-weight transaction) pays that extra price. I.e. it then creates a market for relayers between: delivering & queuing (low weight, higher fee, lower reward) or delivering & processing (high weight, lower fee, higher reward), so the relayers are now responsible for finding an optimum balance between this two sides and given current blockchain dynamics (other transactions waiting in the `target` blockchain transaction pool).

modules/message-lane/src/lib.rs
tomusdrw: > It is already done - please see how `ByWeightMessageDispatch` works with passed `max_dispatch_weight`.

No, it's not. The `Error` here will early-exit and cause the entire top-level dispatch to fail. I meant to actually take messages from the queue & dispatch them in case the error happens to utilize the `weight` that would otherwise be wasted.

tomusdrw:
> if we have two lanes, one for `Headers` sync

I'm a bit confused, why would `Headers` sync be part of a message lane? I thought that `Headers` sync is lower-level, and processing message from lanes actaully require the headers to be synced already. Or did you mean like a Light Client to some chain `C` (on chain `A`) that actually has all the headers relayed over some other chain `B`? I don't think this is in scope for now.

> So basically - delivery-transactions is always dispatching its messages. Weight of delivery tx is always `DELIVERY_BASE_WEIGHT + sum(dispatch-message-weight)` (i.e. relayer no longer controls its weight). We don't need to dispatch queued messages from `on_intialze`. Wdyt?

If you think we can add the queueing as a feature later on, let's do that. It seems though that the split between delivery and processing becomes redundant in such case as well.
I'm also okay with keeping queueing as a feature, but making it configurable (initially set to "false") and we can revisit that in a subsequent PR where we work on the rewards / incentives. The queuing seems super useful and I'm sure is the right way forward, but in the context of this PR it indeed might cause more bad than good, cause it takes the attention out of the important stuff.
I leave it up to you to decide which path is easier for you to follow.

svyatonik:
> I'm a bit confused, why would Headers sync be part of a message lane? I thought that Headers sync is lower-level, and processing message from lanes actaully require the headers to be synced already. Or did you mean like a Light Client to some chain C (on chain A) that actually has all the headers relayed over some other chain B? I don't think this is in scope for now.

Please treat this as an example of `MessageDispatch` implementation. And yes - what I mean is that A->B->C app. Another example may be e.g. some lottery that is only processing incoming "stakes" every 1024 blocks, ...We have a scope of delivering and dispatching `Call`s, not these specific app, ofc. 

> If you think we can add the queueing as a feature later on, let's do that. It seems though that the split between delivery and processing becomes redundant in such case as well, which may cause even more work. I'm also okay with keeping queueing as a feature, but making it configurable (initially set to "false") and we can revisit that in a subsequent PR where we work on the rewards / incentives. The queuing seems super useful and I'm sure is the right way forward, but in the context of this PR it indeed might cause more bad than good, cause it takes the attention out of the important stuff. I leave it up to you to decide which path is easier for you to follow.

I think queueing support has huge impact on pallets API and all that delivery/dispatch weights/fees scheme. So it should be decided now and forever imo. If you leave this decision to me, then I suggest to remove that functionality at all. I'll open a separate PR for removing then.

Final note for future: queueing may still be implemented by custom `MessageDispatch` implementation. But removing this concept from message-lane module seems to solve multiple problems we're discussing here.

primitives/message-lane/src/source_chain.rs
HCastano: So in this case I think it's fine to use a one letter generic (as is Rust convention) since it's only used as a type for an argument called `payload`. When you see `fn foo(payload: P)` it's clear what the generic is for.

I agree that this isn't a blanket rule, and we should be thoughtful when naming our generics. But we should start by following Rust conventions.

modules/message-lane/src/lib.rs
HCastano: It looked like they could be related. The comment has to do with a delivery transaction, which to me is something that `send_message()` would send.

modules/message-lane/src/lib.rs
svyatonik: The proof is just storage proof && shouldn't fail if some of messages were already delivered. But you're right - I should have utilize `max_dispatch_weigth` even if proof was wrong.

This is irrelevant now.

modules/message-lane/src/by_weight_dispatch.rs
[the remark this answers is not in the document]
svyatonik: Irrelevant

modules/message-lane/src/by_weight_dispatch.rs
```diff
@@ -0,0 +1,363 @@
...
+	fn max_dispatch_weight(&self) -> Weight {
+		<T as frame_system::Trait>::MaximumExtrinsicWeight::get()
+	}
+
+	fn enqueue_lane(&mut self, lane: LaneId) {
+		QueuedInboundLanes::<I>::append(lane);
+	}
+
+	fn take_next_queued_lane(&mut self) -> Option<LaneId> {
+		QueuedInboundLanes::<I>::mutate(|lanes| if lanes.is_empty() { None } else { Some(lanes.remove(0)) })
```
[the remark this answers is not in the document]
svyatonik: Irrelevant

modules/message-lane/src/by_weight_dispatch.rs
```diff
@@ -0,0 +1,363 @@
...
+
+	/// If there's enough weight left, dispatch the message. Otherwise, enqueue it
+	/// for later processal.
+	pub fn dispatch(
+		&mut self,
+		message: Message<T::Payload, T::MessageFee>,
+	) -> MessageResult<T::Payload, T::MessageFee> {
+		let dispatch_weight = self.dispatch_weight(&message);
+		if dispatch_weight > self.storage.max_dispatch_weight() {
+			return MessageResult::Processed(0);
```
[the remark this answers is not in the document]
svyatonik: Irrelevant

modules/message-lane/src/lib.rs
```diff
@@ -90,19 +159,237 @@ decl_module! {
 		/// Deposit one of this module's events by using the default implementation.
 		fn deposit_event() = default;
 
+		/// Block initialization.
+		fn on_initialize(now: T::BlockNumber) -> Weight {
+			let mut dispatcher = T::QueuedMessageDispatch::with_any_weight();
```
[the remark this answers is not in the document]
svyatonik: Irrelevant

modules/message-lane/src/lib.rs
svyatonik: As I've been writing in all other comments, I disagree with your points on hardcoding/decoupling logic. But I see no benefits in continuing this discussion, because it isn't progressing at all && we're just wasting time here. I'll turn on agnostic mode for this kind of comments here && apply your suggestions.

modules/message-lane/src/lib.rs
svyatonik: Irrelevant

modules/message-lane/src/lib.rs
svyatonik: I still don't understand you - why `send_message()` sends delivery transaction?

modules/message-lane/src/lib.rs
tomusdrw: Let's leave it as-is now, I need to sleep over this again, cause I think I'm starting to see your concerns, and re-review after we removed the queue.

modules/call-dispatch/src/lib.rs
```diff
@@ -86,7 +82,11 @@ decl_module! {
 impl<T: Trait> MessageDispatch<T::MessageId> for Module<T> {
 	type Message = (SpecVersion, Weight, <T as Trait>::Call);
 
-	fn dispatch(bridge: InstanceId, id: T::MessageId, message: Self::Message) -> Weight {
+	fn dispatch_weight(message: &Self::Message) -> Weight {
+		message.1
+	}
+
+	fn dispatch(bridge: InstanceId, id: T::MessageId, message: Self::Message) {
```
tomusdrw, line 89: Why don't we return `Weight` any more? I think we should still report back that lower amount of weight was used.

modules/message-lane/src/lib.rs
tomusdrw: Fine to leave it as-is, we can revisit later.

modules/message-lane/src/lib.rs
```diff
@@ -96,59 +148,163 @@ decl_module! {
...
+					err,
+				);
+
+				Error::<T, I>::InvalidMessagesProof
+			})?;
+
+			// verify that relayer is paying actual dispatch weight
+			//
+			// strict equality is used here, because we do not want relayer to declare more weight than
+			// it is necessary => blocking other transactions from being included into this block
```
tomusdrw, line 244: I don't think the strict equality is needed here. One can simply send `frame_system::fill_block` trasactions with arbitrary weight. They would obviously fail quickly, since the sender is not `root`, but still you can consume any block resources.

That said, I don't see any reasons to actually relax this requirement so it's good as is.

modules/message-lane/src/lib.rs
```diff
@@ -96,59 +148,163 @@ decl_module! {
...
+
+				return Err(Error::<T, I>::InvalidMessagesDispatchWeight.into());
+			}
+
+			// dispatch messages
+			let total_messages = messages.len();
+			let mut valid_messages = 0;
+			for message in messages {
+				let mut lane = inbound_lane::<T, I>(message.key.lane_id);
+				if lane.receive_message::<T::MessageDispatch>(message.key.nonce, message.data) {
```
tomusdrw, line 265: We should be tracking actual weight usage here to lower the final cost of the transaction. That on it's own gives the relayers a small "tip" for their service.

modules/message-lane/src/lib.rs
```diff
@@ -96,59 +148,163 @@ decl_module! {
...
+					target: "runtime",
+					"Rejecting invalid messages delivery proof: {:?}",
+					err,
+				);
+
+				Error::<T, I>::InvalidMessagesDeliveryProof
+			})?;
+
+			let mut lane = outbound_lane::<T, I>(lane_id);
+			let received_range = lane.confirm_delivery(nonce);
```
tomusdrw, line 295: We are going to need a hook here to be able to track relayers who submitted the messages for the fee payout purposes. Can be addressed in #316 
(Note we don't care who submits `delivery_proof`, only who actually delivered&dispatched messages on the target chain)

tomusdrw (approved):
lgtm. Let's get it merged asap.

modules/call-dispatch/src/lib.rs
svyatonik: We do not have anything to spent remaining weight now. Or do you meant we should refund relayer in this case?

tomusdrw: Yeah, I meant refunding.

modules/message-lane/src/lib.rs
svyatonik: Updated

modules/call-dispatch/src/lib.rs
svyatonik: Nvm, I've found answer in another comment

modules/message-lane/src/lib.rs
svyatonik: Let's fill a separate issue for that - I'm thinking if it can be exploited like that:
1) relayer itself submits message with weight = `1_000_000` that he knows will actually take only `1_000` weight units on the target chain;
2) he's the only relayer && he delivers both message and confirmation;
3) he gets that refund on the target chain (out of nowhere);
4) he gets the `delivery_and_dispatch_fee` on the source chain.

Probably I'm missing something obvious, but it needs another round of thinking anyway imo.

svyatonik: #377

primitives/message-lane/src/target_chain.rs
svyatonik: It should have near-zero overhead. So either take weight directly from message or make some quick lookup.

modules/message-lane/src/lib.rs
svyatonik: Irrelevant

modules/message-lane/src/lib.rs
HCastano: I didn't fully understand the `send` mechanics. Tomek cleared things up for me :)

HCastano (approved):

PR #340, opened 2020-09-08
Add Front-end docker

Tbaut:
TODO:
- [x] This needs first https://github.com/paritytech/bridge-ui/pull/23 and https://github.com/paritytech/bridge-ui/pull/22
- [x] And then we need to change/tweak the ENV/ARG to make sure the UI connects to the local docker nodes.


deployments/rialto/Front-end.Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM node:12 as build-deps
+
+ENV SUBSTRATE_PROVIDER wss://wss.rialto.brucke.link
+ENV ETHEREUM_PROVIDER http://rpc.rialto.brucke.link:8545
+ENV EXPECTED_ETHEREUM_NETWORK_ID 105
```
tomusdrw, line 5:
```suggestion
```

deployments/rialto/Front-end.Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+# install tools and dependencies
+RUN set -eux; \
+	apt-get install -y git
+
+# clone UI repo
+RUN cd /usr/src/ && git clone https://github.com/paritytech/bridge-ui.git
+WORKDIR /usr/src/bridge-ui
+RUN yarn
+RUN yarn build:docker
```
tomusdrw, line 15:
```suggestion
ENV SUBSTRATE_PROVIDER wss://wss.rialto.brucke.link
ENV ETHEREUM_PROVIDER http://rpc.rialto.brucke.link:8545
ENV EXPECTED_ETHEREUM_NETWORK_ID 105
RUN yarn build:docker
```

This will allow the cache to be re-used for previous steps even if `ENV` changes.

tomusdrw: Does the `ENV` variables always override `.env` files?

deployments/rialto/docker-compose.yml
```diff
@@ -215,3 +215,9 @@ services:
       - "4567:4567"
     depends_on:
       - grafana-dashboard
+  front-end:
+    build:
+      context: .
+      dockerfile: ./Front-end.Dockerfile
+    ports:
+      - "80:80"
```
tomusdrw, line 223:
```suggestion
      - "8080:80"
```
Otherwise we will need `sudo` to run it.

deployments/rialto/docker-compose.yml
```diff
@@ -215,3 +215,9 @@ services:
       - "4567:4567"
     depends_on:
       - grafana-dashboard
+  front-end:
+    build:
+      context: .
+      dockerfile: ./Front-end.Dockerfile
```
tomusdrw, line 221: I think we should add `environment` here and by default set the values to `localhost`:
```
  environment:
      SUBSTRATE_PROVIDER:  ${UI_SUBSTRATE_PROVIDER:-ws://localhost:9933}
      ETHEREUM_PROVIDER: ${UI_ETHEREUM_PROVIDER:-http://localhost:8545}	      
      EXPECTED_ETHEREUM_NETWORK_ID: 105
```

tomusdrw (commented):
nice!

deployments/rialto/Front-end.Dockerfile
Tbaut: My goto document about those things is https://vsupalov.com/docker-arg-env-variable-guide/#setting-env-values
It says that ARG receive the `.env` values, and that you can use that to set the ENV. But setting the ENV directly should be fine and simply not take care about any `.env`. I guess this can be tested quickly once the UI PR are in.

deployments/rialto/docker-compose.yml
Tbaut: This is beyond my knowledge :smile: 
From where would `${UI_SUBSTRATE_PROVIDER:-ws://localhost:9933}` get the `UI_SUBSTRATE_PROVIDER` variable? is the `-ws://localhost:9933` a default then?

deployments/rialto/Front-end.Dockerfile
tomusdrw: I rather meant the UI build, not the `Docker`. To pass `ENV` to docker file build, you can put an `.env` file next to it, or you can just pass them via CLI indeed. But that's a docker thing. Then when we build UI we take the ENV variables that are set and we apply them to the code. We read another `.env` file (the one that is checked out from the repo) and my question was if running:
```
$ SUBSTRATE_PROVIDER=http://localhost yarn build:docker
```
causes the frontend `.env` file variable to be overwritten by the one provided in the CLI.

deployments/rialto/docker-compose.yml
tomusdrw: So the `docker-compose` file can pass ENV variables, just like you would do (via CLI flags) when building `docker` image from CLI. The `environment` section will pass these ENV variables to the build, so they can be read by the `Dockerfile` that you created.

Secondly `docker-compose` supports `.env` files (again :exploding_head:), so when you run `docker-compose up` it's going to read `.env` file and substitute `${UI_SUBSTRATE_PROVIDER}` with a value found in environment. If that's not set we will fall back to default `ws://localhost:9933` value.

I request that, because the way we configure the rest of the whole environment (all services in `docker-compose`) is done via compose-specific `.env` file, so I expect the UI should be the same. By default we want it to just connect to `localhost` so that when you run `$ docker-compose up` the UI at `localhost:8080` would automatically connect to nodes that has been started at `:8545` and `:9933`.

Hope this `ENV` inception makes any sense :)

Tbaut: Wow, thanks a lot for the explanations. every time I touch such things it's :exploding_head:  for me, I get it.. and forget it a couple weeks later :roll_eyes: 
So in the end that should do the trick right?:
```suggestion
      dockerfile: ./Front-end.Dockerfile
      environment:
        SUBSTRATE_PROVIDER:  ${UI_SUBSTRATE_PROVIDER:-ws://localhost:9933}
        ETHEREUM_PROVIDER: ${UI_ETHEREUM_PROVIDER:-http://localhost:8545}	      
        EXPECTED_ETHEREUM_NETWORK_ID: 105
```

deployments/rialto/Front-end.Dockerfile
Tbaut: yes, the environment variable is overriding the `.env` e.g `SUBSTRATE_PROVIDER=blabla yarn build` resulted in a bundled js file containing "blabla".

deployments/rialto/docker-compose.yml
tomusdrw: Yup!

HCastano:
This would be a cool addition. One thing I'd like to change though is that in the main `docker-compose` we pull an image from Docker Hub instead of building it ourselves.  If we go with that approach then these Dockerfiles should be part of the `bridge-ui` repo, and we can add a GH Action to build and publish the images either nightly or on tags.

cc @General-Beck for the image publishing

deployments/rialto/docker-compose.yml
```diff
@@ -215,3 +215,13 @@ services:
...
+      environment:
+        SUBSTRATE_PROVIDER:  ${UI_SUBSTRATE_PROVIDER:-ws://localhost:9933}
```
tomusdrw, line 223: Can you also please add these updated env variables to the docs here: https://github.com/paritytech/parity-bridges-common/blob/27f26a677a7e50b5030fbd7ada8be650ba22d69f/deployments/README.md#environment-variables

Possibly with a comment about what they do.

Tbaut: Sure, just added that.

tomusdrw (approved):

tomusdrw:
> This would be a cool addition. One thing I'd like to change though is that in the main `docker-compose` we pull an image from Docker Hub instead of building it ourselves. If we go with that approach then these Dockerfiles should be part of the `bridge-ui` repo, and we can add a GH Action to build and publish the images either nightly or on tags.
> 
> cc @General-Beck for the image publishing

We can actually build using the existing `Dockerfile`, which pulls the repo inside the container. Then our repo would just maintain the latest commit hash that should be built and published to docker hub. @HCastano are you okay that being done as a follow up PR though?

Tbaut:
Don't merge yet please, it's not building correctly. I'll fix that now

Tbaut:
The whole `docker-compose up` works well now, but the UI does not connect to the local nodes. So we're doing something wrong with these ENV etc :(
Quick way to see, the console of the UI is showing what it connects to.

svyatonik:
> The whole docker-compose up works well now, but the UI does not connect to the local nodes

Aren't you asking UI to connect to itself here:
```docker
    environment:
      SUBSTRATE_PROVIDER: ${UI_SUBSTRATE_PROVIDER:-ws://localhost:9933}
      ETHEREUM_PROVIDER: ${UI_ETHEREUM_PROVIDER:-http://localhost:8545}
      EXPECTED_ETHEREUM_NETWORK_ID: 105
```

I mean that localhost in this case is the container where `Front-end.Dockerfile` is deployed. Obviously, noone is listening on ws:9933 and http:8545 there, because they're on another containers. And by exposing Sub/Eth ports in "ports" section like here:
```
  poa-node-arthur: &poa-node
...
    ports:
      - "8545:8545"
```
 we're making them accessible as http://localhost:8545 when referencing **from the host machine**. So imo the fix should be like that:
```docker
    environment:
      SUBSTRATE_PROVIDER: ${UI_SUBSTRATE_PROVIDER:-ws://bridge-node-alice:9933}
      ETHEREUM_PROVIDER: ${UI_ETHEREUM_PROVIDER:-http://poa-node-arthur:8545}
      EXPECTED_ETHEREUM_NETWORK_ID: 105
```

tomusdrw:
The connection happens in the browser (on the host) and not really inside the Docker container.

> Quick way to see, the console of the UI is showing what it connects to.

What does it connect to then? Maybe lets change the defaults to identify the layer that is broken :)

svyatonik:
Gotcha :)

Tbaut:
I've been fighting for an hour or so.. I feel I've tried everything.
> What does it connect to then?

It's connected to the default from the app. 
```js
Connecting to Substrate node: wss://wss.rialto.brucke.link
Connecting to Ethereum node: http://rpc.rialto.brucke.link:8545
```

From my understanding, it means that it doesn't get the environment variables correctly when it does the `yarn build` [here](https://github.com/paritytech/parity-bridges-common/pull/340/files#diff-c5ef811429ac485461c0cfef4236b5a0R15)

The environment variables should be set by the `environment` in `docker-compose`, but when I `RUN echo $SUBSTRATE_PROVIDER` in the dockerfile, it's empty.

deployments/rialto/Front-end.Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+# install tools and dependencies
+RUN set -eux; \
+	apt-get install -y git
+
+# clone UI repo
+RUN cd /usr/src/ && git clone https://github.com/paritytech/bridge-ui.git
+WORKDIR /usr/src/bridge-ui
+RUN yarn
+ENV SUBSTRATE_PROVIDER wss://wss.rialto.brucke.link
```
Tbaut, line 12: Also in my tests I've removed those, the default are already set in the UI and in the docker-compose.

tomusdrw: Then it explains why its broken. in Dockerfile you have to be explicit about ENV.

Tbaut: What do you mean with explicit, when I tried `ENV SUBSTRATE_PROVIDER` without default, hoping it would pick up what was passed by environment, docker complained "ENV must have two arguments".

Tbaut:
What I just pushed, using ARG to pass the values to ENV works as expected. I'll integrate slava suggestions to get the right addresses.

deployments/rialto/Front-end.Dockerfile
tomusdrw: You must always provide a default, cause it might not be set.

Tbaut:
it's now working as expected, 9944 wasn't exposed on the substrate node. I'll see if I can use ENV without ARG.

deployments/rialto/Front-end.Dockerfile
Tbaut: I was partly wrong in my previous [comment](https://github.com/paritytech/parity-bridges-common/pull/340#issuecomment-689623338). The values were either the UI's default, or the Dockerfile's default ENV.

If I keep those ENV here, they replace anything, and don't pick-up what was set in the `docker-compose`.
```
ENV SUBSTRATE_PROVIDER aaaaaws://bridge-node-alice:9944
ENV ETHEREUM_PROVIDER aaaaahttp://poa-node-arthur:8545
```

results in the UI getting those "aaa*" values.

tomusdrw:
Ah, I think I was wrong to suggest using `environment` in `docker-compose.yml`. This will only set the environment variables for the `run` time of the docker container, not the build time. Using `build-args` is the way to go then.

deployments/rialto/Bridge.Dockerfile
```diff
@@ -81,5 +81,7 @@ ENV HEALTH=$HEALTH
 HEALTHCHECK --interval=2m --timeout=10s \
   CMD curl -f $HEALTH || exit 1
 
+EXPOSE 9944 9933
```
HCastano, line 84: FYI, this might not do what you think it does. The `EXPOSE` command is [more for documentation](https://docs.docker.com/engine/reference/builder/#expose), and doesn't actually expose the ports for you.

deployments/README.md
```diff
@@ -107,7 +107,7 @@ dashboard can be accessed at `http://localhost:9090`. The Grafana dashboard can
 
 Here is an example `.env` file which is used for production deployments and network updates. For
 security reasons it is not kept as part of version control. When deploying the network this
-file should be correctly populated and kept in the [`rialto`](./rialto) folder.
+file should be correctly populated and kept in the [`rialto`](./rialto) folder. The `UI_*_PROVIDER` variables let you define the urls of the nodes that the user interface will connect to. Per default, it will connect to the local substrate and ethereum nodes, using their default RPC port. UI_EXPECTED_ETHEREUM_NETWORK_ID should be set to the ethereum network id. This is used by the user interface to prevent users from connecting their Metamask account to another network.
```
HCastano, line 110: Lines should be less than 100 characters in length. Maybe this is something we should enforce in our editor config actually...

deployments/rialto/docker-compose.yml
```diff
@@ -215,3 +215,14 @@ services:
       - "4567:4567"
     depends_on:
       - grafana-dashboard
+  front-end:
```
HCastano, line 218:
```suggestion

  front-end:
```

deployments/rialto/docker-compose.yml
```diff
@@ -215,3 +215,14 @@ services:
...
+    build:
+      context: .
+      dockerfile: ./Front-end.Dockerfile
+      args:
+        SUBSTRATE_PROVIDER: ${UI_SUBSTRATE_PROVIDER:-ws://bridge-node-alice:9944}
+        ETHEREUM_PROVIDER: ${UI_ETHEREUM_PROVIDER:-http://poa-node-arthur:8545}
+        EXPECTED_ETHEREUM_NETWORK_ID: ${UI_EXPECTED_ETHEREUM_NETWORK_ID:-105}
+
```
HCastano, line 226:
```suggestion
```

HCastano:
The Substrate node is exposed properly btw. The Grafana dashboard is picking up data from it. So there's either something wrong in the front-end Docker file, or app. I'll try and dig into it a bit

HCastano:
I ended up changing `SUBSTRATE_PROVIDER` to `ws://localhost:9944` and I got the UI to connect to our Substrate node. I'm not entirely sure why we can't use the internal Docker IP address here, since it works fine for the connection to the Ethereum nodes. 

Maybe it's because the connection string is used from the browser, which is external to the Docker setup?

General-Beck:
you don't mind if I start implementing this idea on Monday. I'm still on vacation. but if it's urgent, I can try it tomorrow. PM me in matrix pls

deployments/rialto/Bridge.Dockerfile
Tbaut: Oh, thanks for the info. Didn't know that.
```suggestion
```

Tbaut:
> Maybe it's because the connection string is used from the browser, which is external to the Docker setup?

Absolutely, I was brain fried yesterday. The UI we are connected to is the 8080 external to dockers. From the docker-compose, I can see that all the `poa-node-arthur` exposes 8545, and `bridge-node-alice` exposes 9944 to our local machine. Using localhost for the UI is then working as expected.

deployments/README.md
```diff
@@ -108,6 +108,11 @@ dashboard can be accessed at `http://localhost:9090`. The Grafana dashboard can
 Here is an example `.env` file which is used for production deployments and network updates. For
 security reasons it is not kept as part of version control. When deploying the network this
 file should be correctly populated and kept in the [`rialto`](./rialto) folder.
+The `UI_SUBSTRATE_PROVIDER` variable let you define the url of the nodes that the user interface
+will connect to. `UI_ETHEREUM_PROVIDER` is used only as a guidance for users to connect to
+Metamask to the right Ethereum network. `UI_EXPECTED_ETHEREUM_NETWORK_ID` should be set to the
+Ethereum network id. This is used by the user interface to prevent users from connecting their
+Metamask to an unexpected network.
```
Tbaut, line 115:
```suggestion
The `UI_SUBSTRATE_PROVIDER` variable lets you define the url of the Substrate node that the user interface
will connect to. `UI_ETHEREUM_PROVIDER` is used only as a guidance for users to connect
Metamask to the right Ethereum network. `UI_EXPECTED_ETHEREUM_NETWORK_ID`  is used by
the user interface as a fail safe to prevent users from connecting their Metamask extension to an
unexpected network.
```

HCastano:
> you don't mind if I start implementing this idea on Monday. I'm still on vacation. but if it's urgent, I can try it tomorrow. PM me in matrix pls

If you're on vacation you shouldn't be looking through GitHub! Go relax 🏖️

deployments/rialto/Front-end.Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+RUN cd /usr/src/ && git clone https://github.com/paritytech/bridge-ui.git
+WORKDIR /usr/src/bridge-ui
+RUN yarn
+ARG SUBSTRATE_PROVIDER
+ARG ETHEREUM_PROVIDER
+ARG EXPECTED_ETHEREUM_NETWORK_ID
+
+ENV SUBSTRATE_PROVIDER $SUBSTRATE_PROVIDER
+ENV ETHEREUM_PROVIDER $ETHEREUM_PROVIDER
+ENV EXPECTED_ETHEREUM_NETWORK_ID $EXPECTED_ETHEREUM_NETWORK_ID
```
HCastano, line 17: I removed these and everything still worked, what are they being used for right now?

HCastano (approved):
Tried it out locally and it all seems to work 😃. One comment about the `ENV` variables, but looks good otherwise.

Like I first suggested I want to see a published front-end image used here, but let's do that in a follow up PR.

deployments/rialto/Front-end.Dockerfile
Tbaut: When you say it still worked, you mean it still connected to the local nodes (the Subsrate one I mean because the Ehtereum Env variable is actually only used in the Error message for users to point their Metamask).

The `ENV $ARG` is to my knowledge the only way to pass any dynamic value form docker-compose to a Docker image. I've seen that in several places I think, more notably 
https://vsupalov.com/docker-arg-env-variable-guide/#setting-env-values

tomusdrw (approved):

tomusdrw:
Thanks a ton @Tbaut !

PR #341, opened 2020-09-08
Prepare separate runtime for testing sub2sub bridge

svyatonik:
This PR changes our `bin` folder layout to:
- `bin/rialto-runtime` - is runtime that has everything required to sync with Rialto-PoA. In future we'll add modules required to sync with chains that are running `bin/millau-runtime` runtime;
- `bin/millau-runtime` - is the `bin/rialto-runtime`, minus all PoA-related modules; 
- `bin/node` - is a library that provides single `run` function that starts Substrate node. Declares two features ("millau" and "rialto") that are used to select runtime;
- `bin/millau-node` - is a simple binary that is only calls `run()` function of `bin/node`, compiled with `bin/millau-runtime`;
- `bin/rialto-node` - is a simple binary that is only calls `run()` function of `bin/node`, compiled with `bin/rialto-runtime`.

I've also extracted `ShiftSessionManager` into separate `pallet-shift-session-manager`. Millau runtime and node are not mentioned in deployments, docs or whatever just because we have nothing functional yet.


deployments/rialto/Bridge.Dockerfile
```diff
@@ -38,8 +39,9 @@ WORKDIR /parity-bridges-common
 ARG BRIDGE_REPO=https://github.com/paritytech/parity-bridges-common
 RUN git clone $BRIDGE_REPO /parity-bridges-common && git checkout master
 
+# TODO: remove || true once PR is merged? Reason: if branch has new projects (rialto-bridge-node) and master has no this project yet (it has bridge-node), build fails
```
svyatonik, line 42: We may leave this `|| true`, or not. For now, I've marked it as to-be-removed in the future.

deployments/rialto/OpenEthereum.Dockerfile
```diff
@@ -18,7 +18,8 @@ RUN update-ca-certificates && \
 
 ENV PATH="/root/.cargo/bin:${PATH}"
 ENV LAST_RUST_UPDATE="2020-08-17"
-RUN rustup update stable && \
+RUN rustup toolchain uninstall stable && \
```
svyatonik, line 21: Without this change (and similar changes in other 2 files), I've been getting something like "failed to copy docs/html" when running docker-compose. Could rollback this fix, if it's only me

bin/node/millau-runtime/src/lib.rs
```diff
@@ -0,0 +1,475 @@
...
+
+// Parity Bridges Common is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! The Substrate Node Template runtime. This can be compiled with `#[no_std]`, ready for Wasm.
```
HCastano, line 17:
```suggestion
//! The Millau node runtime. This can be compiled with `#[no_std]`, ready for Wasm.
```

Looks like we also left the "Node Template" comment in the Rialto runtime, lol

deployments/rialto/docker-compose.git.yml
```diff
@@ -46,22 +46,22 @@ services:
         BRIDGE_HASH: ${RELAY_BRIDGE_HASH:-master}
         PROJECT: ethereum-poa-relay
 
-  bridge-node-alice: &bridge-node
+  rialto-bridge-node-alice: &rialto-bridge-node
```
HCastano, line 49: We'll need another deployment folder for Millau (not necessary in this PR), which will have it's own Docker Compose file. With different `docker-compose` files we don't need to have `rialto-*` in the names (they'll get added by docker based off the folder name, which in this case is `rialto`).

HCastano (commented):
Can I recommend a folder structure closer to this?

```
bin
└── node
    ├── millau
    └── rialto
└── runtime
    ├── millau
    └── rialto
 └── scripts
    └── init.sh
```

It makes the split between the `node` and the `runtime` a little clearer.

I would also prefer to see the shift-session manager pallet changes in a different PR. I know they're small but I think it would be better to make sure they're properly reviewed instead of just being a small part of a much larger set of changes.

svyatonik:
> It makes the split between the node and the runtime a little clearer.

Otoh, we may create chain-specific folders (millau, rialto) instead. In any case - there aren't many folders right now, so imo just prefixing the folder name is enough for now.

> I would also prefer to see the shift-session manager pallet changes in a different PR. I know they're small but I think it would be better to make sure they're properly reviewed instead of just being a small part of a much larger set of changes.

There aren't many changes - it just moves the code outside of runtime, so that it is usable from both runtimes. But I've opened #342 for that

deployments/rialto/docker-compose.git.yml
svyatonik: The idea is to get rid of simple `bridge-node`, `bridge-runtime` everywhere. Now we don't have "node" and "runtime", we have either Rialto, or Millau "-node" and "-runtime".

About deployment - I'm not yet sure, but I'd prefer to bridge real Rialto-Sub chain - i.e. the chain that is syncing Rialto-PoA headers with Millau-Sub chain. There are two benefits imo:
1) we'll see how `HeaderChain`, `MessageLane`, `CurrencyExchange` transactions are competing for block in-the-wild;
2) we have (almost) complete application that'll work on top of `MessageLane` module - i.e. we may change our currency-exchange-transactions-relay to work with `MessageLane` module of Millau-Sub. So that it'll:
2.1) will read proof-of-exchange transactions from Rialto-PoA node;
2.2) will submit Rialto-Sub **calls** as **messages** to Millau-Sub `MessageLane` module;
2.3) message-lane relay would finally submit these messages to the Rialto-Sub `MessageLane` module.

HCastano: Yeah maybe we need to rethink our deployments a little. Maybe something like this?

```
deployments
├── rialto // Eth <-> Sub
├── millau // Sub <->  Sub
└── tbd    // Eth <-> Sub <-> Sub
```

Each one of these would have their own compose files, leveraging the same published images. This doesn't have to be done now (and can't since Sub<->Sub is still not a thing), but something to keep in mind.

In this setup though I don't think any of the nodes aside from the `tbd` ones would need to have prefixes in the names.

HCastano:
> > It makes the split between the node and the runtime a little clearer.
> 
> Otoh, we may create chain-specific folders (millau, rialto) instead. In any case - there aren't many folders right now, so imo just prefixing the folder name is enough for now.

If you take a look at the Polkadot repo there are multiple runtimes which are split in a similar way to what I've proposed. Sure, we don't have as many runtimes as that repo, but I think the split between node and runtime is clearer there.

If you don't accept my suggestion, which is fine, I think we should try and get rid of the awkward top level `node` folder, or at least move the other folders to be at the same level. 

```
bin
├── millau-node
├── millau-runtime
├── ...
├── node
└── scripts
```

deployments/rialto/docker-compose.git.yml
svyatonik: I think what we need is `deployments/rialto` and `deployments/millau` with two `docker-compose` files, which will be using the same docker network => they both should have prefixes. But it is undecided => I do not want to talk about it now.

What's wrong with having `rialto-` prefix? Why we're using `bridge-node-` prefix then? I don't want to see `bridge-node` and `bridge-runtime` anymore, simply because we do not have `bridge-node` and `bridge-runtime` crates anymore.

deployments/rialto/OpenEthereum.Dockerfile
svyatonik: FTR - this hasn't actually solved my problem && I still have to `rustup toolchain uninstall stable` first :/ But since I've see no problems with this on CI, I'm considering this as my local issue.

deployments/rialto/docker-compose.git.yml
HCastano: Right now if I do `docker ps -a` I get back container names that look like this:

```
rialto_bridge-node-alice_1
```

If we add the `rialto` prefix our logs will end up looking like `rialto_rialto-bridge-node-alice_1` which is kinda awkward. If we leave them as they are in the compose file now they'll be prefixed correctly based off their folder.

deployments/rialto/docker-compose.yml
```diff
@@ -93,13 +93,13 @@ services:
       - "9618:9616"
     depends_on: *all-nodes
 
-  bridge-node-alice: &bridge-node
-    image: paritytech/bridge-node
+  rialto-bridge-node-alice: &rialto-bridge-node
+    image: paritytech/rialto-bridge-node
```
HCastano, line 97: I just tried the deployment locally and this change is actually a problem right now. We don't have any published images called `paritytech/rialto-bridge-node`, they're still published as `paritytech/bridge-node`. This means that when I do `docker-compose pull` this fails. Let's change this back to `bridge-node`, and maybe next week when General-Beck comes back we'll get some new images published.

deployments/rialto/docker-compose.git.yml
svyatonik: Ok, this makes sense. I'm removing `rialto-` prefix from container name, so logs should not have this duplicated prefix. I've added network aliases for default network instead (later we'll probably switch to some persistent network => we only have to change that "default:" in `docker-compose.yml`). I've also added optional `hostname:` association (thought it'll solve problem initially).

deployments/rialto/docker-compose.yml
svyatonik: I don't think anything should be done here - images are auto-published from master every day, right? So once this PR is merged, it'll be auto-fixed next day.

bin/millau-runtime/src/lib.rs
```diff
@@ -0,0 +1,475 @@
+// Copyright 2019-2020 Parity Technologies (UK) Ltd.
```
tomusdrw, line 1: I wonder if instead of duplicating the runtime code it makes more sense to have it conditionally compiled? Also couldn't we just do with a different genesis state?

tomusdrw (commented):
Looks good apart from Hernando's grumbles. I also wonder if it wouldn't be more sensible to actually conditionally compile the runtime and extract primitives used by the relayer into a "runtime-common" crate or something (which is not conditionally compiled for millau/rialto)

deployments/rialto/docker-compose.yml
HCastano: Yeah, good point. It escaped me that published package names would be whatever is in the `publish-docker.yml` file

deployments/rialto/docker-compose.git.yml
```diff
@@ -46,22 +46,22 @@ services:
...
       context: .
       dockerfile: ./Bridge.Dockerfile
       args:
         BRIDGE_HASH: ${NODE_BRIDGE_HASH:-master}
-        PROJECT: bridge-node
+        PROJECT: rialto-bridge-node
 
-  bridge-node-bob:
-    <<: *bridge-node
+  rialto-bridge-node-bob:
```
HCastano, line 57: These names need the `rialto` prefix removed. Otherwise when we merge the two files (if we want to use the Git one for instance) Docker doesn't realize that we want to override nodes and we end up with a ton of duplicated nodes

bin/millau-runtime/src/lib.rs
HCastano: I don't know if we can quite get everything conditionally compiled properly. From what I remember some macros, like `construct_runtime!`, don't support conditional compilation within them

deployments/rialto/docker-compose.git.yml
HCastano: I haven't used `networks` in Docker yet, so I don't know how that'll look in practice, but on the surface it looks alright

HCastano (approved):
As long as the `docker-compose` comment I made gets fixed I think this is ready to go. It's a bit of a gamble if we'll break the Docker image builds, but hopefully it goes alright :P

Also, don't forget to address the `|| true` TODO

svyatonik:
> Also, don't forget to address the || true TODO

It should be done in follow-up PR, or this one will never succeed.

PR #343, opened 2020-09-09
Resurrect substrate StorageProofChecker

svyatonik:
I've just reused some already available types and methods (like `StorageProof`, `into_memory_db()` and `read_trie_value()`)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4087,7 +4087,9 @@ version = "0.1.0"
 dependencies = [
  "hash-db",
  "serde",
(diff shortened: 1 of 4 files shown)


svyatonik:
One thing to remember when exposing something like `fn verify_proof(proof, keys) -> Values` is that we'll have multiple keys proved by single proof => the function should accept `keys`, not `key` to avoid cloning, etc

tomusdrw (approved):
lgtm

PR #345, opened 2020-09-10
Make message-lane Event generic

svyatonik:
(otherwise (non-default instances?) can't be plugged into runtime)

### DIFF
--- a/modules/message-lane/src/lib.rs
+++ b/modules/message-lane/src/lib.rs
@@ -49,7 +49,7 @@ mod mock;
 /// The module configuration trait
 pub trait Trait<I = DefaultInstance>: frame_system::Trait {
 	/// They overarching event type.
-	type Event: From<Event> + Into<<Self as frame_system::Trait>::Event>;
+	type Event: From<Event<Self, I>> + Into<<Self as frame_system::Trait>::Event>;
 	/// Message payload.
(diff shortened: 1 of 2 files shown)


modules/message-lane/src/lib.rs
```diff
@@ -75,13 +75,17 @@ decl_storage! {
 }
 
 decl_event!(
-	pub enum Event {
+	pub enum Event<T, I=DefaultInstance> where
```
HCastano, line 78:
```suggestion
	pub enum Event<T, I = DefaultInstance> where
```

HCastano (approved):

modules/message-lane/src/lib.rs
tomusdrw: Weird that it isn't covered by `cargo fmt`.

tomusdrw (approved):

modules/message-lane/src/lib.rs
HCastano: I think this is because we're inside a macro, so it doesn't really make sense to format code since you could break the code the macro is expecting

HCastano:
Our CI runners are stuck, but I'm going to merge this since the only change between now and then last time the CI was green as adding spaces around `=`

PR #347, opened 2020-09-11
Add Diego's key.

### LINKED ISSUE #346: Add Dev Account
While testing #340 I found myself moving funds from Arthur, Bertha, or Carlos to another MetaMask account. This is because those three accounts are being used for other purposes (e.g submitting CE transactions). This messes up the account nonces and confuses MetaMask, leaving these accounts unusable. 

We should add another account (whose name obviously starts with a D :P) which has some amount of funds endowed, but is not used for anything else.

tomusdrw:
Resolves #346


HCastano (approved):

HCastano:
It would be nice to see the role of each authority mentioned in the Rialto deployment README. Something for a different PR though

PR #350, opened 2020-09-14
Fix Rialto test-poa genesis state root

svyatonik:
this should (hopefully) fix the deployment issue


tomusdrw:
Ah, it's related to #347? Is there a way we can prevent this in the future?
Maybe a tool to generate ETH genesis hash based on the chainspec?

tomusdrw (approved):

svyatonik:
I think the feature we need is "comments-in-json" :)

HCastano:
I thought #347 was pretty trivial, didn't realize it would break our deployment upon merging, whoops!

@svyatonik how did you get the new `state_root`?

HCastano (approved):

svyatonik:
By starting PoA node && calling RPC

PR #351, opened 2020-09-14
Relay receiving + processing confirmations

### LINKED ISSUE #334: Update relay to support DummyOrdered receiving and processing confirmations
related #213 #318

Once #318 is in, it should be updated with additional races - `MessageReceiving` and `MessageProcessing`.

svyatonik:
closes #334


svyatonik:
(will no longer need separate processing confirmations => should be reworked a bit)

svyatonik:
(disregard my previous comment)

relays/ethereum/src/message_race_delivery.rs
```diff
@@ -135,34 +136,53 @@ where
...
 			target_nonce: Default::default(),
+			max_nonces_to_relay_in_single_tx,
 			_phantom: Default::default(),
 		}
 	}
 }
 
-impl<P: MessageLane> RaceStrategy<SourceHeaderIdOf<P>, TargetHeaderIdOf<P>, P::MessageNonce, P::MessagesProof>
-	for MessageDeliveryStrategy<P>
+impl<SHN, SHH, THN, THH, Nonce, Proof> RaceStrategy<HeaderId<SHH, SHN>, HeaderId<THH, THN>, Nonce, Proof>
```
HCastano, line 174: Yeah, so I know I was the one that suggested not using full names for generics, but in this case this makes the code more confusing since now you need to think about what these are acronyms for (it's not clear from anything in the function body)

relays/ethereum/src/message_race_delivery.rs
```diff
@@ -135,34 +136,60 @@ where
...
+	<P as MessageLane>::SourceHeaderHash,
+	<P as MessageLane>::TargetHeaderNumber,
+	<P as MessageLane>::TargetHeaderHash,
+	<P as MessageLane>::MessageNonce,
+	<P as MessageLane>::MessagesProof,
+>;
+
+/// Nonces delivery strategy.
+#[derive(Debug)]
+pub struct DeliveryStrategy<SourceHeaderNumber, SourceHeaderHash, TargetHeaderNumber, TargetHeaderHash, Nonce, Proof> {
```
tomusdrw, line 151: Why do we split the generic parameters instead of simply using `<P: MessageLane>`? The extra bounds could be added via `where` clause, no?

relays/ethereum/src/message_race_delivery.rs
```diff
@@ -229,18 +266,19 @@ impl<P: MessageLane> RaceStrategy<SourceHeaderIdOf<P>, TargetHeaderIdOf<P>, P::M
...
 		// 3) we can't deliver new nonce until header, that has emitted this nonce, is finalized
 		// by target client
 		let nonces_begin = self.target_nonce + 1.into();
 		let best_header_at_target = &race_state.target_state.as_ref()?.best_peer;
 		let mut nonces_end = None;
-		for i in 0..MAX_MESSAGES_TO_RELAY_IN_SINGLE_TX {
-			let nonce = nonces_begin + i.into();
+		let mut i = Zero::zero();
+		while i < self.max_nonces_to_relay_in_single_tx {
+			let nonce = nonces_begin + i;
```
tomusdrw, line 277:
```suggestion
		let mut nonce = nonces_begin;
		let mut end = nonces_begin + self.max_nonces_to_relay_in_single_tx;
		while nonce < end {
```

Wouldn't that be a bit simpler?

tomusdrw (approved):
lgtm!

relays/ethereum/src/message_race_delivery.rs
svyatonik: Because when `DeliveryStrategy` is used in message-delivery-race, then the `race.source = lane.source && race.target = lane.target`. When used in delivery-confirmations-race, then the `race.source = lane.target && race.target = lane.source`.

tomusdrw: Okay, fair! We could implement `struct FlipSourceTarget<P: MessageLane>; impl MessageLane for FlipSourceTarget<P> {}` though, couldn't we and then just wrap it?

svyatonik: This would work, but `MessageLane` is not only about source/target chain headers hashes/numbers. So it doesn't seems right.

PR #353, opened 2020-09-15
Verify GRANDPA justifications from runtime

svyatonik:
slightly adapted (no `HashMap`s, `HashSet`s, no excess cloning, tests) version of https://github.com/paritytech/substrate/blob/master/client/finality-grandpa/src/justification.rs

cc @HCastano


HCastano:
Nice, I'll review this tomorrow :)

modules/substrate/src/justification.rs
```diff
@@ -0,0 +1,329 @@
...
+	/// Justification has invalid authority singature.
+	InvalidAuthoritySignature,
+	/// The justification has precommit for the header that has no route from the target header.
+	InvalidPrecommitAncestryProof,
+	/// The justification has 'unused' headers in its precommit ancestries.
+	InvalidPrecommitAncestries,
+}
+
+/// Verify that justification, that is generated by given authority set, finalizes given header.
+pub fn verify_justification<Header: HeaderT>(
```
tomusdrw, line 48: Where would we verify that the `finalized_target` actually builds on top of the latest finalized block we know of? I guess we need the ancestry hashes for that?

modules/substrate/src/justification.rs
```diff
@@ -0,0 +1,329 @@
...
+				return Err(Error::InvalidPrecommitAncestryProof);
+			}
+		}
+	}
+
+	let ancestry_hashes = justification
+		.votes_ancestries
+		.iter()
+		.map(|h: &Header| h.hash())
+		.collect();
```
tomusdrw, line 109: To avoid collecting to a `Vec` we can just `.zip` with `visited_hashes` and then verify in a loop.

modules/substrate/src/justification.rs
```diff
@@ -0,0 +1,329 @@
...
+			}
+			match self.ancestry.get(&current_hash).cloned() {
+				Some(parent_hash) => {
+					current_hash = parent_hash;
+					route.push(current_hash);
+				}
+				_ => return Err(GrandpaError::NotDescendent),
+			}
+		}
+		route.pop(); // remove the base
```
tomusdrw, line 161: Hmm, why does the comment say "remove the base", if we never add the base to `route` since we exit the loop early in case `current_hash == base`? I guess this will remove `base.parent_hash`, no?

modules/substrate/src/justification.rs
```diff
@@ -0,0 +1,329 @@
...
+	Header::Number: finality_grandpa::BlockNumberOps,
+{
+	fn ancestry(&self, base: Header::Hash, block: Header::Hash) -> Result<Vec<Header::Hash>, GrandpaError> {
+		let mut route = Vec::new();
+		let mut current_hash = block;
+		loop {
+			if current_hash == base {
+				break;
+			}
+			match self.ancestry.get(&current_hash).cloned() {
```
tomusdrw, line 153: Does it make sense to collect into a `Map` at all? It seems that we can just walk the `&[Header]` slice backwards, until we find the `block` and then start collecting the `route` until we get to `base`. Or is it possible that `&[Header]` contains a bunch of forks as well?

modules/substrate/src/justification.rs
```diff
@@ -0,0 +1,329 @@
...
+				_ => return Err(GrandpaError::NotDescendent),
+			}
+		}
+		route.pop(); // remove the base
+
+		Ok(route)
+	}
+
+	fn best_chain_containing(&self, _block: Header::Hash) -> Option<(Header::Hash, Header::Number)> {
+		None
```
tomusdrw, line 167: I guess this is not implemented, but we just don't need it right? Wouldn't it make more sense to do `unreachable!()` here?

tomusdrw (commented):
Looks good! Few questions left in comments.

modules/substrate/src/justification.rs
svyatonik: This is outside of scope of this function. This function is verifying that justification, that is generated by given authority set, finalizes given header (`finalized_target`).

modules/substrate/src/justification.rs
svyatonik: Using `zip()`assumes that both iterators are sorted in the same order. I don't know if it is true.

modules/substrate/src/justification.rs
svyatonik: If we exit loop early, then the vec would be empty && pop will be a noop, which, I believe is the expected result

modules/substrate/src/justification.rs
svyatonik: I don't know if there are some guarantees about ordering headers in this vec. Keep in mind that the justification may be incorrect/malformed.

modules/substrate/src/justification.rs
tomusdrw: Ah, fair enough. I thought we compare `Vec`s.

modules/substrate/src/justification.rs
tomusdrw: Okay. Better safe than sorry.

modules/substrate/src/justification.rs
tomusdrw: I was thinking about the non-empty case, but on a second look I got it.. The resulting vec will contain `[block.parent_hash, base]` after we exit the loop, because we go through the headers in reverse order (i.e. we push `next_block(base).parent_hash` as the last step in the loop, and then exit when `next_block(base).parent_hash == current_hash == base`, so we need to `pop` the base out of the vec.

tomusdrw (approved):

HCastano (approved):
Looks good!

Two comments (mainly so I hopefully don't forget). I think it would be useful to have this as part of the primitives, so that this code can be re-used by different components like the built-in, or even external projects. Actually, now that I think about it, what's stopping us from making these `no_std` changes directly in Substrate and pulling the code in from there?

Secondly, some of the test helpers here could also be merged with some helpers in the Substrate pallet (once that's merged).

PR #355, opened 2020-09-16
Support user-provided origins in Call-dispatch module

### LINKED ISSUE #335: Support user-provided origins in Call-dispatch module
related #211 #300

Original idea by @tomusdrw : instead of originating all requests by [bridge-account](https://github.com/paritytech/parity-bridges-common/pull/300/files#diff-185ee8dcbded1e37d6080f3951330d70R148) we should instead allow users on the source side to provide origin account on target chain. So every message (i.e. encoded `Call`) will be bundled with `signature` over `<something>` (encoded call? encoded call + nonce? ...).

We still need to keep `bridge-account`. It'll play another role - if `MessageLane::send_message()` is called by the **source chain** root, then we are dispatching this call with origin = `bridge-account` on the target chain.

We need to be careful with possible replay attacks. The possible solution is to include `SourceAccountId` in the signed message (again - cheers to @tomusdrw ):
```
The message would contain (Call, SourceAccountId, TargetSignature) where for instance for ed25519/sr25519 TargetSignature = TargetPubKey + TargetSignatureOver(Call ++ SourceAccountId).
On the source chain you only verify SourceAccountId = ensure_signed(origin) and on the target you verify the signature. So it just won't dispatch if it's invalid.
```

svyatonik:
closes #335

### DIFF
--- a/modules/call-dispatch/src/lib.rs
+++ b/modules/call-dispatch/src/lib.rs
@@ -28,13 +28,20 @@
 
 use bp_message_dispatch::{MessageDispatch, Weight};
 use bp_runtime::{bridge_account_id, InstanceId, CALL_DISPATCH_MODULE_PREFIX};
+use codec::{Decode, Encode};
 use frame_support::{
-	decl_event, decl_module,
+	decl_event, decl_module, decl_storage,
 	dispatch::{Dispatchable, Parameter},
 	traits::Get,
 	weights::{extract_actual_weight, GetDispatchInfo},
+	RuntimeDebug,
 };
-use sp_runtime::DispatchResult;
+use frame_system::{ensure_root, ensure_signed, RawOrigin};
+use sp_runtime::{
+	traits::{BadOrigin, IdentifyAccount, Verify},
+	DispatchResult,
+};
+use sp_std::{marker::PhantomData, prelude::*};
 
 /// Spec version type.
 pub type SpecVersion = u32;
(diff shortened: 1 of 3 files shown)


modules/call-dispatch/src/lib.rs
```diff
@@ -43,14 +50,52 @@ pub type SpecVersion = u32;
 /// Weight of single deposit_event() call.
 const DEPOSIT_EVENT_WEIGHT: Weight = 0;
 
+/// Origin of the call on the target chain.
+#[derive(RuntimeDebug, Encode, Decode, Clone)]
+pub enum CallOrigin<SourceChainAccountPublic, TargetChainAccountPublic, TargetChainSignature> {
+	/// Call is originated from bridge account, which is (designed to be) specific to
+	/// the single deployed instance of the messages bridge (message-lane, ...) module.
+	/// It is assumed that this account is not controlled by anyone and has zero balance
+	/// (unless someone would make transfer by mistake?).
```
tomusdrw, line 59:
```suggestion
	/// (unless someone would make transfer by mistake?).
	/// If we trust the source chain to allow sending calls with that origin in case they originate from source chain `root` account (default implementation),
	/// `BridgeAccount` represents the source-chain-root origin on the target chain and can be used to send and authorize
	/// "control plane" messages between the two runtimes. 
```

modules/call-dispatch/src/lib.rs
```diff
@@ -60,75 +105,117 @@ pub trait Trait: frame_system::Trait {
...
+			));
 			return DEPOSIT_EVENT_WEIGHT;
 		}
 
+		// prepare dispatch origin
+		let origin_account = match message.origin {
+			CallOrigin::BridgeAccount => bridge_account_id(bridge, CALL_DISPATCH_MODULE_PREFIX),
+			CallOrigin::RealAccount(source_public, target_public, target_signature) => {
+				let mut signed_message = Vec::new();
+				message.call.encode_to(&mut signed_message);
```
tomusdrw, line 196: I was thinking that `spec_version` might be required to be here as well, but after a deeper thought it's not needed, because:
1. We verify the whole `MessagePayload` objects indirectly, by verifying the proof of inclusion on the target chain (i.e. no tampering possible here)
2. Since we add `source_public` to the `signed_message` here, it's only possible for the `source_public` account on the source chain to add messages to the outbound queue.
3. So it means that only the original sender can replay the same `origin(with signature) + call` on the source chain against a different `spec_version` or `weight`.

modules/call-dispatch/src/lib.rs
```diff
@@ -60,75 +105,117 @@ pub trait Trait: frame_system::Trait {
...
+				if !target_signature.verify(&signed_message[..], &target_account) {
+					frame_support::debug::trace!(
+						"Message {:?}/{:?}: origin proof is invalid. Expected account: {:?} from signature: {:?}",
+						bridge,
+						id,
+						target_account,
+						target_signature,
+					);
+					Self::deposit_event(RawEvent::MessageSignatureMismatch(bridge, id));
+					return DEPOSIT_EVENT_WEIGHT;
```
tomusdrw, line 209: Shouldn't we charge a bit more due to signature verification?

modules/call-dispatch/src/lib.rs
```diff
@@ -147,6 +234,39 @@ impl<T: Trait> MessageDispatch<T::MessageId> for Module<T> {
 	}
 }
 
+/// Verify payload of the message at the sending side.
+pub fn verify_sending_message<
```
tomusdrw, line 238: A call to this function should be added to some kind of payload verification in the message delivery pallet, right?
I mean that the message delivery upon `add_message` call should delegate the payload verification to the dispatcher, and that's where we'd call this function?
I guess we don't have a trait for that and it's not really plugged into the runtime, so that's why it's not used apart from the tests?

tomusdrw (approved):
lgtm!

modules/call-dispatch/src/lib.rs
svyatonik: Yes - it should be verified during `send_message()` call. That's what's `LaneMessageVerifier` from #339 is for

modules/call-dispatch/src/lib.rs
svyatonik: This (returned `Weight`) is actually removed in #339

tomusdrw (approved):

HCastano:
Seems okay, but I only took a quick look. Will take another look tomorrow

tomusdrw (approved):
lgtm, awaiting Hernando's review.

HCastano (approved):

Issue #344, opened 2020-09-09
Change relay name and source files layout


svyatonik:
origin: https://github.com/paritytech/parity-bridges-common/pull/318#discussion_r483393192

Suggestions:
1) relay should live in smth like `/relays/generic-relay` or simply `/relay` folder, because it is no longer eth2sub-specific, as it has been in the very beginning;
2) we should stop calling it "ethereum-poa-relay";
3) group source files by type of relay. Something like: `relay/src/exchange`, `relay/src/headers` and `relay/src/messages` + some top-level modules for connecting to nodes, etc.

This ^^^ is written in assumption that we will keep using single binary to relay everything (which isn't optimal imo, but we're going this way currently).

PR #357, opened 2020-09-16
Extract (headers, exchange, messages) relay loops into separate crates

svyatonik:
related #344 

The relay binary itself is not renamed, because imo its future is undecided yet - either we have one binary for all chains as @HCastano suggested, or we'll have separate binary for every pair of chains. I do not want to start this discussion now, so no binary renames, no chain-specific crates, ...


relays/utils/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+
+[dependencies]
+async-std = "1.6.2"
+backoff = "0.2"
+futures = "0.3.5"
+log = "0.4.11"
+sysinfo = "0.15"
+
+# Substrate dependencies
+substrate-prometheus-endpoint = { version = "0.8.0-rc6", git = "https://github.com/paritytech/substrate.git", tag = "v2.0.0-rc6" }
```
HCastano, line 16: To avoid long lines, can you use the `[dependencies.substrate-prometheus-endpoint]` style for imports?

relays/exchange-relay/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+
+[dependencies]
+async-std = "1.6.2"
+async-trait = "0.1.40"
+backoff = "0.2"
+futures = "0.3.5"
+log = "0.4.11"
+num-traits = "0.2"
+parking_lot = "0.11.0"
+relay-utils = { path = "../utils" }
```
HCastano, line 16: Of the new `.toml` files this is the only one without a newline at the end of the file

relays/exchange-relay/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+
+// Parity Bridges Common is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Relaying currency-exchange application specific data.
```
HCastano, line 17: Can you beef up the description a bit?

relays/headers-relay/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+
+// Parity Bridges Common is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Relaying headers.
```
HCastano, line 17: Same here, this doesn't tell a reader much.

relays/messages-relay/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+
+// Parity Bridges Common is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Relaying messages over message lanes.
```
HCastano, line 17: This also makes some assumptions about the readers previous knowledge. I'd write this imagining that the reader is a new person coming into the codebase and doesn't know what "lanes" are for example.

HCastano (commented):
Looks fine, I like the split. I've left some comments about documentation.

This is a good time to update the README, since the "Project Layout" and "Bridge Relay" sections are already a bit out of date. I think that should be part of this PR, but I'm okay having it as a follow up as well.

relays/utils/Cargo.toml
svyatonik: They're hard to read. When it's just one-line, I could easily see all dependencies on one page. When every dependency takes up 6 lines, it's easy to follow what crate really depends on. I'd prefer to have one-liners everywhere. But here I'm just starting with relay libs + bin.

svyatonik:
Let's revisit README after we have `sub2sub` working. Given that the repo is growing, it seems too detailed. Imo we should concentrate on what is this and how to run it instead of enumerating files and folders. This gives no useful information at all and only adds another-thing-to-maintain.

relays/utils/Cargo.toml
tomusdrw: > I'd prefer to have one-liners everywhere.

Me too actually. The list of deps (names) is imho way more important than the rest of the line, it has to be always the same anyway (same version / git repo / etc).

relays/ethereum/src/ethereum_types.rs
```diff
@@ -34,6 +35,10 @@ pub type Transaction = web3::types::Transaction;
 /// Ethereum header type.
 pub type Header = web3::types::Block<H256>;
 
+/// Ethereum header type used in headers sync.
+#[derive(Clone, Debug, PartialEq)]
+pub struct EthereumSyncHeader(pub Header);
```
tomusdrw, line 40: If this is only a type-safety wrapper, why not have `Deref` implemented?

relays/ethereum/src/main.rs
```diff
@@ -266,7 +264,16 @@ fn substrate_sync_params(matches: &clap::ArgMatches) -> Result<SubstrateSyncPara
 		eth_sign: ethereum_signing_params(matches)?,
 		metrics_params: metrics_params(matches)?,
 		instance: instance_params(matches)?,
-		sync_params: HeadersSyncParams::substrate_sync_default(),
+		sync_params: HeadersSyncParams {
+			max_future_headers_to_download: MAX_FUTURE_HEADERS_TO_DOWNLOAD,
+			max_headers_in_submitted_status: MAX_SUBMITTED_HEADERS,
+			// since we always have single Substrate header in separate Ethereum transaction,
```
tomusdrw, line 270: Shouldn't the comment be the other way around? I.e. "we always have only one Ethereum header in each Substrate transaction"?

relays/headers-relay/src/sync_loop_tests.rs
```diff
@@ -18,30 +18,30 @@
...
 	time::Duration,
 };
 
-type TestNumber = u64;
-type TestHash = u64;
-type TestExtra = u64;
-type TestCompletion = u64;
-type TestHeaderId = HeaderId<TestHash, TestNumber>;
-type TestQueuedHeader = QueuedHeader<TestHeadersSyncPipeline>;
+pub type TestNumber = u64;
```
tomusdrw, line 33: It's the test code, but maybe still worth adding docs for at least some of the stuff?

relays/headers-relay/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+//! Relaying source chain headers to target chain. This module provides entrypoint
+//! that starts reading new headers from source chain and submit these headers as
+//! module/contract transactions to the source chain. Module/contract on the target
+//! chain is a light-client of the source chain. All other trustless bridge
+//! applications are built using this light-client, so running headers-relay is
+//! essential for running all other bridge applications.
+
+// required for futures::select!
+#![recursion_limit = "1024"]
+
```
tomusdrw, line 26:
```suggestion
#![warn(missing_docs)]
```
?

relays/exchange-relay/src/lib.rs
```diff
@@ -0,0 +1,24 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Relaying [`currency-exchange`](../pallet_bridge_currency_exchange/index.html) application
+//! specific data. Currency exchange application allows exchanging tokens between bridged chains.
+//! This module provides entrypoints for crafting and submitting (single and multiple)
+//! proof-of-exchange-at-source-chain transaction(s) to target chain.
```
tomusdrw, line 20:
```suggestion
//! proof-of-exchange-at-source-chain transaction(s) to target chain.

#![warn(missing_docs)]
```
?

relays/messages-relay/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+
+//! Relaying [`message-lane`](../pallet_message_lane/index.html) application specific
+//! data. Message lane allows sending arbitrary messages between bridged chains. This
+//! module provides entrypoint that starts reading messages from given message lane
+//! of source chain and submits proof-of-message-at-source-chain transactions to the
+//! target chain. Additionaly, proofs-of-messages-delivery are sent back from the
+//! target chain to the source chain.
+
+// required for futures::select!
+#![recursion_limit = "1024"]
```
tomusdrw, line 25:
```suggestion
#![recursion_limit = "1024"]
#![warn(missing_docs)]
```

tomusdrw (approved):
lgtm!

relays/ethereum/src/main.rs
svyatonik: It is about `sub2eth`, so comment was correct some time ago. Now we may have up to 4 sub headers in single eth tx => it is obsolete

relays/exchange-relay/src/lib.rs
```diff
@@ -0,0 +1,26 @@
...
+
+// Parity Bridges Common is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Relaying [`currency-exchange`](../pallet_bridge_currency_exchange/index.html) application
```
HCastano, line 17: I was curious as to whether or not the `index.html` path was needed, and it turns out that nightly Rust has had intra-doc links (so regular Rust paths) for a couple of years now. However, they're _just_ being stabilized now. I think we should see this on stable in the next one or two releases

relays/headers-relay/src/headers.rs
```diff
@@ -14,11 +14,17 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
 
+//! Headers queue - the intermediate buffer that is filled when headers are read
+//! from the source chain. Headers are removed from the queue once they became
```
HCastano, line 18:
```suggestion
//! from the source chain. Headers are removed from the queue once they become
```

relays/headers-relay/src/headers.rs
```diff
@@ -14,11 +14,17 @@
...
+//! known to the target chain. Inside, there are several sub-queues, where headers
+//! may stay until source/target chain state isn't updated. When header reach
```
HCastano, line 20:
```suggestion
//! may stay until source/target chain state isn't updated. When a header reaches the
```

relays/headers-relay/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Relaying source chain headers to target chain. This module provides entrypoint
+//! that starts reading new headers from source chain and submit these headers as
+//! module/contract transactions to the source chain. Module/contract on the target
```
HCastano, line 19:
```suggestion
//! module/contract transactions to the target chain. Module/contract on the target
```

relays/utils/Cargo.toml
HCastano: Maybe it's just me then, but I really don't like how these long lines look on (for example) Polkadot's `Cargo.toml` files. It's especially unhelpful when the feature flags are declared at the end and you need to scroll around to see what features a particular crate has.

That being said, I think we'll get off the GitHub versions in a few weeks and start using versions directly off crates.io. At that point I wouldn't mind in-line dependencies.

HCastano (approved):
> Let's revisit README after we have `sub2sub` working. Given that the repo is growing, it seems too detailed. Imo we should concentrate on what is this and how to run it instead of enumerating files and folders. This gives no useful information at all and only adds another-thing-to-maintain.

Sure. I think that going forward we should move all the file level description we had the README into the files themselves, but still provide a high level folder overview (`modules`, `relays`, `primitives`, etc). This at least gives a newcomer a _sense_ of where to start poking around, and then once they dig in they can see more detailed docs.

Thanks for beefing up the docs by the way!

PR #359, opened 2020-09-18
Extract minimal ethereum client

svyatonik:
related #344

This extracts minimal ethereum client code from relay binary to separate crate. So relay binary now only has eth-related code that is only valid in `eth2sub` and `sub2eth` context. This PR isn't touching any logic - just moving the code + adopting existing code to that move.

Unrelated to this PR, but the final goal is to also extract `rialto-client` and `mullau-client` => then we may combine clients crates and generic-relay crates (headers-relay, messages-relay, exchange-relay) in any way we'd agree on - all-in-one relay, per-chain-pair relay all-in-one relay with features, ...

### DIFF
--- a/relays/ethereum-client/Cargo.toml
+++ b/relays/ethereum-client/Cargo.toml
@@ -6,7 +6,6 @@ edition = "2018"
 license = "GPL-3.0-or-later WITH Classpath-exception-2.0"
 
 [dependencies]
-async-trait = "0.1.40"
 codec = { package = "parity-scale-codec", version = "1.3.4" }
 ethereum-tx-sign = "3.0"
 hex = "0.4"
--- a/relays/ethereum-client/Cargo.toml
+++ b/relays/ethereum-client/Cargo.toml
@@ -8,10 +8,10 @@ license = "GPL-3.0-or-later WITH Classpath-exception-2.0"
 [dependencies]
 codec = { package = "parity-scale-codec", version = "1.3.4" }
 ethereum-tx-sign = "3.0"
+headers-relay = { path = "../headers-relay" }
 hex = "0.4"
 jsonrpsee = { git = "https://github.com/svyatonik/jsonrpsee.git", branch = "shared-client-in-rpc-api", default-features = false, features = ["http"] }
 log = "0.4.11"
-headers-relay = { path = "../headers-relay" }
 parity-crypto = { version = "0.6", features = ["publickey"] }
 relay-utils = { path = "../utils" }
 web3 = "0.13"
--- a/relays/ethereum-client/Cargo.toml
+++ b/relays/ethereum-client/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "relay-ethereum-client"
+version = "0.1.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+license = "GPL-3.0-or-later WITH Classpath-exception-2.0"
+
+[dependencies]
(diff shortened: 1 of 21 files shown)


HCastano (approved):
I think there are some quirks with the folder structure/naming (e.g `ethereum/src/ethereum_client.rs` and `ethereum-client/src/client.rs`) but I'm gonna wait until the Substrate client gets split (or I guess Rialto/Millau is what you want to split next?) before making any comments on it.

I like where this is going though!

relays/ethereum/src/ethereum_client.rs
```diff
@@ -240,7 +87,7 @@ pub trait EthereumHighLevelRpc: EthereumRpc {
 }
 
 #[async_trait]
-impl EthereumHighLevelRpc for EthereumRpcClient {
+impl<T: EthereumClient> EthereumHighLevelRpc for T {
```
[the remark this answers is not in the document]
svyatonik: I can't say if it's better practice or not. But in this case it is implement-EthereumHighLevelRpc-just-for-one-type vs implement-EthereumHighLevelRpc-for-all-types-that-implement-EthereumClient. Probably we don't need `trait EthereumHighLevelRpc` at all, but I don't want to touch too much code in this PR.

relays/ethereum-client/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+license = "GPL-3.0-or-later WITH Classpath-exception-2.0"
+
+[dependencies]
+async-trait = "0.1.40"
+codec = { package = "parity-scale-codec", version = "1.3.4" }
+ethereum-tx-sign = "3.0"
+hex = "0.4"
+jsonrpsee = { git = "https://github.com/svyatonik/jsonrpsee.git", branch = "shared-client-in-rpc-api", default-features = false, features = ["http"] }
+log = "0.4.11"
+headers-relay = { path = "../headers-relay" }
```
tomusdrw, line 15: Would you mind sorting the deps?

relays/ethereum-client/src/lib.rs
```diff
@@ -0,0 +1,91 @@
...
+			port: 8545,
+		}
+	}
+}
+
+/// The API for the supported Ethereum RPC methods.
+///
+/// Cloning client is a lightweight operation that only clones internal references.
+#[async_trait]
+pub trait Client: 'static + Send + Sync + Clone {
```
tomusdrw, line 61: I wonder if it makes sense to have this trait at all. Why not expose the `client::Client` directly with a bunch of `pub async fn` methods?
Do we have a different implementation for tests?

tomusdrw: The only reason I see for the trait is to move it to a yet another crate so that we can keep the dependency tree light. Now we import the `trait` but it always comes with a specific implementation, which kind of defeats the purpose of having the trait in the first place.

tomusdrw (approved):
+1 for extraction although I think we either need to get rid of the trait or split the trait from actual concrete implementation to keep the dependency tree light.

relays/ethereum-client/src/lib.rs
svyatonik: I think initially maybe we were going to use some custom implementation in tests? But we now have higher-level traits that are actually customized for tests (`SourceClient`/`TargetClient`) => I agree that removing `Client` trait is a good deed. Removed

tomusdrw (approved):

PR #364, opened 2020-09-22
Update schedule for deny

General-Beck:
'* * * * 0' -> '0 0 * * *'

### DIFF
--- a/.github/workflows/deny.yml
+++ b/.github/workflows/deny.yml
@@ -3,7 +3,7 @@ name:                   Cargo deny
 on:
   pull_request:
   schedule:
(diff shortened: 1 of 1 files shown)


HCastano (approved):
To confirm, you want to change from running it once a week to running this nightly? I know this is what the change _does_, but I want to confirm that this is the change you _want_.

What's the motivation behind this?

General-Beck:
take a look at the history. once a minute every week is too much. Let's do it once a week then.
`'0 0 * * 0'`

tomusdrw (approved):

PR #365, opened 2020-09-23
Extract minimal Rialto-Sub and Millau-Sub clients

svyatonik:
Code in `relays/ethereum` is now specific to eth2sub && sub2eth, where only supported Sub is currently Rialto-Sub chain (we may add support for other chains as well, but it isn't what I want to do now).
Code in `relays/substrate-client` is a minimal `Client<Chain>` - i.e. it is client that is able to talk with selected Substrate-based chains. There are two builtin chains there (activated with cargo features) - millau and rialto.


Cargo.lock
```diff
@@ -2617,7 +2612,7 @@ dependencies = [
 [[package]]
 name = "jsonrpsee"
 version = "1.0.0"
-source = "git+https://github.com/svyatonik/jsonrpsee.git?branch=shared-client-in-rpc-api#99fb015dd9d678392dab8e90efdbd9b4a6b82fd9"
+source = "git+https://github.com/svyatonik/jsonrpsee.git?branch=shared-client-in-rpc-api#8e7fe826517a21c99e7868e056b793e5c6749fce"
```
svyatonik, line 2615: There's another small change in our fork of jsonrpsee. [This commit](https://github.com/svyatonik/jsonrpsee/commit/bc0d938217fd0c34e8dcaa08d034ab157c87b363) allows specifying using associated types in RPC methods. I.e.:
```rust
jsonrpsee::rpc_api! {
	pub(crate) Substrate<C: Chain> {
		#[rpc(method = "chain_getHeader", positional_params)]
		fn chain_get_header(block_hash: Option<C::Hash>) -> C::Header;
        }
}
```

Without it, I had to specify all used types explicitly, instead of using 'types container' (`Chain`).

svyatonik:
Next PR will be about relay itself (sub2sub specifically). I kindly ask you to leave your opinion about future of `ethereum-poa-relay` relay here. The possibly options I consider are:
1) rename it to generic relay and use it to serve all sub2eth, eth2sub and sub2sub bridges. I don't think putting everything in a single binary is a best option - e.g. we won't have message-lanes in sub2eth and we do not want (at least initially) currency exchange in sub2sub;
2) rename it to generic relay and turn on/off support for chains/applications using cargo features. Imo it is even worse than (1);
3) add separate relay for serving Millau headers/messages relay (Millau-Sub <-> Rialto-Sub). I guess later we may extend this binary to support other chains as well - but this is yet to be decided.

relays/ethereum/src/instances.rs
```diff
@@ -33,83 +33,84 @@ use rialto_runtime::Call;
...
-				.collect(),
-		);
-
-		rialto_runtime::Call::BridgeKovan(pallet_call)
+impl BridgeInstance {
+	/// Used to build a `Call` for importing signed headers to a Substrate runtime.
+	pub fn build_signed_header_call(&self, headers: Vec<QueuedEthereumHeader>) -> Call {
+		match *self {
+			Self::TestPoA => {
+				let pallet_call = rialto_runtime::BridgeEthPoACall::import_signed_headers(
```
tomusdrw, line 49: The `pallet_call` seems identical for `TestPoA` and `Kovan` afaict, can't we extract it?

relays/ethereum/src/instances.rs
```diff
@@ -33,83 +33,84 @@ use rialto_runtime::Call;
...
 /// but slightly different `Call` in order to be synchronized.
-pub trait BridgeInstance: Send + Sync + std::fmt::Debug {
-	/// Used to build a `Call` for importing signed headers to a Substrate runtime.
-	fn build_signed_header_call(&self, headers: Vec<QueuedEthereumHeader>) -> Call;
-	/// Used to build a `Call` for importing an unsigned header to a Substrate runtime.
-	fn build_unsigned_header_call(&self, header: QueuedEthereumHeader) -> Call;
-	/// Used to build a `Call` for importing peer transactions to a Substrate runtime.
-	fn build_currency_exchange_call(&self, proof: Proof) -> Call;
+#[derive(Debug, Clone, Copy)]
+pub enum BridgeInstance {
```
tomusdrw, line 37: The reason against `enum` was that it's not extensible - i.e. it's not easy for an external party to add it's own rialto-like runtime with more instances of the bridge.
IMHO ideally the runtime would be generic on the `sync_loop` level, so that other chains can use the library and initialize the event loop with a different runtime.
That said, since we have the runtime hardcoded already I believe having this enum doesn't matter that much. I'd leave it up to you - I'm fine with just creating an issue for this.

relays/ethereum/src/instances.rs
```diff
@@ -33,83 +33,84 @@ use rialto_runtime::Call;
...
-			into_substrate_ethereum_header(header.header()),
-			into_substrate_ethereum_receipts(header.extra()),
-		);
-
-		rialto_runtime::Call::BridgeKovan(pallet_call)
+	/// Used to build a `Call` for importing an unsigned header to a Substrate runtime.
+	pub fn build_unsigned_header_call(&self, header: QueuedEthereumHeader) -> Call {
+		match *self {
+			Self::TestPoA => {
+				let pallet_call = rialto_runtime::BridgeEthPoACall::import_unsigned_header(
```
tomusdrw, line 85: Same here, the `pallet_call` seems the duplicated.

relays/ethereum/src/ethereum_deploy_contract.rs
```diff
@@ -102,11 +100,11 @@ pub fn run(params: EthereumDeployContractParams) {
...
 /// Prepare initial header.
 async fn prepare_initial_header(
-	sub_client: &SubstrateRpcClient,
+	sub_client: &SubstrateClient<Rialto>,
 	sub_initial_header: Option<Vec<u8>>,
-) -> Result<(SubstrateHeaderId, Vec<u8>), String> {
+) -> Result<(RialtoHeaderId, Vec<u8>), String> {
 	match sub_initial_header {
-		Some(raw_initial_header) => match SubstrateHeader::decode(&mut &raw_initial_header[..]) {
+		Some(raw_initial_header) => match rialto_runtime::Header::decode(&mut &raw_initial_header[..]) {
```
tomusdrw, line 107: Why not keep all the types in a single type (I'm thinking `Rialto` which implements `Chain`)? That would simplfy migration to a fully generic code in the future.

relays/substrate-client/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+
+pub use crate::chain::{Chain, TransactionSignScheme};
+pub use crate::client::Client;
+pub use crate::error::{Error, Result};
+
+#[cfg(feature = "millau")]
+pub use crate::millau::Millau;
+
+#[cfg(feature = "rialto")]
+pub use crate::rialto::Rialto;
```
tomusdrw, line 40: Isn't it possible to split this into it's own crates? Instead of activating a feature, the crate would simply have two deps: `substrate-client` and the runtime it uses. If the crate is generic it could just import `substrate-client` and then accept any `Chain` object.
Mid-term I think it would make sure we don't hardcode even more runtime-specific things into the relay code.

relays/substrate-client/src/rpc.rs
```diff
@@ -0,0 +1,43 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! The most generic Substrate node RPC interface.
+
+// The compiler doesn't think we're using the
+// code from rpc_api!
+#![allow(dead_code)]
```
tomusdrw, line 21: I supposed it's only in the case where none of the features is enabled (i.e. we are just building the crate from the workspace). Splitting to crates, as suggested in another comment, should solve that.

svyatonik: No, it's for some internal enum that is created by jsonrpsee::rpc_api macro

relays/substrate-client/src/lib.rs
svyatonik: It is possible - I just wanted to avoid too many crates. So idea is that we have basic crate with basic definitions + for basic chain that we use. Everyone else may implement their own `trait Chain` to get `Client` functionality. Happy to split it if you think it's better.

> Mid-term I think it would make sure we don't hardcode even more runtime-specific things into the relay code.

I don't get it - how split will help? :)

relays/ethereum/src/instances.rs
svyatonik: I will try to revert this change - I can't remember why exactly I had to change that :)

svyatonik: Having completely runtime-agnostic (you have similar concerns here and in other comments as well) relay is possible - it's just not a part of this PR. I've extracted minimal client mostly to avoid duplicating RPC code.

Even though we have had `trait BridgeInstance`, it was returning `rialto_runtime::Call` :) So having enum imo is the same level of extensibility :) But as I said I'll try to revert this change - I'm not sure why I need to do that :)

relays/ethereum/src/ethereum_deploy_contract.rs
svyatonik: Most of required types are already there. This PR isn't about making existing relay generic - it is about making some RPC-related code generic so that we may reuse the same code in sub2sub. We have separate issue #61 for making this relay more or less generic.

relays/ethereum/src/instances.rs
svyatonik: Reverted

relays/ethereum/src/instances.rs
svyatonik: Reverted

relays/ethereum/src/instances.rs
svyatonik: Reverted.

relays/ethereum/src/ethereum_sync_loop.rs
```diff
@@ -158,19 +159,27 @@ impl SourceClient<EthereumHeadersSyncPipeline> for EthereumHeadersSource {
...
 struct SubstrateHeadersTarget {
 	/// Substrate node client.
-	client: SubstrateRpcClient,
+	client: SubstrateClient<Rialto>,
 	/// Whether we want to submit signed (true), or unsigned (false) transactions.
 	sign_transactions: bool,
 	/// Substrate signing params.
-	sign_params: SubstrateSigningParams,
+	sign_params: RialtoSigningParams,
+	/// Bridge instance used in eth2sub sync.
```
HCastano, line 167:
```suggestion
	/// Bridge instance used in Ethereum to Substrate sync.
```

HCastano (commented):
I'm gonna take another look tomorrow and try running the relay with the new changes. I've got mixed feelings about how some of the stuff got moved and renamed, but I need to think about it a bit more before I give any intelligent feedback.

I think that it would be okay to have two relay binaries: `ethereum-relay` and `substrate-relay`. The Ethereum relay would be used for Eth2Sub, Sub2Eth and ideally support different instances. The Substrate relay would be for Sub2Sub and be configurable again through some instances flag.

HCastano:
> There's another small change in our fork of jsonrpsee. This commit allows specifying using associated types in RPC methods. I.e.:
> 
> jsonrpsee::rpc_api! {
> 	pub(crate) Substrate<C: Chain> {
> 		#[rpc(method = "chain_getHeader", positional_params)]
> 		fn chain_get_header(block_hash: Option<C::Hash>) -> C::Header;
>         }
> }
> 
> Without it, I had to specify all used types explicitly, instead of using 'types container' (Chain).

Is this something that should/can be merged upstream?

relays/ethereum/src/instances.rs
```diff
@@ -44,9 +44,9 @@ pub trait BridgeInstance: Send + Sync + std::fmt::Debug {
 
 /// Corresponds to the Rialto instance used in the bridge runtime.
 #[derive(Default, Clone, Debug)]
-pub struct Rialto;
+pub struct TestPoA;
```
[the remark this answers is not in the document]
svyatonik: To avoid confusion - there's type with the same `substrate-client`.

svyatonik:
> Is this something that should/can be merged upstream?

No

relays/ethereum/src/instances.rs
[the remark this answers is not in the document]
HCastano: I'd prefer to change the type in `substrate-client` to `RialtoRuntime` and keep this instance type as `Rialto`

HCastano:
> > Is this something that should/can be merged upstream?
> 
> No

Why not?

svyatonik:
> Why not?

In general, yes - I want this functionality in `jsonrpsee` or whatever we want to use. In practice: (1) I haven't checked if there any caveats - I've just changed macro to work in our case (2) this is change that's made on top of other change that we don't want to backport (3) jsonrpsee seems to have no owner (aka unmaintained) atm.

relays/ethereum/src/instances.rs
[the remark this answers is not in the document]
svyatonik: I'll change this to RialtoPoA for now - if you'll decide to use something else, it could be changed. So now (in near future) we'll have 3 chains:
- Millau-Sub - chain that has all modules for sub2sub sync with Rialto-Sub;
- Rialto-Sub - chain that has all modules for sub2sub sync with Millau-Sub and all modules to sync with Rialto-PoA;
- Rialto-PoA - Etherum PoA chain.

For simplicity, I suggest to use Millau/Rialto both to identify bridges and their 'main' substrate-based chains. Other chains should be referenced with some suffix - e.g. RialtoPoA. Or we may use some names as you've suggested, but it isn't a part of this PR.

tomusdrw (approved):
lgtm, although would still prefer crates than conditional compilation.

relays/substrate-client/src/lib.rs
svyatonik: > To make sure everything actually works properly you should run multiple separate builds with different combination of feature flags

Ok, this makes sense - I'll extract to separate crates then

primitives/ethereum-poa/src/lib.rs
```diff
@@ -549,7 +549,7 @@ pub fn step_validator<T>(header_validators: &[T], header_step: u64) -> &T {
 
 sp_api::decl_runtime_apis! {
 	/// API for querying information about headers from the Rialto Bridge Pallet
-	pub trait RialtoHeaderApi {
+	pub trait RialtoPoAHeaderApi {
```
tomusdrw, line 552: Luckily we don't have the bridge deployed anywhere :) BTW why do we hardcode the bridge name in the runtime API? This feels like it could be something more generic like: "PoAHeaderApi" - just curious, no changes required here.

relays/substrate-client/Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+millau-runtime = { path = "../../bin/millau-runtime", optional = true }
+rialto-runtime = { path = "../../bin/rialto-runtime", optional = true }
+
+# Substrate Dependencies
+
+frame-system = { version = "2.0.0-rc6", tag = 'v2.0.0-rc6', git = "https://github.com/paritytech/substrate.git" }
+pallet-transaction-payment = { version = "2.0.0-rc6", tag = 'v2.0.0-rc6', git = "https://github.com/paritytech/substrate.git", optional = true }
+sp-core = { version = "2.0.0-rc6", tag = 'v2.0.0-rc6', git = "https://github.com/paritytech/substrate.git" }
+sp-keyring = { version = "2.0.0-rc6", tag = 'v2.0.0-rc6', git = "https://github.com/paritytech/substrate.git", optional = true }
+sp-runtime = { version = "2.0.0-rc6", tag = 'v2.0.0-rc6', git = "https://github.com/paritytech/substrate.git" }
```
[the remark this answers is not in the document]
tomusdrw: Until we have an automated check & formatting tool for this, I'm reluctant to address that. We have `cargo fmt` to avoid biases towards formatting in `.rs` files, and I really think we should do that for `.toml` as well, otherwise it's pretty futile.
Added an issue for this: https://github.com/paritytech/parity-bridges-common/issues/378

relays/substrate-client/src/lib.rs
svyatonik: done

relays/substrate-client/src/chain.rs
```diff
@@ -0,0 +1,87 @@
...
+};
+use sp_std::fmt::Debug;
+
+/// Substrate-based chain from minimal relay-client point of view.
+pub trait Chain {
+	/// The block number type used by the runtime.
+	type BlockNumber: Parameter
+		+ Member
+		+ MaybeSerializeDeserialize
+		+ Debug
```
tomusdrw, line 34: It seems that some of these bounds are already part of `Parameter` or `Member`. Do we really need them explicitly?

relays/substrate-client/src/client.rs
```diff
@@ -0,0 +1,136 @@
...
+/// Substrate client type.
+pub struct Client<C: Chain> {
+	/// Substrate RPC client.
+	client: RpcClient,
+	/// Genesis block hash.
+	genesis_hash: C::Hash,
+}
+
+#[cfg(feature = "std")]
+impl<C: Chain> std::fmt::Debug for Client<C> {
```
tomusdrw, line 44: Why can't we derive it and enforce `Chain` to have `Debug` as well?

relays/substrate-client/src/client.rs
```diff
@@ -0,0 +1,136 @@
...
+
+/// Substrate client type.
+pub struct Client<C: Chain> {
+	/// Substrate RPC client.
+	client: RpcClient,
+	/// Genesis block hash.
+	genesis_hash: C::Hash,
+}
+
+#[cfg(feature = "std")]
```
tomusdrw, line 43: I don't think this crate is `no_std` compatible, don't see a feature like that.
```suggestion
```

tomusdrw (approved):

relays/substrate-client/src/chain.rs
svyatonik: It's a plain copy of `frame_system::Trait`. Actually we don't need all of these trait, but will probably need some of these later.

primitives/ethereum-poa/src/lib.rs
svyatonik: Because: (1) we have two instances of PoA bridge module in the runtime (2) API is implemented for runtime => we can't implement the same API twice => we need to include instance name into API.

relays/substrate-client/src/client.rs
svyatonik: `jsonrpsee::Client` is not `Debug`

HCastano (approved):
Looks good! I like the crates more than having things feature flagged. The naming thing I brought up can definitely be changed later.

HCastano:
CI failures related to #379. I don't want to be too hasty with the merge, so I'd rather wait until we fix our CI to get this in.

PR #372, opened 2020-09-23
Split Rialto and Millau Nodes

HCastano:
While trying to integrate the Substrate pallet into a node for testing I realized that it would be a bit messy since we'd have to feature flag some things in the `chain_spec` to prevent duplicate imports from the Rialto runtime and the Millau runtimes. To avoid this I'm proposing a change where each node can be built standalone using the Cargo `--package` flag. This is nicer than adding feature flags in the code and also passing them in the CLI (since we have to specify the `--manifest-path` to build a specific node).


HCastano:
I've added a new `node` strategy to the `check-test-build` matrix instead of just copying the benchmark check for each node. I wonder if this will kill our CI time...

svyatonik (approved):

tomusdrw (commented):
Ugh, that's a net change of +800 lines and it seems a bunch of things is duplicated between these two nodes (means twice as much updates in say `service.rs` file with every substrate update). Isn't there a way to still extract and share the common part?

As suggested in one of Slava's PRs I'd be up to have everything generic over some `Chain` implementation - instead of importing `rialto_runtime::{Block, etc}` you simply have `Rialto: Chain` and then get things from this one type via `BlockOf<Chain>`.

tomusdrw (changes requested):
decided to request some change after looking closer at the two `chain_spec.rs` files, they are basically identical, the only difference is the `rialto_runtime` vs `millau_runtime` imports.

HCastano:
> decided to request some change after looking closer at the two `chain_spec.rs` files, they are basically identical, the only difference is the `rialto_runtime` vs `millau_runtime` imports.

They look basically identical now because we haven't hooked up the Substrate pallet at all into the runtime. They will start to diverge once that happens.

However, I agree that we need a way to split some of the common code out, so I'll take and look and see how I can improve this.

HCastano:
I think that making this more like Polkadot, in the sense that you can pick your chain using CLI options, would be ideal. Not sure if I'll have time to do this week but we'll see.

svyatonik:
I'm not working on this right now => can't raise any arguments here. But at some point in the future I want to make Millau and Rialto to use different crypto/hashers/primitives/RPCs just to show that relay/pallet is able to bridge heterogeneous chains. Imo in this case, two separate nodes binaries are better than one. Polkadot and Kusama are twins && will always stay the same. The future of Rialto and Millau is not set yet imo.

HCastano:
Slava convinced me that it's worthwhile to continue on the two separate node approach, at least for now. I've moved some stuff around so the folder structure is now like this:

```
bin
├── millau
│  ├── node
│  └── runtime
└── rialto
   ├── node
   └── runtime
```

As far as Tomek's concerns go, we might be able to share code between `cli.rs`, `command.rs`, and `chain_spec.rs`, but I think it's too early to do that. Maybe the code won't change much now that Substrate is a bit more stable, but we'll need to wait a bit to see that play out. We might also decide to customize the options for each node a bunch, as was hinted by Slava, in which case it wouldn't be worth it to combine them.

If you have a strong opinion about it though I'm open to working towards the shared code.

svyatonik (approved):
You've removed `scripts/init.sh` because we don't need it anymore?

HCastano:
> You've removed `scripts/init.sh` because we don't need it anymore?

Yeah I don't think it's used at all.

tomusdrw (approved):
> As far as Tomek's concerns go, we might be able to share code between cli.rs, command.rs, and chain_spec.rs, but I think it's too early to do that. 

I completely disagree with this reasoning, but don't want to block this PR. It's much easier to duplicate code than to de-duplicate it. If we start off without shared crate (`common`) it's a huge hassle to bring one up later on (a lot of unrelated changes and just big upfront mental "cost"), so having things duplicated in the beginning legitimizes more duplication in the future.

PR #374, opened 2020-09-23
Remove Cargo Check with Nightly Toolchain

HCastano:
I recently learned that [Clippy runs `cargo check` explicitly](https://stackoverflow.com/questions/57449356/is-cargo-clippy-a-superset-of-cargo-check) as part of its execution. This means we can get rid of our own `cargo check` runs. Well, actually only the one on `+nightly` since we can't run Clippy on stable in our repo.

Also, since it's not totally clear from the diff I moved the Clippy CI stage right after our `cargo check` stage, I didn't move the test stage at all.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -30,6 +30,7 @@ jobs:
         - /cache:/cache
     env:
       RUST_BACKTRACE:              full
+      SKIP_WASM_BUILD:             1
       SCCACHE_REDIS:               redis://172.17.0.1/0
       SCCACHE_IDLE_TIMEOUT:        0
       CMAKE_C_COMPILER_LAUNCHER:   sccache
@@ -70,8 +71,6 @@ jobs:
         if:                        matrix.toolchain == 'nightly'
         run:                       rustup component add clippy --toolchain ${{ matrix.toolchain }}
       - name:                      Clippy
-        env:
-          SKIP_WASM_BUILD:         1
         uses:                      actions-rs/cargo@master
         if:                        matrix.toolchain == 'nightly'
         with:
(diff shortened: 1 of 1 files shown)


HCastano:
Umm I'm not sure why Clippy is complaining about our dependencies now. If any of you guys know feel free to push a fix. If not I'll take a deeper look tomorrow.

.github/workflows/rust.yml
```diff
@@ -81,6 +76,14 @@ jobs:
           command:                 clippy
           toolchain:               ${{ matrix.toolchain }}
           args:                    --all-targets -- -D warnings
+
+## Test stage
+      - name:                      Testing rust-${{ matrix.toolchain }}
```
tomusdrw, line 81: The reason for tests being before clippy was to get quicker feedback about failing tests instead of getting a bunch of annoying `clippy` complaints :) But I guess we will still get that from the `stable` job.

tomusdrw (approved):
lgtm!

tomusdrw:
Ugh, clippy is complaining about `wasm-builder` project :/ Can we run `cargo check -p wasm-builder` before running clippy to maybe prevent that? Or maybe we should run clippy with `SKIP_WASM_BUILD=1`?

General-Beck (approved):

.github/workflows/rust.yml
HCastano: I think it depends on which pipeline complains first. I suspect Clippy may be fast to fail than the tests (since we need to wait for _everything_ to compile first). If we find this annoying in our workflow we can change the order back to the original one

HCastano:
Testing it locally and it fails when run with `-D warnings`. I think running `cargo +nightly check` beforehand avoided the problem since Clippy wasn't encountering these "bad" crates in its first run.

I don't fully understand this yet, but I'll take another look tomorrow 🙃

HCastano:
So it looks like having `cargo +nightly check` run first is fine since all the build artifacts are cached anyways. When we get around to running `cargo +nightly clippy` it takes a really short amount of time (like less than a minute) since it's not re-checking stuff. This change isn't helpful at all, so I'm going to close the PR.

PR #375, opened 2020-09-23
Integrate Grandpa Proof Checker into Substrate Pallet

### LINKED ISSUE #370: Plug Grandpa Finality Checker into Substrate Pallet
The Substrate pallet isn't actually verifying Grandpa justifications right now. To do this we should use the code in `modules/substrate/src/justifications.rs`. This integration should be carefully tested as we don't want to accept invalid justifications.

One thing worth exploring here is whether or not we can move the justification code into a primitives crate (eg. `bp-substrate`) and use it in different parts of the project, like the Ethereum built-in crate.

Related to: #296.

HCastano:
Wires up the `justification` module from #353 into the Substrate pallet. It also does a little bit of cleanup with the primitives crate and some of the testing utilities.

Kinda closes #370. I say kinda because in that issue I mention that we should have extra testing. After thinking about it a bit it looks like Slava's tests cover pretty reasonable scenarios, and testing them in the `verifier` code would just end up duplicating them but with a lot more setup.

As for having a justification verification crate mentioned in #370, I'll do that on Monday.

This is ready for an initial look though.


modules/substrate/src/verifier.rs
```diff
@@ -176,10 +184,18 @@ where
 		}
 
 		let current_authority_set = self.storage.current_authority_set();
-		let is_finalized = check_finality_proof(&header, &current_authority_set, &proof.0);
-		if !is_finalized {
-			return Err(FinalizationError::UnfinalizedHeader);
-		}
+		let voter_set = VoterSet::new(current_authority_set.authorities).expect(
```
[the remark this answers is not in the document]
svyatonik: (probably add test for this case?)

modules/substrate/src/justification.rs
```diff
@@ -230,26 +207,44 @@ mod tests {
...
+		let mut precommits = vec![];
+		let mut votes_ancestries = vec![];
+
+		// I'm using the same header for all the voters since it doesn't matter as long
+		// as they all vote on blocks _ahead_ of the one we're interested in finalizing
+		for (id, _weight) in authorities {
+			let signer = extract_keyring(&id);
+			let precommit = signed_precommit(signer, header_id(header_index + 1), round, set_id);
+			precommits.push(precommit);
+			votes_ancestries.push(test_header((header_index + 1).into()));
```
svyatonik, line 226: You never change `header_index` => are you sure that you need to duplicate same header in `votes_ancestries` 3 times? Not an issue, though.

svyatonik (approved):

modules/substrate/src/lib.rs
```diff
@@ -31,43 +31,95 @@
...
 
 #[cfg(test)]
 mod mock;
 
-type Hash<T> = <T as HeaderT>::Hash;
-type Number<T> = <T as HeaderT>::Number;
-
-pub trait Trait: frame_system::Trait {}
+pub trait Trait: frame_system::Trait {
+	/// A type that fulfulls the abstract idea of what a Substrate header is.
```
tomusdrw, line 54:
```suggestion
	/// A type that fulfills the abstract idea of what a Substrate header is.
```

modules/substrate/src/lib.rs
```diff
@@ -31,43 +31,95 @@
...
+	// See here for more info:
+	// https://crates.parity.io/sp_runtime/traits/trait.Header.html
+	type BridgedHeader: HeaderT<Number = Self::BridgedBlockNumber, Hash = Self::BridgedBlockHash>;
+
+	/// A type that fulfills the abstract idea of what a Substrate block number is.
+	// Constraits come from the associated Number type of `sp_runtime::traits::Header`
+	// See here for more info:
+	// https://crates.parity.io/sp_runtime/traits/trait.Header.html#associatedtype.Number
+	//
+	// Note that the `AsPrimitive<usize>` trait is required by the Grandpa justification
```
tomusdrw, line 64: Why is it `usize`? We shouldn't be using that in the runtime, should be at least `u64`.

modules/substrate/src/lib.rs
```diff
@@ -31,43 +31,95 @@
...
+		+ sp_std::hash::Hash
+		+ Ord
+		+ Copy
+		+ MaybeDisplay
+		+ Default
+		+ SimpleBitOps
+		+ Codec
+		+ AsRef<[u8]>
+		+ AsMut<[u8]>
+		+ MaybeMallocSizeOf
```
tomusdrw, line 94: A bunch of these traits is already part of `Member`. There is also `Parameter` trait in the substrate repo, do we really need to list all of them here?

modules/substrate/src/lib.rs
```diff
@@ -31,43 +31,95 @@
...
+		+ EncodeLike;
+
+	/// A type that fulfills the abstract idea of what a Substrate hasher (a type
+	/// that produces hashes) is.
+	// Constraits come from the associated Hashing type of `sp_runtime::traits::Header`
+	// See here for more info:
+	// https://crates.parity.io/sp_runtime/traits/trait.Header.html#associatedtype.Hashing
+	type BridgedBlockHasher: HashT;
```
tomusdrw, line 102: Don't we need to enforce `Output` to be `Self::BridgedBlockHash`?

modules/substrate/src/verifier.rs
```diff
@@ -33,6 +35,7 @@ use sp_std::{prelude::Vec, vec};
 ///
 /// For a Substrate based chain using Grandpa this will
 /// be an encoded Grandpa Justification.
+#[derive(Debug)]
```
tomusdrw, line 38: Should be `RuntimeDebug` from `frame_support`

modules/substrate/src/lib.rs
HCastano: I didn't make the rules, I just follow them (trait definition [here](https://github.com/paritytech/finality-grandpa/blob/master/src/lib.rs#L169))

modules/substrate/src/lib.rs
HCastano: I wasn't sure if there was already a trait in Substrate that captured this. I think Slava did a similar thing in a recent relay refactor PR. I can take a look at `Member` and `Parameter` though

modules/substrate/src/justification.rs
HCastano: I was under the impression that we needed one header per authority for the vote ancestries. Looks like the tests pass if I just have one header in there though. I'll need to think about what more closely matches what Grandpa does

modules/substrate/src/verifier.rs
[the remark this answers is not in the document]
HCastano: > This proof is invalid - you never check that the set is correct.

Then the proof panicking would've done it's job to indicate a bug :P Yep I'll do it

modules/substrate/src/lib.rs
HCastano: While we can clean up _some_ trait bounds which are duplicated in both `Member` and `Parameter` I don't think it's worth it. We explicitly point the reader to the trait bounds in `sp_runtime::traits::Header` since that's where they were taken from and we should be consistent with that.

In the future though we definitely should move all the traits bounds to a more central location where we can hide it behind a single trait.

HCastano:
I did a little bit of work towards moving the justification code into a module within `primitives/substrate` but I didn't really like how the test code was being shared between different places. I decided to revert those commits.

PR is ready for another look.

svyatonik (approved):

modules/substrate/src/lib.rs
tomusdrw: `sp_runtime::traits::Header` just has excessive bounds, probably no one really took a look at cleaning them up at some point, when some of the traits got merged into `Member`.

IMHO repeating the same bounds multiple time is just super dumb (basically like writing `Clone + Clone`), and the fact that they are repeated in substrate codebase is not a compelling argument for that either. That said I'd rather get the PR merged than go back and forth about it:)

modules/substrate/src/verifier.rs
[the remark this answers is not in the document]
tomusdrw: The proof is still incorrect. The fact we have a bug is obvious (since we panicked). The proof should include a bunch of statements that indicate why it won't happen. Something like: `"VoterSet correctness is verified during import, before writing to storage; qed"`

tomusdrw (approved):
lgtm!

PR #383, opened 2020-09-28
Split payload types

svyatonik:
as noted in #https://github.com/paritytech/parity-bridges-common/pull/339#issuecomment-690089390, this solves following problems:
1) on delivery/dispatch level, `Payload` is split into opaque `MessagePayload` and decodable `DispatchPayload`. Now we are able to deliver messages with improperly encoded payload. E.g. if payload is `Call` and `Call` format has been changed on the target chain, then only dispatch will fail, but the message will be marked as delivered;
2) on module/API level, `Payload` and `Fee` types are split into `Inbound` and `Outbound` payload/fee accordingly. E.g. if payload is `Call`, then it's different for outbound and inbound messages, because `Call`s are dispatched on different chains.


modules/message-lane/src/lib.rs
```diff
@@ -238,6 +243,9 @@ decl_module! {
 				Error::<T, I>::InvalidMessagesProof
 			})?;
 
+			// try to decode message payloads
+			let messages: Vec<_> = messages.into_iter().map(Into::into).collect();
```
tomusdrw, line 247: With #384 implemented, we should move the decoding part after weight lower bound checking and add an extra step to compare the declared weight with the actual one.

modules/message-lane/src/mock.rs
```diff
@@ -210,17 +213,36 @@ impl SourceHeaderChain<TestPayload, TestMessageFee> for TestSourceHeaderChain {
...
-impl MessageDispatch<TestPayload, TestMessageFee> for TestMessageDispatch {
-	fn dispatch_weight(message: &Message<TestPayload, TestMessageFee>) -> Weight {
-		message.data.payload.1
+impl MessageDispatch<TestMessageFee> for TestMessageDispatch {
+	type DispatchPayload = TestPayload;
+
+	fn dispatch_weight(message: &DispatchMessage<TestPayload, TestMessageFee>) -> Weight {
+		match message.data.payload.as_ref() {
+			Ok(payload) => payload.1,
+			Err(_) => 0,
```
tomusdrw, line 222: I think in a production implementation we might still want to charge for the byte-length of the payload. Without that you may send a message that you know would fail to be decoded and have the bytes included in the chain cheaply (especially if we do refunds).

primitives/message-lane/src/target_chain.rs
```diff
@@ -37,20 +56,43 @@ pub trait SourceHeaderChain<Payload, Fee> {
...
 
 	/// Called when inbound message is received.
 	///
 	/// It is up to the implementers of this trait to determine whether the message
 	/// is invalid (i.e. improperly encoded, has too large weight, ...) or not.
-	fn dispatch(message: Message<Payload, Fee>);
+	fn dispatch(message: DispatchMessage<Self::DispatchPayload, Fee>);
+}
+
+impl<DispatchPayload: Decode, Fee> From<Message<Fee>> for DispatchMessage<DispatchPayload, Fee> {
```
tomusdrw, line 82: I'd put these closer to the `struct` definitions, but that's just the personal taste.

modules/message-lane/src/lib.rs
```diff
@@ -60,41 +61,45 @@ pub trait Trait<I = DefaultInstance>: frame_system::Trait {
...
 
+	/// Payload type of outbound messages. This payload is dispatched on the bridged chain.
+	type OutboundPayload: Parameter;
+	/// Message fee type of outbound messages. This fee is paid on this chain.
+	type OutboundMessageFee: Parameter;
+
+	/// Payload type of inbound messages. This payload is dispatched on this chain.
+	type InboundPayload: Decode;
+	/// Message fee type of inbound messages. This fee is paid on the bridged chain.
+	type InboundMessageFee: Decode;
```
tomusdrw, line 78: We only need this to be able to decode the messages, right? I.e. it's not really used on the target chain?

tomusdrw (approved):
lgtm modulo #384 which is logged separately

modules/message-lane/src/mock.rs
svyatonik: I'm not sure I understand. The `delivery_and_dispatch_fee` is paid by `submitter` in advance on the source chain. This function is called on the **target chain** when `relayer` submits deivery-transaction. The fee is already paid => bytes are already 'included' in the target chain.

Probably you mean that we should check that on the source chain instead?

modules/message-lane/src/lib.rs
svyatonik: It isn't used now. But I had an idea (see "Some ideas for future PRs" of #339) on how to use it to build something like exchange rate between bridged chains weights/fee tokens. I.e. on the target chain you know:
1) weight `W` of the message;
2) fee `TF = WeightToFee(W)` in `TargetFeeTokens` that needs to be paid for this message;
3) fee `SF` in `SourceFeeTokens` that has been paid for delivery-and-dispatch of this message.

So if you have enough samples, you could (roughly) estimate conversion rate between `SourceFeeTokens` and `TargetFeeTokens` like that `ConversionRate = SourceFeeTokens / TargetFeeTokens`. This information could be used later in `LaneMessageVerifier::verify_message` on **this** chain to check if fee paid by the submitter is enough. So we may start (i.e. configure verifier) with some fixed rate and then it'll start slowly changing towards actual market rate.

So yes - not used now, but maybe used later.

HCastano (approved):

PR #385, opened 2020-09-29
Reward relayers for dispatching messages

svyatonik:
related #316
follow up of #339 (II)

This PR introduces scheme which rewards relayers that have delivered messages to the target chain. There are no any transfers - reward logic is hidden behind `MessageDeliveryAndDispatchPayment` trait, so both "the winner takes all" and election scheme from #316 can be applied.

Main change is that inbound lane state now holds set of relayers which have delivered last messages. Since proof-of-delivery is that state, the set is included into the "delivery race" and is relayed back to the source chain.

The main drawback of this approach is that if there are too may relayed messages, this set may grow indefinitely. This is solved by introducing additional config param `MaxUnconfirmedMessagesAtInboundLane`. If relayers are not confirming delivery back to the source chain for too long, and there are `MaxUnconfirmedMessagesAtInboundLane` unconfirmed messages in the inbound lane, then it stops accepting new messages && waits until proof-of-outbound-lane-state will be included into the messages-proof. When updated outbound lane state is received, all rewarded relayers ids are removed from the relayers set. This would require minimal changes into "delivery race" logic of the relayer - I'll add those changes in separate PR.


modules/message-lane/src/mock.rs
```diff
@@ -176,13 +212,20 @@ impl TestMessageDeliveryAndDispatchPayment {
...
+	/// Returns true if given fee has been paid by given submitter.
 	pub fn is_fee_paid(submitter: AccountId, fee: TestMessageFee) -> bool {
 		frame_support::storage::unhashed::get(b":message-fee:") == Some((submitter, fee))
 	}
+
+	/// Returns true if given relayer has been rewarded with given balance. The reward-paid flag is
+	/// cleared after the call.
+	pub fn is_reward_paid(relayer: AccountId, fee: TestMessageFee) -> bool {
+		let key = (b":relayer-reward:", relayer, fee).encode();
+		frame_support::storage::unhashed::take::<bool>(&key).is_some()
```
HCastano, line 224: Is using functionality from the `unhashed` module only useful for testing, or does it have applications beyond that?

HCastano (approved):
Looks fine, but I'd wait until Tomek takes a look before merging

modules/message-lane/src/lib.rs
```diff
@@ -262,13 +278,31 @@ decl_module! {
...
+							target: "runtime",
+							"Received lane {:?} state update: latest_confirmed_nonce={}",
+							lane_id,
+							updated_latest_confirmed_nonce,
+						);
+					}
+				}
+
+				for message in lane_data.messages {
+					debug_assert_eq!(message.key.lane_id, lane_id);
```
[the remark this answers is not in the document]
svyatonik: Can remove. But changing to something that'll work in release isn't desired, because this `group_by` is actually happens in the same pallet function.

primitives/message-lane/src/target_chain.rs
```diff
@@ -16,11 +16,23 @@
...
+use crate::{LaneId, Message, MessageData, MessageKey, OutboundLaneData};
 
-use codec::{Decode, Error as CodecError};
+use codec::{Decode, Encode, Error as CodecError};
 use frame_support::{weights::Weight, Parameter, RuntimeDebug};
-use sp_std::{fmt::Debug, prelude::*};
+use sp_std::{collections::btree_map::BTreeMap, fmt::Debug, prelude::*};
+
+/// Proved messages from the source chain.
+pub type ProvedMessages<Message> = BTreeMap<LaneId, ProvedLaneMessages<Message>>;
```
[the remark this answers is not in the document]
svyatonik: When you relay messages to the target chain, they aren't simply encoded messages. Instead we relay storage-proof of these messages. `ProvedMessages` are the messages that are proved by that proof (of proved to be in that proof). If you have any other identifier idea - please suggest.

modules/message-lane/src/mock.rs
svyatonik: It is also used for some keys that are shared by all runtimes (aka 'well_known_keys') - i.e. ":code", ":max_heap_pages", ...

primitives/message-lane/src/lib.rs
```diff
@@ -70,14 +70,34 @@ pub struct Message<Fee> {
 }
 
 /// Inbound lane data.
-#[derive(Default, Encode, Decode, Clone, RuntimeDebug, PartialEq)]
-pub struct InboundLaneData {
+#[derive(Encode, Decode, Clone, RuntimeDebug, PartialEq, Eq)]
+pub struct InboundLaneData<RelayerId> {
+	/// Identifiers of relayers and messages that they have delivered (ordered by message nonce).
+	/// It is guaranteed to have at most N entries, where N is configured at module level. If
```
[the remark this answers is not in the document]
svyatonik: I've added another small comment to the code

modules/message-lane/src/inbound_lane.rs
```diff
@@ -57,7 +87,13 @@ impl<S: InboundLaneStorage> InboundLane<S> {
 			return false;
 		}
 
+		// if there are more unconfirmed messages that we may accept, reject this message
```
tomusdrw, line 90:
```suggestion
		// if there are more unconfirmed messages than we may accept, reject this message
```

primitives/message-lane/src/target_chain.rs
```diff
@@ -49,14 +61,35 @@ pub trait SourceHeaderChain<Fee> {
...
+	) -> Result<ProvedMessages<DispatchMessage<DispatchPayload, Fee>>, Self::Error> {
+		Self::verify_messages_proof(proof).map(|messages_by_lane| {
+			messages_by_lane
+				.into_iter()
+				.map(|(lane, lane_data)| {
+					(
+						lane,
+						ProvedLaneMessages {
+							lane_state: lane_data.lane_state,
+							messages: lane_data.messages.into_iter().map(Into::into).collect(),
```
tomusdrw, line 86: We don't verify here that `messages` are actually coming from the same `lane`. While in theory it shouldn't happen, having this data duplicated may just lead to mistakes. Maybe let's change `Message/DispatchMessage` to only contain `nonce` (instead of entire `key`) and add the lane information only during actual `receive_message`?

modules/message-lane/src/lib.rs
```diff
@@ -262,13 +278,31 @@ decl_module! {
...
+
+					total_messages += 1;
+					if lane.receive_message::<T::MessageDispatch>(relayer_id.clone(), message.key.nonce, message.data) {
+						valid_messages += 1;
```
tomusdrw, line 304: Why don't we reject the entire transaction here? Previously we would get `false` here in case there was a nonce mismatch, now it may also happen if there is simply too many unconfirmed deliveries. I feel this is not something that is super desirable, but at the same time can't find any significant problems with the current approach.

modules/message-lane/src/lib.rs
```diff
@@ -66,6 +66,11 @@ pub trait Trait<I = DefaultInstance>: frame_system::Trait {
 	/// confirmed. The reason is that if you want to use lane, you should be ready to pay
 	/// for it.
 	type MaxMessagesToPruneAtOnce: Get<MessageNonce>;
+	/// Maximal number of messages in the 'unconfirmed' state at inbound lane. Unconfirmed
+	/// message at inbound lane is the message that has been: sent, delivered and dispatched.
+	/// Its delivery confirmation is still pending. This limit is introduced to bound maximal
+	/// number of relayers-ids in the inbound lane state.
+	type MaxUnconfirmedMessagesAtInboundLane: Get<MessageNonce>;
```
tomusdrw, line 73: While this works I really doubt it's the right abstraction. It not only limits the number of "unconfirmed messages", but it actually also limits the number of messages you can dispatch in a single transaction (even though it's a single relayer).

A really simple optimisation would be to store `VecDeque<(StartNonce, Option<EndNonce>, Relayer)>` instead of duplicating the relayer for each nonce it delivered. We really expect the relayers to dispatch multiple messages at once, so IMHO it makes sense to optimize that. Thanks to this, if a relayer delivers a batch, it will only occupy one slot in that collection, so the `MaxUncofirmedMessagesAtInboundeLane` wouldn't restrict the message delivery that much.

modules/message-lane/src/lib.rs
tomusdrw: I was initially assuming that the delivery proof is going to be simply the transaction proof, but I've realised is that this makes the `verify_messages_delivery_proof` function harder to implement and potentially increase the size of the proof unnecessarily.
It seems that you assume that the delivery proof is actually going to be simply a storage proof, we don't necessarily care that much.

modules/message-lane/src/lib.rs
```diff
@@ -286,7 +320,7 @@ decl_module! {
 		#[weight = 0] // TODO: update me (https://github.com/paritytech/parity-bridges-common/issues/78)
 		pub fn receive_messages_delivery_proof(origin, proof: MessagesDeliveryProofOf<T, I>) -> DispatchResult {
 			let _ = ensure_signed(origin)?;
```
tomusdrw, line 322: We might consider incentivising this separately. Currently it's assumed that the relayer who delivered the messages is interested to get rewards. However, we should make sure that other relayers are not going to get blocked by him (which currently is the case).
We might split the total fee between the relayers who delivered a message and the submitter of the delivery proof. CCing #316 here.

tomusdrw (commented):
Looks good, few concerns in comments.

modules/message-lane/src/inbound_lane.rs
```diff
@@ -57,7 +87,13 @@ impl<S: InboundLaneStorage> InboundLane<S> {
...
+		if self.storage.max_unconfirmed_messages() == data.relayers.len() as MessageNonce {
```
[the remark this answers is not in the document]
svyatonik: Yes, it was original idea from #339. There were multiple problems with this approach. Can't remember all, since a week has already passed, but here are some examples:
1) if you have some sort of queue (store relayers of last 1024 messages), it'll mean that every confirmation transaction after 1024 messages is guaranteed to have these additional >32kb bytes, unless you'll find a way to prove that reward has been paid (i.e. what is presented in this PR);
2) if you're going to introduce periods in blocks instead (store relayers + number of messages that they have relayed in last 1024 blocks), it'll require additional storage entry + separate transaction type to deliver that. Also - size of confirmation transaction may grow indefinitely (because we don't limit number of messages/1024 blocks);
3) since confirmation transaction requires source chain to know about target block' finality, it means that reward requires: (3.a) GRANDPA authorities to finalize target chain blocks in timely fashion (3.b) header relayers to relay target chain blocks and finality to source chain and (3.c) message relayer to submit delivery transaction. So there are too many other players (GRANDPA authorities, headers relayers, other relayers, nodes txpools) that may block claiming reward;
4) if you're going to introduce claim-reward transaction for relayers that have missed their opportunity (i.e. GRANDPA authorities have failed to finalize target chain blocks in timely fashion), it would mean that you'll need some way to mark reward as claimed (i.e. store unclaimed rewards separately OR change how pruning works).

primitives/message-lane/src/source_chain.rs
```diff
@@ -87,11 +87,14 @@ pub trait LaneMessageVerifier<Submitter, Payload, Fee> {
 /// So to be sure that any non-altruist relayer would agree to deliver message, submitter
 /// should set `delivery_and_dispatch_fee` to at least (equialent of): sum of fees from (2)
 /// to (4) above, plus some interest for the relayer.
-pub trait MessageDeliveryAndDispatchPayment<AccountId, Balance> {
+pub trait MessageDeliveryAndDispatchPayment<SubmitterId, RelayerId, Balance> {
```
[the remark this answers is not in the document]
svyatonik: I think we may remove it.

primitives/message-lane/src/target_chain.rs
svyatonik: > We don't verify here that messages are actually coming from the same lane. While in theory it shouldn't happen, having this data duplicated may just lead to mistakes. Maybe let's change Message/DispatchMessage to only contain nonce (instead of entire key) and add the lane information only during actual receive_message?

`verify_messages_proof` could return anything it wants - it can e.g. replace all messages payloads with `BalancesCall::transfer(svyatonik, 1 DOT)`. So you just need to trust it.

We can't remove `lane` from the MessageKey, because in this way relayer can just submit messages from Lane1 to Lane2.

modules/message-lane/src/lib.rs
svyatonik: > Why don't we reject the entire transaction here? Previously we would get false here in case there was a nonce mismatch, now it may also happen if there is simply too many unconfirmed deliveries. I feel this is not something that is super desirable, but at the same time can't find any significant problems with the current approach.

Because, as it was previously, transactions may overlap. Tx1[msg1, msg2, msg3] from relayer1 may be mined earler than Tx2[msg1, msg2, msg3, msg4] from relayer2, but it still may bring useful data (msg4). Relayer has all means to determine when it need to include `proof-of(outbound_lane.state)` into delivery transaction.

modules/message-lane/src/lib.rs
svyatonik: > While this works I really doubt it's the right abstraction. It not only limits the number of "unconfirmed messages", but it actually also limits the number of messages you can dispatch in a single transaction (even though it's a single relayer).

If you care about possible number of mesages in a single transaction, make this at least `MaxBlockWeight / MaxExtrinsicWeight`. But in practice it should be much larger, or we will be blocked by GRANDPA authorities + headers relays.

> A really simple optimisation would be to store VecDeque<(StartNonce, Option<EndNonce>, Relayer)> instead of duplicating the relayer for each nonce it delivered. We really expect the relayers to dispatch multiple messages at once, so IMHO it makes sense to optimize that. Thanks to this, if a relayer delivers a batch, it will only occupy one slot in that collection, so the MaxUncofirmedMessagesAtInboundeLane wouldn't restrict the message delivery that much.

Yes - this may be done, but since it is only a potential optimization - let's file a separate issue for that.

modules/message-lane/src/lib.rs
svyatonik: > We might consider incentivising this separately. Currently it's assumed that the relayer who delivered the messages is interested to get rewards. However, we should make sure that other relayers are not going to get blocked by him (which currently is the case).

Other relayers always have an opportunity to unblock themselves. But I agree that incentivisation hook may be done here. Actually, probably adding `confirmation_transaction_submitter` to `pay_relayer_reward()` would be enough.

modules/message-lane/src/lib.rs
svyatonik: > MaxBlockWeight / MaxExtrinsicWeight

This is obviously wrong, sorry. Should be something like `MaxBlockWeight / Avg(MaxExtrinsicWeight)`. Nevertheless, it is still relevant when talking about messages/tx.

tomusdrw: Extracted as #406

primitives/message-lane/src/target_chain.rs
tomusdrw: My point was that this information is now duplicated. We already have lane id in the returned map, why do we need it in the message itself?

svyatonik: Ok - will remove from trait

svyatonik: done

modules/message-lane/src/lib.rs
```diff
@@ -331,24 +384,38 @@ fn outbound_lane<T: Trait<I>, I: Instance>(lane_id: LaneId) -> OutboundLane<Runt
...
 		self.lane_id
 	}
 
-	fn data(&self) -> InboundLaneData {
-		InboundLanes::<I>::get(&self.lane_id)
+	fn max_unconfirmed_messages(&self) -> MessageNonce {
+		T::MaxUnconfirmedMessagesAtInboundLane::get()
+	}
+
+	fn data(&mut self) -> InboundLaneData<T::InboundRelayer> {
```
tomusdrw, line 405: Couldn't we use like a `RefCell` to avoid adding `&mut` here?

tomusdrw (approved):
lgtm!

PR #386, opened 2020-09-29
Integrate Rialto <-> Millau message lanes into Millau/Rialto runtimes

svyatonik:
The integration code is mostly in `bin/runtime-common/src/messages.rs` with runtime-specific type aliases in `bin/millau/runtime/src/rialto_messages.rs` and `bin/rialto/runtime/src/millau_messages.rs`.

### DIFF
--- a/bin/millau/runtime/src/rialto_messages.rs
+++ b/bin/millau/runtime/src/rialto_messages.rs
@@ -54,7 +54,7 @@ pub struct WithRialtoMessageBridge;
 impl MessageBridge for WithRialtoMessageBridge {
 	const INSTANCE: InstanceId = *b"rlto";
 
-	const RELAYER_INTEREST_PERCENT: u32 = 10;
+	const RELAYER_FEE_PERCENT: u32 = 10;
 
 	type ThisChain = Millau;
 	type BridgedChain = Rialto;
--- a/bin/millau/runtime/src/rialto_messages.rs
+++ b/bin/millau/runtime/src/rialto_messages.rs
@@ -31,37 +31,37 @@ use frame_support::{
 };
 use sp_trie::StorageProof;
 
-/// Message payload for Rialto -> Millau messages.
-pub type ToMillauMessagePayload = messages::source::FromThisChainMessagePayload<WithMillauMessageBridge>;
-
-/// Message verifier for Rialto -> Millau messages.
-pub type ToMillauMessageVerifier = messages::source::FromThisChainMessageVerifier<WithMillauMessageBridge>;
-
 /// Message payload for Millau -> Rialto messages.
-pub type FromMillauMessagePayload = messages::target::FromBridgedChainMessagePayload<WithMillauMessageBridge>;
+pub type ToRialtoMessagePayload = messages::source::FromThisChainMessagePayload<WithRialtoMessageBridge>;
 
-/// Call-dispatch based message dispatch for Millau -> Rialto messages.
-pub type FromMillauMessageDispatch = messages::target::FromBridgedChainMessageDispatch<
-	WithMillauMessageBridge,
+/// Message verifier for Millau -> Rialto messages.
+pub type ToRialtoMessageVerifier = messages::source::FromThisChainMessageVerifier<WithRialtoMessageBridge>;
+
+/// Message payload for Rialto -> Millau messages.
+pub type FromRialtoMessagePayload = messages::target::FromBridgedChainMessagePayload<WithRialtoMessageBridge>;
+
+/// Call-dispatch based message dispatch for Rialto -> Millau messages.
+pub type FromRialtoMessageDispatch = messages::target::FromBridgedChainMessageDispatch<
+	WithRialtoMessageBridge,
 	crate::Runtime,
 	pallet_bridge_call_dispatch::DefaultInstance,
 >;
 
 /// Millau <-> Rialto message bridge.
 #[derive(RuntimeDebug, Clone, Copy)]
-pub struct WithMillauMessageBridge;
+pub struct WithRialtoMessageBridge;
 
-impl MessageBridge for WithMillauMessageBridge {
-	const INSTANCE: InstanceId = *b"mllu";
+impl MessageBridge for WithRialtoMessageBridge {
+	const INSTANCE: InstanceId = *b"rlto";
 
 	const RELAYER_INTEREST_PERCENT: u32 = 10;
 
-	type ThisChain = Rialto;
-	type BridgedChain = Millau;
+	type ThisChain = Millau;
+	type BridgedChain = Rialto;
 
 	fn maximal_dispatch_weight_of_message_on_bridged_chain() -> Weight {
 		// we don't want to relay too large messages + keep reserve for future upgrades
-		bp_millau::MAXIMUM_EXTRINSIC_WEIGHT / 2
+		bp_rialto::MAXIMUM_EXTRINSIC_WEIGHT / 2
 	}
 
 	fn weight_of_delivery_transaction() -> Weight {
@@ -91,19 +91,6 @@ impl MessageBridge for WithMillauMessageBridge {
 	}
 }
 
-/// Rialto chain from message lane point of view.
-#[derive(RuntimeDebug, Clone, Copy)]
-pub struct Rialto;
-
-impl messages::ChainWithMessageLanes for Rialto {
-	type AccountId = bp_rialto::AccountId;
-	type Signer = bp_rialto::AccountSigner;
-	type Signature = bp_rialto::Signature;
-	type Call = crate::Call;
-	type Weight = Weight;
-	type Balance = bp_rialto::Balance;
-}
-
 /// Millau chain from message lane point of view.
 #[derive(RuntimeDebug, Clone, Copy)]
 pub struct Millau;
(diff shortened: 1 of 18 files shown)


HCastano (commented):
I haven't finished looking at everything yet. Looks alright so far though

bin/rialto/runtime/src/millau_messages.rs
```diff
@@ -0,0 +1,157 @@
+// Copyright 2019-2020 Parity Technologies (UK) Ltd.
```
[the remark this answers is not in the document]
svyatonik: Right.

bin/runtime-common/src/messages.rs
```diff
@@ -0,0 +1,538 @@
...
+		) -> Result<(), Self::Error> {
+			let minimal_fee_in_bridged_tokens =
+				estimate_message_dispatch_and_delivery_fee::<B>(payload, B::RELAYER_INTEREST_PERCENT)
+					.ok_or("Overflow when computing minimal required message delivery and dispatch fee")?;
+
+			// compare with actual fee paid
+			let actual_fee_in_bridged_tokens =
+				B::this_chain_balance_to_bridged_chain_balance(*delivery_and_dispatch_fee);
+			if actual_fee_in_bridged_tokens < minimal_fee_in_bridged_tokens {
+				return Err("Too low fee paid");
```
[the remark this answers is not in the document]
svyatonik: It should be `&'static str`

bin/runtime-common/src/messages.rs
```diff
@@ -0,0 +1,538 @@
...
+	type Signer: Decode;
+	/// Signature type used on the chain.
+	type Signature: Decode;
+	/// Call type on the chain.
+	type Call: Decode;
+	/// Type of weight that is used on the chain. This would almost always be a regular
+	/// `frame_support::weight::Weight`. But since meaining of weight on different chains
+	/// may be different, the `WeightOf<>` construct is used to avoid confusion between
+	/// different weights.
+	type Weight: From<frame_support::weights::Weight>;
```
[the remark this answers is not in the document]
svyatonik: I think (as stated in comment) that most of times it'll be `type Weight = frame_support::weight::Weight;`, without any wrappers. And there's already bult-in implememntation `impl Into<u128> for frame_support::weight::Weight { ... }`. This conversion should be explicit imo.

bin/runtime-common/src/messages.rs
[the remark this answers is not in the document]
HCastano: Ah gotcha. I'd change it to `Err("Fee paid by submitter is too low.")` then

bin/rialto/runtime/src/millau_messages.rs
```diff
@@ -0,0 +1,157 @@
...
+
+// Parity Bridges Common is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Everything required to serve Millau <-> Rialto message lanes.
```
HCastano, line 17: I'd be a little more specific here

bin/millau/runtime/src/rialto_messages.rs
```diff
@@ -0,0 +1,157 @@
...
+
+// Parity Bridges Common is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Everything required to serve Millau <-> Rialto message lanes.
```
HCastano, line 17: Same comment here about being specific

primitives/millau/src/lib.rs
```diff
@@ -21,11 +21,17 @@
...
 use sp_core::Hasher as HasherT;
-use sp_runtime::traits::BlakeTwo256;
+use sp_runtime::{
+	traits::{BlakeTwo256, IdentifyAccount, Verify},
+	MultiSignature, MultiSigner,
+};
 use sp_std::prelude::*;
 
+/// Maximal weight of single Millau extrinsic.
+pub const MAXIMUM_EXTRINSIC_WEIGHT: Weight = 1_300_000_000_000;
```
HCastano, line 33: How did you get this number? It should have some documentation around the process

modules/message-lane/src/instant_payments.rs
```diff
@@ -0,0 +1,75 @@
+// Copyright 2019-2020 Parity Technologies (UK) Ltd.
```
HCastano, line 1: I'm not sure if `instant_payments` is the best name for this since from what it looks like its main purpose is to pay relayers for message delivery

primitives/rialto/src/lib.rs
```diff
@@ -21,11 +21,17 @@
...
 use sp_core::Hasher as HasherT;
-use sp_runtime::traits::BlakeTwo256;
+use sp_runtime::{
+	traits::{BlakeTwo256, IdentifyAccount, Verify},
+	MultiSignature, MultiSigner,
+};
 use sp_std::prelude::*;
 
+/// Maximal weight of single Rialto extrinsic.
+pub const MAXIMUM_EXTRINSIC_WEIGHT: Weight = 1_300_000_000_000;
```
HCastano, line 33: Same question here

HCastano (commented):
Almost there, I left a few questions/comments

bin/runtime-common/src/messages.rs
```diff
@@ -0,0 +1,539 @@
...
+		SignerOf<BridgedChain<B>>,
+		SignerOf<ThisChain<B>>,
+		SignatureOf<ThisChain<B>>,
+		CallOf<ThisChain<B>>,
+	>;
+
+	/// Message payload for Bridged -> This messages.
+	pub struct FromBridgedChainMessagePayload<B: MessageBridge>(pub(crate) FromBridgedChainDecodedMessagePayload<B>);
+
+	impl<B: MessageBridge> Decode for FromBridgedChainMessagePayload<B> {
```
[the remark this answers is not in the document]
svyatonik: Correct. This custom `Decode` impl should yet to be tested in real environment, though (after we'll have everything else ready). The assumption is that:
1) the sender would submit tx with `BridgedChain::Call::encode()`, because `message-lane` module expects `Vec<u8>`;
2) when message payload will be encoded by `message-lane` module, it'll be prefixed with vector lenght;
3) so on receiving side we'll need to skip that length and proceed directly to wrapped encoded `Call`

bin/runtime-common/src/messages.rs
```diff
@@ -0,0 +1,539 @@
...
+			impl CheckedMul for $name {
+				fn checked_mul(&self, other: &Self) -> Option<Self> {
+					self.0.checked_mul(other.0).map(Self)
+				}
+			}
+		};
+	}
+
+	impl_wrapped_balance!(ThisChainBalance);
+	impl_wrapped_balance!(BridgedChainBalance);
```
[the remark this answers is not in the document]
svyatonik: To test that the module is able to work with different balance types on different chains. I have had a few typos (because of copy-paste) in this code, so I decided it would be better to be sure that we're using single balance types in expressions. The same thing should probably be done with the weight later.

bin/runtime-common/src/messages.rs
```diff
@@ -0,0 +1,539 @@
...
+				weight: 100,
+				origin: pallet_bridge_call_dispatch::CallOrigin::BridgeAccount,
+				call: ThisChainCall::Transfer,
+			}
+		);
+	}
+
+	#[test]
+	fn message_fee_is_checked_by_verifier() {
+		const EXPECTED_MINIMAL_FEE: u32 = 2640;
```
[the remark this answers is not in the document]
svyatonik: It's couple of lines below:
```rust
			source::estimate_message_dispatch_and_delivery_fee::<OnThisChainBridge>(
				&payload,
				OnThisChainBridge::RELAYER_FEE_PERCENT,
			),
```

primitives/millau/src/lib.rs
svyatonik: I just took actual value from `Millau` runtime and moved it here along with its comments (none).

modules/message-lane/src/instant_payments.rs
svyatonik: Its main purpose is to pay relayers just after proof-of-delivery has been received (instantly).

primitives/millau/src/lib.rs
svyatonik: I've actually replaced computation formula with constant, which may be treated as comment. So I've added some more constants to primitives.

bin/runtime-common/src/messages.rs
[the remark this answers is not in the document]
HCastano: Can you make a note of it in the code? I can definitely see myself touching this code in the future and changing it to `u128` without remembering why it was done this way initially

modules/message-lane/src/instant_payments.rs
HCastano: I think that until we have a distinction between instant and non-instant payments for relayers this should be called `relayer_payments` :)

primitives/millau/src/lib.rs
HCastano: Having a formula for `MAXIMUM_EXTRINSIC_WEIGHT` is much better, thanks! Now the question becomes how did we choose `MAXIMUM_BLOCK_WEIGHT`? It looks like it's '2 * WEIGHT_PER_SECOND` which is from Substrate and translates to ~2 seconds of compute. Is that right?

HCastano (approved):
A bit of the documentation of constants could be improved, but other than that looks good!

primitives/millau/src/lib.rs
svyatonik: Yes, I think you're right - it is roughly 2 sec, iiuc

modules/message-lane/src/instant_payments.rs
svyatonik: (1) it isn't only about payments to relayer (2) it's an implementation of trait with specific that payments are made instantly. So it should not be called `relayr_payments`

bin/rialto/runtime/src/millau_messages.rs
svyatonik: I don't think I have anything to add here. This file contains types to allow sending messages between Millau and Rialto. Nothing to add here

PR #387, opened 2020-09-30
Get substrate dependencies from crates io

### LINKED ISSUE #20: Update Substrate dependencies to use crates.io versions

svyatonik:
closes #20 

The only drawback is that we're losing single test, which was testing that the message-lane RPCs are actually generating verifiable storage proofs. I was unable to take `substrate-test-runtime-client` crate from git + all others from crates.io.


tomusdrw (approved):
:tada:

svyatonik:
Looks like this update breaks our metrics + alerts => I'll try to find why

svyatonik:
Looks like some temp local issue

HCastano:
Cool! @svyatonik did you use a script, or do this manually?

svyatonik:
Manually :) We need to do that only once, right?

PR #394, opened 2020-10-02
Subscribe to justifications in Millau->Rialto headers sync

### LINKED ISSUE #71: Finalize Substrate headers using non-persistent Justifications
context: https://github.com/paritytech/parity-bridges-common/pull/69#pullrequestreview-396601216

svyatonik:
closes #71

~I'm opening this PR mostly as TODO for myself, because it is too early to plug-in this now.~

Subscribing to (ephemeral) justifications is essential to bridge Substrate-based chains. Our test chains are configured to generate persistent justifications every 4 blocks (or better to reword - sessions have 4 blocks length). So blocks are finalized almost instantly. But on real chains, we won't see persistent justifications that often => to finalize blocks faster we need to subscribe to ephemeral justifications.

The solution is to have this on-top of regular headers sync - every 30 seconds 'maintain' function is called, which reads pending justifications from Millau node and, if required, submits these to the Rialto node.


relays/substrate/src/millau_headers_to_rialto.rs
```diff
@@ -96,8 +101,71 @@ impl TargetClient<MillauHeadersToRialto> for RialtoTargetClient {
...
+							Err(error) => {
+								log::warn!(
+									target: "bridge",
+									"Failed to decode justification from Millau: {:?}",
+									error,
+								);
+							}
+						}
+					}
+					Poll::Pending => break,
```
tomusdrw, line 157: Is this correct? Current `maybe_next_millau_justification` would be lost?

relays/substrate/src/millau_headers_to_rialto.rs
```diff
@@ -96,8 +101,71 @@ impl TargetClient<MillauHeadersToRialto> for RialtoTargetClient {
 	}
 }
 
+/// Maintainment procedure of Millau -> Rialto headers.
+struct MillauHeadersToRialtoMaintain {
+	millau_justifications: Arc<Mutex<JustificationsSubscription>>,
```
tomusdrw, line 106: Why does it have to be `Arc<Mutex` it doesn't seem to be sharing the state.

tomusdrw (commented):
I know it's still in Draft, but had few questions already ;)

relays/substrate/src/millau_headers_to_rialto.rs
svyatonik: `JustificationsSubscription::next()` requires `&mut self`.

relays/substrate/src/millau_headers_to_rialto.rs
svyatonik: I can't say for sure now - will check before making it ready for review :)

svyatonik: Now I'm pretty sure nothing is lost :)

HCastano (commented):
This is looking cool

relays/substrate/src/headers_maintain.rs
```diff
@@ -0,0 +1,383 @@
...
+	P::Number: Decode + From<C::BlockNumber>,
+	P::Hash: Decode + From<C::Hash>,
+	P: SubstrateHeadersSyncPipeline<Completion = Justification, Extra = ()>,
+{
+	async fn maintain(&self, sync: &mut HeadersSync<P>) {
+		// lock justifications before doing anything else
+		let mut justifications = match self.justifications.try_lock() {
+			Some(justifications) => justifications,
+			None => {
+				// this should never happen, as we use single-thread executor
```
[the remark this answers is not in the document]
svyatonik: `async` + `Subscription::next` that requires `&mut self`

relays/substrate/src/headers_maintain.rs
```diff
@@ -0,0 +1,383 @@
...
+			// we won't submit next justifications as we need to submit previous justifications
+			// first
+			break;
+		}
+
+		// if we know that the header is already synced (it is known to the target node), let's
+		// select it for submission
+		if sync.headers().status(&target) == HeaderStatus::Synced {
+			selected_justification = Some((target, justification));
+			continue;
```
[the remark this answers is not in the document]
svyatonik: Yes - that's the idea. We do not need to submit all justifications, only last one.

relays/substrate/src/headers_maintain.rs
```diff
@@ -0,0 +1,383 @@
...
+		// finally - return justification back to the queue
+		queue.push_back((target, justification));
+		break;
+	}
+
+	selected_justification
+}
+
+/// Returns best finalized source header on the target chain.
+async fn best_finalized_header_id<P, C>(client: &Client<C>) -> Result<HeaderIdOf<P>, SubstrateError>
```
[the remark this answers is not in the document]
svyatonik: It is valid only in context of headers sync, so shouldn't be a part of `relay-substrate-client`

relays/headers-relay/src/sync_loop.rs
```diff
@@ -387,9 +408,16 @@ pub fn run<P: HeadersSyncPipeline, TC: TargetClient<P>>(
...
+			if maintain_required && source_client_is_online && target_client_is_online {
+				log::debug!(target: "bridge", "Maintaining headers sync loop");
+				maintain_required = false;
+				sync_maintain.maintain(&mut sync).await;
+			}
+
 			// If the target client is accepting requests we update the requests that
 			// we want it to run
-			if target_client_is_online {
+			if !maintain_required && target_client_is_online {
```
HCastano, line 420: What problems would we run into if this `!maintain_required` check weren't here?

relays/headers-relay/src/sync_loop.rs
```diff
@@ -476,7 +504,7 @@ pub fn run<P: HeadersSyncPipeline, TC: TargetClient<P>>(
 
 			// If the source client is accepting requests we update the requests that
 			// we want it to run
-			if source_client_is_online {
+			if !maintain_required && source_client_is_online {
```
HCastano, line 507: Same question here

relays/substrate/src/headers_maintain.rs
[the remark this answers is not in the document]
HCastano: Gotcha. Would you mind making a note explaining that in a comment?

HCastano (commented):
I'd like to see some of the naming around "maintain" improved. For example, `maintain_required` doesn't really explain what it's checking. It's not limited to this, as I've pointed out with a trait and a struct

relays/headers-relay/src/sync_loop.rs
svyatonik: The general principle of headers sync is that there's at most 1 future that works with source/target client at any moment. So we can't run maintain if client is busy with other requests. If we'll keep source/target client running other requests, maintain may never happen.

svyatonik:
> I'd like to see some of the naming around "maintain" improved. For example, maintain_required doesn't really explain what it's checking

`maintain_required` is not checking anything. If it is true, then we need to call sync maintain.

> It's not limited to this, as I've pointed out with a trait and a struct

I don't see much point in changing `SyncMaintain` to `MaintainSync`. The `MaintainFinality` is even worse, because it brings unnecessary finality notion to the headers sync.

relays/substrate/src/headers_maintain.rs
```diff
@@ -0,0 +1,383 @@
...
+					P::SOURCE_NAME,
+					target,
+					error,
+				),
+			}
+		}
+	}
+}
+
+impl<P> Justifications<P>
```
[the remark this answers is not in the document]
HCastano: I'd still like to see this closer to the struct definition unless you have a particular reason why it's better here

HCastano (approved):
> > It's not limited to this, as I've pointed out with a trait and a struct
> 
> I don't see much point in changing `SyncMaintain` to `MaintainSync`. The `MaintainFinality` is even worse, because it brings unnecessary finality notion to the headers sync.

Okay, I think it's fair that `MaintainFinality` isn't a good name for this. I do think `MaintainSync` is a bit better than `SyncMaintain` though because it give a better indication of the behaviour that is expected by the trait.

Looking at that trait specifically, I still think it's unclear what "maintain" means. Maybe that's by design, but if it is that's not clear. Would "maintaining" include things like pruning old headers? Or should it only be used to track the head of the chain? I think we should either change the name of the trait and the related structs/vars to be more specific, or give a better indication of _what_ something implementing the trait should be doing through doc comments.

While I would like to see more clarity around the language in this PR, the code does look good, so let's get this moving

svyatonik:
> Looking at that trait specifically, I still think it's unclear what "maintain" means. Maybe that's by design, but if it is that's not clear. Would "maintaining" include things like pruning old headers? Or should it only be used to track the head of the chain? I think we should either change the name of the trait and the related structs/vars to be more specific, or give a better indication of what something implementing the trait should be doing through doc comments.

That is by design. On `headers-sync` level `SyncMaintain` is abstract. Headers sync is self-sufficient - we're syncing headers, uploading extra and completion data. This works for sub2eth, eth2sub and sub2sub && everything is fine.

But if we need to do something else during that sync, then we have that `SyncMaintain` callback that is called periodically with some guarantees (that there are no other concurrent futures that are using the source/target clients). In sub2sub case it is used to read and submit justifications. But you're right - it may be used e.g. for pruning old headers or doing something else. So it is abstract callback with access to internal headers sync state && which is periodically called.

relays/substrate/src/headers_maintain.rs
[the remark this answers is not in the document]
svyatonik: My habit is to write like that:
```
<Public-Structs-Declarations>
<First-Class-Private-Structs-Declarations>
<Public-Structs-And-Functions-Implementations>
<First-Class-Private-Structs-Implementations>
<Private-Structs-And-Functions-Declarations-And-Implementations>
<Tests>
```

This isn't a super-strict rule, but I see much more point in seeing relations between data structs on single screen than seeing full screen of code. Imo it gives you better overview of what's this file is about. I'm not trying to convince you that this is better, though :)

Issue #397, opened 2020-10-05
Implement APIs for MessageLane <-> SubstrateBridge pallets integration


svyatonik:
The minimal API would be function that given `header_hash`, `StorageProof` and `keys: Vec<StorageKey>` will:
1) ensure that the header is finalized;
2) read optional `StorageValue` from the proof and will return `Vec<Option<StorageValue>>`.

PR #399, opened 2020-10-05
Millau -> Rialto messages relay

svyatonik:
~This PR should be minimal working Millau -> Rialto messages relay. Right now it won't work because #386 is still pending.~

This PR is minimal working Millau -> Rialto messages relay. Right now I'm unable to confirm if it works, because some underlying components are still missing. There are some things that needs to be fixed in follow up PRs: #432 #433 #457 . They are not critical for test deployments, that's why I have extracted them as a separate issues.


HCastano (commented):
I'm having a bit of a hard time following the changes in this PR. There's a few things I feel like I'm missing:

- Where should I start reviewing?
- Is there's anything particularly important to look out for? 
- What's the overall approach to getting the message relay working?
- What's the expected usage?
- Etc.

You mention that there are "underlying components missing". What components are missing and why? What part of the relay doesn't work because of it? There's also a few TODOs. Which ones are going to be addressed in this PR vs. later PRs? For some TODOs its unclear

svyatonik:
> Where should I start reviewing?
> Is there's anything particularly important to look out for?
> What's the overall approach to getting the message relay working?
> What's the expected usage?

The `message-lane` relay is originally introduced in #318 . This PR implements traits from #318, required to serve Millau -> Rialto lanes. You start relay by running `millau-messages-to-rialto` subcommand of the `substrate-relay`. The single relay instance is serving single lane.

Some implementation details of Millau -> Rialto message lane relay:
- since all the chains we're going to bridge in near future will be based on the `message-lane` + `call-dispatch` modules, the most of traits-implementation code would be the same. So the `relays/substrate/src/messages_source.rs` contains implementation of `messages_relay::SourceClient` for such chains and the `relays/substrate/src/messages_target.rs` is the implementation of `messages_relay::TargetClient`;
- everything that is specific for Millau -> Rialto lanes is in the `relays/substrate/src/millau_messages_to_rialto.rs`.

There's additional change included in this PR that is required for Millau -> Rialto message lane relay. It is in these two commits: https://github.com/paritytech/parity-bridges-common/pull/399/commits/2fe0ac355d4f57e817205209a44cff982239e27e and https://github.com/paritytech/parity-bridges-common/pull/399/commits/795d30aa407c13b1c5a8f14383572917a45ef111. Basically - we have had the same delivery strategy for both delivery and confirmation races before. It has been moved from `relays/messages-relay/src/message_race_delivery.rs` to `relays/messages-relay/src/message_race_strategy.rs`. It simply includes all pending messages/confirmations in the batch.

But this original concept needs to be changed now, because: (1) scheduled messages were removed from `message-lane` module and (2) outbound lane state proof is now required to be included into messages proof sometimes. So there's now custom implementation of the `RaceStrategy` for message delivery loop. It is the `MessageDeliveryStrategy` from the `relays/messages-relay/src/message_race_delivery.rs`.

> You mention that there are "underlying components missing". What components are missing and why?

Atm the last required component afaik is #397 . It is missing because noone works on that.

> What part of the relay doesn't work because of it?

I haven't claimed that some parts of relay doesn't work because of this :) The whole message delivery + dispatch pipeline isn't working => I can't guarantee that relay works, because all its transactions would fail.

> There's also a feew TODOs. Which ones are going to be addressed in this PR vs. later PRs? For some TODOs its unclear

Right, thanks! I have been checking the code for leftover TODOs before moving to ready-for-review state, but missed 3 of them :) Fixed in last commits. So there are now 4 TODOs left: #432 #433 #457 #468

relays/substrate/src/messages_source.rs
```diff
@@ -0,0 +1,221 @@
...
+#[async_trait]
+impl<C, P, M> SourceClient<P> for SubstrateMessagesSource<C, P, M>
+where
+	C: Chain,
+	C::Header: DeserializeOwned,
+	C::Index: DeserializeOwned,
+	<C::Header as HeaderT>::Number: Into<u64>,
+	P: MessageLane<
+		MessageNonce = MessageNonce,
+		MessagesProof = (Weight, (HashOf<C>, StorageProof, LaneId, MessageNonce, MessageNonce)),
```
HCastano, line 94: I think it would be worthwhile to typedef this

relays/substrate/src/messages_source.rs
```diff
@@ -0,0 +1,221 @@
...
+		let tx = self
+			.tx_maker
+			.make_messages_receiving_proof_transaction(generated_at_block, proof)
+			.await?;
+		self.client.submit_extrinsic(Bytes(tx.encode())).await?;
+		Ok(())
+	}
+}
+
+/// Read client state.
```
HCastano, line 184: Doc comment same as function name

relays/substrate/src/messages_source.rs
```diff
@@ -0,0 +1,221 @@
...
+
+	async fn latest_generated_nonce(
+		&self,
+		id: SourceHeaderIdOf<P>,
+	) -> Result<(SourceHeaderIdOf<P>, P::MessageNonce), Self::Error> {
+		let encoded_response = self
+			.client
+			.state_call(
+				// TODO: https://github.com/paritytech/parity-bridges-common/issues/457
+				"OutboundLaneApi_latest_generated_nonce".into(),
```
HCastano, line 122: We should do something about these hardcoded Runtime API strings, we already saw how easy they were to break last week

relays/messages-relay/src/message_lane_loop.rs
```diff
@@ -97,6 +114,11 @@ pub trait TargetClient<P: MessageLane>: Clone {
 		&self,
 		id: TargetHeaderIdOf<P>,
 	) -> Result<(TargetHeaderIdOf<P>, P::MessageNonce), Self::Error>;
+	/// Get nonce of latest confirmed message.
```
HCastano, line 117:
```suggestion

	/// Get nonce of latest confirmed message.
```

relays/substrate/src/messages_source.rs
HCastano: Also, why does the proof have the weight included in it?

relays/substrate/src/messages_source.rs
```diff
@@ -0,0 +1,221 @@
...
+	let encoded_best_finalized_peer_on_self = self_client
+		.state_call(
+			best_finalized_peer_on_self_method,
+			Bytes(Vec::new()),
+			Some(self_best_finalized_header_hash),
+		)
+		.await?;
+	let decoded_best_finalized_peer_on_self: (BridgedHeaderNumber, BridgedHeaderHash) =
+		Decode::decode(&mut &encoded_best_finalized_peer_on_self.0[..]).map_err(SubstrateError::ResponseParseFailed)?;
+	let peer_on_self_best_finalized_id = HeaderId(
```
HCastano, line 212: Maybe we can add a `From<(Num, Hash)>` implementation for `HeaderId` so this can just be an `.into()` call

relays/substrate/src/messages_source.rs
```diff
@@ -0,0 +1,221 @@
...
+use sp_trie::StorageProof;
+use std::{marker::PhantomData, ops::RangeInclusive};
+
+/// Substrate client as Substrate messages source.
+pub struct SubstrateMessagesSource<C: Chain, P, M> {
+	client: Client<C>,
+	tx_maker: M,
+	lane: LaneId,
+	instance: InstanceId,
+	_marker: PhantomData<P>,
```
HCastano, line 43: Why do we use a `PhantomData` field for `P` instead of a real field of type `P`?

HCastano (commented):
I really appreciate the background and guidance on how to approach this PR. It makes reviewing the code a lot easier. That said, I didn't get through a lot of it today 😅

> You start relay by running millau-messages-to-rialto subcommand of the substrate-relay

May I suggest something like `./relay messages --from=millau --to=rialto`? Although I think this would require support for multiple networks and a bit of a refactor of the CLI UI...

> Atm the last required component afaik is #397 . It is missing because noone works on that.

If this is important to get done let us know and I'm sure someone can find time to do it

relays/substrate/src/messages_source.rs
svyatonik: Because weight needs to be submitted with proof.

relays/substrate/src/messages_source.rs
svyatonik: P is not meant to be instantiated

relays/substrate/src/messages_source.rs
svyatonik: Why do you think `.into()` is better than `HeaderId()`? The `HeaderId` is here, because tuples were not intuitive - you need some context to understand what's this tuple is. Going back to `(...).into()` doesn't help imo

relays/substrate/src/messages_source.rs
svyatonik: (I'll add a typedef, though :) )

svyatonik:
> May I suggest something like ./relay messages --from=millau --to=rialto? Although I think this would require support for multiple networks and a bit of a refactor of the CLI UI...

I don't like it, because it requires relay to be fully symmetric (i.e. now we have Millau -> Rialto, but Rialto -> Millau is still missing). Also there'll be problem if we'll ever need to add chain-specific arguments with default values (like `maximal_balance_decrease` for balance guard from #470). You may add something like that as an additional option in future PRs.

> If this is important to get done let us know and I'm sure someone can find time to do it

I think all issues with Millau release milestone are equally important - no need to reprioritize or something :) The fact that I can't test something doesn't mean that I can't work on that.

relays/substrate/src/messages_source.rs
HCastano: It seems a little silly to me that we're breaking up this tuple (`decoded_best_finalized_peer_on_self`) just to turn it back into a `HeaderId`

relays/substrate/src/messages_source.rs
HCastano: > Because weight needs to be submitted with proof.

I was gonna ask "why" again, but it looks like you answered it with the doc comments 👍

relays/messages-relay/src/message_race_delivery.rs
```diff
@@ -164,280 +232,87 @@ type MessageDeliveryStrategy<P> = DeliveryStrategy<
...
+		// https://github.com/paritytech/parity-bridges-common/issues/432
+		// TODO: message lane loop works with finalized blocks only, but we're submitting transactions that
+		// are updating best block (which may not be finalized yet). So all decisions that are made below
+		// may be outdated. This needs to be changed - all logic here bust be built on top of best blocks.
+
+		// The receiving race is responsible to deliver confirmations back to the source chain. So if
+		// there's a lot of unconfirmed messages, let's wait until it'll be able to do its job.
+		let latest_received_nonce_at_target = target_nonces.latest_nonce;
+		let latest_confirmed_nonce_at_source = source_nonces.confirmed_nonce.expect(CONFIRMED_NONCE_PROOF);
+		let confirmatins_missing = latest_received_nonce_at_target.checked_sub(&latest_confirmed_nonce_at_source);
```
HCastano, line 280:
```suggestion
		let confirmations_missing = latest_received_nonce_at_target.checked_sub(&latest_confirmed_nonce_at_source);
```

relays/messages-relay/src/message_race_delivery.rs
```diff
@@ -164,280 +232,87 @@ type MessageDeliveryStrategy<P> = DeliveryStrategy<
...
+		match confirmatins_missing {
+			Some(confirmatins_missing) if confirmatins_missing > self.max_unconfirmed_nonces_at_target => {
+				log::debug!(
+					target: "bridge",
+					"Cannot deliver any more messages from {} to {}. Too many unconfirmed nonces at target: target.latest_received={:?}, source.latest_confirmed={:?}, max={:?}",
```
HCastano, line 285: `RustFmt` doesn't do a good job with string literals, but it would be easier to read if this was split on multiple lines

relays/messages-relay/src/message_race_delivery.rs
HCastano, line 280: Is there a problem if we have a problem if `latest_confirmed_nonce_at_source` > `latest_received_nonce_at_target`? This seems like something that shouldn't happen

HCastano (commented):
I'm still working my way through the PR

relays/substrate/src/messages_source.rs
```diff
@@ -0,0 +1,225 @@
...
+use sp_trie::StorageProof;
+use std::{marker::PhantomData, ops::RangeInclusive};
+
+/// Intermediate message proof returned by the source Substrate node. Includes everything
+/// required to submit to the target node: cumulative dispatch weight of bundled messages and
+/// the proof itself.
+pub type SubstrateMessagesProof<C> = (Weight, (HashOf<C>, StorageProof, LaneId, MessageNonce, MessageNonce));
+
+/// Substrate client as Substrate messages source.
+pub struct SubstrateMessagesSource<C: Chain, P, M> {
```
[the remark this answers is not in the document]
svyatonik: `MessageLane` is actually `P` across the code :) That's because it has been `MessagePipeline` in the beginning. So this would only add more confusion imo.

relays/messages-relay/src/message_race_loop.rs
```diff
@@ -61,36 +58,47 @@ type SourceClientState<P> = ClientState<<P as MessageRace>::SourceHeaderId, <P a
...
 		&self,
 		at_block: P::SourceHeaderId,
-	) -> Result<(P::SourceHeaderId, P::MessageNonce), Self::Error>;
+	) -> Result<(P::SourceHeaderId, ClientNonces<P::MessageNonce>), Self::Error>;
 	/// Generate proof for delivering to the target client.
 	async fn generate_proof(
 		&self,
 		at_block: P::SourceHeaderId,
 		nonces: RangeInclusive<P::MessageNonce>,
+		additional_proof_required: bool,
```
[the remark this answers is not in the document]
svyatonik: I've replaced `additional_proof_required` with `ProofParameters`. It's just a protocol between delivery strategy and source client.

relays/messages-relay/src/message_race_loop.rs
```diff
@@ -104,20 +112,21 @@ pub trait TargetClient<P: MessageRace> {
...
+	/// Called when nonces are updated at target node of the race.
+	fn target_nonces_updated(
 		&mut self,
-		nonce: MessageNonce,
+		nonces: ClientNonces<MessageNonce>,
 		race_state: &mut RaceState<SourceHeaderId, TargetHeaderId, MessageNonce, Proof>,
 	);
 	/// Should return `Some(nonces)` if we need to deliver proof of `nonces` (and associated
 	/// data) from source to target node.
+	/// The bool flag will be passed as `additional_proof_required` argument to the `generate_proof`.
```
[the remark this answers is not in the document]
svyatonik: Irrelevant

relays/messages-relay/src/message_race_loop.rs
```diff
@@ -61,36 +58,47 @@ type SourceClientState<P> = ClientState<<P as MessageRace>::SourceHeaderId, <P a
 /// State of race target client.
 type TargetClientState<P> = ClientState<<P as MessageRace>::TargetHeaderId, <P as MessageRace>::SourceHeaderId>;
 
+/// Nonces on the race client.
+#[derive(Debug, Clone)]
+pub struct ClientNonces<MessageNonce> {
+	/// Latest nonce that is known to the client.
+	pub latest_nonce: MessageNonce,
+	/// Latest nonce that is confirmed to the bridged client. This nonce only makes
```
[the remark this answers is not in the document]
svyatonik: Right now it is only `Some` in message delivery race.

relays/messages-relay/src/message_race_loop.rs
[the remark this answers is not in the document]
HCastano: I don't think the `ProofParameters` associated type really addresses my concern here. We still have `type ProofParameters = bool`, so it only shifts the `bool` elsewhere. Imo it's still unclear as to what the `bool` should be or what it represents

HCastano (approved):
Looks okay, but I would get another review before merging

relays/messages-relay/src/message_race_loop.rs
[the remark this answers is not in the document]
svyatonik: Then, if we'll ever add another race, this comment would need to be updated. I doubt that going through all code comments every time you change something is ok. So I'd like to avoid any race-specific comments in race-agnostic structs, functions, etc

relays/messages-relay/src/message_race_loop.rs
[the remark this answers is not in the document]
svyatonik: Well, as I said, it is part of protocol between strategy and client. That's the only place where you actually treat `ProofParameters` as `bool`. In both places this `bool` has a name. I think it's enough

relays/messages-relay/src/message_race_delivery.rs
svyatonik: It should not happen, right. In theory :)

HCastano:
@tomusdrw Can you review this?

relays/messages-relay/src/message_race_receiving.rs
```diff
@@ -138,35 +149,44 @@ struct ReceivingConfirmationsRaceTarget<P: MessageLane, C> {
...
-	) -> Result<(SourceHeaderIdOf<P>, P::MessageNonce), Self::Error> {
+	) -> Result<(SourceHeaderIdOf<P>, ClientNonces<P::MessageNonce>), Self::Error> {
 		let result = self.client.latest_confirmed_received_nonce(at_block).await;
 		if let Some(metrics_msg) = self.metrics_msg.as_ref() {
 			if let Ok((_, source_latest_confirmed_nonce)) = result.as_ref() {
 				metrics_msg.update_source_latest_confirmed_nonce::<P>(*source_latest_confirmed_nonce);
 			}
 		}
-		result
+		result.map(|(at_block, latest_confirmed_received_nonce)| {
```
tomusdrw, line 170: Same here, what's the reason for not doing `?` right after `.await`?

relays/substrate/src/messages_source.rs
tomusdrw: IMHO it's worth to at least create constants and maybe even group in a single module, so that there is one single place we need to update in the future. This also encourages re-use if we ever run into using the same Runtime API call in different places

relays/substrate/src/messages_source.rs
tomusdrw: I'd rather be for `HeaderId::from_tuple(...)` instead of `.into()`

tomusdrw (approved):
lgtm! A few grumbles left in comments, but nothing serious.

relays/substrate/src/messages_source.rs
svyatonik: This hasn't helped last time we had a name change. Probably better would be to introduce replacement for `decl_runtime_apis! {}` macro (that would actually call `decl_runtime_apis` from within)  that will generate string constants from method names, or even auto-generate `jsonrpsee` client for these methods. Re this particular string(s) - they're going to change anyway as a part of #457 in near future - I'll extract constants there.

relays/messages-relay/src/message_lane_loop.rs
```diff
@@ -192,15 +208,41 @@ pub fn run<P: MessageLane>(
 
 			match result {
 				Ok(()) => break,
-				Err(failed_client) => {
-					async_std::task::sleep(reconnect_delay).await;
+				Err(failed_client) => loop {
+					async_std::task::sleep(params.reconnect_delay).await;
 					if failed_client == FailedClient::Both || failed_client == FailedClient::Source {
-						source_client = source_client.reconnect();
+						source_client = match source_client.clone().reconnect().await {
```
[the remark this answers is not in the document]
svyatonik: There are conditions on when to reconnect source/target. So we may need to reconnect to source, or we may need to reconnect to target, or we may need to reconnect to both. The loop isn't that large - the largest part is `log::warn` call, which we would need anyway. I don't think that introducing join + additional calls (`.left()`, `.right`, `.and_then`, `ready`) would make this loop easier to read/much smaller.

PR #400, opened 2020-10-06
Message lane loop metrics

svyatonik:


tomusdrw (approved):
lgtm!
we might consider adding additional metrics regarding messages themselves:
- pending messages weight (or some sort of sum of weight to be dispatched/delivered)
- message dispatch status (failed / successful)

svyatonik:
> pending messages weight (or some sort of sum of weight to be dispatched/delivered)
> message dispatch status (failed / successful)

Errrr...That is something that should be provided either by node, or by some other component (not relayer) imo :) Probably there are some solutions that may read data from runtime (storage, events) and provide as Prometheus endpoint?

PR #401, opened 2020-10-06
Add GenesisConfig for Millau and Rialto Runtimes

HCastano:
Small follow up from #389. I've added the genesis config for each of the nodes, albeit with dummy values.

Draft since there's a few small things I want to clean up which I didn't have time for today.


bin/millau/node/src/chain_spec.rs
```diff
@@ -164,3 +165,12 @@ fn testnet_genesis(
 		}),
 	}
 }
+
+fn load_rialto_bridge_config() -> Option<BridgeRialtoConfig> {
+	Some(BridgeRialtoConfig {
+		initial_header: Some(millau_runtime::rialto::initial_header()),
```
svyatonik, line 171: It should be `rialto_runtime::initial_header()` instead :)

svyatonik: Nvm - I think I've misunderstood you design :)

HCastano: Yeah each runtime has modules which correspond to that particular chain's configuration. So the Millau chain has a file called `rialto.rs` where the configuration params for the Rialto chain are defined. If you think this is confusing I can put all that code is a common "config" folder or something. What do you think?

bin/millau/runtime/src/rialto.rs
```diff
@@ -0,0 +1,57 @@
...
+///
+/// Note that this doesn't have to be the "genesis" authority set, as the
+/// pallet can be configured to start from any height.
+pub fn initial_authority_set() -> AuthoritySet {
+	let set_id = 0;
+	let authorities = vec![
+		(AuthorityId::from_slice(&[1; 32]), 1),
+		(AuthorityId::from_slice(&[2; 32]), 1),
+		(AuthorityId::from_slice(&[3; 32]), 1),
+	];
```
HCastano, line 55: I wanted to get the real authorities + weights from Rialto at genesis but wasn't able to do so. I know I need to get the storage values under `GRANDPA_AUTHORITIES_KEY` using the `state_getStorage` RPC call but I wasn't able to figure out how

bin/millau/node/src/chain_spec.rs
svyatonik: No, it's perfectly fine :)

bin/millau/runtime/src/rialto.rs
svyatonik: There are two alternatives:
1) manually craft that value - initial authorities are Alice, Bob, Charlie, Dave, Eve and weight is 1 for every authority;
2) we have `substrate_relay::Client::grandpa_authorities_set()` - you may call it manually, or just run regulat Rialto -> RialtoPoA headers relay && when PoA contract is deployed, relay is tracing hex-encoded authorities set.

bin/millau/runtime/src/rialto.rs
```diff
@@ -0,0 +1,57 @@
...
+// To get this we first need to call the `chain_getBlockHash` RPC method, and then
+// we can use the result from that and call the `chain_getBlock` RPC method to get
+// the rest of the info.
+//
+// In this case we've grabbed the genesis block of the Rialto Substrate chain.
+pub fn initial_header() -> Header {
+	Header {
+		parent_hash: Default::default(),
+		number: Default::default(),
+		state_root: hex!("8fc2e3a8216b9f70fe2bc6804febb3f5192cb4046db5d546d1ad499a29bbbf27").into(),
```
svyatonik, line 39: I was getting different values locally && then realized we have a circular dependency here. Right now you have `Default::default()` in `rialto-runtime/src/millau.rs` && you may guess `state_root` of Rialto genesis block. But once you fix `state_root` in `rialto-runtime/src/millau.rs`, the `Rialto::genesis::state_root` would change and you will need to update it in this file => `Millau::genesis::state_root` will change and so on.

What we need is a separate 'start-headers-sync' step similar to `eth-deploy-contract` we have in sub2eth sync. So my suggestion is to:
1) since we have started with Millau -> Rialto headers sync (see PRs list), let's:
1.1) change `state_root` to `Default::default()` in this file;
1.2) change `state_root` from `Default::default()` to actual value in `bin/rialto/runtime/src/millau.rs` (this would allow us to start Millau -> Rialto headers and messages relay);
2) in next PRs change substrate pallet to:
2.1) support starting with empty configuration - i.e. no `initial_header`, no `initial_genesis_set`, ...;
2.2) support "initialize" sudo-call that will accept `GenesisConfig`
3) add additional function to relay - `initialize_rialto_headers_to_millau` that will submit tx with that sudo call.

bin/rialto/runtime/src/millau.rs
```diff
@@ -0,0 +1,47 @@
...
+/// The first header known to the pallet.
+///
+/// Note that this does not need to be the genesis header of the Millau
+/// chain since the pallet may start at any arbitrary header.
+pub fn initial_header() -> Header {
+	Header {
+		parent_hash: Default::default(),
+		number: Default::default(),
+		state_root: Default::default(),
+		extrinsics_root: Default::default(),
```
svyatonik, line 34:
```suggestion
		state_root: hex!("756ba147f373121b1c26281c51ec914d3f2790a064ad7728c2fa2935d0afe893").into(),
		extrinsics_root: hex!("03170a2e7597b7b7e3d84c05391d139a62b157e78786d8c082f29dcf4c111314").into(),
```

bin/rialto/runtime/src/millau.rs
```diff
@@ -0,0 +1,47 @@
...
+	}
+}
+
+/// The first set of Grandpa authorities known to the pallet.
+///
+/// Note that this doesn't have to be the "genesis" authority set, as the
+/// pallet can be configured to start from any height.
+pub fn initial_authority_set() -> AuthoritySet {
+	let set_id = 0;
+	let authorities = vec![(AuthorityId::from_slice(&[1; 32]), 1)];
```
svyatonik, line 45:
```suggestion
	let authorities = vec![
		(AuthorityId::from_slice(&hex!("88dc3417d5058ec4b4503e0c12ea1a0a89be200fe98922423d4334014fa6b0ee")), 1),
		(AuthorityId::from_slice(&hex!("d17c2d7823ebf260fd138f2d7e27d114c0145d968b5ff5006125f2414fadae69")), 1),
		(AuthorityId::from_slice(&hex!("439660b36c6c03afafca027b910b4fecf99801834c62a5e6006f27d978de234f")), 1),
		(AuthorityId::from_slice(&hex!("5e639b43e0052c47447dac87d6fd2b6ec50bdd4d0f614e4299c665249bbd09d9")), 1),
		(AuthorityId::from_slice(&hex!("1dfe3e22cc0d45c70779c1095f7489a8ef3cf52d62fbd8c2fa38c9f1723502b5")), 1),
	];
```

bin/rialto/runtime/src/millau.rs
svyatonik: (please check this, though - maybe I have some other changes that affect this state in #398)

svyatonik (commented):
LGTM - let's just [fix that in this PR](https://github.com/paritytech/parity-bridges-common/pull/401/files#r501628560) && merge this.

I was able to relay 9 Millau headers to Rialto using this PR and #398 :) Then it got stuck, because substrate pallet works slightly differently than eth contract - will try to fix it asap.

bin/rialto/runtime/src/millau.rs
HCastano: I got the same `extrinsics_root` but a different `state_root`. What I'm going to do is leave my values for this PR and you can change them to ones that work in #398

bin/millau/runtime/src/rialto.rs
HCastano: > manually craft that value - initial authorities are Alice, Bob, Charlie, Dave, Eve and weight is 1 for every authority

I didn't do this because I didn't know what the default weight was for authorities. I guess it is indeed `1` though.

Also, when looking at our Rialto deployment I saw that every time the authorities change only three are selected, not five. I wasn't sure if during genesis our authority list should only consist of the three authorities chosen at the time, or the "full list" which we've set up in our `docker-compose`.

bin/millau/runtime/src/rialto.rs
HCastano: I opened  #407 to track this.

bin/millau/runtime/src/rialto.rs
svyatonik: At genesis it is the full list, then (starting from session 2) only 3 -of-5 authorities are selected

svyatonik (approved):

PR #405, opened 2020-10-08
Millau -> Rialto headers sync - docker deployment

svyatonik:
should be finished after #398 

The idea is not to touch (almost) existing rialto deployment. The only thing that changes is that you need to create docker network before - something like `docker network create bridge-deployments`. Then rialto should work as before.

Millau compose will 'connect' to the same network and start Millau nodes + `millau-headers-to-rialto-relay` (later - reverse and message relays)


deployments/millau/docker-compose.yml
```diff
@@ -0,0 +1,136 @@
...
+    environment:
+      RUST_LOG: rpc=trace,bridge=trace
+    ports:
+      - "19616:9616"
+    depends_on: &all-nodes
+      - bridge-node-alice
+      - bridge-node-bob
+      - bridge-node-charlie
+      - bridge-node-dave
+      - bridge-node-eve
```
HCastano, line 17: Do we need all these nodes for an initial deployment?

deployments/millau/docker-compose.yml
```diff
@@ -0,0 +1,136 @@
...
+      - --node-key=0f900c89f4e626f4a217302ab8c7d213737d00627115f318ad6fb169717ac8e0
+      - --rpc-cors=all
+      - --unsafe-rpc-external
+      - --unsafe-ws-external
+    volumes:
+      - ./bridge-config:/config
+    environment: &bridge-env
+      RUST_LOG: runtime=trace,rpc=debug,txpool=trace
+    ports:
+      - "19933:9933"
```
HCastano, line 41: I'd start this at `20033` and increase the others accordingly

HCastano (commented):
Tbh I'm not sure if this is the right approach to the Millau deployment. I don't like that it's not standalone in the same way that the Rialto deployment is. We shouldn't have to spin up the the Rialto deployment and connect to the correct network for this deployment to work.

What I would like to see from the Millau deployment is the following:
- With one `docker-compose` command we can deploy a network
- Overrides for different networks through `docker-compose` configs, e.g `docker-compose.rialto.yml` or `docker-compose.kusama.yml`

Also, for whatever approach we end up going with can you also write up some documentation in the deployment guide for running a network?

svyatonik:
> Tbh I'm not sure if this is the right approach to the Millau deployment. I don't like that it's not standalone in the same way that the Rialto deployment is.

You could always run something like `docker-compose -f deployments/millau/docker-compose.yml -f deployments/rialto/docker-compose.yml config` to get consolidated deployment config. But having Millau config in separate file seems better option to me than having multiple level overrides.

That said, I'm not using our dockers at all and don't want to stand firm here :) Feel free to introduce your configuration instead.

tomusdrw:
I think it would be cool to use `extends` for this. Basically we move all services to a `common-services.yml` file and then rialto and millau compose files both just extend the basic values.

HCastano:
> I think it would be cool to use `extends` for this. Basically we move all services to a `common-services.yml` file and then rialto and millau compose files both just extend the basic values. See https://docs.docker.com/compose/extends/#understand-the-extends-configuration

@tomusdrw I looked into `extends` and it looks like a good solution. However, support for it was dropped for Compose v3, which is what we use. So unless we want to downgrade our compose files (which I don't think is a good idea) using override files is the way to go (at least with my current understanding of Docker deployments)

Issue #371, opened 2020-09-23
Make Substrate Pallet Storage Abstraction Deeper


HCastano:
The Substrate pallet provides an abstraction to its storage through the `BridgeStorage` trait. Right now this is a very thin wrapper around direct storage access (basically just a pass through). It would be nice to have this module provide a more rich feature set, while keeping the interface simple.

In the future different bridge pallets should be able to implement this interface and provide consistent functionality.

tomusdrw:
Referring mostly to this comment: https://github.com/paritytech/parity-bridges-common/pull/409#discussion_r509366627

I think we have some misunderstanding about what "storage abstraction" is supposed to be doing.
IMO the abstraction is introduced if you don't care about specific details and you want these details to potentially be replaced by something else. Here I meant "substrate-storage abstract", where the main goal is to decouple substrate-specific storage items from the actual light client implementation.
I was pushing for this abstraction to be as low-level (as thin) as possible, cause storage reads and writes are the most expensive things we do in the runtime development, so we should really keep them at the minimal level.

>  It would be nice to have this module provide a more rich feature set, while keeping the interface simple.

Here I feel that you want some "abstraction" to simplify the repetitive tasks that you do in your domain-specific logic (bridge-pallet specific logic). To me this is more about extracting the common patterns into a separate, but concrete struct, if you keep the "low-level storage abstraction" minimal, you can easily build a more complex "domain storage abstraction" on top of it (for instance inserting a header and updating multiple substrate-storage values in one go), but it doesn't mean that this shared logic should be traitified, and what's more it really should not mean that we extend the minimal low-level abstraction to encompass this.
If it contains domain-specific logic (bridge-pallet specific) it's not a substrate-storage abstraction any more for me - the traits should really be minimal, it doesn't make sense to put traits everywhere where we say "abstraction", where in reality we rather mean "code simplficiation / deduplication" - this can safely go to a concrete struct that builds up on the abstract low-level storage.

tomusdrw:
I believe this is not that relevant anymore, since we've opted for a simpler `pallet-bridge-grandpa` approach with only finality proofs. I feel that with BEEFY Light Client (#1223) and especially Dynamic BEEFY Bridging (#1225) we would not need to worry about storage abstraction that much.

Issue #367, opened 2020-09-23
Implement an Efficient Ancestry Checker


HCastano:
In the Substrate pallet we verify ancestry by walking back through every header between the `child` and `ancestor`. While it doesn't happen often, Grandpa has been known to finalize really long chains (upwards of 30,000 blocks). Walking through this many headers in the runtime is inefficient and we should come up with a scheme which only needs to verify a subset of the chain.

Slava suggested using a [Skip List](https://en.wikipedia.org/wiki/Skip_list), which seems like alright to me.

For more info see [here](https://github.com/paritytech/parity-bridges-common/pull/296#pullrequestreview-490374230).

PR #409, opened 2020-10-09
Support Tracking Forks in Substrate Pallet

### LINKED ISSUE #369: Fork Management for Substrate Pallet
The Substrate pallet doesn't handle any forks at the moment. If we try and import two valid headers at the same height we will reject the second one, even though that may eventually become part of the canonical chain. We should be able to track different forks and resolve them correctly once we receive a finalized header.

Related to #296.

HCastano:
While it would be nice to live in a world where we only get blocks from a single, canonical chain, this isn't the world we live in. 
Since we may get multiple blocks which are potential candidates to be part of the canonical chain we need to accept them all (ish) until we can make a decision as to whether or not they're part of the canonical chain (when one of their children gets finalized). This PR introduces support for managing header imports and finality proof imports across different forks.

The main "user facing" (via Runtime APIs) change that comes out of this PR is that methods for getting the best headers and headers which require finality proofs return lists of headers that may live on different forks. The code that creates these lists may do a lot of iterating depending on the state of the chain. While I don't think it's a problem right now (Runtime API calls are somewhat trusted, right?) it would be good to try and get some limits introduced so we don't have unbounded loops.

I've also introduced a dedicated module for reasoning about our header and finality proof imports (`forks.rs`). I recommend that you take a look at the tests there first when reviewing this PR. If those tests are wrong then the logic in the pallet is wrong.

Right now there are some redundant tests between the `forks` module and the `verifier` module. This is because the `forks` module doesn't have as fine grained knowledge of the headers its importing as `verifier` does. I will try and see if I can remove the redundancies though.

Closes #369


modules/substrate/src/verifier.rs
```diff
@@ -127,19 +130,27 @@ where
...
 		// we need to make a note of it.
 		//
-		// TODO: This assumes that we can only have one authority set change pending at a time.
-		// This is not strictly true as Grandpa may schedule multiple changes on a given chain
-		// if the "next next" change is scheduled after the "delay" period of the "next" change
+		// Note: This assumes that we can only have one authority set change pending at a time.
+		// While this is not strictly true of Grandpa (it can have multiple pending changes, even
+		// across forks), this assumption simplifies our tracking of authority set changes.
 		let requires_justification = if let Some(change) = self.storage.scheduled_set_change() {
+			// We don't want to accept a header which schedules an authority set change
```
svyatonik, line 143: It's still a draft - just wanted to clarify. I thought that we're going to support single pending change **per fork**, no? In current implementation it seems that anyone could craft an invalid (because we don't verify headers) header that is scheduling some change and then pallet will stuck waiting for justification of this/next header forever. I thought that we're not going to have more than 1 scheduled change per fork - i.e. if this fork needs justification for header `H`, then we'll reject all headers that are trying to continue this fork until we'll have a justification for `H`. At the same time, the other fork may still be active.

HCastano: Yeah there's still some confusion around this on my end. You're right, we should only allow one pending change _per fork_.

HCastano:
@svyatonik can you check the last commit I pushed and go through those tests to see if the behaviour I wrote out is the expected behaviour? If you have time I'd also appreciate it if you added a commit with some scenarios you think we should cover which are missing. It would help me get this all straight so I can implement it properly :)

modules/substrate/src/verifier.rs
```diff
@@ -924,4 +924,89 @@ mod tests {
...
+			//   \ [2']
+			//
+			// Order: 1, 2, 2', F2, F2'
+			//
+			// Not allowed to finalize 2'
+			// Can't import finality proof since not ancestor of 2
+			//
+			// ---
+			//
+			//   / [2: S|1] <- [3]
```
svyatonik, line 952: Iirc if change is scheduled in 1 block, then it'll be enacted at 3 => block 3 requires justification. I think you mean to have `S|0` here instead. Then everything looks fine.

modules/substrate/src/verifier.rs
```diff
@@ -924,4 +924,89 @@ mod tests {
...
+			//
+			// Order: 1, 2, 3, 2', 3'
+			//
+			// We should not be allowed to import 3
+			//	 Ancestor requires justification since it schedules a change
+			// It should be fine to import 2', and 3'
+			//
+			// ---
+			//
+			//   / [2: S|1] <- [3]
```
svyatonik, line 963: Same here - iirc N is normally 0, which means that the block that is scheduling changes also enacts changes && they start finalizing starting from next block. So should be `S|0` instead.

modules/substrate/src/verifier.rs
```diff
@@ -924,4 +924,89 @@ mod tests {
...
+			// [1] <- [2': S|1] <- [3']
+			//
+			// Order: 1, 2, 2', 3, 3'
+			//
+			// Should not be allowed to import 3 or 3'
+			// Will resolve fork depending on which finality proof for {2|2'} we get first
+			// If we get a proof for 2, how do we prevent 3' from being imported?
```
svyatonik, line 970: You already have this check iirc - you're checking that justification target is strictly larger than best finalized block. But in any case - if we have justifications for competing forks, it means that the chain is doomed.

modules/substrate/src/verifier.rs
```diff
@@ -924,4 +924,89 @@ mod tests {
...
+			//    We're blocked by the scheduled change
+			//    Also check if newly imported blocks are ancestors of `last_finalized`?
```
svyatonik, line 972: Iirc you're already doing this :) But again - if there are >2/3 malicious/misbehaving authorities, then we can't do anything with that.

modules/substrate/src/verifier.rs
```diff
@@ -924,4 +924,89 @@ mod tests {
...
+			//
+			// ---
+			//
+			// [1] <- [2: S|1, F] <- [3: E] <- [4]
+			//
+			// Order: 1, 2, F2, 3, 4
+			//
+			// Since 3 enacts a change, should we wait for it to be finalized
```
svyatonik, line 980: Yes, I suggest to do reject it. Iiuc we may still accept headers that are not scheduling changes, but imo it doesn't worth it. With #394 finality delays should normally be low => we won't lose much time to provide justification for 3

modules/substrate/src/verifier.rs
```diff
@@ -924,4 +924,89 @@ mod tests {
...
+			// before allowing 4 to be imported?
+			//
+			// ---
+			//
+			//                  / [3': E] <- [4']
+			// [1] <- [2: S|1, F] <- [3: E] <- [4]
+			//
+			// Order: 1, 2, F2, 3, 4, 3', 4'
+			//
+			// Only allow import of {3, 3'} after we get F2
```
svyatonik, line 990: This test is slightly invalid given confusion with `S|1` and `S|0`. But the general rule is: if there's some ancestor `A` that is enacting new set, we need to reject justifications for any of its descendants until we'll get justification for `A` itself.

svyatonik:
I've left some comments. Imo the test set you've suggested is enough, even redundant sometimes.

I'm not going to affect any your implementation/design decisions, but just in case - [that's how solidity contract resolves that](https://github.com/svyatonik/substrate-bridge-sol/blob/master/substrate-bridge.sol#L69). Basically - you store `previous_signal_header: Option<(HeaderId)>` with every header in the storage. So you only need single storage read (that happens anyway) to determine if header requires justification. So probably something like that should be done in the pallet.

Please also keep in mind that relay needs IDs of all headers that require justifications. So once you import header that enacts new set (starting from the next block), you need to remember it.

Also there's another important optimization required (well, desired) for relay after this PR will be merged. Given vec of headers that we're going to submit, pallet need to return headers that will be rejected because their ancestors require justification. Without that optimization, relay will submit `import_header` transactions that will be rejected by the pallet => lose its funds. So pallet needs API (not the `Call` - just API) like that:
```rust
fn headers_that_will_be_rejected(headers_to_import: Vec<BridgedHeader>) -> Vec<HeaderId> {
  <return all headers that will be rejected if pallet will import headers from headers_to_import by-order>
}
```

The contract [has API like that](https://github.com/svyatonik/substrate-bridge-sol/blob/master/substrate-bridge.sol#L142).

HCastano:
Thanks for the feedback on the tests. I've updated them a bit to address some of the issues.

> I'm not going to affect any your implementation/design decisions, but just in case - [that's how solidity contract resolves that](https://github.com/svyatonik/substrate-bridge-sol/blob/master/substrate-bridge.sol#L69). Basically - you store `previous_signal_header: Option<(HeaderId)>` with every header in the storage. So you only need single storage read (that happens anyway) to determine if header requires justification. So probably something like that should be done in the pallet.

I like this approach, so I think I'll implement something similar here.

> Please also keep in mind that relay needs IDs of all headers that require justifications. So once you import header that enacts new set (starting from the next block), you need to remember it.

Yeah, I'll need to change the code to return headers on different forks which may require justifications. And should it only be headers which enact changes (what is it now), or should their unfinalized ancestors be included as well? I think it should be the first.

> Also there's another important optimization required (well, desired) for relay after this PR will be merged. Given vec of headers that we're going to submit, pallet need to return headers that will be rejected because their ancestors require justification. Without that optimization, relay will submit import_header transactions that will be rejected by the pallet => lose its funds. So pallet needs API (not the Call - just API) like that:

Can you log an issue with the details so I don't forget to do this?

modules/substrate/src/lib.rs
```diff
@@ -222,18 +236,19 @@ impl<T: Trait> Module<T> {
...
+	/// These headers require proofs because they enact authority set changes.
+	pub fn requires_justification() -> Vec<BridgedHeader<T>> {
 		let storage = PalletStorage::<T>::new();
-		let hash = storage.unfinalized_header()?;
-		let imported_header = storage.header_by_hash(hash).expect(
-			"We write a header to storage before marking it as unfinalized, therefore
-			this must always exist if we got an unfinalized header hash.",
-		);
-		Some(imported_header.header)
+		let hashes = storage.unfinalized_headers();
```
svyatonik, line 244: Probably use some other name? There may be a ton of unfinalized headers but only few of these would enact new authorities set => would require implicit justifications. Looking at the code it seems that the function is returning all `unfinalized_headers` here :)

modules/substrate/src/verifier.rs
```diff
@@ -127,22 +134,39 @@ where
...
-			change.height == *header.number()
+		// Note: This assumes that we can only have one authority set change pending per fork at a
+		// time. While this is not strictly true of Grandpa (it can have multiple pending changes,
+		// even across forks), this assumption simplifies our tracking of authority set changes.
+		let mut signal_hash = parent_header.signal_hash;
+		let scheduled_change = find_scheduled_change(&header);
+
+		// Check if our fork is expecting an authority set change
+		let requires_justification = if let Some(hash) = signal_hash {
+			if let Some(pending_change) = self.storage.scheduled_set_change(hash) {
```
svyatonik, line 149: nit: multiple nested `if`s are hard to read, especially when they're accompanied with `else`s. Since you have `unreachable!()` in `else` branch anyway, maybe just change to `.expect()` here?

modules/substrate/src/lib.rs
```diff
@@ -61,25 +59,36 @@ pub(crate) type BridgedBlockHash<T> = HashOf<<T as Trait>::BridgedChain>;
...
 decl_storage! {
 	trait Store for Module<T: Trait> as SubstrateBridge {
+		/// The number of the highest block(s) we know of.
+		ChainTipHeight: BridgedBlockNumber<T>;
 		/// Hash of the header at the highest known height.
-		BestHeader: BridgedBlockHash<T>;
+		///
+		/// If there are multiple headers at the same "best" height
+		/// this will contain all of their hashes.
+		BestHeaders: Vec<BridgedBlockHash<T>>;
```
svyatonik, line 75: Not necessarily a part of this PR, but if we want pruning this should probably be:
```rust
HeadersAtHeight: map hasher(identity) BridgedBlockNumber<T> => Vec<BridgedBlockHash<T>>;
```

instead of `BestHeaders`. Or do you have other ideas on how to organize pruning?

modules/substrate/src/lib.rs
```diff
@@ -309,18 +329,44 @@ impl<T: Trait> BridgeStorage for PalletStorage<T> {
...
+				<BestHeaders<T>>::append(hash);
+				<ChainTipHeight<T>>::put(current_height);
+			}
+			Ordering::Less => {
+				// This is fine. We can still have a valid header, but it might just be on a
+				// different fork and at a lower height than the "best" overall header.
+			}
+		}
+
+		if header.requires_justification {
```
svyatonik, line 353: Probably think about changing that as a part of #371 - if `requires_justification` is `true` then the entry is inserted into `RequiresJustification` in this function. But it is `false`, then it is not removed here, but instead in a separate `update_best_finalized`. That is not super-intuitive

modules/substrate/src/lib.rs
```diff
@@ -309,18 +329,44 @@ impl<T: Trait> BridgeStorage for PalletStorage<T> {
...
 
 	fn write_header(&mut self, header: &ImportedHeader<BridgedHeader<T>>) {
-		let hash = header.header.hash();
+		use core::cmp::Ordering;
+
+		let hash = header.hash();
+		let current_height = header.number();
+		let best_height = <ChainTipHeight<T>>::get();
+
+		match current_height.cmp(&best_height) {
```
svyatonik, line 338: Q: what if we are finalizing header that has best height less than `ChainTipHeight`, but all headers in the `BestHeaders` collection are now abandoned (i.e. they have forked from finalized branch before best finalized header)

modules/substrate/src/verifier.rs
```diff
@@ -258,19 +279,25 @@ where
 		// new authority set change. When we finalize the header we need to update the current
 		// authority set.
 		if header.requires_justification {
+			let signal_hash_proof = "When we import a header we only mark it as
```
svyatonik, line 282: Can't add comment to unchanged line, so leaving here :)

Could probably be changed as a part of #367 . Now you can change this line above:
```rust
let requires_justification = ancestors.iter().skip(1).find(|h| h.requires_justification);
```

to something that will look read scheduled change from `header.signal_hash` and will check if this change should be enacted by the `header`. So you will only need ancestry to update `is_finalized` flag. Later, when we will add skip-list (or whatever else), we may remove ancestry reads at all => the function would be `O(1)` instead of `O(N)` now.

svyatonik (commented):
Looks good at first glance, but needs a fix for the `MillauHeadersToRialto` (`best_block` -> `best_blocks`). Will give another look tomorrow

modules/substrate/src/lib.rs
```diff
@@ -246,11 +261,11 @@ pub trait BridgeStorage {
 	/// Write a header to storage.
 	fn write_header(&mut self, header: &ImportedHeader<Self::Header>);
 
-	/// Get the header at the highest known height.
-	fn best_header(&self) -> ImportedHeader<Self::Header>;
+	/// Get the header(s) at the highest known height.
+	fn best_headers(&self) -> Vec<ImportedHeader<Self::Header>>;
```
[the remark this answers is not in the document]
svyatonik: Best block number is not enough - relayer would also need best block hash (or a fn to query hash by number).

primitives/millau/src/lib.rs
```diff
@@ -55,11 +55,13 @@ sp_api::decl_runtime_apis! {
...
 	pub trait MillauHeaderApi {
-		/// Returns number and hash of the best block known to the bridge module.
+		/// Returns number and hash of the best blocks known to the bridge module.
+		///
+		/// Will return multiple headers if there are many headers at the same "best" height.
 		///
 		/// The caller should only submit an `import_header` transaction that makes
 		/// (or leads to making) other header the best one.
-		fn best_block() -> (BlockNumber, Hash);
+		fn best_blocks() -> Vec<(BlockNumber, Hash)>;
```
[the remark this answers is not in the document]
HCastano: > So I guess using constants for API methods names is still fragile :) This change breaks `MillauHeadersToRialto` relay.

Whoops, yeah it is! Will fix.

modules/substrate/src/lib.rs
```diff
@@ -246,11 +261,11 @@ pub trait BridgeStorage {
...
 
-	/// Update the header at the highest height.
-	fn update_best_header(&mut self, hash: <Self::Header as HeaderT>::Hash);
+	/// Update the header(s) at the highest height.
+	fn update_best_headers(&mut self, hash: <Self::Header as HeaderT>::Hash);
```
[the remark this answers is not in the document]
HCastano: Yeah I rolled that into `write_headers()` and forgot to remove it

modules/substrate/src/lib.rs
HCastano: I was kicking the pruning can down the road, which is why I opened a separate issue for it, lol

modules/substrate/src/lib.rs
HCastano: Good point. I'm still struggling a bit with getting the abstractions right for storage. I think this specific concern should be something that can be fixed fairly easily in this PR

modules/substrate/src/verifier.rs
```diff
@@ -68,6 +68,13 @@ pub enum ImportError {
 	/// the authority weights being empty or overflowing the `AuthorityWeight`
 	/// type.
 	InvalidAuthoritySet,
+	/// This header is not allowed to be imported since an ancestor requires a finality proof.
+	///
+	/// This can happen if an ancestor is supposed to enact an authority set change.
```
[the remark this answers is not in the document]
HCastano: No. We require finality proofs for headers _enacting_ changes, not those _scheduling_ changes

modules/substrate/src/forks.rs
```diff
@@ -0,0 +1,508 @@
+// Copyright 2020 Parity Technologies (UK) Ltd.
```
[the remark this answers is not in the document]
HCastano: I guess we _could_, but does it provide any tangible benefit?

modules/substrate/src/lib.rs
```diff
@@ -309,18 +329,44 @@ impl<T: Trait> BridgeStorage for PalletStorage<T> {
...
+		let proof = "A header must have been written at genesis, therefore this must always exist";
+		<BestHeaders<T>>::get()
+			.iter()
+			.map(|h| self.header_by_hash(*h).expect(proof))
+			.collect()
 	}
 
-	fn update_best_header(&mut self, hash: BridgedBlockHash<T>) {
-		<BestHeader<T>>::put(hash)
+	fn update_best_headers(&mut self, hash: BridgedBlockHash<T>) {
```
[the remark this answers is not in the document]
HCastano: I forgot to remove this method, its functionality is now part of `write_headers()`

modules/substrate/src/lib.rs
```diff
@@ -309,18 +329,44 @@ impl<T: Trait> BridgeStorage for PalletStorage<T> {
...
 
-	fn best_header(&self) -> ImportedHeader<BridgedHeader<T>> {
-		let hash = <BestHeader<T>>::get();
-		self.header_by_hash(hash)
-			.expect("A header must have been written at genesis, therefore this must always exist")
+	fn best_headers(&self) -> Vec<ImportedHeader<BridgedHeader<T>>> {
+		let proof = "A header must have been written at genesis, therefore this must always exist";
+		<BestHeaders<T>>::get()
+			.iter()
+			.map(|h| self.header_by_hash(*h).expect(proof))
```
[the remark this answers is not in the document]
HCastano: I'm also not happy with the storage abstraction at the moment, which is why #371 is open. However, I disagree with this:

> the methods should be one-to-one mappings of what we have in decl_storage

I think the storage abstraction should provide a few simple methods which provide a lot of functionality (e.g have pallet logic). Each implementation of the pallet will need to do things like write headers to storage, but any other internal accounting they want to do should be up to them.

I don't think this PR is the right place to discuss this though. I'd move the conversation over to #371.

primitives/millau/src/lib.rs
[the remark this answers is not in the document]
HCastano: I changed it to only return the `last()` best header. Why does it not matter which one we choose from the list?

modules/substrate/src/lib.rs
HCastano: Right now we'd continue to return the height of the abandoned fork as the best. I made a note about this in one of the `forks` tests saying that with pruning we'd be eventually consistent. Tomek didn't let that fly, so I gotta change it

modules/substrate/src/lib.rs
```diff
@@ -361,9 +403,9 @@ impl<T: Trait> BridgeStorage for PalletStorage<T> {
 		CurrentAuthoritySet::put(new_set)
 	}
 
-	fn enact_authority_set(&mut self) -> Result<(), ()> {
-		if <NextScheduledChange<T>>::exists() {
-			let new_set = <NextScheduledChange<T>>::take()
+	fn enact_authority_set(&mut self, signal_hash: BridgedBlockHash<T>) -> Result<(), ()> {
+		if <NextScheduledChange<T>>::contains_key(signal_hash) {
+			let new_set = <NextScheduledChange<T>>::take(signal_hash)
```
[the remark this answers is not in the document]
HCastano: The reason I did this is because `take()` requires decoding the storage value, whereas `contains_key()` doesn't. This would help in the case where we don't have the key.

In `verifier` we only call this method in the happy-path, so maybe this optimization isn't really worth it in practice.

primitives/millau/src/lib.rs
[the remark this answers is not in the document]
svyatonik: There are 2 cases for relay:
1) `source.best_header_id` == `target.best_header_id`: then relayer may sleep for a while;
2) if `source.best_header_id` != `target.best_header_id`: then the first thing relayer will do is a check if `source.best_header_id` is known to the `target`. If it is known, then again - the job is done and it'll sleep. If not, it'll start submitting missing headers.

So even if `fn_best_header()` will not return `source.best_header_id`, then relayer will do at most 1 action - it'll check if `source.best_header_id` is known to the target. And then it'll sleep again anyway.

modules/substrate/src/lib.rs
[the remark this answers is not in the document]
HCastano: This has to do with a discussion from a few weeks ago on fork-choice rules. What I got out of it was that since we don't know what the fork-choice rule is for any given Substrate chain we should just default to picking the longest chain

modules/substrate/src/lib.rs
```diff
@@ -309,18 +329,44 @@ impl<T: Trait> BridgeStorage for PalletStorage<T> {
...
+		let current_height = header.number();
+		let best_height = <ChainTipHeight<T>>::get();
+
+		match current_height.cmp(&best_height) {
+			Ordering::Equal => {
+				<BestHeaders<T>>::append(hash);
+			}
+			Ordering::Greater => {
+				<BestHeaders<T>>::kill();
+				<BestHeaders<T>>::append(hash);
```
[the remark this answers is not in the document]
HCastano: Answered elsewhere

modules/substrate/src/lib.rs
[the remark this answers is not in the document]
HCastano: > I don't like that we have to do so many storage reads here. For some cases we later on just need number and hash instead of entire header, this indicates to me that the storage layout is not optimal for the use case we have.

After looking at the usage of `BestHeaders` you're right, we basically only call this when we want header IDs, not full headers. I've changed that

modules/substrate/src/forks.rs
```diff
@@ -0,0 +1,508 @@
...
+// Order: 1, 2, 2', F2, 3, 3'
+//
+//   / [2'] <- [3']
+// [1] <- [2: F] <- [3]
+//
+// In our current implementation we're allowed to keep building on fork 2 for as long as our hearts'
+// content. However, we'll never be able to finalize anything on that fork. We'd have to check for
+// ancestry with `best_finalized` on every import which will get expensive.
+//
+// I think this is fine as long as we run pruning every so often to clean up these dead forks.
```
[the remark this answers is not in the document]
HCastano: I haven't gotten around to addressing this just yet, I need to think about it

modules/substrate/src/forks.rs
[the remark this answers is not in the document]
tomusdrw: Cleaner separation between production and test code imho. When I saw `forks.rs` I initially thought this module will be responsible for forks handling, and it turned out just to be a bunch of tests.

modules/substrate/src/forks.rs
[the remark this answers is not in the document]
tomusdrw: Let's log this out as an issue then and follow up later on.

modules/substrate/src/verifier.rs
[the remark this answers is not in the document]
tomusdrw: I thought we require finality for both, but if we don't it's still okay.

modules/substrate/src/lib.rs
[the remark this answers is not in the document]
tomusdrw: But if you `take` and there is `None`, you don't have to decode anything (`contains_key==false`), if `contains_key==true` you decode the value anyway, so there is actaully one storage read (existence check) that you save with just `take`.

modules/substrate/src/lib.rs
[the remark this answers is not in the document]
tomusdrw: Slava explained that to me off-band, basically the sync algo the relayer implements currently requires the best block header hash to work correctly. We don't really care if it's one or multiple though.

modules/substrate/src/lib.rs
svyatonik: What I mean is that when you'll start work on pruning, you may need to change this again (at least the way how pruning is organized in sibling pallet needs this change). So probably worth to think about pruning now. But it is up to you.

modules/substrate/src/lib.rs
```diff
@@ -96,15 +105,18 @@ decl_storage! {
...
+			<BestHeaders<T>>::put(vec![initial_hash]);
+			<BestFinalized<T>>::put(initial_hash);
 			<ImportedHeaders<T>>::insert(
-				initial_header.hash(),
+				initial_hash,
 				ImportedHeader {
 					header: initial_header,
 					requires_justification: false,
 					is_finalized: true,
+					signal_hash: None,
```
svyatonik, line 119: Are you sure it should be `None`. I think it should be `Some(initial_hash)` if `config.first_scheduled_change.is_some()`

svyatonik: Also probably worth to add something like `assert!(first_scheduled_change.height > initial_header.number)`

modules/substrate/src/lib.rs
```diff
@@ -222,18 +232,19 @@ impl<T: Trait> Module<T> {
...
 	}
 
-	/// Return the latest header which enacts an authority set change
-	/// and still needs a finality proof.
+	/// Returns a list of headers which require finality proofs.
 	///
-	/// Will return None if there are no headers which are missing finality proofs.
-	pub fn requires_justification() -> Option<BridgedHeader<T>> {
+	/// These headers require proofs because they enact authority set changes.
+	pub fn require_justifications() -> Vec<BridgedHeader<T>> {
```
svyatonik, line 238: Probably also worth to change to:
```rust
pub fn require_justifications() -> Vec<(BridgedBlockNumber<T>, BridgedBlockHash<T>)> {}
```
? Or do you think some clients will need **headers**, not the **headers ids**? This isn't critical in any way, though

modules/substrate/src/lib.rs
svyatonik: I agree with the other Tomek comment that it isn't super-critical now, so may be fixed later, in a separate issue

modules/substrate/src/lib.rs
svyatonik: Actually, I just wanted to note that this struct needs some clearer API :) The way you've changed that adds 2 storage ops (read + write) for (almost) every header that we update. And now we update headers that we're finalizing. So when finalizing 10 headers it would be 10 additional reads + 10 additional writes :)

Let's leave it as is for now, but definitely it is something to think about when working on #371

svyatonik: (btw - I think `contains_key` call is redundant - you may just call `remove` in else)

svyatonik (approved):
I haven't reviewed tests, but everything else looks good now.

relays/substrate/src/headers_target.rs
```diff
@@ -88,10 +88,14 @@ where
 		let data = Bytes(Vec::new());
 
 		let encoded_response = self.client.state_call(call, data, None).await?;
-		let decoded_response: (P::Number, P::Hash) =
+		let decoded_response: Vec<(P::Number, P::Hash)> =
 			Decode::decode(&mut &encoded_response.0[..]).map_err(SubstrateError::ResponseParseFailed)?;
 
-		let best_header_id = HeaderId(decoded_response.0, decoded_response.1);
+		let best_header = decoded_response.last().expect(
```
svyatonik, line 94: I don't think this expect conforms to our expect/unwrap rules :) You're in relay code and get data from external component (node -> runtime pallet). There's a lot of components that may fail and it'll lead to empty vector here. So this expect is the same as using `Decode::decode().expect("pallet properly encodes hashes and numbers; qed")` couple of lines above. I'd return some error here instead. But it's just a mumble

modules/substrate/src/forks.rs
```diff
@@ -0,0 +1,508 @@
...
+				}
+			}
+			(Type::Finality(num, fork_id), expected_result) => {
+				let header = map[fork_id]
+					.iter()
+					.find(|h| h.number == *num)
+					.expect("Trying to finalize block that doesn't exist");
+
+				// TODO: Tests pass when I create multiple justifications even though I don't change
+				// these, is that a problem?
```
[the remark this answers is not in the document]
HCastano: I filed #430 for this

modules/substrate/src/lib.rs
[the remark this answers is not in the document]
HCastano: I just took a look at the storage code and you're right. When you `take()` it doesn't decode unless there was a value at the given key

modules/substrate/src/lib.rs
HCastano: I think returning HeaderIds is fine. This will also save us some trips to storage

modules/substrate/src/lib.rs
HCastano: You're right, good catch

relays/substrate/src/headers_target.rs
HCastano: I did have second thoughts when `expect()`-ing yesterday, but figured it was fine. You're right though, we shouldn't be relying on cross-component assumptions like that, so I've changed it to an error. I _might_ be abusing `codec::Error` a little, but I didn't want to introduce a whole new error variant for this

modules/substrate/src/forks.rs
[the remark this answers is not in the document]
HCastano: That makes sense to me. I tried splitting it out and ran into a bunch of errors since I'm using a lot of private types in the tests. I'll look into it more tomorrow - but if I have to expose a bunch of private types to get the compiler happy I think I'll just leave it where it is (maybe with a rename to `fork_tests.rs`)

modules/substrate/src/lib.rs
HCastano: Pruning will likely be the next thing I work on so I'm not concerned about making changes now

modules/substrate/src/forks.rs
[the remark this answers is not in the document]
HCastano: Logged as #454.

modules/substrate/src/forks.rs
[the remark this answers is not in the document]
HCastano: I took another look, and it's a little gross to split into a proper integration test folder right now. I've opted to rename the module to `fork_tests.rs`. Pallets in Substrate typically have a `tests.rs` module within `src/` anyways, so I don't think it's that big a mental jump

svyatonik (approved):

modules/substrate/src/forks.rs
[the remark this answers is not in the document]
tomusdrw: I was thinking about `src/tests` directory not top-level `tests` (meant for integraiton testing). `tests` module would have both `mock.rs` and `fork.rs` files. I'm okay with the prefix for now, but a bit afraid that this becomes a pattern and we will have a mix of test and src files directly in the `src` directory which would become messy.

modules/substrate/src/lib.rs
```diff
@@ -311,18 +343,42 @@ impl<T: Trait> BridgeStorage for PalletStorage<T> {
...
 
-	fn best_header(&self) -> ImportedHeader<BridgedHeader<T>> {
-		let hash = <BestHeader<T>>::get();
-		self.header_by_hash(hash)
-			.expect("A header must have been written at genesis, therefore this must always exist")
+		if header.requires_justification {
+			<RequiresJustification<T>>::insert(hash, current_height);
+		} else if <RequiresJustification<T>>::contains_key(hash) {
+			<RequiresJustification<T>>::remove(hash);
+		}
```
tomusdrw, line 371:
```suggestion
		} else {
			<RequiresJustification<T>>::remove(hash);
		}
```
To minimize storage access - if the key is not there it's a no-op anyway and at least we don't call into the host twice.

modules/substrate/src/verifier.rs
```diff
@@ -163,12 +185,16 @@ where
...
 				let scheduled_change = ScheduledChange {
 					authority_set: next_set,
 					height,
 				};
 
-				self.storage.schedule_next_set_change(scheduled_change);
+				// Note: It's important that the signal hash is updated if a header schedules a
+				// change or else we end up with inconsistencies in other places.
+				signal_hash = Some(hash);
+				self.storage.schedule_next_set_change(hash, scheduled_change);
```
tomusdrw, line 197: IMHO storage writes should be bundled together at the very end of that function. It's too easy to add another condition right after this line with `?` at the end which might have catastrophic consequences.

modules/substrate/src/verifier.rs
```diff
@@ -258,19 +277,25 @@ where
 		// new authority set change. When we finalize the header we need to update the current
 		// authority set.
 		if header.requires_justification {
+			let signal_hash_proof = "When we import a header we only mark it as
+			`requires_justification` if we have checked that it contains a signal hash. Therefore
+			this must always be valid.";
+
+			let enact_set_proof =
```
tomusdrw, line 284: nit: For each of this proofs you could just use `const` instead of `let`.

relays/substrate/src/headers_target.rs
```diff
@@ -90,10 +90,15 @@ where
 		let data = Bytes(Vec::new());
 
 		let encoded_response = self.client.state_call(call, data, None).await?;
-		let decoded_response: (P::Number, P::Hash) =
+		let decoded_response: Vec<(P::Number, P::Hash)> =
 			Decode::decode(&mut &encoded_response.0[..]).map_err(SubstrateError::ResponseParseFailed)?;
 
-		let best_header_id = HeaderId(decoded_response.0, decoded_response.1);
+		let best_header = decoded_response.last().ok_or_else(|| {
```
tomusdrw, line 96: Do we assume that the headers are sorted in some way? If we do, this assumption is not really properly documented. If we don't care about the ordering why do we waste time to encode all of them and just take the last one? The `RuntimeApi` should be tailored for our relayers use case.

tomusdrw (approved):
couple of grumbles, but nothing critical I think.

relays/substrate/src/headers_target.rs
HCastano: > Do we assume that the headers are sorted in some way?

Not really. The `last()` header is just the last header that's been imported into the pallet (at the best height).

> If we don't care about the ordering why do we waste time to encode all of them and just take the last one

There's [this discussion](https://github.com/paritytech/parity-bridges-common/pull/409#discussion_r509053999) where Slava goes over why the ordering doesn't matter.

> The RuntimeApi should be tailored for our relayers use case.

I agree to a certain extent, but I don't see the downside to providing a bit of flexibility with the pallet in that it can return a few of the best headers. The goal of this PR wasn't to change relayer logic, so that change was really the bare minimum to make sure we didn't break the relayer. Going forward the relayer can be updated to use multiple of the best headers in its sync procedure.

modules/substrate/src/verifier.rs
HCastano: I agree, but it's a little annoying to change at the moment. I'll make a note of it though

relays/substrate/src/headers_target.rs
tomusdrw: > I don't see the downside to providing a bit of flexibility with the pallet in that it can return a few of the best headers. 

IMHO the increased maintenance cost* of the code and the fact that we've spent 3 rounds of comments discussing the purpose of this change is enough to justify that if the change is not strictly required for this PR and is there only for some quite vague "future improvements", we should not introduce it in the first place.
I.e. in the current context it introduces more confusion than adds value, especially given that pruning is not implemented and it might simply be reporting some weird data.
If we want to update the relayer to support this new sync procedure, we should just create an issue for this and then apply this change when it's really going to be used.

*It seems that currently we have some extra logic to update `BestHeaders` in `write_header`, a bunch of iterators to return in `RuntimeApi` and extra changes in the relayer (different runtime api to call), just to discard all, expect for one random header in that collection.

That said, I'm fine to keep it, just for the sake of getting the PR merged faster, but going forward I think we should really focus only on things that are actually useful at present or we have a good enough case that it is going to be needed in the near future either by us or external developers (pallet as a library).

HCastano: I still think the pallet should provide all the best headers and it's the relays choice whether or not they want to use them. Handing back one kinda arbitrary header at the pallet level doesn't seem like intuitive behaviour for anybody building on top of us. If we are going to return a sole header we need to really think about what guarantees we want to provide to users of the pallet.

I'm interested to hear @svyatonik's opinion on this since he's driving the relay development.

HCastano:
Despite the kinda on-going discussion about best headers I'm going to merge this. We can address the result of that discussion at a later date.

PR #410, opened 2020-10-10
Add ChainTime associated type

### LINKED ISSUE #51: Check header timestamp against timestamp module

adoerr:
Resolves #51


modules/ethereum/src/import.rs
```diff
@@ -31,13 +31,15 @@ use sp_std::{collections::btree_map::BTreeMap, prelude::*};
 /// we have NOT imported.
 /// Returns error if fatal error has occured during import. Some valid headers may be
 /// imported in this case.
-pub fn import_headers<S: Storage, PS: PruningStrategy>(
+#[allow(clippy::too_many_arguments)]
```
adoerr, line 34: I actually totally agree with clippy here. Some sort of 'import_ctx' struct should be introduced. This would kind of restructure the whole solution, though. So maybe this could/should be done as a separate issue.

modules/ethereum/src/import.rs
```diff
@@ -79,20 +82,22 @@ pub type FinalizedHeaders<S> = Vec<(HeaderId, Option<<S as Storage>::Submitter>)
 /// has returned true.
 ///
 /// Returns imported block id and list of all finalized headers.
-pub fn import_header<S: Storage, PS: PruningStrategy>(
+#[allow(clippy::too_many_arguments)]
```
adoerr, line 85: Same here.

modules/ethereum/src/lib.rs
```diff
@@ -326,6 +328,25 @@ pub trait PruningStrategy: Default {
 	fn pruning_upper_bound(&mut self, best_number: u64, best_finalized_number: u64) -> u64;
 }
 
+/// ChainTime represents the runtime on-chain time
+pub trait ChainTime: Default {
```
tomusdrw, line 332: Why do we need the `Default` bound?

modules/ethereum/src/lib.rs
```diff
@@ -326,6 +328,25 @@ pub trait PruningStrategy: Default {
...
+	/// Is a header timestamp ahead of the current on-chain time.
+	///
+	/// Check whether 'timestamp' is ahead (i.e greater than) the current on-chain
+	/// time. If so, return 'true', 'false' otherwise.
+	fn header_is_ahead(&self, timestamp: u64) -> bool;
```
tomusdrw, line 337:
```suggestion
	fn is_timestamp_ahead(&self, timestamp: u64) -> bool;
```
nit: such name would work a bit better for me, `is_` implies it returns `bool` and it just reads a bit better for me. Since we are passing `timestamp` not `header` I'd also change it too. Not super strong on this though, so just a suggestion.

modules/ethereum/src/lib.rs
```diff
@@ -366,6 +387,8 @@ pub trait Trait<I = DefaultInstance>: frame_system::Trait {
 	type FinalityVotesCachingInterval: Get<Option<u64>>;
 	/// Headers pruning strategy.
 	type PruningStrategy: PruningStrategy;
+	/// ChainTime is a provider for the runtime on-chain time
```
tomusdrw, line 390:
```suggestion
	/// Header timestamp verification against current on-chain time.
```

modules/ethereum/src/verification.rs
```diff
@@ -554,6 +574,27 @@ mod tests {
 		assert_ne!(default_verify(&header), Err(Error::TimestampOverflow));
 	}
 
+	#[test]
+	fn verifies_chain_time() {
+		// header is behind
+		let header = HeaderBuilder::with_parent(&genesis())
+			.timestamp(i32::min_value() as u64)
+			.sign_by(&validator(1));
+		assert_ne!(chain_time_verify(&header), Err(Error::HeaderTimestampIsAhead));
```
tomusdrw, line 583: I think the test would be more valuable if it did `assert_eq` on a concrete expected `Ok/Err` value. Now it might be passing, but for instance just because the header itself is completely malformed.

modules/ethereum/src/lib.rs
```diff
@@ -44,6 +44,8 @@ mod benchmarking;
 
 #[cfg(test)]
 mod mock;
+#[cfg(test)]
+mod mock_chain_time;
```
tomusdrw, line 48: I think I'd rather make `mock.rs` file a module and have `mock::chain_time` there to prevent too many mocks from polluting `src` directory, wdyt?

bin/rialto/runtime/src/kovan.rs
```diff
@@ -134,6 +135,17 @@ impl BridgePruningStrategy for PruningStrategy {
 	}
 }
 
+/// ChainTime provider
+#[derive(Default)]
```
tomusdrw, line 139:
```suggestion
#[derive(Default, RuntimeDebug)]
```

bin/rialto/runtime/src/kovan.rs
```diff
@@ -134,6 +135,17 @@ impl BridgePruningStrategy for PruningStrategy {
 	}
 }
 
+/// ChainTime provider
```
tomusdrw, line 138:
```suggestion
/// PoA Header timestamp verification against `Timestamp` pallet.
```

tomusdrw (commented):
Looks good to me! Couple of tiny grumbles left in the comments.

modules/ethereum/src/lib.rs
svyatonik: Would it be enough just to add `ConstChainTime` to existing `mock.rs` file?

modules/ethereum/src/lib.rs
adoerr: An instance of `ChainTime` is needed for the Dispatchables. Reasoning is that the `Default` bound would save everyone from having to implement `new()`. Or am I missing something here?

modules/ethereum/src/lib.rs
adoerr: Agreed `is_` is more canonical.

modules/ethereum/src/verification.rs
adoerr: Good point!

modules/ethereum/src/lib.rs
adoerr: Started with that. Just adding a `TestConstantChainTimeRuntime` to `mock.rs`. Ended up with some duplicate items. Since I don't like having multiple `mock` files myself, I will give it another try.

modules/ethereum/src/import.rs
adoerr: Done

modules/ethereum/src/lib.rs
```diff
@@ -326,6 +328,25 @@ pub trait PruningStrategy: Default {
 	fn pruning_upper_bound(&mut self, best_number: u64, best_finalized_number: u64) -> u64;
 }
 
+/// ChainTime represents the runtime on-chain time
+pub trait ChainTime: Default {
+	/// Is a header timestamp ahead of the current on-chain time.
+	///
+	/// Check whether 'timestamp' is ahead (i.e greater than) the current on-chain
+	/// time. If so, return 'true', 'false' otherwise.
```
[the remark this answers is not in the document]
adoerr: Done

modules/ethereum/src/lib.rs
svyatonik: Why would you need new test runtime (`TestConstantChainTimeRuntime`) at all? :) Could you just add `type ChainTime = ConstChainTime;` to the existing `TestRuntime`? :)

adoerr: My goal was, to first of all, not break any existing tests. I don't know, whether the constant time chosen will interfere with any assumptions of the existing test. Hence, the new runtime. Can give it a try with just one runtime and see how it goes.

svyatonik: I think most of tests will be fine. The rest needs to be adopted. I think it is better to have single test runtime - having separate runtime for every check is too much imo :)

adoerr: additional runtime removed.

modules/ethereum/src/lib.rs
```diff
@@ -326,6 +326,25 @@ pub trait PruningStrategy: Default {
...
+	///
+	/// Check whether `timestamp` is ahead (i.e greater than) the current on-chain
+	/// time. If so, return `true`, `false` otherwise.
+	fn is_timestamp_ahead(&self, timestamp: u64) -> bool;
+}
+
+/// ChainTime implementation for the empty type.
+///
+/// This implementation will allow a runtime without the timestamp pallet to use
+/// the empty type as it's ChainTime associated type.
```
HCastano, line 341:
```suggestion
/// the empty type as its ChainTime associated type.
```

HCastano (approved):
Looks good to me. Some small comments from Tomek should be fixed before merging though

bin/rialto/runtime/src/rialto_poa.rs
```diff
@@ -109,6 +110,17 @@ impl TPruningStrategy for PruningStrategy {
 	}
 }
 
+/// ChainTime provider
+#[derive(Default)]
+pub struct ChainTime;
+
+impl TChainTime for ChainTime {
```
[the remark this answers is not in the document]
adoerr: I was following the example of `PruningStrategy`. Will use `TChainTime` in both cases.

modules/ethereum/src/lib.rs
adoerr: Done

svyatonik (approved):
Thank you!

adoerr:
I did improve the unit test to assert against a concrete import context. Can I still push that change to the PR?

svyatonik:
@adoerr Sure!

modules/ethereum/src/verification.rs
adoerr: Done

tomusdrw (approved):
Awesome, thanks!

PR #418, opened 2020-10-13
Optimize max messages at inbound lane

### LINKED ISSUE #406: Optimize `MaxUnconfirmedMessagesAtInboundLane`
See details: https://github.com/paritytech/parity-bridges-common/pull/385#discussion_r500283566

MaciejBaj:
The issue was suggested by @tomusdrw and @svyatonik in PR #385 (see a [comment](https://github.com/paritytech/parity-bridges-common/pull/385#discussion_r500283566)). It involves optimising max messages at inbound lane to support batch messages sent by relayers. 

Currently, the lane is exhausted by the number of different messages with no differentiation by single or batch dispatch. With a default `MaxUnconfirmedMessagesAtInboundLane = 16`, a single relayer can exhaust the lane capacity by dispatching 16 single messages.

This PR adds the support for dispatching messages in batches. Relayer can now dispatch 16 messages in batch and still leave 15 slots empty in the inbound lane.

Note: The batch equivalent `imbound_lane::receive_message` is not yet implemented. Functionality was tested by hardcoding data in [tests](https://github.com/paritytech/parity-bridges-common/compare/reward-relayers...MaciejBaj:optimize_max_messages?expand=1#diff-7fc5f13dcb43043e9955075e99fcb798bbd1929ba74ff379a7db94ce4541f0a4R219).

Closes #406.


primitives/message-lane/src/lib.rs
```diff
@@ -79,7 +79,10 @@ pub struct InboundLaneData<RelayerId> {
 	/// 2) all incoming messages are rejected if `proof-of(outbound-lane.state).latest_received_nonce` is
 	///    equal to `this.latest_confirmed_nonce`.
 	/// Given what is said above, all nonces in this queue are in range (latest_confirmed_nonce; latest_received_nonce].
-	pub relayers: VecDeque<(MessageNonce, RelayerId)>,
+	/// Also, when a relayer sends a single message, both of MessageNonces are the same.
+	/// When relayer sends messages in a batch, the first arg is the lowest nonce, second arg the highest nonce.
+	/// Multiple dispatches from the same relayer one are allowed.
+	pub relayers: VecDeque<(MessageNonce, MessageNonce, RelayerId)>,
```
MaciejBaj, line 85: We should probably think of a new limit now per max messages per batch or reuse `MaxUncofirmedMessagesAtInboundeLane` to defacto set overall msgs capacity at `MaxUncofirmedMessagesAtInboundeLane * MaxUncofirmedMessagesAtInboundeLane`

modules/message-lane/src/inbound_lane.rs
```diff
@@ -61,14 +61,24 @@ impl<S: InboundLaneStorage> InboundLane<S> {
...
 			.relayers
 			.front()
-			.map(|(nonce, _)| *nonce <= data.latest_confirmed_nonce)
+			.map(|(_, nonce_high, _)| *nonce_high <= data.latest_confirmed_nonce)
 			.unwrap_or(false)
 		{
 			data.relayers.pop_front();
 		}
+		/// Secondly, update all of the records with lower nonce equal to new confirmed nonce
+		for relayer in data.relayers.iter_mut() {
```
svyatonik, line 74: Am I right that there must be at most 1 entry that falls under `*nonce_low <= data.latest_confirmed_nonce` condition? Basically you may only check `front()`

svyatonik (commented):
Looks good, thank you! But there's small confusion about the original issue, that needs to be fixed :)

> Note: The batch equivalent imbound_lane::receive_message is not yet implemented. Functionality was tested by hardcoding data in tests.

The transaction (runtime `Call`) `fn receive_messages_proof()` already supports messages batch delivery :) So nothing should be changed in module API'. And probably we shouldn't change `InboundLane` API too - existing API is good enough.

The idea is that when `InboundLane::receive_message()` is called, you check if previous message(s) has also been delivered by the same relayer. If so, you're simply extending the nonces range. If not, you should add entry to the `relayers` deque.

Hope it makes sense :)

MaciejBaj:
> Looks good, thank you! But there's small confusion about the original issue, that needs to be fixed :)
> 
> > Note: The batch equivalent imbound_lane::receive_message is not yet implemented. Functionality was tested by hardcoding data in tests.
> 
> The transaction (runtime `Call`) `fn receive_messages_proof()` already supports messages batch delivery :) So nothing should be changed in module API'. And probably we shouldn't change `InboundLane` API too - existing API is good enough.
> 
> The idea is that when `InboundLane::receive_message()` is called, you check if previous message(s) has also been delivered by the same relayer. If so, you're simply extending the nonces range. If not, you should add entry to the `relayers` deque.
> 
> Hope it makes sense :)

Thanks for clarifying. I wanted to go this way at the first place and merge the records for each relayer while dispatching (in `imbound_lane::receive_message`) but then each relayer would be a distinct value in the relayers set and I guess a `HashMap` would have been better, not `VecDeque` as initially suggested in the comment. And if that's the case then for each relayer we should still keep the distinct dispatches (like now), bc what's in the case if we receive single messages in that order: 
[`relayer_a -> msg #1`, `relayer_b -> msg #2`, `relayer_a -> msg #3`]. 
As of the current PR state the deque looks: `[1, 1, relayer_a], [2, 2, relayer_b], [3, 3, relayer_a]`.
Before it was `[1, relayer_a], [2, relayer_b], [3, relayer_a]` and with your suggested changes: `[1, 3, relayer_a], [2, relayer_b]`, which is wrong bc relayer_a never dispatched the message #2. 

Lmk how to proceed and if I'm missing something.

svyatonik:
`VecDeque` is used because we need to confirm messages in-order (or we'll need some complex struct on the source chain). So `HashMap` won't work here :/

I'm not suggesting to have `[1, 3, relayer_a], [2, relayer_b]`. You're right - that is obviously wrong. What I'm suggesting is the code like that in the `InboundLane::receive_message`:
```rust
if data.relayers.last().map(|(_, _, last_relayer)| last_relayer == current_relayer).unwrap_or(false) {
  data.relayers.back_mut().1 = message_nonce;
} else {
  data.relayers.push_back((message_nonce, message_nonce, current_relayer));
}
```

So you can only update **last entry** in the deque if relayer is the same. And since messages are accepted in-order, relayers `VecDeque` will also have relayers sorted by delivered message nonce. In your example it'll be `[1, 1, relayer_a], [2, 2, relayer_b], [1, 1, relayer_a]`.

So this optimization is ideal for single relayer case - then we'll always have at most 1 entry in the deque. If there are more than one relayer, there can be any number of entries (up to `MaxUnconfirmedMessagesAtInboundLane`)

MaciejBaj:
Got it, thanks for clarifying @svyatonik. Will do as suggested 👍

modules/message-lane/src/inbound_lane.rs
MaciejBaj: You're right because the messages don't overlap for relayers I can just check the front. Updated in the latest commit.

modules/message-lane/src/inbound_lane.rs
```diff
@@ -61,14 +62,25 @@ impl<S: InboundLaneStorage> InboundLane<S> {
...
 		}
+		// Secondly, update the next record with lower nonce equal to new confirmed nonce if needed.
+		// Note: There will be max. 1 record to update as we don't allow messages from relayers to overlap.
+		if data
+			.relayers
+			.front()
+			.map(|(nonce_low, _, _)| *nonce_low <= data.latest_confirmed_nonce)
+			.unwrap_or(false)
+		{
+			data.relayers.front_mut().unwrap().0 = data.latest_confirmed_nonce + 1;
```
HCastano, line 82: We can't have calls to `unwrap()` in runtime code. The last thing we need is inconsistent storage after a panic. Since it looks like we check `front()` exists before entering this block you can replace this with an `expect()` and write why this can't panic

modules/message-lane/src/inbound_lane.rs
```diff
@@ -93,7 +105,18 @@ impl<S: InboundLaneStorage> InboundLane<S> {
...
 		data.latest_received_nonce = nonce;
-		data.relayers.push_back((nonce, relayer));
+
+		if data
+			.relayers
+			.back()
+			.map(|(_, _, last_relayer)| *last_relayer == relayer)
+			.unwrap_or(false)
+		{
+			data.relayers.back_mut().unwrap().1 = nonce;
```
HCastano, line 115: Same thing about the `unwrap()` here

modules/message-lane/src/inbound_lane.rs
```diff
@@ -61,14 +62,25 @@ impl<S: InboundLaneStorage> InboundLane<S> {
...
 		{
 			data.relayers.pop_front();
 		}
+		// Secondly, update the next record with lower nonce equal to new confirmed nonce if needed.
+		// Note: There will be max. 1 record to update as we don't allow messages from relayers to overlap.
+		if data
+			.relayers
+			.front()
+			.map(|(nonce_low, _, _)| *nonce_low <= data.latest_confirmed_nonce)
+			.unwrap_or(false)
```
HCastano, line 80: I see this more as a `filter()` than a `map()` operation, so what do you think about doing `filter().is_some()` instead? More legible, or nah? I know the original code had the `map().unwrap_or(false)` pattern, so don't worry about changing it, just curious

HCastano (commented):
Some small comments, but it's getting there :)

modules/message-lane/src/inbound_lane.rs
MaciejBaj: Pro tip! I switched `unwrap()` to `expect()` in both places.

modules/message-lane/src/inbound_lane.rs
```diff
@@ -216,6 +270,84 @@ mod tests {
...
+			assert!(lane.receive_message::<TestMessageDispatch>(
+				TEST_RELAYER_B,
+				2,
+				message_data(REGULAR_PAYLOAD).into()
+			));
+		});
+	}
+
+	#[test]
+	fn correctly_receives_following_messages_from_two_relayers_alternately() {
```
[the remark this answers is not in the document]
MaciejBaj: Yep, that's a good point, `two_relayers_alternately` covers `two_relayers`. I added this text to ensure that case we discussed [here](https://github.com/paritytech/parity-bridges-common/pull/418#issuecomment-708351529) works. Will remove the `two_relayers` test.

modules/message-lane/src/inbound_lane.rs
```diff
@@ -113,9 +136,14 @@ mod tests {
...
 		inbound_lane,
-		mock::{message_data, run_test, TestMessageDispatch, TestRuntime, REGULAR_PAYLOAD, TEST_LANE_ID, TEST_RELAYER},
+		mock::{
+			message_data, run_test, AccountId, TestMessageDispatch, TestRuntime, REGULAR_PAYLOAD, TEST_LANE_ID,
+			TEST_RELAYER,
+		},
 		DefaultInstance, RuntimeInboundLaneStorage,
 	};
+	const TEST_RELAYER_A: AccountId = TEST_RELAYER;
+	const TEST_RELAYER_B: AccountId = TEST_RELAYER + 1;
```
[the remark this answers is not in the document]
MaciejBaj: I declared them in mock and used across the message-lane test suite.

modules/message-lane/src/inbound_lane.rs
```diff
@@ -216,6 +270,84 @@ mod tests {
...
+	fn correctly_receives_same_messages_from_two_relayers() {
+		run_test(|| {
+			let mut lane = inbound_lane::<TestRuntime, _>(TEST_LANE_ID);
+			assert!(lane.receive_message::<TestMessageDispatch>(
+				TEST_RELAYER_A,
+				1,
+				message_data(REGULAR_PAYLOAD).into()
+			));
+			assert_eq!(
+				false,
```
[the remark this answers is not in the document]
MaciejBaj: I corrected the test name.

modules/message-lane/src/inbound_lane.rs
MaciejBaj: Addressed.

modules/message-lane/src/inbound_lane.rs
```diff
@@ -203,19 +236,126 @@ mod tests {
...
 			assert_eq!(lane.storage.data().latest_received_nonce, 0);
 		});
 	}
 
+	#[test]
+	fn fails_to_receive_messages_above_max_limit_per_lane() {
+		run_test(|| {
+			let mut lane = inbound_lane::<TestRuntime, _>(TEST_LANE_ID);
+			let max_nonce = <TestRuntime as crate::Trait>::MaxUnconfirmedMessagesAtInboundLane::get();
+			for current_nonce in 1..max_nonce {
```
svyatonik, line 286: Imo this test fails because it receives 'nonces' in range `[1; max_nonce)` and then with `max_nonce + 1`. So the message with `nonce = max_nonce` => we're trying to receive messages out-of-order => `receive_message` returns false.

So this test isn't actually about `max_limit_per_lane` :) Could you, please, slightly fix it to actually fail because it has more entries than `MaxUnconfirmedMessagesAtInboundLane`? Iiuc the only thing to change is to receive messages from different relayers && then to relay message with `max_nonce` (not `max_nonce + 1`).

modules/message-lane/src/inbound_lane.rs
MaciejBaj: I feel like how it was done before as it expresses how the deque is organised - by oldest messages at the front thus we start from the front the pop only until the set is updated. `filter()` would go over all of the values ignoring the fact they're ordered, so maybe let's keep it as it is?

modules/message-lane/src/lib.rs
```diff
@@ -335,22 +335,24 @@ decl_module! {
...
-					}
-
-					let message_data = OutboundMessages::<T, I>::get(MessageKey {
-						lane_id,
-						nonce,
-					}).expect("message was just confirmed; we never prune unconfirmed messages; qed");
+				for (nonce_low, nonce_high, relayer) in lane_data.relayers {
+					for nonce in nonce_low..nonce_high + 1 {
+						// we may receive the same confirmation multiple times
+						if nonce_high < received_range.0 {
```
svyatonik, line 341: 1) I think you wanted to compare `nonce`, not `nonce_high` here, right?
2) let's move before the inner loop instead. Something like that:
```rust
let nonces_begin = sp_std::cmp::max(nonce_low, received_range.0);
let nonces_end = sp_std::cmp::min(nonce_high, received_range.1);
for nonce in nonces_begin..=nonces_end {
 ...
}
```
The second check (`nonces_end`) isn't absolutely necessary iirc - just in case...

MaciejBaj: This is really good. I updated the code accordingly. Thanks for spotting the mistake with `if nonce_high < received_range.0` - you're absolutely right it was supposed to be `nonce`. What it should be is actually a check 
```
if nonce_low > nonce_end {
 continue;
}
```
to avoid rewarding the relayers that dispatched messages ahead of the received range. The inverse case for received ranges from before the dispatched messages range are already discarded by `lane.confirm_delivery()`.

modules/message-lane/src/inbound_lane.rs
MaciejBaj: Thanks for spotting, I corrected the test 👍

modules/message-lane/src/lib.rs
```diff
@@ -336,12 +336,13 @@ decl_module! {
...
-					for nonce in nonce_low..nonce_high + 1 {
-						// we may receive the same confirmation multiple times
-						if nonce_high < received_range.0 {
-							continue;
-						}
-
+					let nonce_begin = sp_std::cmp::max(nonce_low, received_range.0);
+					let nonce_end = sp_std::cmp::min(nonce_high, received_range.1);
+					// don't proceed if this messages entry is ahead of received range.
+					if nonce_low > nonce_end {
```
svyatonik, line 342: This check is actually redundant now :) It is done by the for loop. Let's remove it probably?

MaciejBaj: You're right, bc of the range.

svyatonik (approved):
Looks good now, thanks!

modules/message-lane/src/inbound_lane.rs
HCastano: Well we're performing both the `map()` and `filter()` calls on an `Option` after calling `front()`, so we wouldn't be going through all the values.

Again, I was just curious as to what you thought, keep the code as is

modules/message-lane/src/mock.rs
```diff
@@ -115,7 +115,13 @@ impl Trait for TestRuntime {
 }
 
 /// Account id of test relayer.
-pub const TEST_RELAYER: AccountId = 100;
+pub const TEST_RELAYER_A: AccountId = 100;
+
+/// Account id additional test relayer - B.
```
HCastano, line 120:
```suggestion
/// Account id of additional test relayer - B.
```

modules/message-lane/src/mock.rs
```diff
@@ -115,7 +115,13 @@ impl Trait for TestRuntime {
...
+pub const TEST_RELAYER_B: AccountId = 101;
+
+/// Account id additional test relayer - C.
```
HCastano, line 123:
```suggestion
/// Account id of additional test relayer - C.
```

HCastano (approved):
Looks good!

I'm not sure what's up with the CI, but I don't think it's your fault. I'll look into it though

modules/message-lane/src/inbound_lane.rs
```diff
@@ -61,14 +62,28 @@ impl<S: InboundLaneStorage> InboundLane<S> {
...
+			.relayers
+			.front()
+			.map(|(nonce_low, _, _)| *nonce_low <= data.latest_confirmed_nonce)
+			.unwrap_or(false)
+		{
+			data.relayers
+				.front_mut()
+				.expect("already ensured front relayer exists in preceding check")
+				.0 = data.latest_confirmed_nonce + 1;
+		}
```
tomusdrw, line 86:
```suggestion
		match data.relayers.front_mut() {
		   Some(ref mut (nonce_low, _, _)) if *nonce_low < data.latest_confirmed_nonce => {
		     *nonce_low = data.latest_confirmed_nonce + 1;
		   }
		   _ => {},
		}
```

I find this code much more readable.

modules/message-lane/src/inbound_lane.rs
```diff
@@ -93,7 +108,21 @@ impl<S: InboundLaneStorage> InboundLane<S> {
...
+			.map(|(_, _, last_relayer)| *last_relayer == relayer)
+			.unwrap_or(false)
+		{
+			data.relayers
+				.back_mut()
+				.expect("already ensured back relayer exists in preceding check")
+				.1 = nonce;
+		} else {
+			data.relayers.push_back((nonce, nonce, relayer));
+		}
```
tomusdrw, line 124: Here I'd prefer a match as well, but I think you'll run into lifetime problems, so only something like this might work:
```rust
let push_new = match data.relayers.back_mut() {
  Some(ref mut (first_nonce, last_nonce, last_relayer)) if last_relayer == &relayer => {
    *last_nonce = nonce;
    false
  },
  _ => true,
};
if push_new {
  data.relayers.push_back((nonce, nonce, relayer));
}

File: modules/message-lane/src/lib.rs [hunk 16]
```diff
@@ -335,22 +335,22 @@ decl_module! {
...
-
-					let message_data = OutboundMessages::<T, I>::get(MessageKey {
-						lane_id,
-						nonce,
-					}).expect("message was just confirmed; we never prune unconfirmed messages; qed");
+				for (nonce_low, nonce_high, relayer) in lane_data.relayers {
+					let nonce_begin = sp_std::cmp::max(nonce_low, received_range.0);
+					let nonce_end = sp_std::cmp::min(nonce_high, received_range.1);
+					// loop won't proceed if current entry is ahead of received range (begin > end).
+					for nonce in nonce_begin..nonce_end + 1 {
```
tomusdrw, line 342: This may panic if there is no overlap between `(nonce_low, nonce_high)` and `received_range`. Do we validate somewhwre that we won't end up with `nonce_begin > nonce_end`?

svyatonik: wdym? `for _ in 200..100 { }` isn't panicking

tomusdrw: Oh, you're right. I've confused the range itself with slice indexing with an incorrect range. Previous code was not using `received_range.1` though, it seems we didn't filter out correctly or assumed it's done in some other place?

Also I guess it's safe to assume that values are `.0` < `.1`?

svyatonik: > Oh, you're right. I've confused the range itself with slice indexing with an incorrect range. Previous code was not using received_range.1 though, it seems we didn't filter out correctly or assumed it's done in some other place?

Yest - I've commented before that using `received_range.1` isn't strictly necessary, because it is already filtered by `InboundLane`. But just in case :)

> Also I guess it's safe to assume that values are .0 < .1?

Yeah - these values are coming from internal code (`InboundLane`). So no need to reverify

MaciejBaj: Yep, we touched on the range handling the `.0 > .1` case in the comments [here](https://github.com/paritytech/parity-bridges-common/pull/418#discussion_r505378092).

modules/message-lane/src/inbound_lane.rs
MaciejBaj: Addressed.

primitives/message-lane/src/lib.rs
tomusdrw: max number of messages in batch is naturally limited by weight, I don't think we need an extra limit. `MaxUnconfirmedMessagesAtInboundLane` is there to make sure we don't end up having too many items in this collection.

I wonder what happens now if the collection is full (i.e. `.len() == MaxUnconfirmed..`) - is the last relayer still allowed to submit more messages? If yes then it's a potential griefing attack for the relayers. We should disallow submitting any more, even if it doesn't grow the collection - can we add a test for this?

tomusdrw (commented):
LGTM, I'd like to see one more test, where:
1. We populate `relayers` with `MaxUnconfirmedMessagesAtInboundLane` items
2. The last relayer tries to submit one more message.
3. This doesn't end up with `relayers` collection growing, but it should still be disallowed, so that the relayer doesn't end up the only one who can still submit more stuff.

MaciejBaj:
https://github.com/paritytech/parity-bridges-common/pull/418#pullrequestreview-511667612 
I covered the mentioned scenario with the highest relayer posting new messages in the existing test checking dispatches above max limit + added comments. Worked as expected 👍

tomusdrw (approved):

tomusdrw:
Awesome, thanks!

modules/message-lane/src/lib.rs
bkchr: `nonce_begin..=nonce_end` would have done the same :see_no_evil:

bkchr: As + 1.

PR #420, opened 2020-10-15
Pin Rust Nightly Version

HCastano:
We're currently running into problems building Substrate due to the latest Rust 1.49 nightly releases. I've pinned the nightly compiler version to the one used by Polkadot for the `v0.8.25` release from ~a week ago.


HCastano:
🤠

Issue #214, opened 2020-07-17
XCM dispatcher


tomusdrw:
Intepret the payload of a message as XCM: https://github.com/paritytech/xcm-format and then dispatch to the right pallets.

tomusdrw:
See: https://github.com/paritytech/polkadot/blob/e1a5c6b66f31166127b19af1fea95fd6aa833589/xcm/xcm-executor/src/lib.rs#L35

svyatonik:
This is started in #1379, but it is far from the production-ready code

svyatonik:
Should be done now

PR #422, opened 2020-10-16
send_message over lane is now only available to other modules

svyatonik:
Previously it was available to regulat tx submitters as well. We have discussed it recently. The idea is that regular submitter may submit tx to target chain directly, so this is unnecessary.

### DIFF
--- a/modules/message-lane/src/lib.rs
+++ b/modules/message-lane/src/lib.rs
@@ -147,82 +147,6 @@ decl_module! {
 		/// Deposit one of this module's events by using the default implementation.
 		fn deposit_event() = default;
 
-		/// Send message over lane.
-		#[weight = 0] // TODO: update me (https://github.com/paritytech/parity-bridges-common/issues/78)
-		pub fn send_message(
-			origin,
-			lane_id: LaneId,
-			payload: T::OutboundPayload,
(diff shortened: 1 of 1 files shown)


svyatonik:
The interesting thing is how that "caller-pallet" API would be looking:
1) it'll have regular API like `fn transfer_funds_to_other_chain()` and will encode target chain `Call` by itself. But the same thing (encode call) should be done by the submitter, because it needs to sign over call. And if encoding is done differently (because of runtime upgrade or because pallet-call => target chain mapping logic is complicated), then the signature will be wrong and message will be rejected;
2) it'll have API that accepts pre-encoded target chain `Call`. But then the only thing that the pallet does is simply redirects its own calls to the `message-lane` module. The same thing could be done e.g. by filtering messages by the `LaneMessageVerifier`, which is imo simpler than having separate pallet.

So probably just close this PR in favor of `LaneMessageVerifier` that will check that the `Call` that is being relayed matches lane semantics?

tomusdrw:
Hmm, that's indeed a problem, but only with `Call::decode` dispatch. I'd keep both APIs with different verifiers (or just a flag), so that when we switch to #214 the apps can encode their intent (message payload) on their own.

svyatonik:
Ok, then let's keep it as is.

Issue #457, opened 2020-10-26
Chain-specific runtime message-delivery API


svyatonik:
Right now message delivery APIs (`OutboundLaneApi` and `InboundLaneApi`) are not chain-specific and thus the runtime can't provide APIs to bridge more than one chain. The way we have solved this in headers sync is that we have separate API trait for every chain (i.e. `RialtoHeaderApi` and `MillauHeaderApi`). The same should be done here - i.e. we should have `RialtoOutboundLaneApi`, `RialtoInboundLaneApi`, `MillauOutboundLaneApi` and `MillauInboundLaneApi`.

The alternative implementation is to add additional argument (`instance: InstanceId`) to all methods. It is hard-to implement in the runtime (without some macro helpers), but it is more convenient for the relay. So probably worth trying this option and then in separate PR also migrate `HeaderApi` to that version.

tomusdrw:
Related: #446

Issue #446, opened 2020-10-23
Make relayer loops generic


tomusdrw:
Currently for each combination of chains we need to prepare a separate headers & messages sync loop.
There are two main reasons for that:
1. We need to encode a `Call` to a specific runtime when creating the transaction and add the right `SignedExtensions` data.
2. `RuntimeApi`s exposed by the pallets are specific to chains.

For (1) the solution would be to create transactions based on `Metadata` somehow or via RPC. (1) has also a very unwanted consequence of being dependent on the entire runtime code - that could be alleviated by extracting some of the runtime code which allows transaction creation.

For (2) we should probably just add some identifier to the pallet instances and have them share the `RuntimeApi`, but add an extra parameter with instance id.

tomusdrw:
A potentially useful pointer: https://github.com/scs/substrate-api-client/blob/master/src/extrinsic/mod.rs#L41

tomusdrw:
Partially done via recent CLI refactoring (#849), we might keep this around to make the relayer based on metadata only (scale-info stuff).

EmmanuellNorbertTulbure:
@svyatonik is this done? Is still something needed to be done?

serban300:
> For (1) the solution would be to create transactions based on Metadata somehow or via RPC

Looking on [subxt-codegen](https://docs.rs/subxt-codegen/latest/subxt_codegen) and [subxt](https://github.com/paritytech/subxt) to see how the generated code based on metadata would look like and if there's anything that would help us.

serban300:
Building calls based on metadata done in #1812

PR #458, opened 2020-10-26
Integrate message-lane module RPCs into Rialto/Millau nodes

svyatonik:
required for `messages-relay`

(as a side note - this is how alternate implementation of #457 would look like). As we add more bridged chains, the match inside `RialtoMessageLaneKeys`/`MillauMessageLaneKeys` would grow


bin/rialto/runtime/src/millau_messages.rs
```diff
@@ -16,21 +16,41 @@
...
+}
+
+/// Storage key of the Rialto -> Millau message lane state in the runtime storage.
+pub fn outbound_lane_data_key(lane: &LaneId) -> StorageKey {
+	StorageKey(OutboundLanes::<DefaultInstance>::storage_map_final_key(*lane))
+}
+
+/// Storage key of the Millau -> Rialto message lane state in the runtime storage.
+pub fn inbound_lane_data_key(lane: &LaneId) -> StorageKey {
+	StorageKey(InboundLanes::<Runtime>::storage_map_final_key(*lane))
```
svyatonik, line 51: If someone could point me to the issue/typo with `OutboundLanes`/`InboundLanes`, I would be very grateful. As you can see, I have to use `::<DefaultInstance>::` with `OutboundLanes`, but `::<Runtime>::` with `InboundLanes`. All other options lead to compilation error. Both maps are almost identical, so I'm not sure - why's that? Also, I guess, it'll lead to incorrect storage keys returned => incorrect proofs. But I can't prove that right now.

bin/rialto/node/src/service.rs
```diff
@@ -28,13 +28,16 @@
 // =====================================================================================
 // =====================================================================================
```
HCastano, line 29: Same thing here

bin/rialto/runtime/src/millau_messages.rs
HCastano: I think the difference has to do with the `OutboundLanes` map not requiring access to `Runtime`'s associated types (via `T`). This means it just needs something that satisfies the `Instance` type. On the other hand since `InboundLanes` uses `T::InboundRelayer` it requires `Runtime` since that where the associated type comes from. 

What you have now with`InboundLanes::<Runtime>::storage_map_final_key(*lane)` is the same as `<Runtime, DefaultInstance>`, so the types will work out. I'm not sure what problems we may run into if they don't though.

We might want to make these generic over `Instance` so we're not tied to only using the default instance of the pallet.

HCastano (approved):
I'm not a big fan of having to add fields to the `match` for every network we have, but I'm okay moving forward with this for now

bin/rialto/runtime/src/millau_messages.rs
svyatonik: I guess that's the issue, thanks!

bin/millau/node/src/service.rs
```diff
@@ -28,6 +28,8 @@
 // =====================================================================================
 // =====================================================================================
```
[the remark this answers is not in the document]
svyatonik: Why? I'm only modifying the `rpc_extensions_builder` part :) So everything works as before.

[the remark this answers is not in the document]
svyatonik: Ahh, there's also that `MillauMessageLaneKeys` :) I'll move it to `rpc_extensions_builder`

bin/millau/node/src/service.rs
```diff
@@ -161,10 +164,40 @@ pub fn new_full(config: Configuration) -> Result<TaskManager, ServiceError> {
...
 	let rpc_extensions_builder = {
+		// This struct is here to ease update process.
+
+		/// Millau runtime from message-lane RPC point of view.
+		struct MillauMessageLaneKeys;
+
+		impl pallet_message_lane_rpc::Runtime for MillauMessageLaneKeys {
+			fn message_key(&self, instance: &InstanceId, lane: &LaneId, nonce: MessageNonce) -> Option<StorageKey> {
+				match *instance {
+					RIALTO_BRIDGE_INSTANCE => Some(millau_runtime::rialto_messages::message_key(lane, nonce)),
```
tomusdrw, line 175: Seems that at least some of the functions inside `rialto_messages` could easily be made generic, so this implementation could be made simpler - i.e. I think the `instance` match should be made way further in the API (within the module), not here long term.
Just something to keep in mind when working on #446

bin/millau/runtime/src/rialto_messages.rs
```diff
@@ -16,21 +16,41 @@
...
 	RuntimeDebug,
 };
+use pallet_message_lane::{DefaultInstance, InboundLanes, OutboundLanes, OutboundMessages};
+use sp_core::storage::StorageKey;
 use sp_trie::StorageProof;
 
+/// Storage key of the Millau -> Rialto message in the runtime storage.
+pub fn message_key(lane: &LaneId, nonce: MessageNonce) -> StorageKey {
+	let message_key = MessageKey { lane_id: *lane, nonce };
+	let raw_storage_key = OutboundMessages::<Runtime, DefaultInstance>::storage_map_final_key(message_key);
```
tomusdrw, line 40: Why isn't there a `RIALTO_BRIDGE_INSTANCE -> Instance` mapping? All 3 functions added here could stay completely generic over the instance and there would be just single match instead of 3.

tomusdrw: basically we would need:
```
fn dispatch_instance<F, I, R>(id: InstanceId, f: F, params: P) -> R where F: FnOnce<I>(P) -> R, I: Instance {...}
```

tomusdrw (approved):
I imagine that most of the APIs would just pass the instance id further down the line or convert `InstanceId` into a generic call with a proper `Instance` type - this would need a single match instead of matching in every API which is quite error-prone.

Issue #219, opened 2020-07-17
Substrate-dispatch Transactions Generator


tomusdrw:
Write a tool to randomly dispatch some messages to be delivered over the bridge. Should be running on both persistent and ephemeral Millau (and Westend see #218 ) to make sure the bridge is operational.

svyatonik:
I'm going to add subcommand to our substrate relay that will submit `Call(system::remark)` message. This would allow us to test messages pallet and relay. If later we'll add separate balances module for wrapped rialto/millau, it may be extended to support these calls.

svyatonik:
The subcommand is added (`submit-millau-to-rialto-message`), so I'm unassigning myself from this task.

tomusdrw:
Implemented already.

Issue #445, opened 2020-10-23
Runtime upgrade deployment strategy


tomusdrw:
Currently every runtime upgrade will break the relayer, cause the `spec_version` is hardcoded in the transaction generated by the relayer.

While most of the runtime upgrades should not require any changes in the relayer, it's still a manual operation to confirm that.
We could have spec_version being passed as a CLI parameter, in such case the relayers would simply need to be restarted with an updated `spec_version`, but re-compilation (or re-deployment) would not be needed.

Even with some workarounds we should make sure to come up with some strategy about runtime upgrades to ensure the bridge is operational.

We might consider having two relayers running in a "fallback" mode or sth, so that whenever the first one stops producing transactions (because of the spec_version mismatch), the second one takes over (we can pre-deploy it with the new `spec_version`).

Related #538

EmmanuellNorbertTulbure:
Duplicate of #826

PR #460, opened 2020-10-26
SubmitMillauToRialtoMessage subcommand in substrate-relay

svyatonik:
related #219


relays/millau-client/src/lib.rs
```diff
@@ -42,6 +46,68 @@ impl Chain for Millau {
...
+				frame_system::CheckTxVersion::<millau_runtime::Runtime>::new(),
+				frame_system::CheckGenesis::<millau_runtime::Runtime>::new(),
+				frame_system::CheckEra::<millau_runtime::Runtime>::from(sp_runtime::generic::Era::Immortal),
+				frame_system::CheckNonce::<millau_runtime::Runtime>::from(signer_nonce),
+				frame_system::CheckWeight::<millau_runtime::Runtime>::new(),
+				pallet_transaction_payment::ChargeTransactionPayment::<millau_runtime::Runtime>::from(0),
+			),
+			(
+				millau_runtime::VERSION.spec_version,
+				millau_runtime::VERSION.transaction_version,
```
HCastano, line 73: Could these be configurable via the CLI, or is that a bad idea?

relays/substrate/src/cli.rs
```diff
@@ -38,6 +39,52 @@ pub enum Command {
...
+		message: ToRialtoMessage,
+		/// Delivery and dispatch fee.
+		#[structopt(long)]
+		fee: bp_millau::Balance,
+	},
+}
+
+arg_enum! {
+	#[derive(Debug)]
+	/// All possible messages that may be delivered from to Rialto chain.
```
HCastano, line 64:
```suggestion
	/// All possible messages that may be delivered to the Rialto chain.
```

relays/substrate/src/cli.rs
```diff
@@ -38,6 +39,52 @@ pub enum Command {
 		#[structopt(flatten)]
 		prometheus_params: PrometheusParams,
 	},
+	/// Submit message to given Rialto -> Millau lane.
+	SubmitMillauToRialtoMessage {
+		#[structopt(flatten)]
+		millau: MillauConnectionParams,
+		#[structopt(flatten)]
+		millau_sign: MillauSigningParams,
```
HCastano, line 47: Is there any way to give example defaults here?

relays/substrate/src/cli.rs
```diff
@@ -38,6 +39,52 @@ pub enum Command {
...
+		#[structopt(flatten)]
+		rialto_sign: RialtoSigningParams,
```
HCastano, line 49: Same question here

HCastano (approved):

HCastano:
CI errors don't look like they're related to your PR. I think we need to bump Substrate

relays/millau-client/src/lib.rs
svyatonik: It has to be solved as a part of #445 . But generally speaking, encoding/tx format/... may change between versions, and you should only create transactions of the runtime you are compiled with. Unless you're sure that nothing has actually changed, which we can't prove => better not to provide spec versions using cli. Right now all relay transactions are created that way.

relays/substrate/src/cli.rs
svyatonik: I remember Tomek has had some problems with default values when he worked on eth2sub/sub2eth deployments, which he has mentioned in #162. So I haven't added any default values (except for prometheus cfg).

Having default values is imo better than not having them. But if it causes problems, then let's add all arguments explicitly.

PR #464, opened 2020-10-26
Rework Compose Setup for Millau Deployment

### LINKED ISSUE #216: Millau testnet compose file
Set-up a Millau testnet (both persistent and emphemeral) docker compose files. This should be a Substrate<>Substrate test network with headers relayers both sides & currency-exchange & some message dispatch & delivery protocol.

HCastano:
This is a different approach to what Slava presented in #405. The idea here is that we have a bunch of "base" Docker Compose files which we use to build up more complicated services (e.g the bridge). The base compose files will be networks comprised of homogenous nodes. These networks can then be combined with additional components like relays to become a bridge. 

The network compose files can be used to deploy their own individual networks like so:

```bash
docker-compose -f rialto-nodes.yml up
```

To give you an example of how this would work consider the Eth<->Sub bridge. With this new approach we would have two "base" compose files. One for the Ethereum PoA nodes and another for the Rialto Substrate nodes. We can combine these two networks and add the `eth2sub` relay, `sub2eth` relay, `tx-generator`, etc in a third compose file. Deploying the bridge could then be as easy as:

```bash
docker-compose -f ./docker-compose.yml -f ../nodes/rialto-nodes.yml -f ../nodes/eth-poa-nodes.yml up
```

I think this approach gives us some flexibility for adding and deploying bridges to different networks.

Since the Compose commands get a little gross, I've added a new script to help with deployments. It also replaces the old `update.sh` script. One thing worth noting here is that I've gotten rid of the option to update/build from Git. I figured we can do this locally ourselves with `docker build` and then manually replace images within the Compose files we're interested it. The main win of this approach is that we only need to build an image once instead of five+ times.

One confession before I finish though. The Ethereum PoA network can't be spun up standalone at the moment. It requires a `--config` param that's set in the `environment` Compose option. We can't have a default local/dev config and then later override it with the `eth-poa-sub` config because Compose merges entries instead of replacing them (for multiline entries anyways). If you guys have any ideas of how to address this, let me know, otherwise I think it's tough luck for people that want to run a standalone Ethereum network (for now anyways).

Would close #216.


HCastano:
Some small things I still need to fix:

1. I'm running into issues with the `grafana-matrix-notify` component, it can't find a the `GranfanaMatrix.Dockerfile` for some reason
2. I need to update the `state_root` to get the relay working
3. The `update.sh` script needs to be updated since I've removed Git builds

HCastano:
> I'm running into issues with the grafana-matrix-notify component, it can't find a the GranfanaMatrix.Dockerfile for some reason

I'm still running into this and I don't know what's wrong. Both the `eth-poa-sub` and `rialto-millau` deployments use the same `monitoring` Compose file and get called from the same place (root of `deployments`) directory. To reproduce just do `./run rialto-millau`.

tomusdrw:
You have to specify `--project-directory .` in the CLI, otherwise the paths are relative to the first `docker-compose.yml` file passed in `-f` (hence `Front-end.Dockerfile` works).

HCastano:
> You have to specify --project-directory . in the CLI, otherwise the paths are relative to the first docker-compose.yml file passed in -f (hence Front-end.Dockerfile works).

Turns out I had some images cached, which explained why it was failing for one bridge but not the other. Once I sorted that out, Tomek was right - relative paths were the way to fix this. However, the consequence from this is that we now have to specify paths for `volumes` and `entrypoints` from the root of the `deployments` folder. Not the most inconvenient thing, but something to be aware of.

Since we don't have Rialto node images with the most updated `state_root`, I've changed it in this PR. I tested the Railto-Millau deployment with a local Rialto image I built. The nodes and relayer run fine...kinda. The relayer is stuck with the following error:

```
2020-10-30 20:17:22 +0000 ERROR bridge Error retrieving best known Millau header from Rialto node: Request(Request(Error { code: InternalError, message: "Unknown error occurred", data: Some(String("Client(Execution(MethodNotFound(\"MillauHeaderApi_best_blocks\")))")) })). Retrying in 0.415399441
```
I looked through the Runtime APIs and all the names match up, so I'm not entirely sure why it's unhappy. Maybe @svyatonik would know?

Once the issue with the relayer is fixed I think all I have to do is clean up the `run` script and update the `README` a little and we should be good to go.

svyatonik:
> I looked through the Runtime APIs and all the names match up, so I'm not entirely sure why it's unhappy. Maybe @svyatonik would know?

So the problem is with what you've build locally? Probably you're using millau node instead of rialto? I have no other ideas :/

HCastano:
> > I looked through the Runtime APIs and all the names match up, so I'm not entirely sure why it's unhappy. Maybe @svyatonik would know?
> 
> So the problem is with what you've build locally? Probably you're using millau node instead of rialto? I have no other ideas :/

It was a caching issue. After I cleared the Docker cache and rebuilt the Rialto images locally it synced headers 🎉

HCastano:
FYI: When reviewing, I think the best place to start is the deployment `README`.

deployments/README.md
```diff
@@ -1,77 +1,89 @@
...
+networks: `eth-poa-sub` and `rialto-millau`.
 
 ```bash
-cd rialto
-docker-compose pull   # Get the latest images from the Docker Hub
-docker-compose build  # This is going to build images
-docker-compose up     # Start all the nodes
-docker-compose up -d  # Start the nodes in detached mode.
-docker-compose down   # Stop the network.
+./run.sh eth-poa-sub
```
tomusdrw, line 55: Can we make this two separate scripts? It's easier to just look at the files in directory than to figure out possible options from bash case-matching. The shared part could be extracted out.

deployments/README.md
```diff
@@ -1,77 +1,89 @@
...
 
-### Docker-Compose Deployment
+These networks can be deployed using our [`./run.sh`](./run.sh) script. We need to call the script
+from the root of the `deployments` folder since it assumes that for some paths.
 
-To run a full network with two-way bridge functionality and cross-chain transfers you
-may use the `docker-compose.yml` file in the [`rialto`](./rialto) folder. This will pull
-images from the Docker Hub for all the components.
+The first argument it takes is the name of the bridge you want to run. Right now we only support two
+networks: `eth-poa-sub` and `rialto-millau`.
```
tomusdrw, line 52:
```suggestion
networks: `poa-rialto` and `rialto-millau`.
```

IMHO it should be called `poa-rialto` not `eth-poa-sub`. `poa` being just a codename for the `ETH PoA` test network. The name is already used like that in services definitions (e.g. `poa-node-arthur` and `rialto-node-alice`)

deployments/README.md
```diff
@@ -119,12 +189,7 @@ GRAFANA_ADMIN_PASS=admin_pass
 GRAFANA_SERVER_ROOT_URL=%(protocol)s://%(domain)s:%(http_port)s/
 GRAFANA_SERVER_DOMAIN=server.domain.io
 MATRIX_ACCESS_TOKEN="access-token"
-WITH_GIT=1   # Optional
 WITH_PROXY=1 # Optional
```
tomusdrw, line 192: I think our current deployment was actually using `GIT` version to make sure we always run the latest code, but I'm not 100% sure. I guess we can let it go though.

deployments/bridges/eth-poa-sub/docker-compose.yml
```diff
@@ -0,0 +1,101 @@
...
+      dockerfile: ./bridges/eth-poa-sub/Front-end.Dockerfile
+      args:
+        SUBSTRATE_PROVIDER: ${UI_SUBSTRATE_PROVIDER:-ws://localhost:9944}
+        ETHEREUM_PROVIDER: ${UI_ETHEREUM_PROVIDER:-http://localhost:8545}
+        EXPECTED_ETHEREUM_NETWORK_ID: ${UI_EXPECTED_ETHEREUM_NETWORK_ID:-105}
+    ports:
+      - "8080:80"
+
+  # Note: These are being overridden from the top level `monitoring` compose file.
+  prometheus-metrics:
```
tomusdrw, line 89: I now see why monitoring is required. Can we just provide some dummy image (maybe `scratch`?) to make sure this service is complete that is later on overwritten by including `monitoring.yml`?

deployments/bridges/eth-poa-sub/docker-compose.yml
```diff
@@ -0,0 +1,101 @@
...
+  # We provide an override for this particular node since this is a public facing
+  # node which we use to connect from things like Polkadot JS Apps.
+  rialto-node-charlie:
+    environment:
+      VIRTUAL_HOST: rialto.bridges.test-installations.parity.io,wss.rialto.brucke.link
+      VIRTUAL_PORT: 9944
+      LETSENCRYPT_HOST: rialto.bridges.test-installations.parity.io,wss.rialto.brucke.link
+      LETSENCRYPT_EMAIL: admin@parity.io
+
+  relay-eth2sub:
```
tomusdrw, line 24:
```suggestion
  relay-headers-poa2rialto:
```
Let's use consistent naming everywhere now.

deployments/bridges/eth-poa-sub/docker-compose.yml
```diff
@@ -0,0 +1,101 @@
...
+    image: paritytech/ethereum-poa-relay
+    entrypoint: /config/relay-eth2sub-entrypoint.sh
```
tomusdrw, line 26:
```suggestion
    entrypoint: /relay/headers-poa2rialto-entrypoint.sh
```
?

deployments/bridges/rialto-millau/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
+version: '3.5'
+services:
+  relay-millau-headers-to-rialto:
```
tomusdrw, line 3:
```suggestion
  relay-headers-millau2rialto
```
or `relay-headers-millau-rialto` would be my preference. I think we should also be consistent with naming now.

deployments/bridges/rialto-millau/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+    image: paritytech/substrate-relay
+    entrypoint: /config/relay-millau-headers-to-rialto-entrypoint.sh
```
tomusdrw, line 5: Same here. I think we can actually have `/entrypoints/relay-headers-millau-rialto.sh`  or `/relay/headers-millau-rialto.sh`

deployments/bridges/rialto-millau/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+    volumes:
+      - ./bridges/rialto-millau/scripts:/config
```
tomusdrw, line 7:
```suggestion
      - ./bridges/rialto-millau/entrypoints:/entrypoints
```
?

deployments/monitoring/GrafanaMatrix.Dockerfile
```diff
@@ -8,6 +8,7 @@ RUN mkdir $APP_HOME
 WORKDIR $APP_HOME
 
 RUN git clone https://github.com/ananace/ruby-grafana-matrix.git $APP_HOME
+RUN git checkout 0d662b29633d16176291d11a2d85ba5107cf7de3
```
tomusdrw, line 11: Can you add a comment here? Is there something that changed and is not compatible now?

deployments/run.sh
```diff
@@ -0,0 +1,90 @@
...
+#
+# `./run.sh rialto-millau update`
+#
+# Once you've stopped having fun with your deployment you can take it down with:
+#
+# `./run.sh rialto-millau stop`
+
+set -xeu
+
+RIALTO=' -f ./networks/docker-compose.rialto.yml'
```
tomusdrw, line 20:
```suggestion
RIALTO=' -f ./networks/rialto.yml'
```
IMHO this stuff is redundant, the deployment-level README explains that these files are docker-compose files, we've touch on this off-band, but I'm still not convinced what the value of adding this redundancy is :)

deployments/run.sh
```diff
@@ -0,0 +1,90 @@
...
+MILLAU=' -f ./networks/docker-compose.millau.yml'
+ETH_POA=' -f ./networks/docker-compose.eth-poa.yml'
+MONITORING=' -f ./monitoring/docker-compose.yml'
+
+BRIDGE=''
+NETWORKS=''
+case "$1" in
+	eth-poa-sub)
```
tomusdrw, line 28: As in some other comment, I think this might rather be two separate bash scripts with a shared part, but not super strong on this, since we only have two bridges now.

deployments/scripts/run_eth2sub.sh
```diff
@@ -1,19 +0,0 @@
-#!/bin/bash
```
tomusdrw, line 1: Can we add some reference to https://github.com/svyatonik/parity-bridges-common.test in the README given these scripts are now removed?

tomusdrw (commented):
Looks good, a bunch of small improvements, some are my personal taste (so not that important), but I'd make sure we are consistent in naming things.

deployments/README.md
HCastano: No, it's using the nightly Docker images

deployments/monitoring/GrafanaMatrix.Dockerfile
HCastano: I explained it in the commit message of this change, but I can add a comment.

tl;dr new commits to master and running it is a little different now

deployments/run.sh
HCastano: I don't really see the value in different scripts at this point in time. Especially since the only split in logic is choosing two strings

deployments/scripts/run_eth2sub.sh
HCastano: I'd rather see us develop some simplified test/dev scripts for the repo, and leave the more complicated setups (like some in that repo) as something do be done with Compose.

deployments/README.md
```diff
@@ -1,77 +1,89 @@
...
-Its main purpose is to make sure that basic PoA<>Substrate bridge operation works.
-The network is being reset every now and then without a warning.
+### Running and Updating Deployments
+We currently support two bridge deployments
+1. Ethereum PoA to Rialto Substrate
+2. Rialto Substrate to Millau Substrate
 
-### Docker-Compose Deployment
+These networks can be deployed using our [`./run.sh`](./run.sh) script. We need to call the script
+from the root of the `deployments` folder since it assumes that for some paths.
```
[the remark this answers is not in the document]
HCastano: Why do we need this `>/dev/null 2>&1 ` after the `dirname` call?

This is kinda gross, but definitely useful here

deployments/README.md
HCastano: Addressed elsewhere

deployments/run.sh
HCastano: All the docs I've read on override files have the prefix, but since we basically only run these with the `-f` flag anyways it doesn't matter. I've removed the prefix from the names

deployments/bridges/eth-poa-sub/docker-compose.yml
HCastano: We can't do this since `scratch` isn't actually an image. We _could_ make a dummy image `FROM scratch` and use that, but I'd rather just make the monitoring mandatory at that point

deployments/README.md
```diff
@@ -1,77 +1,89 @@
...
+               -f <monitoring>.yml up
+```
+
+If you want to see how the Compose commands are actually run, check out the source code of the
+[`./run.sh`](./run.sh).
 
-## Rialto
+One thing worth noting is that we have a _monitoring_ Compose file. This adds support for Prometheus
+and Grafana. We cover these in more details in the [Monitoring](#monitoring) section. At the moment
+the monitoring Compose file is _not_ optional, and must be included for bridge deployments.
```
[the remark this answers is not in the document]
HCastano: Addressed elsewhere

deployments/README.md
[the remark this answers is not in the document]
tomusdrw: You don't, since we want to `cd` to that directory (not to get it in the variable) you only need:
```
cd "$( dirname "${BASH_SOURCE[0]}" )"
```

deployments/bridges/eth-poa-sub/docker-compose.yml
tomusdrw: We can just use a well established image like:
```
image: alpine:3.6
```

Let's create an issue for this if you refuse to fix this in that PR.

deployments/monitoring/GrafanaMatrix.Dockerfile
tomusdrw: > I explained it in the commit message of this change, but I can add a comment.

Reading commit messages for individual changes should not be required for a proper review, what's more - going back to the PR history to understand why a line is there in the code shouldn't be required either.

Since we squash-merge PRs `git blame` for that line is useless as well.

deployments/scripts/run_eth2sub.sh
tomusdrw: I'm not suggesting to include any of these scripts, just to add a one line in the README to point to this repo since it might be helpful for someone who is not willing to or can't run docker / docker-compose.

deployments/run.sh
tomusdrw: > All the docs I've read on override files have the prefix

It's not really an override, we compose a file from multiple other files - `-f networks/rialto.yml` adds a bunch of unique services, so is `-f networks/millau.yml`. `bridges/` stuff is more like an override.

deployments/bridges/eth-poa-sub/docker-compose.yml
HCastano: Logged here: #484

tomusdrw (approved):

PR #469, opened 2020-10-28
Return cumulative dispatch weight of messages from the `messageLane_proveMessages`

### LINKED ISSUE #468: Change `messageLane_proveMessages` to return cumulative dispatch weights of messages
required for #399

(most probably will make PR for this today)

svyatonik:
...and also implement `InboundLaneApi`/`OutboundLaneApi` for Millau and Rialto runtimes. Apparently that's another missing component from #399 :)

closes #468


bin/rialto/runtime/src/lib.rs
```diff
@@ -463,7 +464,7 @@ construct_runtime!(
 		BridgeKovanCurrencyExchange: pallet_bridge_currency_exchange::<Instance2>::{Module, Call},
 		BridgeMillau: pallet_substrate_bridge::{Module, Call, Storage, Config<T>},
 		BridgeCallDispatch: pallet_bridge_call_dispatch::{Module, Event<T>},
-		BridgeMillauMessageLane: pallet_message_lane::{Module, Call, Event<T>},
+		BridgeRialtoMessageLane: pallet_message_lane::{Module, Call, Event<T>},
```
tomusdrw, line 467: Why do we have `BridgeRialto` in the `rialto` runtime? I thought we would rather have `Millau` here, no? (we do have `BridgeMillau` for instance)

primitives/message-lane/src/lib.rs
```diff
@@ -126,6 +126,8 @@ impl Default for OutboundLaneData {
 decl_runtime_apis! {
 	/// Outbound message lane API.
 	pub trait OutboundLaneApi {
+		/// Returns dispatch weight of all messages in given inclusive range.
+		fn messages_dispatch_weight(lane: LaneId, begin: MessageNonce, end: MessageNonce) -> Weight;
```
tomusdrw, line 130: So the relayer will have some strategy of bin-searching the right nonce range to make sure it fits in a single transaction, right?

tomusdrw (approved):
lgtm!

bin/rialto/runtime/src/lib.rs
svyatonik: I can't say why I've changed that :/

primitives/message-lane/src/lib.rs
svyatonik: Something like that, right

tomusdrw (approved):

HCastano (approved):

Issue #436, opened 2020-10-23
Value-spend safety valve for the relayer


tomusdrw:
In case of a bug in relayer code (or general in the bridge), the relayer may end up sending excessive amount of transactions. To prevent this, the relayer should keep track of the amounts spend on transactions and there should be a CLI configuration option, which ensures a per-unit-of-time limit.

For instance we might want to introduce a daily limit or a 6-h rolling average as a safety valve for the relayer.

tomusdrw:
Closed via #470

PR #470, opened 2020-10-28
Substrate relay guards

svyatonik:
implements #436 
related #445 

This PR is introducing two background tasks (guards) which must be started when relay starts. First guard aborts relay when runtime upgrade happens. Second guard aborts relay if relay account has spent too many funds in last 24hrs.

Actually these may be separate processes, but for deploy it'll be easier to run them as threads.

I'm not starting guards for our testnets relay, because it doesn't make sense (we have no runtime upgrades, we have infinite amount of tokens on relay accounts).

### DIFF
--- a/relays/substrate-client/src/chain.rs
+++ b/relays/substrate-client/src/chain.rs
@@ -19,16 +19,22 @@ use crate::client::Client;
 use bp_runtime::Chain as ChainBase;
 use frame_support::Parameter;
 use jsonrpsee::common::{DeserializeOwned, Serialize};
-use sp_core::Pair;
+use num_traits::{CheckedSub, Zero};
+use sp_core::{storage::StorageKey, Pair};
 use sp_runtime::{
 	generic::SignedBlock,
 	traits::{AtLeast32Bit, Dispatchable, MaybeDisplay, MaybeSerialize, MaybeSerializeDeserialize, Member},
 	Justification,
 };
-use sp_std::fmt::Debug;
+use std::{fmt::Debug, time::Duration};
 
 /// Substrate-based chain from minimal relay-client point of view.
 pub trait Chain: ChainBase {
+	/// Chain name.
+	const NAME: &'static str;
+	/// Average block interval.
+	const AVERAGE_BLOCK_INTERVAL: Duration;
+
 	/// The user account identifier type for the runtime.
 	type AccountId: Parameter + Member + MaybeSerializeDeserialize + Debug + MaybeDisplay + Ord + Default;
 	/// Account index (aka nonce) type. This stores the number of previous transactions associated
@@ -38,6 +44,13 @@ pub trait Chain: ChainBase {
 	type SignedBlock: Member + Serialize + DeserializeOwned + BlockWithJustification;
 	/// The aggregated `Call` type.
 	type Call: Dispatchable + Debug;
+	/// Balance of an account in native tokens.
+	type NativeBalance: Parameter + Member + DeserializeOwned + Clone + Copy + CheckedSub + PartialOrd + Zero;
+
+	/// Return runtime storage key for that is storing `pallet_balances::AccountData` of given account.
+	///
+	/// Would panic if chain uses different format to store account data.
+	fn account_data_storage_key(account_id: &Self::AccountId) -> StorageKey;
 }
 
 /// Block with justification.
@@ -62,6 +75,12 @@ pub trait TransactionSignScheme {
 		signer_nonce: <Self::Chain as Chain>::Index,
 		call: <Self::Chain as Chain>::Call,
 	) -> Self::SignedTransaction;
+}
+
+impl BlockWithJustification for () {
+	fn justification(&self) -> Option<&Justification> {
+		None
+	}
 }
 
 impl<Block> BlockWithJustification for SignedBlock<Block> {
--- a/relays/substrate-client/src/chain.rs
+++ b/relays/substrate-client/src/chain.rs
@@ -33,6 +33,8 @@ pub trait Chain: ChainBase {
 	/// Chain name.
 	const NAME: &'static str;
 	/// Average block interval.
+	/// 
+	/// How often blocks are produced on that chain. It's suggested to set this value to match the block time of the chain.
 	const AVERAGE_BLOCK_INTERVAL: Duration;
 
 	/// The user account identifier type for the runtime.
--- a/relays/substrate-client/src/chain.rs
+++ b/relays/substrate-client/src/chain.rs
@@ -44,13 +44,16 @@ pub trait Chain: ChainBase {
 	type SignedBlock: Member + Serialize + DeserializeOwned + BlockWithJustification;
 	/// The aggregated `Call` type.
 	type Call: Dispatchable + Debug;
+}
+
+/// Substrate-based chain with `frame_system::Trait::AccountData` set to
+/// the `pallet_balances::AccountData<NativeBalance>`.
+pub trait ChainWithBalances: Chain {
 	/// Balance of an account in native tokens.
 	type NativeBalance: Parameter + Member + DeserializeOwned + Clone + Copy + CheckedSub + PartialOrd + Zero;
 
-	/// Return runtime storage key for that is storing `pallet_balances::AccountData` of given account.
-	///
-	/// Would panic if chain uses different format to store account data.
-	fn account_data_storage_key(account_id: &Self::AccountId) -> StorageKey;
+	/// Return runtime storage key for getting `frame_system::AccountInfo` of given account.
+	fn account_info_storage_key(account_id: &Self::AccountId) -> StorageKey;
 }
 
 /// Block with justification.
--- a/relays/substrate-client/src/chain.rs
+++ b/relays/substrate-client/src/chain.rs
@@ -33,7 +33,7 @@ pub trait Chain: ChainBase {
 	/// Chain name.
 	const NAME: &'static str;
 	/// Average block interval.
-	/// 
+	///
 	/// How often blocks are produced on that chain. It's suggested to set this value to match the block time of the chain.
 	const AVERAGE_BLOCK_INTERVAL: Duration;
 
--- a/relays/substrate-client/src/client.rs
+++ b/relays/substrate-client/src/client.rs
@@ -17,15 +17,19 @@
 //! Substrate node client.
 
 use crate::chain::Chain;
+use crate::error::Error;
 use crate::rpc::Substrate;
 use crate::{ConnectionParams, Result};
 
+use codec::Decode;
 use jsonrpsee::common::DeserializeOwned;
 use jsonrpsee::raw::RawClient;
 use jsonrpsee::transport::ws::WsTransportClient;
 use jsonrpsee::{client::Subscription, Client as RpcClient};
 use num_traits::Zero;
+use pallet_balances::AccountData;
 use sp_core::Bytes;
+use sp_version::RuntimeVersion;
 
 const SUB_API_GRANDPA_AUTHORITIES: &str = "GrandpaApi_grandpa_authorities";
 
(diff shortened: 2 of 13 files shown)


relays/substrate-client/src/chain.rs
```diff
@@ -19,16 +19,22 @@ use crate::client::Client;
...
 	Justification,
 };
-use sp_std::fmt::Debug;
+use std::{fmt::Debug, time::Duration};
 
 /// Substrate-based chain from minimal relay-client point of view.
 pub trait Chain: ChainBase {
+	/// Chain name.
+	const NAME: &'static str;
+	/// Average block interval.
```
tomusdrw, line 35: Can you expand on the doc here? Is it meant to be equal to block time on the actual chain, or is it some other polling scheme?

relays/substrate-client/src/chain.rs
```diff
@@ -38,6 +44,13 @@ pub trait Chain: ChainBase {
 	type SignedBlock: Member + Serialize + DeserializeOwned + BlockWithJustification;
 	/// The aggregated `Call` type.
 	type Call: Dispatchable + Debug;
+	/// Balance of an account in native tokens.
+	type NativeBalance: Parameter + Member + DeserializeOwned + Clone + Copy + CheckedSub + PartialOrd + Zero;
+
+	/// Return runtime storage key for that is storing `pallet_balances::AccountData` of given account.
+	///
+	/// Would panic if chain uses different format to store account data.
```
tomusdrw, line 52: Why would this panic?

relays/substrate-client/src/guard.rs
```diff
@@ -0,0 +1,370 @@
...
+	async_std::task::spawn(async move {
+		let mut interval_start = env.now();
+		let mut start_balance = None;
+		let mut last_balance = None;
+
+		loop {
+			// if 24 hours have passed, start new interval
+			let current_time = env.now();
+			let duration_since_start = current_time.duration_since(interval_start);
+			if duration_since_start > DAY {
```
tomusdrw, line 98: I thought we will actually have a 24h rolling window instead of full periods. This approach is a bit sub-optimal imho, cause it really depends on when the relay is started. If we have a `maximal_decrease / 2` drop in 23h and then another `maximal_decrease / 2` drop in `1h + 1s` it won't really trigger the check, cause the counter is being reset in the meantime.

relays/substrate-client/src/client.rs
```diff
@@ -100,15 +106,40 @@ where
...
 	}
 
+	/// Return runtime version.
+	pub async fn runtime_version(&self) -> Result<RuntimeVersion> {
+		Ok(Substrate::<C, _, _>::runtime_version(&self.client).await?)
+	}
+
+	/// Return native tokens balance of the account.
+	///
+	/// May panic if called on the chain that is not storing `frame_balances::AccountData`
```
tomusdrw, line 124: It doesn't seem it will panic though. I think it will just return garbage data, no?

tomusdrw (commented):
I think it's still worth to start on the dev relayers, we should file an issue to actually trigger this on the test network to make sure it works.

relays/substrate-client/src/chain.rs
svyatonik: I could change to "Average interval between blocks" :) Not sure how to describe it better. No matter how it would be used (either to poll, or to do any computations), it should be near to average time between blocks on real chain.

relays/substrate-client/src/chain.rs
svyatonik: If chain stores something other than `pallet_balances::AccountData`, then this fn should panic. But actually, maybe it worth additional trait - something like `ChainWithBalancesAccountData`. Will change that

relays/substrate-client/src/guard.rs
svyatonik: Imo it isn't critical in any way. But I agree that your approach is easier to understand. Will change

relays/substrate-client/src/client.rs
svyatonik: It calls `account_data_storage_key`, which may panic

relays/substrate-client/src/chain.rs
svyatonik: Added `ChainWithBalances: Chain`

relays/substrate-client/src/client.rs
svyatonik: Irrelevant

svyatonik:
> I think it's still worth to start on the dev relayers, we should file an issue to actually trigger this on the test network to make sure it works.

I don't think enabling for dev relayers is a good idea now - I can't ever estimate daily transactions cost. And we have never made a runtime upgrade. I've filled #471 to do manual tests later

relays/substrate-client/src/chain.rs
tomusdrw:
```suggestion
	/// Average block interval.
	/// 
	/// How often blocks are produced on that chain. It's suggested to set this value to match the block time of the chain.
```

relays/substrate-client/src/client.rs
tomusdrw: I still don't understand how this would panic, and I suppose it might, since the comment is still here. Can you explain which of the function calls can panic?

relays/substrate-client/src/guard.rs
```diff
@@ -0,0 +1,371 @@
...
+			match current_balance {
+				Ok(current_balance) => {
+					// remember balance
+					balances.push_back((current_time, current_balance));
+
+					// check if difference between current and oldest balance is too large
+					let (oldest_time, oldest_balance) =
+						balances.front().expect("pushed to queue couple of lines above; qed");
+					let balances_difference = oldest_balance.checked_sub(&current_balance);
+					if balances_difference > Some(maximal_decrease) {
```
tomusdrw, line 117: Okay, it's better now, cause we compare with the balances from 24h in the past, but it won't work correctly if the account is topped up in the meantime.

Why can't we store:
```rust
last_balance: Balance;
balance_diffs: Vec<(Timestamp, bool /* sign */, Balance /* difference */)>,
```
and have a logic like this:
```rust
let diff = match current_balance.cmp(self.last_balance) {
  Ordering::Greater => Some((current_time, true, current_balance - self.last_balance)),
  Ordering::Equal => None,
  Ordering::Less => Some((current_time, false, self.last_balance - current_balance)),
};
if let Some(diff) = diff {
  self.balance_diffs.push_back(diff);
}
self.last_balance = current_balance;

let spent = self.balance_diffs.iter()
  .filter(|diff| !diff.1) // only count value drops
  .map(|diff| diff.2)
  .sum();

if spent > maximal_decrease {
 ...
}
```

It's only slightly more complex, but is more straightforward to think about.

tomusdrw (changes requested):
> I don't think enabling for dev relayers is a good idea now - I can't ever estimate daily transactions cost. And we have never made a runtime upgrade. 

I don't see any reason why it shouldn't be enabled. Let's set it arbitrarily - if it's too low we should get an alert that the relayer stopped working (which is good, we will test our monitoring infra), if it's too high it won't be triggered, but at least that code will be running along and we might detect potential cpu issues, etc.

Runtime upgrade is something we might test manually at some point, but I'd rather do it on Millau/Rialto than on someone's private computer.

relays/substrate-client/src/client.rs
svyatonik: Yeah, it should have been removed with `ChainWithBalances`.

relays/substrate-client/src/guard.rs
svyatonik: Well, it "won't work correctly" from your point of view. From mine, it works as it was designed. Your logic also "won't work correctly" if the account balance has been first increased by `maximal_difference` and then decreased by the same amount between two consequent checks (e.g. in the same block). We may go even further and check all events of balances module in all blocks that we see. Then the check will finally would be correct.

But I don't like where it goes. Thing that was designed to be super-simple is now turning into something that both doesn't work as you want and is checking the same thing that has been checked by the original version.

The idea of this guard is not to be super-precise. If you're ok with relayer spending 1-token/24h in regular workload, then I guess losing 2-units is acceptable. But probably relayer should be stopped in this case and you should restart it manually after making sure everything works fine. That's why there's that guard. Not to compute spent sum with absolute accuracy. Just something that is easy to deploy and to maintain. So all 3 options would work && I prefer to revert to original code to keep it simple, or just remove this guard from the relay code && rely on external components. I.e. you could just make script that is started every hour and is transferring max-tokens-to-spend-in-1-hour to relay account.

svyatonik:
> I don't see any reason why it shouldn't be enabled. Let's set it arbitrarily - if it's too low we should get an alert that the relayer stopped working (which is good, we will test our monitoring infra), if it's too high it won't be triggered, but at least that code will be running along and we might detect potential cpu issues, etc.

The reason is simple - I don't want to have relay that aborts just after start. So arbitrary selection is not the right approach. Adding cli options without defaults? Well, maybe, but better to do that once at least we'll have estimations on what's `maximal_decrease` would look like. It'd be even better with weights to estimate real costs on Polkadot/Kusama.

> Runtime upgrade is something we might test manually at some point, but I'd rather do it on Millau/Rialto than on someone's private computer.

Actually having relay compiled with different runtime version is much easier to test than making runtime upgrade. That's what I did manually [here](https://github.com/paritytech/parity-bridges-common/issues/471#issuecomment-719415575). If you still this worth testing on real chain, ok - I could include spec version guard to run in Millau <-> Rialto relays.

relays/substrate-client/src/guard.rs
tomusdrw: > . Thing that was designed to be super-simple is now turning into something that both doesn't work as you want and is checking the same thing that has been checked by the original version.

I have to disagree here - afair we've been talking about alerting when we **spend** too much, not when balance drops (what was implemented). I only said that's it not equal.

> Your logic also "won't work correctly" if the account balance has been first increased by maximal_difference and then decreased by the same amount between two consequent checks (e.g. in the same block). We may go even further and check all events of balances module in all blocks that we see. Then the check will finally would be correct.

Sure, but IMHO that's edge cases. We can assume how the account is going to be used and imho it's 100% reasonable to assume that the relayer account is going to be used as follows:
1. Starts off with some initial balance
2. Spends some to perform it's work
3. At some point you realise it's reaching 0, so you top it up.

So both of your solutions are sub-optimal for the most reasonable workflow of the relayer I can think of.

> I prefer to revert to original code to keep it simple, or just remove this guard from the relay code && rely on external components.

I don't care that much any more, especially if you refuse to enable it for testnet deployments - it's basically dead code for me. Production deployment will most likely require additional steps anyway, so we may just provide a README instead.
I do like the idea of this safety valve though, but just want it to be optimal, not minimal.

tomusdrw (approved):

svyatonik:
I was assuming that the account would have enough money for long-term needs. And delay between (1) and (3), if if'll ever happen, would be huge. So the balance increase is the edge case and the balance drop = funds spend by relayer. Probably you're right && we'll have that other component (script? human?) that will transfer money from some treasury to relay account more often. Now I'm not that sure. I'm only sure that the original version presented here is what I have been suggesting initially. I was never claiming to build something non-trivial that is computing funds spent by relayer on its own transactions.

So iiuc there'll be some external component watching for relay balance (or probably alert?). Then it also worth implementing these guards there.

tomusdrw:
> I was assuming that the account would have enough money for long-term needs. And delay between (1) and (3), if if'll ever happen, would be huge. So the balance increase is the edge case and the balance drop = funds spend by relayer.

Fair enough, let's get that merged, I've overreacted a bit, sorry - this is obviously better than nothing.

svyatonik:
Sorry, @HCastano - I missed that "requested review" sign. If you have any questions/comments - please leave here && I'll try to explain/fix

HCastano:
> Sorry, @HCastano - I missed that "requested review" sign. If you have any questions/comments - please leave here && I'll try to explain/fix

Eh, it's alright. I trust that Tomek's review is sufficient here

Issue #440, opened 2020-10-23
Freeze switch for the header sync and message lanes


tomusdrw:
The pallets should initially have a way to freeze (halt) all operations, later on this "sudo" access can be transferred to `Root` origin (i.e. to the governance), but initially it might be worth to temporarily stop the bridge in case some serious problem is detected.

tomusdrw:
Implemented in #472 and #485

PR #472, opened 2020-10-29
Ability to halt/resume message-lane pallet

svyatonik:
related #440

Afaik we can't inherit module calls and I do not want to add another pallet just to introduce halt/ability to two our pallets => that's why implementation is not reusable (and we can't use macro inside substrate module-related macros). But we'll need the same functionality in the substrate bridge pallet. And it would be better if we can make interface (Calls, config fields names, storage variables names) the same. @HCastano can you, please, take a look at names in this PR && commit that you'll use the same in your pallet? :)

### DIFF
--- a/modules/message-lane/src/lib.rs
+++ b/modules/message-lane/src/lib.rs
@@ -182,7 +182,7 @@ decl_module! {
 		/// Halt all pallet operations. Operations may be resumed using `resume_operations` call.
 		///
 		/// May only be called either by root, or by `ModuleOwner`.
-		#[weight = (0, DispatchClass::Operational)]
+		#[weight = (T::DbWeight::get().reads_writes(1, 1), DispatchClass::Operational)]
 		pub fn halt_operations(origin) {
 			ensure_owner_or_root::<T, I>(origin)?;
 			IsHalted::<I>::put(true);
@@ -191,7 +191,7 @@ decl_module! {
 		/// Resume all pallet operations. May be called even if pallet is halted.
 		///
 		/// May only be called either by root, or by `ModuleOwner`.
-		#[weight = (0, DispatchClass::Operational)]
+		#[weight = (T::DbWeight::get().reads_writes(1, 1), DispatchClass::Operational)]
 		pub fn resume_operations(origin) {
 			ensure_owner_or_root::<T, I>(origin)?;
 			IsHalted::<I>::put(false);
--- a/modules/message-lane/src/lib.rs
+++ b/modules/message-lane/src/lib.rs
@@ -143,7 +143,7 @@ decl_error! {
 decl_storage! {
 	trait Store for Module<T: Trait<I>, I: Instance = DefaultInstance> as MessageLane {
 		/// Optional pallet owner.
-		/// 
+		///
 		/// Pallet owner has a right to halt all pallet operations and then resume it. If it is `None`, then there are no direct ways to halt/resume
 		/// pallet operations, but other runtime methods may still be used to do that (i.e.
 		/// democracy::referendum to update halt flag directly or call the `halt_operations`).
--- a/modules/message-lane/src/lib.rs
+++ b/modules/message-lane/src/lib.rs
@@ -144,9 +144,10 @@ decl_storage! {
 	trait Store for Module<T: Trait<I>, I: Instance = DefaultInstance> as MessageLane {
 		/// Optional pallet owner.
 		///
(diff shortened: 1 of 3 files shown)


HCastano (approved):
Looks good!

modules/message-lane/src/lib.rs
```diff
@@ -137,13 +142,23 @@ decl_error! {
 
 decl_storage! {
 	trait Store for Module<T: Trait<I>, I: Instance = DefaultInstance> as MessageLane {
+		/// Optional pallet owner. Pallet owner has a right to halt all pallet operations
+		/// and then resume it. If it is `None`, then there are no direct ways to halt/resume
```
tomusdrw, line 146:
```suggestion
		/// Optional pallet owner.
		/// 
		/// Pallet owner has a right to halt all pallet operations and then resume it. If it is `None`, then there are no direct ways to halt/resume
```

modules/message-lane/src/lib.rs
```diff
@@ -137,13 +142,23 @@ decl_error! {
...
+		/// pallet operations, but other runtime methods may still be used to do that (i.e.
+		/// democracy::referendum to update halt flag directly or call the `halt_operations`).
+		pub ModuleOwner get(fn module_owner) config(): Option<T::AccountId>;
```
tomusdrw, line 149: Shall we add ownership transfer function? Now we have to rely on `sudo/democracy` to alter that entry.
Something like
```rust
set_owner(origin, new_owner: Option<T::AcountId>) {
  ensure_owner_or_root::<T, I>(origin)?;
  ModuleOwner::put(new_owner);
}

File: modules/message-lane/src/lib.rs [hunk 3]
```diff
@@ -164,6 +179,24 @@ decl_module! {
 		/// Deposit one of this module's events by using the default implementation.
 		fn deposit_event() = default;
 
+		/// Halt all pallet operations. Operations may be resumed using `resume_operations` call.
+		///
+		/// May only be called either by root, or by `ModuleOwner`.
+		#[weight = (0, DispatchClass::Operational)]
```
tomusdrw, line 185: This has at least a cost of one storage read and one storage write, no? If you intend to do benchmarks instead please add a TODO here so that we don't forget.
```suggestion
		#[weight = (T::DbWeights::get().reads_writes(1, 1), DispatchClass::Operational)]
```

modules/message-lane/src/lib.rs
```diff
@@ -164,6 +179,24 @@ decl_module! {
...
+		pub fn halt_operations(origin) {
+			ensure_owner_or_root::<T, I>(origin)?;
+			IsHalted::<I>::put(true);
+		}
+
+		/// Resume all pallet operations. May be called even if pallet is halted.
+		///
+		/// May only be called either by root, or by `ModuleOwner`.
+		#[weight = (0, DispatchClass::Operational)]
```
tomusdrw, line 194: Same here.
```suggestion
		#[weight = (T::DbWeights::get().reads_writes(1, 1), DispatchClass::Operational)]
```

tomusdrw (approved):
lgtm! small grumbles about weights and ownership transfer.

modules/message-lane/src/lib.rs
svyatonik: :) I have removed this function right before opening PR just because I was unable to guess when we'll need it. Most probably this account will be shared in some way and then (when the time will come) we will simply rely on sudo/democracy to transfer ownership to root? Or do you think it is better to have explicit function for this?

modules/message-lane/src/lib.rs
svyatonik: I wasn't going to do any benchmarks. The idea was that this call is kind of emergency and should fit any block. So I just left zero here - I've seen such transactions in Substrate. But probably you're right - maybe someone could spam block with these calls (even though they fail) and paying only basic tx fee.

modules/message-lane/src/lib.rs
tomusdrw: > Most probably this account will be shared in some way

Yes, I assumed that as well.

> we will simply rely on sudo/democracy to transfer ownership to root

Here I actually assumed that WE would just send a transaction setting the owner to `None` - which would make `Root` the only owner. I prefer that cause it doesn't involve democracy - it's purely our decision when we want to get rid of the admin rights.

I'm fine with either though, I think `set_owner` is a bit more flexible though long-term: you can go with single key -> multi-sig -> democracy for instance.

Issue #407, opened 2020-10-08
Support Dynamic Genesis Config Values


HCastano:
Opening this up to track the concern Slava brought up [here](https://github.com/paritytech/parity-bridges-common/pull/401#discussion_r501628560) in #401.

In order to bootstrap the Substrate pallet we need to provide (at minimum) an initial header and authority set. This is done in configuration files that are part of the node's runtime. For example, for the Millau node to bridge (e.g sync headers from) the Rialto network it would need to fill out the initial header and authorities in `millau/runtime/src/rialto.rs`.

If we try and do a two way sync things get complicated. We'd need to change the `millau.rs` configuration file that's part of the Rialto node - which would in turn change the values of the initial header for Rialto. We'd then need to go back to the Millau node and change the `rialto.rs` configuration, which would change the initial header value...you get the point.

We should have some way to inject these initial configuration parameters after the node has started running. Something like a trusted dispatchable that's only accessible to `Sudo` to initialize the system is probably the best way to go.

PR #474, opened 2020-10-30
Test that actual Millau state root is used in Rialto runtime

svyatonik:
Right now almost any PR that is touching runtime modules, primitives or dependencies may change state root of Millau genesis block (because it affects Millau runtime code and genesis storage includes that code). This PR adds test to avoid cases when Millau code has been updated, but its state root isn't updated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -736,6 +736,7 @@ version = "0.1.0"
 dependencies = [
  "bp-runtime",
  "frame-support",
+ "hex-literal 0.3.1",
  "sp-api",
  "sp-core",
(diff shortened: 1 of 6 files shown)


svyatonik:
Ok, it's not that easy - looks like we need to use at least same compiler version(s) + need to use reproducible builds. I'm not sure if reproducible builds are already available.

The other option is to start both chains with headers-sync disabled and then start headers sync using relayer. That's what is described in #407 (originally I was going to used it only to start Rialto -> Millau). I'll close this PR in favor of this soltion.

PR #476, opened 2020-11-02
Use is_finalized in is_finalized_header

svyatonik:
Iiuc there's no pruning now && even if it'll be, most probably it'll be non-instant => we can't rely on `header.number() <= storage.best_finalized_header().number()`

### DIFF
--- a/modules/substrate/src/lib.rs
+++ b/modules/substrate/src/lib.rs
@@ -251,7 +251,7 @@ impl<T: Trait> Module<T> {
 	pub fn is_finalized_header(hash: BridgedBlockHash<T>) -> bool {
 		let storage = PalletStorage::<T>::new();
 		if let Some(header) = storage.header_by_hash(hash) {
-			header.number() <= storage.best_finalized_header().number()
+			header.is_finalized
 		} else {
 			false
 		}


modules/substrate/src/lib.rs
```diff
@@ -251,7 +251,7 @@ impl<T: Trait> Module<T> {
 	pub fn is_finalized_header(hash: BridgedBlockHash<T>) -> bool {
 		let storage = PalletStorage::<T>::new();
 		if let Some(header) = storage.header_by_hash(hash) {
-			header.number() <= storage.best_finalized_header().number()
+			header.is_finalized
```
tomusdrw, line 254: I thought that we don't update `is_finalized` for ancestry blocks, do we? That flag would then only be set for blocks that directly had justifications imported.

svyatonik: Afaik [we are](https://github.com/paritytech/parity-bridges-common/blob/master/modules/substrate/src/verifier.rs#L295), but better wait for Hernando to be sure. I guess it has been deferred to optimize later (as a part of #367 probably)

HCastano: Slava's right here for both points. I was hesitant to do checks directly with `is_finalized` because maybe with something like a skip-list every header won't have that flag set even though they are indeed finalized.

That said, I don't have any strong opinions on this change. We just need to make sure that when we implement #367 we realize that this check may have to change.

tomusdrw: Sounds good, hopefuly the tests would detect that:)

tomusdrw (approved):

PR #477, opened 2020-11-02
Rialto -> Millau headers relay

### LINKED ISSUE #443: Rialto -> Millau headers relay
The relayer needs to be able to support syncing headers from Rialto to Millau. We already have the other direction (see #398), but we also need the other one.
Basically means creating an additional sync loop with types specific to these chains in the relayer code and adding a CLI mode for this.

svyatonik:
closes #443

This is required for relaying Millau -> Rialto messages (because we submit transactions to both Relay and Millau). I've extracted common part to `headers_pipeline.rs`, so that both `millau_headers_to_rialto.rs` and `rialto_headers_to_millau.rs` now only have chain-specific definitions and `Call` composing.


primitives/rialto/src/lib.rs
```diff
@@ -60,6 +60,15 @@ impl Chain for Rialto {
 	type Header = Header;
 }
 
+/// Name of the `RialtoHeaderApi::best_block` runtime method.
```
HCastano, line 63:
```suggestion
/// Name of the `RialtoHeaderApi::best_blocks` runtime method.
```

HCastano (approved):

PR #478, opened 2020-11-03
Move storage keys computation to the message-lane pallet

svyatonik:
required for #397 

We need it for both `prove_*()` RPC calls and for proof verification from runtime (that is implemented in #397. There's an issue with this approach, though - generated `StorageMap` requires to specify `Runtime`. And we need e.g. to verify proofs from `Millau` runtime in `Rialto` runtime. Obviously, we have no access to the `Millau` runtime there. But the runtime isn't actually used to compute storage keys => we may use any runtime. That s good. What is not good, is that if e.g. on `Millau` we will be using `Instance42` to bridge with `Rialto`. Then on `Rialto` we need to have `pallet_message_lane::Trait<Instance42>` to be implemented for the `Rialto` runtime. This isn't super-convenient. The alternative approach is to manually craft storage keys. But it requires more knowledge of Substrate internals (like that the storage key = `Twox128(module_prefix) ++ Twox128(storage_prefix) ++ Hasher(encode(key))`) => that's why I'm not choosing this option.


bin/runtime-common/src/messages.rs
```diff
@@ -88,10 +91,14 @@ pub trait ChainWithMessageLanes {
 	type Weight: From<frame_support::weights::Weight>;
 	/// Type of balances that is used on the chain.
 	type Balance: CheckedAdd + CheckedDiv + CheckedMul + PartialOrd + From<u32> + Copy;
+
+	/// Instance of the message-lane pallet.
+	type MessageLaneInstance: Instance;
```
HCastano, line 96: So this is where you mentioned that we'd need to have `MessageLane<Instance42>` if our bridged chain has a bridge pallet with `Instance42` used for connecting with us?

modules/message-lane/src/lib.rs
```diff
@@ -452,6 +452,42 @@ impl<T: Trait<I>, I: Instance> Module<T, I> {
...
+/// Getting storage keys for messages and lanes states. These keys are normally used when building
+/// messages and lanes states proofs.
+///
+/// Keep in mind that all functions in this module are **NOT** using passed `T` argument, so any
+/// runtime can be passed. E.g. if you're verifying proof from Runtime1 in Runtime2, you only have
+/// access to Runtime2 and you may pass it to the functions, where required. This is because our
+/// maps are not using any Runtime-specific data in the keys.
+///
+/// On the other side, passing correct instance is required. So if proof has been crafted by the
+/// Instance1, you should verify it using Instance1. This is inconvinient if you're using different
```
HCastano, line 464:
```suggestion
/// Instance1, you should verify it using Instance1. This is inconvenient if you're using different
```

modules/message-lane/src/lib.rs
```diff
@@ -1018,4 +1055,34 @@ mod tests {
...
 		});
 	}
+
+	#[test]
+	fn storage_message_key_computed_properly() {
+		// If this test fails, then something has been changed in module storage that is breaking all
+		// previously crafted messages proofs.
+		assert_eq!(
+			storage_keys::message_key::<TestRuntime, DefaultInstance>(&*b"test", 42).0,
+			hex!("87f1ffe31b52878f09495ca7482df1a48a395e6242c6813b196ca31ed0547ea79446af0e09063bd4a7874aef8a997cec746573742a00000000000000").to_vec(),
```
HCastano, line 1065: How did you compute these?

HCastano (approved):
I think it's a little unfortunate that instances have to match across the bridge. I also don't have any good ideas on how to get around it. For an initial bridge deployment I think this is something we can live with since it's going to be very hands on anyways (we need to initialize headers, authority sets, perform runtime upgrades, etc).

Further down the line we can look at this again to see if we can streamline the deployment process by removing this restriction.

bin/runtime-common/src/messages.rs
svyatonik: Not exactly here. Instances are exported from `pallet_message_lane`, so they're available in all runtimes => you won't have any problems here. But when you will call any function from `pallet_message_lane::storage_keys` module with runtime that is not implementing `pallet_message_lane::Trait<Instance42>`, you will see compilation issue.

modules/message-lane/src/lib.rs
svyatonik: Just ran test with `vec![]` instead of `hex!(...)`

Issue #442, opened 2020-10-23
Relayer should support initial header to start from


tomusdrw:
Initially when the substrate pallet is going to be deployed the sync will be temporarily frozen.
At some point we will initialize the pallet with a starting block and enable sync.
Main reason for that is the delay between runtime upgrade proposal and enactment, we can hardcode the starting block in the runtime upgrade, but then after the enactment of upgrade the bridge will be extremely behind the other chain and will have to catch up.

Related: #440 - the relayer should handle the case if header sync is frozen.

HCastano:
Here is the related runtime issue: #407. I think that needs to be tackled before we add support in the relayer.

PR #481, opened 2020-11-03
Allow Substrate Pallet to be Initialized via Dipatchable

### LINKED ISSUE #407: Support Dynamic Genesis Config Values

HCastano:
The idea behind this PR is to allow the Substrate pallet to be initialized through a dispatchable. Initialization involves providing an initial header, authority set, and (optionally) a scheduled change.

Currently, we initialize the pallet through the chain spec of the node. This is a problem since the `state_root` of the initial header changes if we change anything in the node. I've added a dispatchable that is only callable by Root which can bootstrap the pallet. I'll add support for the pallet owner to also be able to initialize it in a different PR tomorrow.

I spent some time trying to test this with Polkadot JS Apps, but wasn't able to due to some type errors. I'm not sure what's wrong, but I'm decently confident it'll work. If anything, we can test it as we implement #442, lol.

Closes #407.


modules/substrate/src/lib.rs
```diff
@@ -117,37 +117,13 @@ decl_storage! {
...
-			<ImportedHeaders<T>>::insert(
-				initial_hash,
-				ImportedHeader {
-					header: initial_header,
-					requires_justification: false,
-					is_finalized: true,
-					signal_hash,
-				},
+			initialize_bridge::<T>(
+				initial_header,
```
svyatonik, line 122: Imo the `initial_header` and `initial_authority_list` must be `Option`s, or probably they should be removed from `GenesisConfig` (along with `initial_set_id` and `first_scheduled_change`). Because passing something that we call `millau_genesis_header` in the code, but we know is not genesis header of Millau is not good :)

svyatonik: Probably let's add struct `InitializationData` and make it `Option`?

svyatonik (approved):

deployments/sub-bridge-types.json
```diff
@@ -0,0 +1,20 @@
+{
```
tomusdrw, line 1: Let's just have a single `json` file with all types please. If it grows to big we can later have it auto-crated by CI from multiple smaller ones, but for now, for simplicity I'd rather have just one.

modules/substrate/src/lib.rs
```diff
@@ -215,6 +191,28 @@ decl_module! {
...
+		// TODO: Update weights [#78]
+		#[weight = 0]
+		pub fn initialize(
+			origin,
+			header: BridgedHeader<T>,
+			authority_list: sp_finality_grandpa::AuthorityList,
+			set_id: sp_finality_grandpa::SetId,
+			scheduled_change: Option<ScheduledChange<BridgedBlockNumber<T>>>,
+		) {
+			let _ = ensure_root(origin)?;
```
tomusdrw, line 213: IMHO to prevent abuses this should only be callable if the bridge is not already initialized. IMHO resetting/re-initializing the state of the bridge should be a separate issue, cause it will most likely involve a bit of cleaning up as well to make sure things stay consistent.

tomusdrw:
```suggestion
			let _ = ensure_owner_or_root(origin)?;
```
Also this should allow the owner, introduced in #472

HCastano: The owner stuff that was implemented in #472 is only for the message lane pallet. The Substrate pallet doesn't have that yet which is why I mentioned this:

> I'll add support for the pallet owner to also be able to initialize it in a different PR tomorrow.

As far as this goes: 

> IMHO to prevent abuses this should only be callable if the bridge is not already initialized. IMHO resetting/re-initializing the state of the bridge should be a separate issue, cause it will most likely involve a bit of cleaning up as well to make sure things stay consistent.

I can add a check in that other PR that this should only be called if the pallet hasn't been initialized.

tomusdrw: > The Substrate pallet doesn't have that yet which is why I mentioned this:

Oh, right. Sorry I missed that.

deployments/sub-bridge-types.json
HCastano: I had to rename a type in the Ethereum pallet for this, but since Rialto has both Ethereum and Substrate bridge pallets (and will thus need both types) I think it's okay

tomusdrw (approved):

modules/substrate/src/lib.rs
```diff
@@ -105,53 +105,18 @@ decl_storage! {
 		// Grandpa doesn't require there to always be a pending change. In fact, most of the time
 		// there will be no pending change available.
 		NextScheduledChange: map hasher(identity) BridgedBlockHash<T> => Option<ScheduledChange<BridgedBlockNumber<T>>>;
+		/// Whether or not the bridge has been initialized.
+		///
+		/// This is important to know to ensure that we don't try and initialize the bridge twice
+		/// and create an inconsistent genesis state.
+		IsInitialized: bool;
```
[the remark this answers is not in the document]
svyatonik: I would rather remove all the is-already-initialized checks at all. Root is here, whether we want this, or not. Instead of calling `initialize()`, root may just update storage values directly, without calling the function. Why would we need to build gate in the middle of the unfenced field? I agree that in "normal programming", we should protect against double-initialization. But here it only takes additional resources - no matter if we read `IsInitialized` or ``CurrentAuthoritySet.

I could even imagine this function evolved to the `reinitialize` that may be called if we have missed too many headers and just want to restart the sync.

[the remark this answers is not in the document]
HCastano: > I'm assuming that you're going to reply that this feels hacky.

You know me too well 😄 

> Since the uninitialized code is hacky already, cause the invariants are broken (for instance ImportedHash does not contain BestFinalized

I'm not entirely sure what you mean here

> Next I'd probably pass since there is no good solution here, for safety it would be good to do initialization check for each dispatchable (and maybe even a runtime call), but it might not be worth the storage read costs.

Let me propose a different solution here. Since we're going to be introducing halting to the pallet, all the dispatchables are going to be checking if the pallet is halted before proceeding. The initial state of the pallet needs to be set as a configuration parameter (either via `GenesisConfig` or `initialize()`). We can enforce that if the pallet is not initialized then it must be halted. Once we initialize the pallet then we can lift the halted state (or keep it halted if that's what's desired).

This would have to be done in the halting PR. What do you think?

HCastano:
Since there's already a PR that depends on this, and I'm about to make another which also does, I'm going to go ahead and merge it. I'm not dismissing the initialization conversation, we should continue it either here or the halting PR.

modules/substrate/src/lib.rs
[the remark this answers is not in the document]
HCastano: I've followed up on this in  #485.

PR #483, opened 2020-11-05
Relay subcommands to initialize substrate bridge pallet

### LINKED ISSUE #442: Relayer should support initial header to start from

svyatonik:
~on top of #481 => draft~
closes #442 

The signer is supposed to be ~pallet owner (when it'll be introduced)~ sudo key.


relays/substrate/src/cli.rs
```diff
@@ -165,6 +189,20 @@ macro_rules! declare_chain_options {
 				#[structopt(long)]
 				pub [<$chain_prefix _signer_password>]: Option<String>,
 			}
+
+			#[doc = $chain " headers bridge initialization params."]
+			#[derive(StructOpt)]
+			pub struct [<$chain BridgeInitializationParams>] {
+				#[doc = "Hex-encoded " $chain " header to initialize bridge with. If not specified, genesis header is used."]
```
tomusdrw, line 196: Can you elaborate on why choose raw values instead of supporting a block number from the source chain?
We query for genesis hash anyway, simply querying for block `X` sounds much easier to implement and I don't see that much value in having arbitrary headers passed in raw form (seems more error prone to me).

svyatonik: If you suggest to replace all arguments with single argument - initial header number, then there's a problem with authorities set id. To compute it, you have two options:
1) go from genesis to specified block && increase set by 1 every time authorities set changes. This would be too slow for Polkadot/Kusama;
2) read `CurrentSetId` from `pallet_grandpa`. This field is marked as private, unfortunately => I can't generate storage key (by calling `pallet_grandpa::CurrentSetId::storage_value_final_key()`) for it and use in storage read RPC functions. I guess it may be read from metadata, but we have no tools to parse metadata atm. It may also be hardcoded, but I'd prefer to avoid it. Or maybe you know another option?

Specifying header number for first two args and leaving set id arg as-is doesn't seems good idea.

tomusdrw: Hmm, okay, so currently the only option is to read this via UI and put to CLI directly? Could we possibly make `CurrentSetId` public to make it easy to generate the storage key? CC @andresilva

andresilva: `CurrentSetId` has a getter method, isn't that always accessible regardless of field visibility, i.e. `Grandpa::current_set_id()`? Otherwise we can expose this but I'd like to make sure we only expose getter visibility (so that we can't mutate this field from other pallets).

svyatonik: > isn't that always accessible regardless of field visibility, i.e. Grandpa::current_set_id()

Yes, it is visible for other runtime pallets. But we need to read it using RPC - e.g. using `state_getStorage` call.

andresilva: Got it! Does this make sense to you: https://github.com/paritytech/substrate/pull/7502 ?

svyatonik: Thanks! :) But it's more q to Tomek - we won't be able to use it anyway right now, because we're bounded to the 2.0 :/ But in the future it may be helpful, I guess.

HCastano (approved):
I agree with Tomek that we need some sort of documentation around initializing the bridge. Would you be able to do that soon Slava?

Also, I'm really not a fan of the command line option UI. I think something like `./relay init headers ...` would be much better, but we'd need to rework some stuff to achieve that.

PR #485, opened 2020-11-05
Allow Substrate Pallet to be Halted

HCastano:
Basically a copy-paste of the `message-lane` code from #472.

I did add some stuff surrounding the initialization of the pallet though. I removed the `IsInitialized` storage entry, and instead have decided to enforce (at least when initializing with `GenesisConfig`) that if we didn't get any data for initialization the pallet is halted.

If you decide to initialize via a dipatchable, we have stopped checking if the pallet has already been initialized. As Root/Owner you should be doing this check (by, for example, seeing if there are any finalized headers) before calling `initialize()`. If you don't, sucks to suck 🤷. Another note on initializing via `initialize()`. If you're doing this (for the first time) it means that you didn't go through the genesis config, which means that the pallet must be halted. The pallet should be unblocked manually with `resume_operations()` afterwards.

### DIFF
--- a/bin/millau/node/src/chain_spec.rs
+++ b/bin/millau/node/src/chain_spec.rs
@@ -159,7 +159,6 @@ fn testnet_genesis(
 			// We'll initialize the pallet with a dispatchable instead.
 			init_data: None,
 			owner: None,
-			is_halted: true,
 		}),
 		pallet_sudo: Some(SudoConfig { key: root_key }),
 		pallet_session: Some(SessionConfig {
--- a/bin/millau/node/src/chain_spec.rs
+++ b/bin/millau/node/src/chain_spec.rs
@@ -158,6 +158,8 @@ fn testnet_genesis(
 		pallet_substrate_bridge: Some(BridgeRialtoConfig {
 			// We'll initialize the pallet with a dispatchable instead.
 			init_data: None,
+			owner: None,
+			is_halted: true,
 		}),
 		pallet_sudo: Some(SudoConfig { key: root_key }),
 		pallet_session: Some(SessionConfig {
--- a/modules/message-lane/src/lib.rs
+++ b/modules/message-lane/src/lib.rs
@@ -148,7 +148,7 @@ decl_storage! {
 		/// `None`, then there are no direct ways to halt/resume pallet operations, but other
 		/// runtime methods may still be used to do that (i.e. democracy::referendum to update halt
 		/// flag directly or call the `halt_operations`).
-		pub ModuleOwner get(fn module_owner) config(): Option<T::AccountId>;
+		pub ModuleOwner get(fn module_owner): Option<T::AccountId>;
 		/// If true, all pallet transactions are failed immediately.
 		pub IsHalted get(fn is_halted) config(): bool;
 		/// Map of lane id => inbound lane data.
@@ -160,6 +160,12 @@ decl_storage! {
 	}
 	add_extra_genesis {
 		config(phantom): sp_std::marker::PhantomData<I>;
+		config(owner): Option<T::AccountId>;
+		build(|config| {
+			if let Some(ref owner) = config.owner {
+				<ModuleOwner<T, I>>::put(owner);
+			}
+		})
 	}
 }
 
--- a/modules/message-lane/src/lib.rs
+++ b/modules/message-lane/src/lib.rs
@@ -196,11 +196,11 @@ decl_module! {
 			match new_owner {
 				Some(new_owner) => {
 					ModuleOwner::<T, I>::put(&new_owner);
-					frame_support::debug::info!(target: "runtime", "Setting pallet Owner to: {:?}", new_owner);
+					frame_support::debug::info!("Setting pallet Owner to: {:?}", new_owner);
 				},
 				None => {
 					ModuleOwner::<T, I>::kill();
-					frame_support::debug::info!(target: "runtime", "Removed Owner of pallet.");
+					frame_support::debug::info!("Removed Owner of pallet.");
 				},
 			}
 		}
@@ -212,7 +212,7 @@ decl_module! {
 		pub fn halt_operations(origin) {
 			ensure_owner_or_root::<T, I>(origin)?;
 			IsHalted::<I>::put(true);
-			frame_support::debug::warn!(target: "runtime", "Stopping pallet operations.");
+			frame_support::debug::warn!("Stopping pallet operations.");
 		}
 
 		/// Resume all pallet operations. May be called even if pallet is halted.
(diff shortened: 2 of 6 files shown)


bin/millau/node/src/chain_spec.rs
```diff
@@ -158,6 +158,8 @@ fn testnet_genesis(
 		pallet_substrate_bridge: Some(BridgeRialtoConfig {
 			// We'll initialize the pallet with a dispatchable instead.
 			init_data: None,
+			owner: None,
```
svyatonik, line 161: Imo it should be `root_key`. But it may be changed in follow up PR. We would need to switch both MillauInRialto and RialtoInMillau pallets to start uninitialized with owner set to //Alice (root_key). Then they should be initialized with #483

modules/substrate/src/lib.rs
```diff
@@ -105,17 +107,32 @@ decl_storage! {
...
+			if let Some(ref owner) = config.owner {
+				<ModuleOwner<T>>::put(owner);
+			}
+
 			if let Some(init_data) = config.init_data.clone() {
 				initialize_bridge::<T>(init_data);
+			} else {
+				// Since the bridge hasn't been initialized we shouldn't allow anyone to perform
+				// transactions.
+				IsHalted::put(true);
```
svyatonik, line 135: This means that we'll need to send two transactions in #483 to actually initialize + start pallet, right?

svyatonik: (I mean - it is ok, just wanted to be sure that it is by design)

bin/millau/node/src/chain_spec.rs
tomusdrw: We do include `sudo` pallet as well, so we can always issue as root, but I agree it would be simpler to set it to `Some(root_key)`

modules/substrate/src/lib.rs
```diff
@@ -105,17 +107,32 @@ decl_storage! {
...
+		/// flag directly or call the `halt_operations`).
+		ModuleOwner get(fn module_owner): Option<T::AccountId>;
+		/// If true, all pallet transactions are failed immediately.
+		IsHalted get(fn is_halted) config(): bool;
 	}
 	add_extra_genesis {
+		config(owner): Option<T::AccountId>;
 		config(init_data): Option<InitializationData<BridgedHeader<T>>>;
 		build(|config| {
+			IsHalted::put(config.is_halted);
```
tomusdrw, line 124: This is a bit weird, why not put `is_halted` to `init_data` instead? There are 3 cases to cover:
- no `init_data` -> bridge halted
- `init_data` -> bridge either halted or not

modules/substrate/src/lib.rs
tomusdrw: IMHO this should be part of `init_data`, so that you can easily initialize & resume the bridge.

modules/substrate/src/lib.rs
```diff
@@ -202,17 +221,48 @@ decl_module! {
 		/// This function is only allowed to be called from a trusted origin and writes to storage
 		/// with practically no checks in terms of the validity of the data. It is important that
 		/// you ensure that valid data is being passed in.
+		///
+		/// There's also nothing stopping Owner or Root from calling this function twice (which
+		/// would be bad). Ensure this isn't the case before submitting the call.
```
tomusdrw, line 226: I'm strongly in favor of adding a check here and then adding `reinitialize` method (please create a separate issue for this).

Re-initialization should be heavily tested to make sure the bridge pallet state maintains consistent if such event occurs.

tomusdrw (changes requested):
I'm strongly in favour of having reinitialization being a separate function, and `initialize` should check that the bridge is not initialized already (simply via checking genesis header) - the function is going to be called so rarely (hopefuly once) that the performance argument doesn't apply at all here.

Not doing this is just bringing a massive footgun in - of course we have sudo/democracy that can do arbitrary storage changes, but the bridge owner might not be a collective and there are multiple examples in blockchain history when trusting yourself (the owner) too much caused serious damage.

modules/substrate/src/lib.rs
HCastano: > This means that we'll need to send two transactions in #483 to actually initialize + start pallet, right?

In its current form, yes that's correct

bin/millau/node/src/chain_spec.rs
HCastano: I'm not sure I understand the rational of setting the owner to be the root key in this case. Both Root and the Owner have the same level of control in the pallet, and if you want the owner to be Root you can achieve this result by just not setting an Owner

tomusdrw: `root_key` is not `Root` account/origin. It's the same key that can control `sudo` afaict, so it should be Alice's key.

HCastano: So from what I understand the `root_key` can be any account. In our case the `root_key` is used as the Sudo (although it doesn't have to be). Since this is the case isn't it still redundant to set `owner` to the `root_key` since the `root_key` is already `Sudo`?

HCastano:
@tomusdrw I removed the footgun

modules/message-lane/src/lib.rs
```diff
@@ -194,8 +194,14 @@ decl_module! {
 		pub fn set_owner(origin, new_owner: Option<T::AccountId>) {
 			ensure_owner_or_root::<T, I>(origin)?;
 			match new_owner {
-				Some(new_owner) => ModuleOwner::<T, I>::put(new_owner),
-				None => ModuleOwner::<T, I>::kill(),
+				Some(new_owner) => {
+					ModuleOwner::<T, I>::put(&new_owner);
+					frame_support::debug::info!(target: "runtime", "Setting pallet Owner to: {:?}", new_owner);
```
tomusdrw, line 199: Why is the `target` set to `runtime`? Why do we need `target` at all?

modules/substrate/src/lib.rs
```diff
@@ -241,8 +245,14 @@ decl_module! {
 		pub fn set_owner(origin, new_owner: Option<T::AccountId>) {
 			ensure_owner_or_root::<T>(origin)?;
 			match new_owner {
-				Some(new_owner) => ModuleOwner::<T>::put(new_owner),
-				None => ModuleOwner::<T>::kill(),
+				Some(new_owner) => {
+					ModuleOwner::<T>::put(&new_owner);
+					frame_support::debug::info!(target: "sub-bridge", "Setting pallet Owner to: {:?}", new_owner);
```
tomusdrw, line 250: Do we really need the `target`? It will default to module name afair, which is probably good enough.

tomusdrw (approved):
Cool thanks! Minor things about logging `target`s left.

modules/message-lane/src/lib.rs
HCastano: That's what everything else in the `message-lane` pallet uses. I didn't want to change all the logging in this PR.

modules/substrate/src/lib.rs
HCastano: I didn't realize they defaulted to anything, I thought you always had to have one

PR #488, opened 2020-11-06
Fixed block where proof is generated

svyatonik:
Proof was generated at best finalized source chain block. but it should have been generated at best source chain block finalized on target node.


tomusdrw (approved):
Are there tests that we could use to cover this to prevent regression in the future?

PR #489, opened 2020-11-06
Print progress of message races

svyatonik:


relays/messages-relay/src/message_race_loop.rs
```diff
@@ -383,3 +392,29 @@ impl<SourceHeaderId, TargetHeaderId, MessageNonce, Proof> Default
...
+
+/// Print race progress.
+fn print_race_progress<P, S>(prev_time: Instant, strategy: &S) -> Instant
+where
+	P: MessageRace,
+	S: RaceStrategy<P::SourceHeaderId, P::TargetHeaderId, P::MessageNonce, P::Proof>,
+{
+	let now_time = Instant::now();
+
+	let need_update = now_time.duration_since(prev_time) > Duration::from_secs(10);
```
tomusdrw, line 404:
```suggestion
	let need_update = now_time.saturating_duration_since(prev_time) > Duration::from_secs(10);
```

Just to be sure since in theory, we don't know how `prev_time` is being updated externally.

tomusdrw (approved):

PR #493, opened 2020-11-09
Fixes for Millau-Rialto Deployment

HCastano:
There are some small changes that need to be made in order to smoothly deploy this bridge. With the changes here I've deployed a network with two way header sync between Rialto and Millau, and I also have messages being sent from Millau to Rialto on a single lane. Something worth noting is that we don't have any way to monitor the relays (e.g through Grafana) just yet. That's gonna be done in future PRs.

As part of the deployment there are two public facing nodes - one for each network respectively. You can connect to the deployed nodes here:

- Rialto: [wss://wss.rialto.brucke.link](wss://wss.rialto.brucke.link)
- Millau: [wss://wss.millau.brucke.link](wss://wss.millau.brucke.link)

You'll need to upload the `types.json` file to Polkadot JS Apps before it shows you anything.


relays/substrate/src/headers_target.rs
```diff
@@ -91,11 +91,11 @@ where
 		let decoded_response: Vec<(P::Number, P::Hash)> =
 			Decode::decode(&mut &encoded_response.0[..]).map_err(SubstrateError::ResponseParseFailed)?;
 
-		let best_header = decoded_response.last().ok_or_else(|| {
-			SubstrateError::ResponseParseFailed(
-				"Parsed an empty list of headers, we should always have at least one.".into(),
-			)
-		})?;
+		const WARNING_MSG: &str = "Parsed an empty list of headers, we should always have at least
```
HCastano, line 94: When we spin up the relays they try and get the best headers before the pallet's have been initilaized, which leads to this.

This gets propagated all the way to [`process_future_result()`](https://github.com/paritytech/parity-bridges-common/blob/master/relays/utils/src/lib.rs#L214) which treats it as an error. Ideally this would be a warning, but I don't know how to make that change without making _everything_ a warning (which I don't think is correct).

bin/rialto/node/src/chain_spec.rs
```diff
@@ -187,6 +187,6 @@ fn load_kovan_bridge_config() -> Option<BridgeKovanConfig> {
 fn load_millau_bridge_config() -> Option<BridgeMillauConfig> {
 	Some(BridgeMillauConfig {
 		init_data: None,
-		owner: Some([0; 32].into()),
+		owner: None,
```
tomusdrw, line 190: Who's going to initialize it then?

deployments/run.sh
```diff
@@ -52,7 +52,7 @@ COMPOSE_ARGS+=$BRIDGE_PATH/.env
 grep -e MATRIX_ACCESS_TOKEN -e WITH_PROXY $BRIDGE_PATH/.env > .env2 && . ./.env2 && rm .env2
 
 if [ ! -z ${MATRIX_ACCESS_TOKEN+x} ]; then
-	sed -i '' -e "s/access_token.*/access_token: \"$MATRIX_ACCESS_TOKEN\"/" ./monitoring/grafana-matrix/config.yml
+	sed -e "s/access_token.*/access_token: \"$MATRIX_ACCESS_TOKEN\"/" ./monitoring/grafana-matrix/config.yml
```
tomusdrw, line 55: Does this really work without `-i` ? I thought that without `--in-place` it would simply output the changes instead of applying them?

relays/substrate/src/headers_target.rs
tomusdrw: Can we delay the relayers to make sure the pallet is initialized?

deployments/bridges/rialto-millau/entrypoints/relay-headers-rialto-to-millau-entrypoint.sh
```diff
@@ -0,0 +1,21 @@
...
+curl -v http://rialto-node-alice:9933/health
+
+/home/user/substrate-relay initialize-rialto-headers-bridge-in-millau \
+	--millau-host millau-node-alice \
+	--millau-port 9944 \
+	--rialto-host rialto-node-alice \
+	--rialto-port 9944 \
+	--millau-signer //Alice \
+
+/home/user/substrate-relay rialto-headers-to-millau \
```
tomusdrw, line 15: If this starts to early maybe the previous command should actually wait until the transaction is included?

tomusdrw (commented):
lgtm!

bin/rialto/node/src/chain_spec.rs
HCastano: Root can initialize

deployments/run.sh
HCastano: You're right, I mixed up the flags. It should be using `-i` instead of `-e`

bin/rialto/node/src/chain_spec.rs
tomusdrw: But the relayer is supposed to call this, no? I mean does it currently go through `sudo` module? Why not set `root_key` as the owner here as well (which would be `//Alice`).

Since this is a testnet/development network I feel this would be much easier to use also from the UI if the calls don't have to be `sudo`-wrapped.

deployments/bridges/rialto-millau/entrypoints/relay-headers-rialto-to-millau-entrypoint.sh
tomusdrw: @HCastano I feel this would be more resilient solution than `sleep 6`, maybe let's add log this in the issue tracker, to improve this at some point?

tomusdrw (approved):
Looks good modulo the bridge `owner` grumble.

deployments/bridges/rialto-millau/entrypoints/relay-headers-rialto-to-millau-entrypoint.sh
HCastano: That would require changes to the relayer code, so yeah it wouldn't be appropriate to introduce that in this PR. I can open an issue

PR #494, opened 2020-11-10
Remove Millau Genesis Config in Rialto Node

HCastano:
The initial Millau data isn't exactly accurate, and now that we have #483 we don't need to have it anymore.

I'm submitting this as an individual fix instead of as part of #493 so that I can have a published Docker image to work with.


HCastano:
I'm going to merge this since the change is trivial and I want the CI to build a new Rialto image for me tonight.

PR #496, opened 2020-11-11
Limit messages weight in batch

### LINKED ISSUE #433: Limit number of messages in delivery transaction by their weight
Right now (in #399) number of messages that are delivered in a single transaction is only limited by their number. Since we have no scheduled dispatch anymore, relay code must be changed to limit number of delivered messages by their cumulative weight.

svyatonik:
closes #433 

New schema supersedes intermediate schema from #469. Basically instead of working with two numbers (`latest_nonce` on source and target nodes), message delivery race now works with mapping `nonce => weight`. So whenever new message appear at the source node, we read its weight. Then, when batch is prepared, we limit batch messages by total weight (+ by count).

I've also replaced `MessageLane::MessageNonce` associated type with `bp_message_lane::MessageNonce`, because it's alwayss now && it takes a time to propagate this type bounds all over the code. If we'll ever need to use it with other similar pallet, then it also makes sense to make lane id generic too.


relays/messages-relay/src/message_race_loop.rs
```diff
@@ -443,7 +467,7 @@ mod tests {
 
 		// target node only knows about source' BEST_AT_TARGET block
 		// source node has BEST_AT_SOURCE > BEST_AT_SOURCE block
```
HCastano, line 469:
```suggestion
		// source node has BEST_AT_SOURCE > BEST_AT_TARGET block
```

relays/messages-relay/src/message_race_receiving.rs
```diff
@@ -173,12 +178,51 @@ where
...
+	fn range_inclusive_works_as_nonces_range() {
+		let range = 20..=30;
+
+		assert_eq!(NoncesRange::begin(&range), 20);
+		assert_eq!(NoncesRange::end(&range), 30);
+		assert_eq!(range.clone().greater_than(10), Some(20..=30));
+		assert_eq!(range.clone().greater_than(19), Some(20..=30));
+		assert_eq!(range.clone().greater_than(20), Some(21..=30));
+		assert_eq!(range.clone().greater_than(25), Some(26..=30));
+		assert_eq!(range.clone().greater_than(29), Some(30..=30));
```
HCastano, line 225: I don't think we need so many points here

relays/messages-relay/src/message_race_strategy.rs
```diff
@@ -17,95 +17,169 @@
...
+				Some(queued_range)
+			} else {
+				// selector returns `Some(range)` if this `range` needs to be requeued
+				selector(queued_range)
+			};
+
+			// requeue (sub) range and update range to deliver
+			match range_to_requeue {
+				Some(range_to_requeue) => {
+					assert!(
```
HCastano, line 111: We should add a message here in case we hit this

HCastano (commented):
I've only gone through half. I can't say I follow everything in the code, but the logic in the tests looks fine to me so far

relays/messages-relay/src/message_lane_loop.rs
```diff
@@ -683,8 +720,11 @@ pub(crate) mod tests {
 					source_tick: Duration::from_millis(100),
 					target_tick: Duration::from_millis(100),
 					reconnect_delay: Duration::from_millis(0),
-					stall_timeout: Duration::from_millis(60),
-					max_unconfirmed_nonces_at_target: 100,
+					stall_timeout: Duration::from_millis(60 * 1000),
+					delivery_params: MessageDeliveryParams {
+						max_unconfirmed_nonces_at_target: 4,
+						max_messages_weight_in_single_batch: 4,
```
tomusdrw, line 726: That seems a bit low for a weight.

relays/substrate/src/millau_messages_to_rialto.rs
```diff
@@ -144,7 +143,10 @@ pub fn run(
 			target_tick: rialto_tick,
 			reconnect_delay,
 			stall_timeout,
-			max_unconfirmed_nonces_at_target: bp_rialto::MAX_UNCONFIRMED_MESSAGES_AT_INBOUND_LANE,
+			delivery_params: messages_relay::message_lane_loop::MessageDeliveryParams {
+				max_unconfirmed_nonces_at_target: bp_rialto::MAX_UNCONFIRMED_MESSAGES_AT_INBOUND_LANE,
+				max_messages_weight_in_single_batch: bp_rialto::MAXIMUM_EXTRINSIC_WEIGHT,
```
tomusdrw, line 148: I think there is an extra fee involved for just the data in a transaction, so we probably need some safety buffer here (i.e. we can't assume that the payload weight can be as high as the `MAXIMUM_EXTRINSIC_WEIGHT`).

relays/messages-relay/src/message_lane_loop.rs
svyatonik: That's enough for test

relays/messages-relay/src/message_race_receiving.rs
svyatonik: I think we need

HCastano: To me it seems that the `10` and `19` checks test the same thing, and same with the `20` and `25` ones

relays/messages-relay/src/message_race_delivery.rs
```diff
@@ -112,10 +116,18 @@ where
 			metrics_msg.update_source_latest_confirmed_nonce::<P>(latest_confirmed_nonce);
 		}
 
+		let new_nonces = if latest_generated_nonce > prev_latest_nonce {
+			self.client
+				.generated_messages_weights(at_block.clone(), prev_latest_nonce + 1..=latest_generated_nonce)
+				.await?
+		} else {
+			BTreeMap::new()
```
HCastano, line 124:
```suggestion
			MessageWeightsMap::new()
```

Might be more accurate here?

relays/messages-relay/src/message_race_delivery.rs
```diff
@@ -290,11 +301,227 @@ impl<P: MessageLane> RaceStrategy<SourceHeaderIdOf<P>, TargetHeaderIdOf<P>, P::M
 		let latest_confirmed_nonce_at_target = target_nonces.confirmed_nonce.expect(CONFIRMED_NONCE_PROOF);
 		let outbound_state_proof_required = latest_confirmed_nonce_at_target < latest_confirmed_nonce_at_source;
 
-		// https://github.com/paritytech/parity-bridges-common/issues/432
-		// https://github.com/paritytech/parity-bridges-common/issues/433
-		// TODO: number of messages must be no larger than:
-		// `max_unconfirmed_nonces_at_target - (latest_received_nonce_at_target - latest_confirmed_nonce_at_target)`
+		// If we're here, then the confirmations race did it job && sending side now knows that messages
```
HCastano, line 304:
```suggestion
		// If we're here, then the confirmations race did its job && sending side now knows that messages
```

relays/messages-relay/src/message_race_strategy.rs
```diff
@@ -17,95 +17,173 @@
...
+	/// delivered in-order. Otherwise the function will panic.
+	pub fn select_nonces_to_deliver_with_selector(
+		&mut self,
+		race_state: &RaceState<
+			HeaderId<SourceHeaderHash, SourceHeaderNumber>,
+			HeaderId<TargetHeaderHash, TargetHeaderNumber>,
+			Proof,
+		>,
+		mut selector: impl FnMut(SourceNoncesRange) -> Option<SourceNoncesRange>,
+	) -> Option<(RangeInclusive<MessageNonce>, ())> {
```
HCastano, line 77: Why's this a tuple if the second argumente is just `()`?

relays/messages-relay/src/message_race_receiving.rs
svyatonik: To me, `20` and `30` are critical points of the function, where functions is changing its behavior. `10` and `25` are points far from these points (one is before range of interest, other one is inside this range), where function has stable behavior.

relays/messages-relay/Cargo.toml
```diff
@@ -11,10 +11,13 @@ async-trait = "0.1.40"
...
 parking_lot = "0.11.0"
 
 # Bridge Dependencies
 
 bp-message-lane = { path = "../../primitives/message-lane" }
 relay-utils = { path = "../utils" }
+
+# Substrate Dependencies
+
+frame-support = "2.0"
```
tomusdrw, line 23: Seems that we only need the `Weight` type, right? Could we just use our own typedef for this? I'd rather avoid depending on substrate too much.

tomusdrw (approved):

relays/messages-relay/Cargo.toml
HCastano: Might be risky if Substrate ever changes their definition of `Weight`, no?

HCastano (approved):

relays/messages-relay/Cargo.toml
tomusdrw: Yeah, but this is something that wouldn't be done unnoticed.

tomusdrw (approved):

PR #502, opened 2020-11-12
Add Grafana Dashboards for Rialto<->Millau Header Relays

HCastano:
This adds two dashboards to the Rialto-Millau deployment. These dashboards visualize the header sync in each direction of the bridge. Additionally, there is support for Matrix alerts. The best way to review this is to run the Rialto-Millau deployment locally and take a look at the two dashboards. After you're happy with those, then you can take a quick glance at the dashboard config files in case anything seems out of place there.


deployments/bridges/rialto-millau/dashboard/grafana/provisioning/dashboards/relay-millau-to-rialto-headers-dashboard.json
```diff
@@ -0,0 +1,694 @@
...
+            },
+            "type": "query"
+          }
+        ],
+        "executionErrorState": "alerting",
+        "for": "5m",
+        "frequency": "5m",
+        "handler": 1,
+        "message": "",
+        "name": "Rialto Block Difference > 5 (TODO: change eth2sub)",
```
tomusdrw, line 53: TODO?

deployments/bridges/rialto-millau/dashboard/grafana/provisioning/dashboards/relay-rialto-to-millau-headers-dashboard.json
```diff
@@ -0,0 +1,694 @@
+{
```
tomusdrw, line 1: Do we really need two separate dashboards? I think it would be more ergonomic to present both information on a single dashboard.

deployments/bridges/rialto-millau/dashboard/grafana/provisioning/dashboards/relay-millau-to-rialto-headers-dashboard.json
tomusdrw: There is a bunch of more in this file.

HCastano: It's still in "Draft" 😉

deployments/bridges/rialto-millau/dashboard/grafana/provisioning/dashboards/relay-rialto-to-millau-headers-dashboard.json
HCastano: I'm unsure if that'll be the case. It might be more confusing to have everything in one. I can play around with it the idea though

HCastano:
I've added two extra dashboards and I want to know what you guys think about them. You can view them yourselves by running the Rialto-Millau deployment locally.

The first (crudely) uses Grafana templating variables to toggle between bridges. Basically it's the same dashboard we have now, but you can toggle between a "Rialto to Millau" view and "Millau to Rialto" view. You can view it by going to the "Sub to Sub Header Sync Dashboard".

Screen Shot 2020-11-16 at 5 26 05 PM

The advantage to this one is that we could get rid of the second dashboard config file we have and just configure the correct stats through the variables. Other than that it doesn't really offer any advantages to our current dashboards.

The second has both sides of the bridge hardcoded into it. In each panel I've added an extra query which contains the other side of the bridge. I think there's a way to do it with templating variables, but my PromQL and Grafana skills aren't quite that good. You can view this one by going to the "Mixed Header Sync Dashboard".

Screen Shot 2020-11-16 at 5 12 06 PM

This second one looks pretty cool with the unified view, but it has the issue that it's kinda hard to read the graphs (for example, look at "Difference Between Source and Target Headers", or "Average System Load").

Maybe what we should do for a unified view is have "headline" stats - so just "Best blocks", "Headers Synced", and eventually (in a different PR) number of messages relayed, etc. in one dashboard, and all the other details (CPU, queued headers, etc) in bridge specific dashboards.

tomusdrw:
> Maybe what we should do for a unified view is have "headline" stats - so just "Best blocks", "Headers Synced", and eventually (in a different PR) number of messages relayed, etc. in one dashboard, and all the other details (CPU, queued headers, etc) in bridge specific dashboards.

Both dashboards look cool, and I think I'm in favour of this approach. Basically there are two purposes we have here:
1. We want to trigger alerts in case things go south, and some graphs (like the block difference) is rather meant for the alert purpose, not for "human consumption"
2. We want to inspect the general state of the two-directional bridge.

So having separate dashboards for these purposes makes sense to me. I'm happy to do this in subsequent PRs though, would be good to have alerting on millau asap.

HCastano:
I ended up finding another issue with the combined dashboard approach. Each panel can only have one alert rule, which means we wouldn't be able to tell which side of the bridge failed. This strengthens the case for having a unified dashboard for highlights, and separate dashboards for details and alerting.

tomusdrw (approved):
LGTM!

PR #503, opened 2020-11-13
Chain specific message lane apis

### LINKED ISSUE #457: Chain-specific runtime message-delivery API

svyatonik:
closes #457 
on top of #496 => draft

I've tried all possible options, including:
1) runtime API that has additional `InstanceId` argument;
2) replacing runtime API with custom RPC methods (that are receiving additional `InstanceId` argument);
3) adding macro like `declare_message_lane_runtime_apis!(Rialto)` (failed, because I have found no way to call proc macro from decl macro);
4) bonus: I've tried to add our own wrapper of `decl_runtime_apis` that will generate these string constants, but apparently there's no way to call proc macro from other proc macro. Probably good idea to have this in original substrate macro in the future


relays/substrate/src/messages_lane.rs
```diff
@@ -0,0 +1,63 @@
...
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+use async_trait::async_trait;
+use bp_message_lane::MessageNonce;
+use codec::Encode;
+use messages_relay::message_lane::{MessageLane, SourceHeaderIdOf, TargetHeaderIdOf};
+use relay_substrate_client::Error as SubstrateError;
+use std::ops::RangeInclusive;
+
+/// Headers sync pipeline for Substrate <-> Substrate relays.
```
HCastano, line 24:
```suggestion
/// Message sync pipeline for Substrate <-> Substrate relays.
```

HCastano (approved):
I think it's a little unfortunate that there's no nice way to do this right now. Although I imagine it wouldn't be too difficult or controversial to add an `Into<String>` implementation (or something similar) to the runtime API macros in Substrate.

As far as having separate runtime APIs goes, I also think it would be nice to see the macros accept a generic argument and that was we can reuse the same API with different types (so in our case different instances of the chains). That might be a bit harder to do though.

tomusdrw (approved):
lgtm!

PR #504, opened 2020-11-13
Avoid Panic When Fetching Info Before Bridge is Initialized

HCastano:
Before we introduced the notion of initialization into the Subtrate bridge pallet we were making some assumptions about data we always expected to exist, such as as a best finalized header. However, with initialization we can now end up in a state where we have a running chain which hasn't yet been initialized and therefore doesn't have a best finalized header yet.

This PR prevents a panic in the runtime if someone tries to get the best finalized header before the pallet is initialized by returning a dummy header.


HCastano:
Looks like the breaks the message lane relay. Feel free to review this, but it should not be merged until I figure out what's wrong.

tomusdrw (commented):
IMO being explicit is not worth the complexity this PR introduces.

The pallet being uninitialized is a total edge case, given that we expect the bridge to run forever, the time it's going to be uninitialized is just irrelevant at all.

With that change, we have to pay the cost of handling `Option` forever and in every case during regular (happy) interaction, while I don't really know what are the potential gains.
It's like handling potential OOM error with every allocation you are trying to do.

IMHO the relayer (or any application I can imagine) doesn't benefit much from distinguishing between:
1. The pallet has not been initialized yet.
2. The pallet is in incorrect state.

Both states can't be resolved by this apps without external intervention, so they can either stop working or keep retrying, hoping the situation resolves soon.

Hence to fix the panic, I'd propose to simply return some dummy value (say `Zero::zero()`) instead of exposing the `Option` to external APIs.

HCastano:
> IMO being explicit is not worth the complexity this PR introduces.
> 
> The pallet being uninitialized is a total edge case, given that we expect the bridge to run forever, the time it's going to be uninitialized is just irrelevant at all.
> 
> With that change, we have to pay the cost of handling `Option` forever and in every case during regular (happy) interaction, while I don't really know what are the potential gains.
> It's like handling potential OOM error with every allocation you are trying to do.
> 
> IMHO the relayer (or any application I can imagine) doesn't benefit much from distinguishing between:
> 
>     1. The pallet has not been initialized yet.
> 
>     2. The pallet is in incorrect state.
> 
> 
> Both states can't be resolved by this apps without external intervention, so they can either stop working or keep retrying, hoping the situation resolves soon.
> 
> Hence to fix the panic, I'd propose to simply return some dummy value (say `Zero::zero()`) instead of exposing the `Option` to external APIs.

You raise some good points here. I'm not entirely sure if returning `Zero::zero()` is the best solution here, but it seems reasonable considering that the pallet would be in an incorrect state at this point in time anyways.

HCastano:
Just tested this out and everything (headers and one way messages) appears to sync properly. I wasn't sure if the dummy header would confuse the relayer when it tried to sync, but it doesn't look like that's the case.

tomusdrw (approved):
:ok_hand: thanks!

PR #509, opened 2020-11-16
Fixed referenced balance in rialto_messages.rs

svyatonik:


bin/millau/runtime/src/rialto_messages.rs
```diff
@@ -104,16 +104,16 @@ impl MessageBridge for WithRialtoMessageBridge {
 		0 // TODO: https://github.com/paritytech/parity-bridges-common/issues/391
 	}
 
-	fn this_weight_to_balance(weight: Weight) -> bp_rialto::Balance {
+	fn this_weight_to_balance(weight: Weight) -> bp_millau::Balance {
 		<crate::Runtime as pallet_transaction_payment::Trait>::WeightToFee::calc(&weight)
 	}
 
-	fn bridged_weight_to_balance(weight: Weight) -> bp_millau::Balance {
+	fn bridged_weight_to_balance(weight: Weight) -> bp_rialto::Balance {
```
tomusdrw, line 111: `this_weight_to_balance` and `bridge_weight_to_balance` are a bit misleading, they return different balance types, yet the name doesn't suggest that.

AFAICT `this_weight_to_balance` is always used with `this_chain_balance_to_bridged_chain_balance`, so instead of 3 methods in the trait can we just have:

```
fn this_weight_to_bridged_balance(weight: Weight) -> bp_rialto::Balance {}
fn bridge_weight_to_bridged_balance(weight: Weight) -> bp_rialto::Balance {}
```
?

tomusdrw (approved):
looks good, but I since we already made a mistake here I think making names a bit more explicit would be nice (as suggested in the comment).

HCastano (approved):

bin/millau/runtime/src/rialto_messages.rs
svyatonik: I don't like the idea of doing multiple conversions (weight -> balance1 -> balance2) in single function. Can change the names, though.

PR #515, opened 2020-11-18
Fixed message signed by relay

svyatonik:


tomusdrw (approved):
Can you mention what the difference exactly is? What type do we `.into()` to?

svyatonik:
It is converted into `bp_millau::AccountSigner` (which is an alias to `MultiSigner`), which is used in the chain (in `call-dispatch` pallet to verify signature). Before it was just sr25519 public, which is used across relay => encoding is different.

tomusdrw:
Clear, thanks!

HCastano (approved):

PR #516, opened 2020-11-18
Fixed storage key used to prove outbound lane state

svyatonik:


tomusdrw (approved):

HCastano (approved):
@svyatonik can you give a bit of context as to how you found out this was broken?

PR #517, opened 2020-11-18
Use (some) different chain primitives in Millau

svyatonik:
The point of this PR is to prove that relay + pallets are able to bridge heterogeneous chains. Millau will be using separate hash, hasher and block number. More (mostly account id, signature and Millau<-> Rialto conversion rate) are to come in next PRs.


tomusdrw (approved):
:clap: That's an awesome change, thanks!

HCastano (approved):
Cool stuff!

PR #519, opened 2020-11-19
Add Derived Account Origins to Dispatcher

### LINKED ISSUE #518: Add `CallOrigin::DerivedAccount` to `Call::decode` dispatcher
The idea is to add a third kind here:
https://github.com/paritytech/parity-bridges-common/blob/5d8abad06a7661136cc004ee8940183fa9728a73/modules/call-dispatch/src/lib.rs#L60
namely `CallOrigin::DerivedAccount(source_public)`.

1. On the source chain:
 - Such origin can only be send if `ensure_signed(origin) == source_public` (similar to how we verify `RealAccount`)
2. On the target chain:
 - Such origin translates to `origin = derive_target_origin(source_public)` (similar to `BridgeAccount`).

This would allow predictable target-chain account ids for source-chain accounts, so in the context of Polkadot<>Kusama bridge you would be able to send someone DOTs knowing their KSM address.

This is an alternative to #448, there we transfer the ASSET between the networks (over the bridge), and with this we rather transfer the INTENT what to do with an asset.

HCastano:
This PR introduces some new origins for accepting and dispatching messages on a target chain. 

When a message is being sent from a `source` chain to a `target` chain, the message must originate from an Origin on the `source` chain. Depending on what this source Origin is, the message will be dispatched from a different Origin on the `target` chain. This PR changes what those origins are to be: `Root`, `Target`, and `Source`. 

The `Root` and `Source` origins on the `target` chain are derived using the origin on the `source` chain. This means that they will (likely) not be controlled by a private key on the `target` chain, which is fine for representing things like pallets and proxies which may want to send messages across the bridge. The `Target` origin uses a "real" account (so one with a private key) on the target chain to dispatch the message.

Closes #518.


primitives/runtime/src/lib.rs
```diff
@@ -52,11 +52,20 @@ pub type InstanceId = [u8; 4];
...
 }
+
+/// Derive an account ID on this chain from a foreign account ID.
+pub fn derive_target_account_id<Public, AccountId>(bridge: InstanceId, source_public: Public) -> AccountId
+where
+	Public: Encode,
+	AccountId: Encode + Decode + Default,
+{
+	let target = (bridge, source_public).using_encoded(blake2_256);
+	AccountId::decode(&mut &target[..]).unwrap_or_default()
```
tomusdrw, line 70: This assumes `AccountId` has the same size as `blake2_256` output, while this is the case for the chains we know of, I don't think it's particularly future proof. I think we should rather have this function return `H256` and force the chains to do the conversion and provide an `.expect` proof for why it won't fail in a place where we operate on concrete `AccountId` types.

modules/call-dispatch/src/lib.rs
```diff
@@ -46,24 +44,41 @@ use sp_std::{marker::PhantomData, prelude::*};
 /// Spec version type.
 pub type SpecVersion = u32;
 
-/// Origin of the call on the target chain.
+/// Origin of a Call when it is dispatched on this (a.k.a target) chain.
```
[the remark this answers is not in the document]
HCastano: The reason why I think this makes sense is because this origin is really only used when a call is actually dispatched on the target chain. The usage on the source side is through `verify_sending_message()`, which is only used to check if the correct origin will be used on the target chain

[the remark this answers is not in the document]
tomusdrw: True, but I think the user on the source chain needs that type though, cause it's part of the payload sent to `message_lane::send_message` function. I guess it makes sense to split the source and target types though and make them more explicit as suggested in the other comment.

primitives/runtime/src/lib.rs
HCastano: I couldn't figure out a good approach to this today. Will work on it again on Monday

bin/millau/runtime/src/lib.rs
```diff
@@ -225,7 +225,7 @@ impl pallet_bridge_call_dispatch::Trait for Runtime {
 	type Event = Event;
 	type MessageId = (bp_message_lane::LaneId, bp_message_lane::MessageNonce);
 	type Call = Call;
-	type SourceChainAccountPublic = MultiSigner;
+	type SourceChainAccountId = MultiSigner;
```
tomusdrw, line 228: We should use `<T as frame_system::Trait>::AccountId` instead of this type, also it's incorrect to use `MultiSigner` here, since it's public keys.

modules/call-dispatch/src/lib.rs
```diff
@@ -38,45 +36,56 @@ use frame_support::{
...
-	/// `SourceChainAccountPublic` account (use the `fn verify_sending_message()`).
-	RealAccount(SourceChainAccountPublic, TargetChainAccountPublic, TargetChainSignature),
+pub enum CallOrigin<SourceChainAccountId, TargetChainAccountPublic, TargetChainSignature> {
+	/// Call originates from the Root origin on the source chain.
+	///
+	/// This is useful if the target chain needs some way of knowing that a call came from a
+	/// priviledged origin on the source chain (maybe to allow a configuration change for example).
+	SourceRoot,
+
+	/// Call originates from an account controlled by a private key on the target chain.
```
tomusdrw, line 60: Where, actually the call originates from `SourceChainAccountId` on the source chain, but we want to dispatch the message using an account controlled by a private key on the target chain.

Maybe it makes sense to add some structure to the docs, where for each variant we always say:
1. What originates the call on the source chain.
2. What origin does it expect to have on the target chain.

modules/call-dispatch/src/lib.rs
```diff
@@ -194,11 +202,11 @@ impl<T: Trait<I>, I: Instance> MessageDispatch<T::MessageId> for Module<T, I> {
 
 		// prepare dispatch origin
 		let origin_account = match message.origin {
-			CallOrigin::BridgeAccount => bridge_account_id(bridge, CALL_DISPATCH_MODULE_PREFIX),
-			CallOrigin::RealAccount(source_public, target_public, target_signature) => {
+			CallOrigin::SourceRoot => derive_target_account_id(bridge, T::SourceChainAccountId::default()),
```
tomusdrw, line 205: Can we prefix this somehow, so there is no collision with `CallOrigin::SourceAccount`?

modules/call-dispatch/src/lib.rs
```diff
@@ -238,35 +247,47 @@ impl<T: Trait<I>, I: Instance> MessageDispatch<T::MessageId> for Module<T, I> {
...
+		CallOrigin::SourceRoot => {
 			ensure_root(sender_origin)?;
 			Ok(None)
 		}
-		CallOrigin::RealAccount(ref this_account_public, _, _) => {
-			let this_chain_account_id = ensure_signed(sender_origin)?;
-			if this_chain_account_id != this_account_public.clone().into_account() {
+		CallOrigin::TargetAccount(ref source_account_id, _, _) => {
+			let source_chain_signer = ensure_signed(sender_origin)?;
+			if source_chain_signer != source_account_id.clone().into_account() {
```
tomusdrw, line 279:
```suggestion
			if source_chain_signer != source_account_id {
```

I think it was fine as it was before, we just doesn't need to convert public -> account id now.

modules/call-dispatch/src/lib.rs
```diff
@@ -238,35 +247,47 @@ impl<T: Trait<I>, I: Instance> MessageDispatch<T::MessageId> for Module<T, I> {
...
 				return Err(BadOrigin);
 			}
 
-			Ok(Some(this_chain_account_id))
+			Ok(Some(source_chain_signer))
+		}
+		CallOrigin::SourceAccount(ref source_account_id) => {
+			let source_chain_signer = ensure_signed(sender_origin)?;
+			if source_chain_signer != source_account_id.clone().into_account() {
```
tomusdrw, line 287: ditto

modules/call-dispatch/src/lib.rs
```diff
@@ -289,6 +311,25 @@ mod tests {
 
 	type MessageId = [u8; 4];
 
+	// These derives are a bit gross but we need all of them to ensure we fit into Substrate's
+	// definition of what an `AccountId` is
+	#[derive(Debug, Encode, Decode, Clone, PartialEq, Eq, Default, PartialOrd, Ord, Serialize, Deserialize)]
+	pub struct TestAccountId(AccountId);
```
tomusdrw, line 317: This doesn't seem right, since we have `TestAccountPublic`, which identifies `AcocuntIds` we should simply be able to use `AccountId` - this extra `IdentifyAccount` shouldn't be necessary.

modules/call-dispatch/src/lib.rs
HCastano: When I was saying "originates" I was writing from the perspective of a call on the target chain. But yeah, I think we could benefit from more clarity here

modules/call-dispatch/src/lib.rs
HCastano: I'm confused as to what "before" you're referring to here. The main changes here were just renaming variables

modules/call-dispatch/src/lib.rs
HCastano: If you're referring to the typedef for `AccountId`, that's not going to work. However, I didn't try using `TestRuntime::AccountId` - that might be a better approach

modules/call-dispatch/src/lib.rs
tomusdrw: The point is that `ensure_signed` returns an `AccountId`, and you have `source_account_id` on the RHS side, so there is no conversion necessary at all.

bin/millau/runtime/src/lib.rs
HCastano: `MultiSigner` implements [`IdentifyAccount`](https://crates.parity.io/sp_runtime/enum.MultiSigner.html#impl-IdentifyAccount), so why isn't it fine to use it here?

tomusdrw: Because we don't want `IdentifyAccount` here :) We want `IdentifyAccount::AccountId` here, since it's a `SourceChainAccountId`, not public key.

modules/call-dispatch/src/lib.rs
HCastano: Looking at it now, the type of `source_account_id` appears to be `TargetChainAccountPublic`, so there's something wonky here. I'll fix it

modules/call-dispatch/src/lib.rs
HCastano: This was required since we were using the wrong type in `verify_message_origin()`

primitives/runtime/src/lib.rs
```diff
@@ -50,13 +50,48 @@ pub type InstanceId = [u8; 4];
...
 	AccountId::decode(&mut &entropy[..]).unwrap_or_default()
 }
+
+/// A trait used to map AccountIds from a source chain to those on a target chain.
+pub trait AccountIdConverter<S, T> {
+	/// Convert the source chain AccountId to a target chain AccountId.
+	///
+	/// A context can be optionally provided in order to provide extra entropy when deriving the
+	/// target account IDs.
+	fn convert(context: Option<&[u8]>, id: S) -> T;
```
HCastano, line 67: I wish the signature could just be `fn convert(S) -> T`, but having the `context` is handy (particularly for having that extra context with the root account derivation). Maybe it makes sense to just make this mandatory?

tomusdrw: TBH I'd rather fallback to previous simple function and just pass an enum indicating whetever we are deriving from an arbitrary account or a root account, something like this:
```
pub enum SourceAccount<T> {
  Root,
  AccountId(T),
}
fn derive_account_id<T>(bridge_id: InstanceId, source_account: SourceAccount<T>) -> H256 {
   match source_account {
      SourceAccount::Root => ("root", bridge_id).using_encoded(blake2_256),
      SourceAccount::AccountId(acc) => ("acc", bridge_id, acc).using_encoded(blake2_256),
   }
}
```
There are multiple reasons why I prefer this:
- `TargetAccountId` conversion should happen externaly (hence we return `H256` here)
- I don't think this should be configurable, it should rather be predictable (see below)
- I expect this function to be re-used (or re-implemented) in UIs or tools, so it should not rely on anything external (like `context`) if it doesn't have to.

primitives/runtime/src/lib.rs
```diff
@@ -50,13 +50,48 @@ pub type InstanceId = [u8; 4];
 /// The `module_prefix` (arbitrary slice) may be used to generate module-level
 /// "system" account, so you could have separate "system" accounts for currency
 /// exchange, message dispatch and other modules.
-///
-/// The account is not supposed to actually exists on the chain, or to have any funds.
-/// It is only used to
 pub fn bridge_account_id<AccountId>(bridge: InstanceId, module_prefix: &[u8]) -> AccountId
```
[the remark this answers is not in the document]
HCastano: Yeah, I was waiting to see how the target to source account derivation played out before I removed this

modules/call-dispatch/src/lib.rs
```diff
@@ -194,11 +217,21 @@ impl<T: Trait<I>, I: Instance> MessageDispatch<T::MessageId> for Module<T, I> {
...
-			CallOrigin::BridgeAccount => bridge_account_id(bridge, CALL_DISPATCH_MODULE_PREFIX),
-			CallOrigin::RealAccount(source_public, target_public, target_signature) => {
+			CallOrigin::SourceRoot => {
+				let encoded_id = derive_account_id::<T::SourceChainAccountId>(bridge, SourceAccount::Root);
+				match T::AccountId::decode(&mut &encoded_id[..]) {
+					Ok(id) => id,
+					Err(_) => {
+						let error = <Error<T, I>>::InvalidAccountId.into();
+						Self::deposit_event(RawEvent::InvalidAccountId(bridge, id, error));
+						return;
```
HCastano, line 227: Tbh I'm not sure how to test the codepath for this right now. The test runtime has an `AccountId` of `u64` which I don't think can ever fail to be decoded. I _could_ introduce a wrapper type and then have a sketchy `Decode` implementation which fails on a specific input, but I don't like that.

tomusdrw: Could we use `UncheckedFrom<H256>` or `From<H256>` or even `TryFrom<H256>` instead of `Decode` to get `AccountId`? I'm not convinced that `Decode` is the right choice here at all, cause usually SCALE can decode something even from garbage data. (like decoding `u64` from `H256`).

Alternatively, let's make the conversion explicit by using `Convert<H256, AccountId>` trait from substrate and passing an extra parameter, so that the runtime can provide a proof that the conversion is always successful.

modules/message-lane/src/instant_payments.rs
```diff
@@ -76,6 +76,7 @@ where
 
 /// Return account id of shared relayers-fund account that is storing all fees
 /// paid by submitters, until they're claimed by relayers.
-fn relayers_fund_account<AccountId: Default + Decode>() -> AccountId {
-	bridge_account_id(NO_INSTANCE_ID, MESSAGE_LANE_MODULE_PREFIX)
+fn relayers_fund_account<AccountId: Default + Encode + Decode>() -> AccountId {
+	let encoded_id = derive_account_id::<AccountId>(NO_INSTANCE_ID, SourceAccount::Account(AccountId::default()));
+	AccountId::decode(&mut &encoded_id[..]).unwrap_or_default()
```
tomusdrw, line 81: Same here, this is actually super dangerous, let's make the conversion explicit and let's force the user to provide one instead of relying on `Decoding` the data.

primitives/runtime/src/lib.rs
```diff
@@ -46,17 +47,31 @@ pub const MESSAGE_LANE_MODULE_PREFIX: &[u8] = b"pallet-bridge/message-lane";
...
+/// Derive an account ID from a foreign account ID.
+///
+/// This function returns an encoded Blake2 hash. It is the responsibility of the caller to ensure
+/// this can be succesfully decoded into an AccountId.
 ///
-/// The account is not supposed to actually exists on the chain, or to have any funds.
-/// It is only used to
-pub fn bridge_account_id<AccountId>(bridge: InstanceId, module_prefix: &[u8]) -> AccountId
+/// The `bridge_id` is used to provide extra entropy when producing account IDs. This helps prevent
+/// AccountId collisions between different bridges on a single target chain.
```
tomusdrw, line 67: Could we add a note here that if the same `InstanceId` is used on different chains though (say Kusama -> Millau and Kusama -> Polkadot bridges), then the derived account id would be the same.
It's kind of obvious, but I'd like to emphasize that, cause it has consequences for both user's privacy and security.

tomusdrw (commented):
looks good! Just the `H256 -> AccountId` conversion should imho be done with type safety to prevent issues.

modules/message-lane/src/instant_payments.rs
HCastano: Lol yeah I meant to make a note about this asking for alternatives to the `_or_default()` call 😅

HCastano: I changed this address to be a configuration option in the `message-lane` pallet instead of being derived on every call. The main implication (from what I can tell) is that now each instance of the pallet will need to choose a `RelayerFundAccount`, which may or may not be the same across instances. Before this change all instances used to use the same account.

I'm not sure if this has any serious implications, do you mind chiming in here @svyatonik? The commit where this change is introduced is here: 88e28a84a07634cd8cd0c45ccf57481aaa5b611b.

modules/message-lane/src/lib.rs
```diff
@@ -95,6 +95,12 @@ pub trait Trait<I = DefaultInstance>: frame_system::Trait {
 	/// Identifier of relayer that deliver messages to this chain. Relayer reward is paid on the bridged chain.
 	type InboundRelayer: Parameter;
 
+	/// AccountId of the shared relayer fund account.
+	///
+	/// This account stores all the fees paid by submitters. Relayers are able to claim these
+	/// funds as at their convenience.
+	type RelayerFundAccount: Get<Self::AccountId>;
```
tomusdrw, line 102: is this really necessary? Can't we pass something like `hash("relayer-fund-account" ++ BridgeInstanceId)` to `AcountIdConverter`? I think it would be better to have this pallet-hardcoded than freely configurable (easier to create tooling).

tomusdrw (approved):
lgtm! Tiny grumble about the relayer fund account, which I don't think needs to be configurable.

modules/message-lane/src/lib.rs
```diff
@@ -645,6 +644,16 @@ fn verify_and_decode_messages_proof<Chain: SourceHeaderChain<Fee>, Fee, Dispatch
 	})
 }
 
+/// AccountId of the shared relayer fund account.
+///
+/// This account stores all the fees paid by submitters. Relayers are able to claim these
+/// funds as at their convenience.
+fn relayer_fund_account_id<T: Trait<I>, I: Instance>() -> T::AccountId {
+	use sp_runtime::traits::Convert;
+	let encoded_id = bp_runtime::derive_relayer_fund_account_id(bp_runtime::NO_INSTANCE_ID);
```
tomusdrw, line 653: Shouldn't that be instance aware? Is there a risk if the fund is shared between two instances of the bridge?

HCastano: This account was previously being derived with `NO_INSTANCE_ID` and `MESSAGE_LANE_MODULE_PREFIX` before. This resulted in the same account across different bridge instances, so no behaviour is being changed here

PR #525, opened 2020-11-23
Ability to send messages from Rialto to Millau

svyatonik:
Some changes are still required (in follow-up PRs) to include this into test deployments

PR #526, opened 2020-11-23
Add more substrate accounts + use one account in one relay

svyatonik:
Changing names would break Substrate cli options (`--alice`, `--bob`, ...) that we use + we'll be unable to use matching keyring anymore.

This PR reserves `Ferdie` account, but it is yet unused - missing components should be added after this PR and #525

### DIFF
--- a/deployments/README.md
+++ b/deployments/README.md
@@ -83,13 +83,13 @@ Rialto authorities are named: `Alice`, `Bob`, `Charlie`, `Dave`, `Eve`.
 Rialto-PoA authorities are named: `Arthur`, `Bertha`, `Carlos`.
 Millau authorities are named: `Alice`, `Bob`, `Charlie`, `Dave`, `Eve`.
 
-Both authorities and following accounts have enough funds on corresponding Substrate chains:
+Both authorities and following accounts have enough funds (for test purposes) on corresponding Substrate chains:
 
 - on Rialto: `Ferdie`, `George`, `Harry`.
 - on Millau: `Ferdie`, `George`, `Harry`.
 
 Names of accounts on Substrate (Rialto and Millau) chains may be prefixed with `//` and used as
-seeds for the `sr25519` keys. This seed may also be used in the signer argument in substrate
+seeds for the `sr25519` keys. This seed may also be used in the signer argument in Substrate
 and PoA relays. Example:
 
 ```bash
--- a/deployments/README.md
+++ b/deployments/README.md
@@ -116,8 +116,8 @@ Following accounts are used when `rialto-millau` bridge is running:
 
 - Millau's `Charlie` signs relay transactions with new Rialto headers;
 - Rialto's `Charlie` signs relay transactions with new Millau headers;
-- Millau's `Dave` signs Millau transactions with messages to Rialto;
-- Rialto's `Dave` signs Rialto transactions with messages to Millau;
+- Millau's `Dave` signs Millau transactions which contain messages for Rialto;
+- Rialto's `Dave` signs Rialto transactions which contain messages for Millau;
 - Millau's `Eve` signs relay transactions with message delivery confirmations from Rialto to Millau;
 - Rialto's `Eve` signs relay transactions with messages from Millau to Rialto;
 - Millau's `Ferdie` signs relay transactions with messages from Rialto to Millau;
--- a/deployments/README.md
+++ b/deployments/README.md
@@ -50,7 +50,7 @@ We currently support two bridge deployments
 These networks can be deployed using our [`./run.sh`](./run.sh) script.
 
 The first argument it takes is the name of the bridge you want to run. Right now we only support two
-networks: `poa-rialto` and `rialto-millau`.
+bridges: `poa-rialto` and `rialto-millau`.
 
 ```bash
 ./run.sh poa-rialto
@@ -79,8 +79,49 @@ not strictly required.
 
 ## General Notes
 
-Substrate authorities are named: `Alice`, `Bob`, `Charlie`, `Dave`, `Eve`, `Ferdie`.
-Ethereum authorities are named: `Arthur`, `Bertha`, `Carlos`.
+Rialto authorities are named: `Alice`, `Bob`, `Charlie`, `Dave`, `Eve`.
+Rialto-PoA authorities are named: `Arthur`, `Bertha`, `Carlos`.
+Millau authorities are named: `Alice`, `Bob`, `Charlie`, `Dave`, `Eve`.
+
+Both authorities and following accounts have enough funds on corresponding Substrate chains:
+
+- on Rialto: `Ferdie`, `George`, `Harry`.
+- on Millau: `Ferdie`, `George`, `Harry`.
+
+Names of accounts on Substrate (Rialto and Millau) chains may be prefixed with `//` and used as
+seeds for the `sr25519` keys. This seed may also be used in the signer argument in substrate
+and PoA relays. Example:
+
+```bash
+./substrate-relay rialto-headers-to-millau \
+	--rialto-host rialto-node-alice \
+	--rialto-port 9944 \
+	--millau-host millau-node-alexander \
+	--millau-port 9944 \
+	--rialto-signer //Harry \
+	--prometheus-host=0.0.0.0
+```
+
+Some accounts are used by bridge components. Using these accounts to sign other transactions
+is not recommended, because this may lead to nonces conflict.
+
+Following accounts are used when `poa-rialto` bridge is running:
+
+- Rialto' `Alice` signs relay transactions with new Rialto-PoA headers;
+- Rialto' `Bob` signs relay transactions with Rialto-PoA -> Rialto currency exchange proofs.
+- Rialto-PoA' `Arthur`: signs relay transactions with new Rialto headers;
+- Rialto-PoA' `Bertha`: signs currency exchange transactions.
+
+Following accounts are used when `rialto-millau` bridge is running:
+
+- Millau' `Charlie` signs relay transactions with new Rialto headers;
+- Rialto' `Charlie` signs relay transactions with new Millau headers;
+- Millau' `Dave` signs Millau transactions with messages to Rialto;
+- Rialto' `Dave` signs Rialto transactions with messages to Millau;
+- Millau' `Eve` signs relay transactions with message delivery confirmations from Rialto to Millau;
+- Rialto' `Eve` signs relay transactions with messages from Millau to Rialto;
+- Millau' `Ferdie` signs relay transactions with messages from Rialto to Millau;
+- Rialto' `Ferdie` signs relay transactions with message delivery confirmations from Millau to Rialto.
 
 ### Docker Usage
 When the network is running you can query logs from individual nodes using:
--- a/deployments/README.md
+++ b/deployments/README.md
@@ -107,21 +107,21 @@ is not recommended, because this may lead to nonces conflict.
 
 Following accounts are used when `poa-rialto` bridge is running:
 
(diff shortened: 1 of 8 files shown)


deployments/README.md
```diff
@@ -79,8 +79,49 @@ not strictly required.
...
+	--rialto-signer //Harry \
+	--prometheus-host=0.0.0.0
+```
+
+Some accounts are used by bridge components. Using these accounts to sign other transactions
+is not recommended, because this may lead to nonces conflict.
+
+Following accounts are used when `poa-rialto` bridge is running:
+
+- Rialto' `Alice` signs relay transactions with new Rialto-PoA headers;
```
tomusdrw, line 110:
```suggestion
- Rialto's `Alice` signs relay transactions with new Rialto-PoA headers;
```
aren't `'s` missing everywhere?

deployments/README.md
```diff
@@ -79,8 +79,49 @@ not strictly required.
...
+- Rialto-PoA' `Bertha`: signs currency exchange transactions.
+
+Following accounts are used when `rialto-millau` bridge is running:
+
+- Millau' `Charlie` signs relay transactions with new Rialto headers;
+- Rialto' `Charlie` signs relay transactions with new Millau headers;
+- Millau' `Dave` signs Millau transactions with messages to Rialto;
+- Rialto' `Dave` signs Rialto transactions with messages to Millau;
+- Millau' `Eve` signs relay transactions with message delivery confirmations from Rialto to Millau;
+- Rialto' `Eve` signs relay transactions with messages from Millau to Rialto;
```
tomusdrw, line 122: I'm not sure I understand what `Rialto's Eve' does. How does that differ from `Dave`?

tomusdrw (commented):
:heart:  the docs, thanks!

deployments/README.md
```diff
@@ -79,8 +79,49 @@ not strictly required.
 
 ## General Notes
```
HCastano, line 81:
```suggestion
```

deployments/README.md
```diff
@@ -79,8 +79,49 @@ not strictly required.
...
+
+Some accounts are used by bridge components. Using these accounts to sign other transactions
+is not recommended, because this may lead to nonces conflict.
+
+Following accounts are used when `poa-rialto` bridge is running:
+
+- Rialto' `Alice` signs relay transactions with new Rialto-PoA headers;
+- Rialto' `Bob` signs relay transactions with Rialto-PoA -> Rialto currency exchange proofs.
+- Rialto-PoA' `Arthur`: signs relay transactions with new Rialto headers;
+- Rialto-PoA' `Bertha`: signs currency exchange transactions.
```
HCastano, line 113: I think we should try and avoid calling these `Rialto-PoA`, and stick with either `PoA` or `Ethereum-PoA`

deployments/README.md
```diff
@@ -79,8 +79,49 @@ not strictly required.
...
+
+Names of accounts on Substrate (Rialto and Millau) chains may be prefixed with `//` and used as
+seeds for the `sr25519` keys. This seed may also be used in the signer argument in substrate
+and PoA relays. Example:
+
+```bash
+./substrate-relay rialto-headers-to-millau \
+	--rialto-host rialto-node-alice \
+	--rialto-port 9944 \
+	--millau-host millau-node-alexander \
```
HCastano, line 99:
```suggestion
	--millau-host millau-node-alice \
```

This is the name used in the Compose deployment

deployments/README.md
```diff
@@ -50,7 +50,7 @@ We currently support two bridge deployments
 These networks can be deployed using our [`./run.sh`](./run.sh) script.
```
HCastano, line 50:
```suggestion
These bridges can be deployed using our [`./run.sh`](./run.sh) script.
```

To be consistent with the other fix

deployments/README.md
```diff
@@ -79,8 +79,49 @@ not strictly required.
 
 ## General Notes
 
-Substrate authorities are named: `Alice`, `Bob`, `Charlie`, `Dave`, `Eve`, `Ferdie`.
-Ethereum authorities are named: `Arthur`, `Bertha`, `Carlos`.
+Rialto authorities are named: `Alice`, `Bob`, `Charlie`, `Dave`, `Eve`.
+Rialto-PoA authorities are named: `Arthur`, `Bertha`, `Carlos`.
+Millau authorities are named: `Alice`, `Bob`, `Charlie`, `Dave`, `Eve`.
```
HCastano, line 84: nit: I'd swap these two lines so have Millau and Rialto beside each other

deployments/README.md
```diff
@@ -79,8 +79,49 @@ not strictly required.
...
+
+Both authorities and following accounts have enough funds on corresponding Substrate chains:
```
HCastano, line 86: I'd be more specific as to what "enough funds" means here. Maybe something like "enough funds for testing"?

deployments/README.md
```diff
@@ -79,8 +79,49 @@ not strictly required.
...
+Following accounts are used when `poa-rialto` bridge is running:
+
+- Rialto' `Alice` signs relay transactions with new Rialto-PoA headers;
+- Rialto' `Bob` signs relay transactions with Rialto-PoA -> Rialto currency exchange proofs.
+- Rialto-PoA' `Arthur`: signs relay transactions with new Rialto headers;
+- Rialto-PoA' `Bertha`: signs currency exchange transactions.
+
+Following accounts are used when `rialto-millau` bridge is running:
+
+- Millau' `Charlie` signs relay transactions with new Rialto headers;
```
HCastano, line 117: There's no need for punctuation at the end of the bullet points (`;` and `.`) - these aren't sentences

deployments/README.md
```diff
@@ -79,8 +79,49 @@ not strictly required.
...
+	--prometheus-host=0.0.0.0
+```
+
+Some accounts are used by bridge components. Using these accounts to sign other transactions
+is not recommended, because this may lead to nonces conflict.
+
+Following accounts are used when `poa-rialto` bridge is running:
+
+- Rialto' `Alice` signs relay transactions with new Rialto-PoA headers;
+- Rialto' `Bob` signs relay transactions with Rialto-PoA -> Rialto currency exchange proofs.
```
HCastano, line 111: Same nit on the punctuation here

deployments/README.md
HCastano: So Dave is used by the relayer that _submits_ a message that a user wants sent across a lane, while Eve is used for the relayer which servers (so actually sends) that message lane. This should be clarified a bit in the docs

deployments/README.md
HCastano: I've seen my Mom (non-native English speaker) do the same thing before. The possessive apostrophe can be pretty confusing

deployments/README.md
```diff
@@ -79,8 +79,49 @@ not strictly required.
...
+Rialto-PoA authorities are named: `Arthur`, `Bertha`, `Carlos`.
+Millau authorities are named: `Alice`, `Bob`, `Charlie`, `Dave`, `Eve`.
+
+Both authorities and following accounts have enough funds on corresponding Substrate chains:
+
+- on Rialto: `Ferdie`, `George`, `Harry`.
+- on Millau: `Ferdie`, `George`, `Harry`.
+
+Names of accounts on Substrate (Rialto and Millau) chains may be prefixed with `//` and used as
+seeds for the `sr25519` keys. This seed may also be used in the signer argument in substrate
```
HCastano, line 92:
```suggestion
seeds for the `sr25519` keys. This seed may also be used in the `signer` argument in Substrate
```

HCastano (commented):
I also appreciate the docs, thanks!

deployments/README.md
svyatonik: While we may (in theory) deploy bridge with any Ethereum-like PoA chain with PoA consensus engine, here we talk about specific chain that has 3 authorities (Arthur, Bertha and Carlos). So I don't think any generic nicknames are suitable here.

deployments/README.md
svyatonik: That's correct - Dave is signing transactions with new messages on the sending side. And Eve is used by relay to sign transactions that deliver these messages to the receiving side. Please suggest changes

deployments/README.md
HCastano: Sure, but in practice we've only got one instance of the PoA network being used. The deployment files and documentation has been moving away from the `Rialto-PoA` name (although it still lives on a little bit in the code)

deployments/README.md
```diff
@@ -79,8 +79,49 @@ not strictly required.
...
+- Rialto's `Alice` signs relay transactions with new Rialto-PoA headers;
+- Rialto's `Bob` signs relay transactions with Rialto-PoA -> Rialto currency exchange proofs.
+- Rialto-PoA's `Arthur`: signs relay transactions with new Rialto headers;
+- Rialto-PoA's `Bertha`: signs currency exchange transactions.
+
+Following accounts are used when `rialto-millau` bridge is running:
+
+- Millau's `Charlie` signs relay transactions with new Rialto headers;
+- Rialto's `Charlie` signs relay transactions with new Millau headers;
+- Millau's `Dave` signs Millau transactions with messages to Rialto;
```
HCastano, line 119:
```suggestion
- Millau's `Dave` signs Millau transactions which contain messages for Rialto;
```

deployments/README.md
```diff
@@ -79,8 +79,49 @@ not strictly required.
...
+- Rialto's `Dave` signs Rialto transactions with messages to Millau;
+- Millau's `Eve` signs relay transactions with message delivery confirmations from Rialto to Millau;
```
HCastano, line 121:
```suggestion
- Millau's `Eve` signs transactions for operating the message lane relay from Rialto to Millau;
```

HCastano (approved):

deployments/README.md
HCastano: I left some suggestion. I'd use those/something similar with Dave, Eve, and Ferdie

deployments/README.md
svyatonik: Well, we have at least two PoA bridges built into Rialto runtime - RialtoPoA and Kovan. We have `RialtoPoAHeaderApi` runtime API. We have `bin/rialto/runtime/src/rialto_poa.rs` file. As I said, I don't think that refering any chain with generic names is ok and if that is done in code, that probably should be fixed.

deployments/README.md
svyatonik: I don't think we have any strict rule about this defined? I could see Substrate is using punctuation in many places (e.g. [here](https://github.com/paritytech/substrate/tree/master/frame/system), [here](https://github.com/paritytech/substrate/tree/master/frame/transaction-payment), [here](https://github.com/paritytech/substrate/tree/master/frame/recovery)), so I don't think it should be changed.

deployments/README.md
svyatonik: It is too generic. Dave is also signing relay transactions for operating the message lane relay

deployments/README.md
HCastano: I just read some posts on the punctuation of bullet lists, and the gist I got was that we _may_ use periods at the end of each point here. The reason is because they're (kinda) complete sentences. However, we should _not_ be using semicolons. The examples you gave are clearly full sentences, which is why the have periods at the end.

If we want to be consistent, which is one approach to addressing this, then we should just use periods. However, I don't really want to spend more time writing about punctuation, so I'll let you decide what's best.

svyatonik: [That's the](https://english.stackexchange.com/questions/75591/semicolons-at-the-end-of-list-items) first line in the "punctuation at bullet lists semicolon" google results, so I'm going with semicolon.

PR #528, opened 2020-11-24
Migrate back to Substrate master

svyatonik:


tomusdrw (approved):

HCastano (approved):
For future readers, the motivation for this move is two fold. 

First, we want to start trying to integrate the pallets from this repo into the Polkadot repo. Second, we'd like to start trying to integrate the lastest Parachain code into this repo. Both of these require that we follow Substrate `master` since the current `v2.0` release isn't up-to-date enough.

PR #529, opened 2020-11-25
Generic SyncHeader type

svyatonik:
(to simplify integration with Kusama/Polkadot)

### DIFF
--- a/relays/substrate-client/src/sync_header.rs
+++ b/relays/substrate-client/src/sync_header.rs
@@ -15,7 +15,7 @@
 // along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
 
 use headers_relay::sync_types::SourceHeader;
-use num_traits::One;
+use num_traits::{CheckedSub, One};
 use relay_utils::HeaderId;
 use sp_runtime::traits::Header as HeaderT;
 
@@ -51,6 +51,11 @@ impl<Header: HeaderT> SourceHeader<Header::Hash, Header::Number> for SyncHeader<
 	}
 
 	fn parent_id(&self) -> HeaderId<Header::Hash, Header::Number> {
-		relay_utils::HeaderId(*self.number() - One::one(), *self.parent_hash())
+		relay_utils::HeaderId(
+			self.number()
+				.checked_sub(&One::one())
+				.expect("should never be called for genesis header"),
+			*self.parent_hash(),
+		)
 	}
 }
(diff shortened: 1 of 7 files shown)


relays/substrate-client/src/sync_header.rs
```diff
@@ -0,0 +1,56 @@
...
+	}
+}
+
+impl<Header: HeaderT> SourceHeader<Header::Hash, Header::Number> for SyncHeader<Header> {
+	fn id(&self) -> HeaderId<Header::Hash, Header::Number> {
+		relay_utils::HeaderId(*self.number(), self.hash())
+	}
+
+	fn parent_id(&self) -> HeaderId<Header::Hash, Header::Number> {
+		relay_utils::HeaderId(*self.number() - One::one(), *self.parent_hash())
```
tomusdrw, line 54: I guess we never use this with genesis header, right? I'd still prefer to be on the safe side here and use `saturating_sub`

relays/substrate-client/src/sync_header.rs
```diff
@@ -0,0 +1,56 @@
...
+use sp_runtime::traits::Header as HeaderT;
+
+/// Generic wrapper for `sp_runtime::traits::Header` based headers, that
+/// implements `headers_relay::sync_types::SourceHeader` and may be used in headers sync directly.
+#[derive(Clone, Debug, PartialEq)]
+pub struct SyncHeader<Header>(Header);
+
+impl<Header> SyncHeader<Header> {
+	/// Extracts wrapped header from self.
+	pub fn extract(self) -> Header {
```
tomusdrw, line 29: I think the usual convention is `into_inner`, wdyt?

relays/substrate-client/src/sync_header.rs
svyatonik: I've added panic-related comment to the method definition. Swallowing error here may lead to hard-to-realize/find issues (hot loops, async hot loops), so panic vs Option/Result. Panic here is imo more appropriate (that's not the critical code).

tomusdrw: Hmm, but it's only going to panic in debug mode, and overflow in `--release`. Maybe better to panic explicitly then?

svyatonik: Right, I completely forgot that

tomusdrw (approved):

HCastano (approved):

PR #532, opened 2020-11-26
Message relay fixes for nightly stuck

svyatonik:
Because of [first issue](https://github.com/paritytech/parity-bridges-common/commit/336c0e2372af123dc0269ef9a10056f3f5f85984) relay may have tried to send several (same) transactions simultaneously => this may have lead to errors like `Error submitting proof Rialto::MessagesDelivery: Request(Request(Error { code: MethodError(1014), message: "Priority is too low: (36797852575 vs 36797852575)", data: Some(String("The transaction has too low priority to replace another transaction already in the pool.")) })). Retrying in 0.48135872`.

The attempt was supposed to be retried (as visible from the error message above). But actually, because of [second issue](https://github.com/paritytech/parity-bridges-common/commit/a1cd15b485fea5b9e9c50c9b6df5e40902b58048), target client was never marked as available for retrying.

Since relay has stall-timeout, loop has been restarted after this timeout has passed. But here comes the [third issue](https://github.com/paritytech/parity-bridges-common/commit/6a8c88be111aea46b795e97d5d880eac5f92620c). Basically - on source chain we have already pruned messages 1 and 2, because they were delivered to the target chain. But source nonces were read before target nonces. The code was assuming that the nonce on the target chain is zero in this case. So we were trying to read weights of messages 1 and 2 from the source node. Obviously, this request has failed and the source node was marked as failing (aka offline). But here comes the second issue - source client was never 'returned back'.

Finally, stall-timeout works only if race (delivery or confirmations) has submitted something, but the nonce has not changed at the race target within stall-timeout interval (i.e. transaction is lost). In this case, all messages were delivered and confirmed after restart and since source client was unavailable to check if there are new nonces at the source (because of second issue), relay was (and still is) thinking that there is nothing to do.

I'll add more alerts to the messages-relay dashboard in next PR - for testnets we're sure that there's at least 1 message / minute. So this may be detected.

### DIFF
--- a/relays/messages-relay/src/message_lane_loop.rs
+++ b/relays/messages-relay/src/message_lane_loop.rs
@@ -103,6 +103,9 @@ pub trait SourceClient<P: MessageLane>: Clone + Send + Sync {
 	) -> Result<(SourceHeaderIdOf<P>, MessageNonce), Self::Error>;
 
 	/// Returns mapping of message nonces, generated on this client, to their weights.
+	///
+	/// Some weights may be missing from returned map, if corresponding messages were pruned at
+	/// the source chain.
 	async fn generated_messages_weights(
 		&self,
 		id: SourceHeaderIdOf<P>,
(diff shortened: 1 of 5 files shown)


HCastano (approved):
I really appreciate the write-up with the various issues!

PR #533, opened 2020-11-26
Alert for case when Millau-to-Rialto messages are not detected by relay

svyatonik:
see #532 for details


HCastano (approved):
I'd wait until #532 is merged before merging this PR, or else where going to be spammed with alerts since the current testnet is broken

PR #535, opened 2020-11-27
Cross-Chain Transfer Generator

HCastano:
This PR adds the ability to send `Transfer` messages using our existing message generator. What this allows us to do is to send a message from the Millau chain indicating that you would like to perform a balance transfer to an account on the Rialto chain. When the message is dispatched on the Rialto chain the specified recipient will be transferred the correct amount of RLT tokens.

The Rialto to Millau side of this will be added after #550 is merged.


tomusdrw:
>  I suspect it has to do with the derived account not having any way to cover fees, but I'm not sure yet.

Yes, that's plausible. I think you will need to top-up the derive accounts first.

tomusdrw:
Related #219

scripts/send-message.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
HCastano, line 1: I used this a bit when testing locally. Might be worth keeping around, but if you guys don't find it useful I'm okay removing it too

bin/rialto/node/src/chain_spec.rs
```diff
@@ -191,3 +193,30 @@ fn load_kovan_bridge_config() -> BridgeKovanConfig {
 		initial_validators: rialto_runtime::kovan::genesis_validators(),
 	}
 }
+
+// We use this to get the account on Rialto (target) which is derived from Millau's (source) Dave
+// account. We do this so we can fund the derived account on Rialto at Genesis to it can pay
+// transaction fees. The reason we can use Rialto's Dave for the `millau_account_id` is because
+// they're both derived from the same seed phrase.
+//
+// In case you need it, the derived account ID is: 5Hg7WQyk8C1FmPzxY3xSjR7S6zZZC5sAL35vMr6NpW17jBhQ
```
tomusdrw, line 202: IMHO it makes more sense to add this as a test case, so that whenever it changes we will clearly see it ;)

bin/rialto/node/src/chain_spec.rs
```diff
@@ -191,3 +193,30 @@ fn load_kovan_bridge_config() -> BridgeKovanConfig {
...
+//
+// Note that this should only be used for testing.
+fn derived_dave_account_id() -> AccountId {
+	use sp_runtime::traits::Convert;
+	let millau_account_id = get_account_id_from_seed::<sr25519::Public>("Dave");
+	let encoded_id = bp_runtime::derive_account_id(*b"mlau", bp_runtime::SourceAccount::Account(millau_account_id));
+	bp_rialto::AccountIdConverter::convert(encoded_id)
```
tomusdrw, line 209: Can we put this to `bp_rialto` as a helper function? I imagine it's going to be quick useful for any tools that build on top of rialto-runtime, and you had to repeat it twice already in this file.

deployments/bridges/rialto-millau/entrypoints/relay-messages-generator-entrypoint.sh
```diff
@@ -8,24 +8,80 @@
...
 	SUBMIT_DELAY_S=`shuf -i 0-$MAX_SUBMIT_DELAY_S -n 1`
 	echo "Sleeping $SUBMIT_DELAY_S seconds..."
 	sleep $SUBMIT_DELAY_S
+}
+
+while true
+do
+	rand_sleep
+	echo "Sending Remark from Millau to Rialto using Target Origin"
+	/home/user/substrate-relay submit-millau-to-rialto-message \
```
tomusdrw, line 30: Can you move this to a variable as well?
```
SUBMIT_MESSAGE="/home/user/substrate-relay submit-millau-to-rialto-message $SHARED_ARGS"
```

relays/substrate/src/cli.rs
```diff
@@ -144,19 +147,33 @@ pub enum Command {
 	},
 }
 
+/// All possible messages that may be delivered to the Rialto chain.
+#[derive(StructOpt, Debug)]
+pub enum ToRialtoMessage {
+	Remark,
```
tomusdrw, line 153: Missing docs for variants

scripts/send-message.sh
tomusdrw: Looks useful, but maybe add some docs what setup is expected for this to work (I presume it simply running the test network locally, right?)

HCastano:
Will fix conflicts tomorrow 😒

HCastano:
I'm going to add the Rialto to Millau side in a different PR just to make sure the diffs are managable

svyatonik (approved):

tomusdrw (approved):

PR #536, opened 2020-11-30
lower limit for message weight

### LINKED ISSUE #534: Verify minimal message weight before accepting it
[Here](https://github.com/paritytech/parity-bridges-common/blob/18a4e98d05b94974ea3da49c03994f4b155d3cc3/bin/millau/runtime/src/rialto_messages.rs#L163) runtime checks that the message being sent does not overflow maximal weight limit for the lane. But right now sender is allowed to set declared message weight to zero (or something that is lower than the minimal required payment for e.g. system::remark call on the target chain) and the message will still be relayed.

Should be fixed both for test and production chains. The runtime should reject messages that have declared weight less than `ExtrinsicBaseWeight` or the `system::remark()` call on the target chain.

svyatonik:
closes #534

### DIFF
--- a/bin/millau/runtime/src/rialto_messages.rs
+++ b/bin/millau/runtime/src/rialto_messages.rs
@@ -88,9 +88,7 @@ impl MessageBridge for WithRialtoMessageBridge {
 	type ThisChain = Millau;
 	type BridgedChain = Rialto;
 
-	fn weight_limits_of_message_on_bridged_chain(
-		message_payload: &[u8],
-	) -> RangeInclusive<Weight> {
+	fn weight_limits_of_message_on_bridged_chain(message_payload: &[u8]) -> RangeInclusive<Weight> {
 		// we don't want to relay too large messages + keep reserve for future upgrades
 		let upper_limit = bp_rialto::MAXIMUM_EXTRINSIC_WEIGHT / 2;
 
--- a/bin/millau/runtime/src/rialto_messages.rs
+++ b/bin/millau/runtime/src/rialto_messages.rs
@@ -93,9 +93,8 @@ impl MessageBridge for WithRialtoMessageBridge {
 		let upper_limit = bp_rialto::MAXIMUM_EXTRINSIC_WEIGHT / 2;
 
 		// given Rialto chain parameters (`TransactionByteFee`, `WeightToFee`, `FeeMultiplierUpdate`),
-		// the minimal weight of the message may be computed as message.length() + minimal extrinsic weight
-		let message_payload_length_weight = Weight::try_from(message_payload.len()).unwrap_or(Weight::MAX);
-		let lower_limit = message_payload_length_weight.saturating_add(bp_rialto::BASE_EXTRINSIC_WEIGHT);
+		// the minimal weight of the message may be computed as message.length()
+		let lower_limit = Weight::try_from(message_payload.len()).unwrap_or(Weight::MAX);
 
 		lower_limit..=upper_limit
 	}
--- a/bin/millau/runtime/src/rialto_messages.rs
+++ b/bin/millau/runtime/src/rialto_messages.rs
@@ -30,6 +30,7 @@ use frame_support::{
 	RuntimeDebug,
 };
 use sp_core::storage::StorageKey;
+use sp_std::{convert::TryFrom, ops::RangeInclusive};
 
 /// Storage key of the Millau -> Rialto message in the runtime storage.
 pub fn message_key(lane: &LaneId, nonce: MessageNonce) -> StorageKey {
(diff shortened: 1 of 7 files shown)


bin/millau/runtime/src/rialto_messages.rs
```diff
@@ -87,9 +88,16 @@ impl MessageBridge for WithRialtoMessageBridge {
...
 
-	fn maximal_dispatch_weight_of_message_on_bridged_chain() -> Weight {
+	fn weight_limits_of_message_on_bridged_chain(message_payload: &[u8]) -> RangeInclusive<Weight> {
 		// we don't want to relay too large messages + keep reserve for future upgrades
-		bp_rialto::MAXIMUM_EXTRINSIC_WEIGHT / 2
+		let upper_limit = bp_rialto::MAXIMUM_EXTRINSIC_WEIGHT / 2;
+
+		// given Rialto chain parameters (`TransactionByteFee`, `WeightToFee`, `FeeMultiplierUpdate`),
+		// the minimal weight of the message may be computed as message.length() + minimal extrinsic weight
+		let message_payload_length_weight = Weight::try_from(message_payload.len()).unwrap_or(Weight::MAX);
```
tomusdrw, line 97: `message_payload` is just `Call`, right? So it doesn't include the transaction overhead (signed extensions, signatures, etc), I think we should add some extra constant to cater for this.
Instead of using `BASE_EXTRINSIC_WEIGHT` I would use something like: the weight of transaction with `System::remark(vec![])` call and have it measured externally and hardcoded.

svyatonik: Right, that's the encoded `Call`. Signed extensions and signatures checks are not a part of the call dispatch, right? So `Call::dispatch()` will not verify any signatures or signed extensions.

Verification will be done by the delivery transaction itself, which will hand the encoded call to the call-dispatch module, which in turn would verify signatures (if required). So signatures verification weight is included into delivery transaction base weight (which is accounted in the dispatch fee through the `weight_of_delivery_transaction()` call).

So here we deal only with pure call weight. Including overhead here doesn't make sense, because it is already accounted.

tomusdrw: Thanks for explanations. I thought this should be the minimal weight of the entire transaction.
We've also discussed off-band that the `BASE_EXTRINSIC_WEIGHT` doesn't need to be added here, since it's rather part of `weight_of_delivery_transaction()`, since it's payed per-extrinsic and not per-dispatch, so each message only requires a minimal weight to cover the length-weight cost.

tomusdrw (approved):

bin/rialto/runtime/src/millau_messages.rs
```diff
@@ -87,9 +88,15 @@ impl MessageBridge for WithMillauMessageBridge {
 	type ThisChain = Rialto;
 	type BridgedChain = Millau;
 
-	fn maximal_dispatch_weight_of_message_on_bridged_chain() -> Weight {
+	fn weight_limits_of_message_on_bridged_chain(message_payload: &[u8]) -> RangeInclusive<Weight> {
 		// we don't want to relay too large messages + keep reserve for future upgrades
-		bp_millau::MAXIMUM_EXTRINSIC_WEIGHT / 2
+		let upper_limit = bp_millau::MAXIMUM_EXTRINSIC_WEIGHT / 2;
+
+		// given Rialto chain parameters (`TransactionByteFee`, `WeightToFee`, `FeeMultiplierUpdate`),
```
HCastano, line 95:
```suggestion
		// given Millau chain parameters (`TransactionByteFee`, `WeightToFee`, `FeeMultiplierUpdate`),
```

bin/millau/runtime/src/rialto_messages.rs
```diff
@@ -87,9 +88,15 @@ impl MessageBridge for WithRialtoMessageBridge {
...
 
-	fn maximal_dispatch_weight_of_message_on_bridged_chain() -> Weight {
+	fn weight_limits_of_message_on_bridged_chain(message_payload: &[u8]) -> RangeInclusive<Weight> {
 		// we don't want to relay too large messages + keep reserve for future upgrades
-		bp_rialto::MAXIMUM_EXTRINSIC_WEIGHT / 2
+		let upper_limit = bp_rialto::MAXIMUM_EXTRINSIC_WEIGHT / 2;
+
+		// given Rialto chain parameters (`TransactionByteFee`, `WeightToFee`, `FeeMultiplierUpdate`),
+		// the minimal weight of the message may be computed as message.length()
+		let lower_limit = Weight::try_from(message_payload.len()).unwrap_or(Weight::MAX);
```
HCastano, line 97: So by having the "default" be `Weight::MAX` we basically invalidate any range returned afterwards since nothing can be contained in the Range `Weight::MAX..=upper_limit`, right?

HCastano (approved):

bin/millau/runtime/src/rialto_messages.rs
svyatonik: Right

Issue #539, opened 2020-11-30
Unbounded loops in message-lane


svyatonik:
Before starting working on weights in message-lane module, we need to have upper bound on all loops in message-lane module. Right now there are four loops in message-lane calls:
- [x] unbounded [messages dispatch loop](https://github.com/paritytech/parity-bridges-common/blob/18a4e98d05b94974ea3da49c03994f4b155d3cc3/modules/message-lane/src/lib.rs#L346) in `fn receive_messages_proof()`. It is unbounded, because we only limit messages by dispatch weight (and dispatch weight may be zero). This makes impossible to compute base weight of delivery transaction. The base weight should not be larger than `TargetChain.MaximalExtrinsicWeight` - [`MaximalMessageWeight`](https://github.com/paritytech/parity-bridges-common/blob/18a4e98d05b94974ea3da49c03994f4b155d3cc3/bin/runtime-common/src/messages.rs#L50), so that we always can deliver single large message. At the same time, we want more than one message in the batch, so it shouldn't be minimal;
- [x] unbounded [relayers reward loop](https://github.com/paritytech/parity-bridges-common/blob/18a4e98d05b94974ea3da49c03994f4b155d3cc3/modules/message-lane/src/lib.rs#L402). It was assumed to be bounded by the [`MaxUnconfirmedMessagesAtInboundLane`](https://github.com/paritytech/parity-bridges-common/blob/18a4e98d05b94974ea3da49c03994f4b155d3cc3/modules/message-lane/src/lib.rs#L84) parameter at the bridged chain, but after #418 it is no longer true. There are several ways to fix this:
 - treat `MaxUnconfirmedMessagesAtInboundLane` as it was designed initially. The optimization may stay, but only as storage optimization (i.e. instead of having N entries for N messages relayed by single relayer, we may have 1 entry for N messages);
 - add another 'strict' upper bound;
 - add some structure that will optimize payouts computation (so we won't need to read N messages to pay reward for these messages);
 - introduce non-instant rewards - rewards may be paid e.g. when message is being pruned (in block initialization or when new message is being sent);

- [x] bounded [prune messages loop](https://github.com/paritytech/parity-bridges-common/blob/18a4e98d05b94974ea3da49c03994f4b155d3cc3/modules/message-lane/src/outbound_lane.rs#L90);
[...]

PR #541, opened 2020-12-01
Limit max number of messages in delivery transaction

svyatonik:
Fixes "unbounded messages dispatch loop" part of #539

### DIFF
--- a/bin/runtime-common/src/messages.rs
+++ b/bin/runtime-common/src/messages.rs
@@ -864,6 +864,7 @@ mod tests {
 
 	#[test]
 	fn message_proof_is_rejected_if_outbound_lane_state_decode_fails() {
+		#[allow(clippy::reversed_empty_ranges)]
 		assert_eq!(
 			target::verify_messages_proof_with_parser::<OnThisChainBridge, _, _>(
 				(Default::default(), StorageProof::new(vec![]), Default::default(), 1, 0),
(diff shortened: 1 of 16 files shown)


modules/message-lane/src/lib.rs
```diff
@@ -82,6 +86,11 @@ pub trait Trait<I = DefaultInstance>: frame_system::Trait {
 	/// transaction#2 with individual messages [3; 4], this would be treated as single "Message" and
 	/// would occupy single unit of `MaxUnconfirmedMessagesAtInboundLane` limit.
 	type MaxUnconfirmedMessagesAtInboundLane: Get<MessageNonce>;
+	/// Maximal number of messages in single delivery transaction. This directly affects the base
+	/// weight of thedelivery transaction.
```
tomusdrw, line 90:
```suggestion
	/// weight of the delivery transaction.
```

bin/runtime-common/src/messages.rs
```diff
@@ -344,71 +346,155 @@ pub mod target {
 	/// Verify proof of Bridged -> This chain messages.
 	pub fn verify_messages_proof<B: MessageBridge, ThisRuntime>(
 		proof: FromBridgedChainMessagesProof<B>,
+		max_messages: MessageNonce,
 	) -> Result<ProvedMessages<Message<BalanceOf<BridgedChain<B>>>>, &'static str>
```
tomusdrw, line 350: Since we have a proper error enum now, why not return it here and implement `Display` instead of `Into<&'static str>`?

tomusdrw (approved):
lgtm!

bin/runtime-common/src/messages.rs
svyatonik: Implementing `Display` would bring things `from std::fmt` to runtime (like `Formatter`, `Result`, `write!`), right? I thought it isn't a good idea, even if possible.

tomusdrw: Oh yeah, good point, we don't want that.

HCastano (approved):

PR #545, opened 2020-12-01
Limit max number of unconfirmed messages at inbound lane

### LINKED ISSUE #539: Unbounded loops in message-lane

svyatonik:
closes #539

Instead of single `MaxUnconfirmedMessagesAtInboundLane` constant now there are two: `MaxUnrewardedRelayerEntriesAtInboundLane` and `MaxUnconfirmedMessagesAtInboundLane`.

Messages are rejected if there are `MaxUnrewardedRelayerEntriesAtInboundLane` unrewarded relayer entries at inbound lane. E.g. if it is 3 and `inbound_lane_state.relayers = vec![(1..=10, relayer1), (11..=20, relayer2), (21..=30, relayer1)]`, then no new messages may be accepted.

Messages are rejected if there are `MaxUnconfirmedMessagesAtInboundLane` unconfirmed messages at inbound lane. E.g. if it is 30 and `inbound_lane_state.relayers = vec![(1..=30, relayer1)`, or `inbound_lane_state.relayers = vec![(1..=10, relayer1), (11..=20, relayer2), (21..=30, relayer1)]`  then no new messages may be accepted.

So the worst case (which is relevant for computing weights) is: paying rewards for Max(`MaxUnconfirmedMessagesAtInboundLane`, `MaxUnrewardedRelayerEntriesAtInboundLane`) messages, where every message is relayed by a separate relayer.

### DIFF
--- a/modules/message-lane/src/inbound_lane.rs
+++ b/modules/message-lane/src/inbound_lane.rs
@@ -31,6 +31,8 @@ pub trait InboundLaneStorage {
 
 	/// Lane id.
 	fn id(&self) -> LaneId;
+	/// Return maximal number of unrewarded relayer entries in inbound lane.
+	fn max_unrewarded_relayer_entries(&self) -> MessageNonce;
 	/// Return maximal number of unconfirmed messages in inbound lane.
 	fn max_unconfirmed_messages(&self) -> MessageNonce;
 	/// Get lane data from the storage.
@@ -97,8 +99,14 @@ impl<S: InboundLaneStorage> InboundLane<S> {
 			return false;
 		}
 
+		// if there are more unrewarded relayer entries than we may accept, reject this message
+		if self.storage.max_unrewarded_relayer_entries() <= data.relayers.len() as MessageNonce {
+			return false;
+		}
+
 		// if there are more unconfirmed messages than we may accept, reject this message
-		if self.storage.max_unconfirmed_messages() <= data.relayers.len() as MessageNonce {
+		let unconfirmed_messages_count = nonce.saturating_sub(data.latest_confirmed_nonce);
+		if unconfirmed_messages_count > self.storage.max_unconfirmed_messages() {
 			return false;
 		}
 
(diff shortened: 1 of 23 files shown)


modules/message-lane/src/inbound_lane.rs
```diff
@@ -97,8 +99,14 @@ impl<S: InboundLaneStorage> InboundLane<S> {
 			return false;
 		}
 
+		// if there are more unrewarded relayer entries than we may accept, reject this message
+		if self.storage.max_unrewarded_relayer_entries() <= data.relayers.len() as MessageNonce {
```
tomusdrw, line 103:
```suggestion
		if data.relayers.len() as MessageNonce >= self.storage.max_unrewarded_relayer_entries() {
```
easier to parse if it matches the docs. Why do we disallow equality here?

svyatonik: Because if there are already max entries, we can't add any more entries if relayer is different. And if relayer is the same, then we're rejecting message because of your suggestion [here](https://github.com/paritytech/parity-bridges-common/pull/418#pullrequestreview-511667612).

HCastano (approved):

tomusdrw (approved):

PR #554, opened 2020-12-02
Message lane benchmarks: start

svyatonik:
This should be a PR with the only benchmark - `send_message_worst_case`, all other benchmarks will be added in follow-up PRs. Right now #551 is required to continue this work => draft.

### DIFF
--- a/modules/message-lane/src/benchmarking.rs
+++ b/modules/message-lane/src/benchmarking.rs
@@ -1,0 +1,87 @@
+// Copyright 2019-2020 Parity Technologies (UK) Ltd.
+// This file is part of Parity Bridges Common.
+
+// Parity Bridges Common is free software: you can redistribute it and/or modify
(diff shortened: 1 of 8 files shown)


modules/message-lane/src/benchmarking.rs
```diff
@@ -0,0 +1,92 @@
...
+	// * maximal number of messages is being pruned during the call;
+	// * message size is maximal for the target chain.
+	send_message_worst_case {
+		let i in 1..100;
+
+		let lane_id = bench_lane_id();
+		let sender = account("sender", i, SEED);
+		T::endow_account(&sender);
+
+		// 'send' messages that are to be pruned when our message is sent
```
HCastano, line 67:
```suggestion
		// Send messages which will be pruned when our benchmarking message is sent
```

HCastano (approved):

PR #558, opened 2020-12-03
Verify Source origin in TargetChainVerifier

### LINKED ISSUE #556: Check Dispatch Origin on Source Chain
As I was working on #535 I realized that at the moment you can send a message which is dispatched from the derived `Root` account on the target chain from any account on the source chain. Obviously we only want the source chain's `Root` account to be able to send these messages, so we need to do some checks on the source chain before sending the message.

The main thing to do here is find the right place to call [verify_message_origin](https://github.com/paritytech/parity-bridges-common/blob/master/modules/call-dispatch/src/lib.rs#L269) in the send message pipeline.

tomusdrw:
Closes #556


modules/message-lane/src/instant_payments.rs
```diff
@@ -37,11 +37,20 @@ where
...
 		relayer_fund_account: &AccountId,
 	) -> Result<(), Self::Error> {
-		Currency::transfer(submitter, relayer_fund_account, *fee, ExistenceRequirement::AllowDeath).map_err(Into::into)
+		match submitter {
+			Sender::Signed(submitter) => {
+				Currency::transfer(submitter, relayer_fund_account, *fee, ExistenceRequirement::AllowDeath)
+					.map_err(Into::into)
+			}
+			Sender::Root | Sender::None => {
+				// fixme: we might want to add root account id to this struct.
```
HCastano, line 50: Sure. Would you just want the Root Id to log it?

HCastano (approved):

modules/message-lane/src/instant_payments.rs
tomusdrw: I was thinking that we can provide an account that should be charged fee when root is sending messages. The treasury for instance.

bin/millau/runtime/src/rialto_messages.rs
```diff
@@ -166,12 +168,19 @@ impl TargetHeaderChain<ToRialtoMessagePayload, bp_rialto::AccountId> for Rialto
 	// - id of the lane we prove state of.
 	type MessagesDeliveryProof = ToRialtoMessagesDeliveryProof;
 
-	fn verify_message(payload: &ToRialtoMessagePayload) -> Result<(), Self::Error> {
+	fn verify_message(
```
svyatonik, line 171: The main idea of splitting `TargetHeaderChain::verify_message` and `LaneMessageVerifier::verify_message` is that you may use single implementation of `TargetHeaderChain` for all lanes (and message-lane pallet instances) that are sending messages to that target chain. So it was supposed to only do things like [that](https://github.com/paritytech/parity-bridges-common/blob/master/primitives/message-lane/src/source_chain.rs#L41). When talking about substrate-based chains, it should check message weight and message size. Everything else is pallet/lane specific and was supposed to be done in `LaneMessageVerifier::verify_message`.

tomusdrw: Yes, but the dispatcher is not lane specific - it's a property of a target chain. Currently it's even questionable if we need both of them - the split seems unnecessary, cause we do both checks [one after the other](https://github.com/paritytech/parity-bridges-common/blob/3be2b72714641389e2e7638b409e1f26c0f80444/modules/message-lane/src/lib.rs#L254-L266).

Okay, now that I've analyzed this a bit more I think I understand what your point is. We can re-use the same `TargetHeaderChain` for multiple `MessageDelivery` implementations and each message delivery mechanism can be plugged into a different dispatch mechanism.

I'll move the check then.

svyatonik: > Yes, but the dispatcher is not lane specific - it's a property of a target chain. Currently it's even questionable if we need both of them - the split seems unnecessary, cause we do both checks one after the other.

Dispatcher may be lane-specific, or even message-specific, it's just our implementation in `bp-runtime-common` assumes that all messages are encoded calls of target chain.

tomusdrw:
@svyatonik moved the verification to `LaneMessageVerifier`, could you take another look?

bin/runtime-common/src/messages.rs
```diff
@@ -149,25 +149,34 @@ pub mod source {
...
 		type Error = &'static str;
 
 		fn verify_message(
-			_submitter: &AccountIdOf<ThisChain<B>>,
+			submitter: &Sender<AccountIdOf<ThisChain<B>>>,
 			delivery_and_dispatch_fee: &BalanceOf<ThisChain<B>>,
 			_lane: &LaneId,
 			payload: &FromThisChainMessagePayload<B>,
 		) -> Result<(), Self::Error> {
+			// Do the dispatch-specific check. We know that Millau uses `CallDispatch`,
```
svyatonik, line 169:
```suggestion
			// Do the dispatch-specific check. We know that the Bridged chain uses `CallDispatch`,
```

svyatonik (approved):

PR #566, opened 2020-12-08
Update Substrate Dependency

HCastano:
This does not update to the latest Substrate just yet, but it is a newest version than what we've been using. This is because there are issues updating to `libp2p` `v0.32.2` (see my comment below) which stem from https://github.com/libp2p/rust-libp2p/issues/1873.

For reviewers, there are two main changes here. First is the `Trait` to `Config` rename, and the second is the changes to weight types (see comment by @adoerr). The second changes are mostly introduced in da2aee66cc1370159ef9387041a8aafa8f493db2, so taking a look at just that commit will reduce a bit of noise introduced by the first set of changes.


adoerr:
For the build issues you are still seeing, you might want to have a look at [Streamline frame_system weight parametrization](https://github.com/paritytech/substrate/commit/332399d16668a6c769f1a7db154bb9ea3b50e61c#diff-156aed3fc4146d01885e6778bcd310d3af15ca03b420b76e955ec1502282540a) especially the changes to `bin/node-template/runtime/src/lib.rs`

HCastano:
> For the build issues you are still seeing, you might want to have a look at [Streamline frame_system weight parametrization](https://github.com/paritytech/substrate/commit/332399d16668a6c769f1a7db154bb9ea3b50e61c#diff-156aed3fc4146d01885e6778bcd310d3af15ca03b420b76e955ec1502282540a) especially the changes to `bin/node-template/runtime/src/lib.rs`

The CI actually gets further than I do - so I haven't gotten around to actually updating our code to match the changes in Substrate yet 😅

HCastano:
I think we need to wait until https://github.com/paritytech/substrate/pull/7696 is merged before this compiles.

mxinden:
@HCastano https://github.com/paritytech/substrate/pull/7696 is merged now. If you can, please pin to libp2p `v0.32.1` as that pins to async-tls `v0.10.2`.

HCastano:
Since this compiles I think it's ready to be reviewed. However, it's worth noting that to get this to the latest version of Substrate (or at least to one which uses the new `libp2p` release) requires updates to some upstream crates. In particular we need updates for crates which use `v0.17` of the `secp256k1` crate since it requires that the `cc` crate is `<= v1.0.41` - making it incompatible with version used in the dependency tree of the new `libp2p` release.

Some crates which require updates include:
- `ethereum-tx-sign`
- `parity-crypto`

I think we should try and bump Substrate again in a week or two once these dependencies are updated.

primitives/millau/src/lib.rs
```diff
@@ -155,6 +160,44 @@ impl sp_runtime::traits::Convert<sp_core::H256, AccountId> for AccountIdConverte
 		hash.to_fixed_bytes().into()
 	}
 }
+
+/// Get a struct which defines the weight limits and values used during extrinsic execution.
+pub fn runtime_block_weights() -> frame_system::limits::BlockWeights {
```
HCastano, line 165: I moved the builder call (and the length builder below) here because I didn't want other parts of the codebase to be pulling in the runtime code as a dependency, but this isn't super nice. Two questions here:

1. Is it okay for me to be pulling stuff in from `frame_system` here?
2. Is there any way for me to get a shared instance of this that I can use in the runtime but also in the relay and message lane crates?

bin/millau/runtime/src/lib.rs
```diff
@@ -217,12 +198,18 @@ impl frame_system::Trait for Runtime {
 	// TODO: update me (https://github.com/paritytech/parity-bridges-common/issues/78)
 	/// Weight information for the extrinsics of this pallet.
 	type SystemWeightInfo = ();
+	/// boop
```
svyatonik, line 201: boop? :)

bin/rialto/runtime/src/lib.rs
```diff
@@ -225,14 +206,20 @@ impl frame_system::Trait for Runtime {
 	// TODO: update me (https://github.com/paritytech/parity-bridges-common/issues/78)
 	/// Weight information for the extrinsics of this pallet.
 	type SystemWeightInfo = ();
+	/// boop
```
svyatonik, line 209: the same

primitives/millau/src/lib.rs
```diff
@@ -68,23 +71,25 @@ impl sp_runtime::traits::Hash for BlakeTwoAndKeccak256 {
...
-/// transactions minus 10% for initialization).
-pub const MAXIMUM_EXTRINSIC_WEIGHT: Weight = MAXIMUM_BLOCK_WEIGHT / 100 * (AVAILABLE_BLOCK_RATIO as Weight - 10);
-/// Maximal size of Millau block.
-pub const MAXIMUM_BLOCK_SIZE: u32 = 2 * 1024 * 1024;
-/// Maximal size of single normal Millau extrinsic (75% of maximal block size).
-pub const MAXIMUM_EXTRINSIC_SIZE: u32 = MAXIMUM_BLOCK_SIZE / 100 * AVAILABLE_BLOCK_RATIO;
+/// Maximum weight of single Millau block.
+///
+/// This represents two seconds of compute assuming a target block time of six seconds.
+pub const MAXIMUM_BLOCK_WEIGHT: Weight = 2 * WEIGHT_PER_SECOND;
```
svyatonik, line 77: The value is now the same as in Rialto. I'd prefer to keep different weight/size limits. The idea was to:
1) change `Millau` to chain that operates differently from `Rialto` (i.e. it has different crypto/hash algorithms, block times, consensus algorithms). This is WIP, but every time I introduce/change some constants, they're a bit different from `Millau`. The direction is to make `Millau` lighter+faster than `Rialto`;
2) that isn't something I work on (and probably there are better ways to do that), but these constant in different chains may be used to compare messages bandwidth under different limits.

So unless you have any arguments against keeping previous constants (or at least having different values in different chains), let's change that.

svyatonik (approved):

bin/millau/runtime/src/lib.rs
HCastano: Ah this is embarrassing 😅 I meant to write actual docs

primitives/millau/src/lib.rs
HCastano: @tomusdrw Hey can you take a look at this? I feel like you have opinions about this 😛

tomusdrw: This looks good.
1. Yes, it actually has less dependencies than `frame-support` anyway.
2. Long term relay should rely on the metadata instead of the primitives directly.

tomusdrw (approved):

PR #567, opened 2020-12-09
Benchmarks for message delivery transaction

svyatonik:
Delivery transaction weight consist of:
A) message(s) dispatch weight;
B) extra weight required for every message (aka message overhead);
C) extra weight required for updating confirmed message (pruning relayers entries) (aka outbound lane state overhead);
D) extra weight of transaction itself (aka transaction overhead).

If I understand correctly, the idea of new automated (and parametrized) weights is that we run benchmarks (on demand or at runtime release) and then we may directly use results (weights) of these benchmarks to compute pre-dispatch weight of the call. This has its advantages (we may have separate weights for Kusama/Polkadot/Rialto/Millau), but at the same time it isn't super-flexible (I haven't found non-hacky way to estimate every of A..D points from above).

The idea of this PR is to have:
1) `receive_single_message_proof` - benchmark that delivers single (near-zero weight) message that is dispatched. So it has weight A+B+D, but given that the dispatch weight is near-zero, we may neglect it, so it is B+D;
2) `receive_two_messages_proof` - benchmark that delivers two (near-zero weight) message that is dispatched. So it has weight (omitting A) 2*B+D. Given benchmarks (1) and (2) we may roughly compute `B = weight(benchmark#2) - weight(benchmark#1)`;
3) `receive_single_message_proof_with_outbound_lane_state` - benchmark that delivers single (near-zero weight) message that is dispatched AND outbound lane state. So it has weight (omitting A) B+C+D. Given benchmarks (1) and (3) we may roughly compute `C = weight(benchmark#3) - weight(benchmark#1)`;
4) given value of `B`, we may compute `D = weight(benchmark#1) - B = weight(benchmark#1) - (weight(benchmark#2) - weight(benchmark#1)) = 2*weight(benchmark#1) - weight(benchmark#2)`.

I'm not including [weight formula](https://github.com/paritytech/parity-bridges-common/blob/597c7a82ed14aa320029cbd0bfe00fe0401db003/modules/message-lane/src/lib.rs#L325) change in this PR, because we need to migrate to `WeightInfo` trait first and I'm currently want to prepare benchmarks for that migration. But the formula may probably be changed to something like:
```rust
#[weight =
	// (D) transaction overhead
	(2*T::WeightInfo::receive_single_message_proof() - T::WeightInfo::receive_two_messages_proof())
		// (B) message overhead * max number of messages
		.saturating_add(
			T::MaxMessagesInDeliveryTransaction::get()
				.saturating_mul(T::WeightInfo::receive_two_messages_proof() - T::WeightInfo::receive_single_message_proof())
		)
		// (C) outbound lane state overhead
		.saturating_add(
			T::WeightInfo::receive_single_message_proof_with_outbound_lane_state() - T::WeightInfo::receive_single_message_proof()
		)
		// (A) dispatch weight
		.saturating_add(
			*dispatch_weight
		)
]
```

There are three caveats with this formula:
1) I'm not sure if auto-weights are accurate enough for low weights. Can there be situation when `receive_single_message_proof() > receive_two_messages_proof()`? I haven't seen that locally, but I assume it may happen on faster machines. So probably we'll need to add some tests to ensure that weights are sorted accordingly;
2) you can see from the suggested formula that the relayer is always charged for `T::MaxMessagesInDeliveryTransaction` messages overhead. Probably it would be better to change that before release - the suggested fix is to add another parameter to the delivery tx - number of bundled messages (similar to `dispatch_weight`) and charge relayer for this amount only;
3) you can see from the suggested formula that the relayer is always charged for outbound lane state overhead. While this also seems unnecessary if we'll add another param (`includes_outbound_lane_state`), I still think that it is better to always charge relayer for that. Reasones:
3.1) these costs are designed to be paid by the submitter anyway. So relayer actually loses nothing if everything is working as designed;
3.2) if tx cost always includes outbound lane state, then relayer has no benefit from not including that info. So rewards confirmations will be synced faster => we'll have less situations when inbound lane rejects messages because its `relayers` vector is too large.


bin/rialto/runtime/src/lib.rs
```diff
@@ -853,6 +858,56 @@ impl_runtime_apis! {
...
+						),
+						call_weight,
+						pallet_bridge_call_dispatch::MessagePayload {
+							spec_version: VERSION.spec_version,
+							weight: call_weight,
+							origin: pallet_bridge_call_dispatch::CallOrigin::<
+								bp_millau::AccountId,
+								MultiSigner,
+								Signature,
+							>::SourceRoot, // TODO: should be signature instead
```
svyatonik, line 906: The idea is to have `CallOrigin::TargetAccount` here. But I have found no way to generate signatures from `no_std` runtime using Substrate crates.

modules/message-lane/src/benchmarking.rs
```diff
@@ -70,23 +95,202 @@ benchmarks_instance! {
 		}
 		confirm_message_delivery::<T, I>(T::MaxMessagesToPruneAtOnce::get());
 
-		let (payload, fee) = T::prepare_message(MessageParams { size_factor: WORST_MESSAGE_SIZE_FACTOR });
+		let (payload, fee) = T::prepare_outbound_message(MessageParams { size_factor: WORST_MESSAGE_SIZE_FACTOR });
 	}: send_message(RawOrigin::Signed(sender), lane_id, payload, fee)
+	verify {
+		assert_eq!(
```
svyatonik, line 101: I may add `test_benchmarks` to this module, but we are not testing benchmarks on CI anyway. So imo it doesn't makes sense. These verifications are also running during benchmark execution.

bin/rialto/runtime/src/lib.rs
svyatonik: Will be solved as a part of #569

modules/message-lane/src/benchmarking.rs
```diff
@@ -70,23 +95,202 @@ benchmarks_instance! {
...
+	// * message is successfully dispatched;
+	// * message requires all heavy checks done by dispatcher.
+	//
+	// This benchmarks gives us an approximation of single message delivery weight. It is similar to the
+	// `weight(receive_two_messages_proof) - weight(receive_single_message_proof)`. So it may be used
+	// to verify that the other approximation is correct.
+	receive_multiple_messages_proof {
+		let i in 1..T::MaxMessagesInDeliveryTransaction::get()
+			.try_into()
+			.expect("too large MaxMessagesInDeliveryTransaction");
```
HCastano, line 212:
```suggestion
			.expect("Value of MaxMessagesInDeliveryTransaction is too large");
```

modules/message-lane/src/benchmarking.rs
```diff
@@ -70,23 +95,202 @@ benchmarks_instance! {
...
+	// * message is successfully dispatched;
+	// * message requires all heavy checks done by dispatcher.
+	//
+	// This benchmarks gives us an approximation of outbound lane state delivery weight. It is similar to the
+	// `weight(receive_single_message_proof_with_outbound_lane_state) - weight(receive_single_message_proof)`.
+	// So it may be used to verify that the other approximation is correct.
+	receive_multiple_messages_proof_with_outbound_lane_state {
+		let i in 1..T::MaxMessagesInDeliveryTransaction::get()
+			.try_into()
+			.expect("too large MaxMessagesInDeliveryTransaction");
```
HCastano, line 242:
```suggestion
			.expect("Value of MaxMessagesInDeliveryTransaction is too large");
```

modules/message-lane/src/benchmarking.rs
```diff
@@ -70,23 +95,202 @@ benchmarks_instance! {
...
+
+	// Benchmark `receive_messages_proof` extrinsic with multiple minimal-weight messages and following conditions:
+	// * proof includes outbound lane state proof;
+	// * inbound lane already has state, so it needs to be read and decoded;
+	// * message is successfully dispatched;
+	// * message requires all heavy checks done by dispatcher.
+	//
+	// This benchmarks gives us an approximation of outbound lane state delivery weight. It is similar to the
+	// `weight(receive_single_message_proof_with_outbound_lane_state) - weight(receive_single_message_proof)`.
+	// So it may be used to verify that the other approximation is correct.
```
HCastano, line 238: Have you tried running the benchmarks and compared them yet?

modules/message-lane/src/lib.rs
```diff
@@ -558,11 +558,16 @@ fn ensure_operational<T: Trait<I>, I: Instance>() -> Result<(), Error<T, I>> {
 
 /// Creates new inbound lane object, backed by runtime storage.
 fn inbound_lane<T: Trait<I>, I: Instance>(lane_id: LaneId) -> InboundLane<RuntimeInboundLaneStorage<T, I>> {
-	InboundLane::new(RuntimeInboundLaneStorage {
+	InboundLane::new(inbound_lane_storage::<T, I>(lane_id))
+}
+
+/// Creates new runtime inbound lane storage.
+fn inbound_lane_storage<T: Trait<I>, I: Instance>(lane_id: LaneId) -> RuntimeInboundLaneStorage<T, I> {
```
HCastano, line 565: Do we benefit much from splitting the lane initialization from the storage initialization? I thought it was fine before

HCastano (approved):
> I'm not sure if auto-weights are accurate enough for low weights. Can there be situation when receive_single_message_proof() > receive_two_messages_proof()? I haven't seen that locally, but I assume it may happen on faster machines. So probably we'll need to add some tests to ensure that weights are sorted accordingly;

As far as _weight calculations_ are concerned, I don't think this can happen since the benchmarks would be computed on the same machine. If we're talking about the network at large, then yeah - but that can happen with any extrinsic being run on a faster machine. What's your concern here though?

I like the PR docs btw. I think it would be nice to see a succinct version of them in the benchmarking code to give a bit of context to any future readers of the file

bin/runtime-common/src/messages_benchmarking.rs
```diff
@@ -0,0 +1,144 @@
...
+use pallet_message_lane::benchmarking::MessageProofParams;
+use sp_core::Hasher;
+use sp_runtime::traits::Header;
+use sp_std::prelude::*;
+use sp_trie::{read_trie_value_with, trie_types::TrieDBMut, Layout, MemoryDB, Recorder, StorageProof, TrieMut};
+
+/// Generate ed25519 signature to be used in `pallet_brdige_call_dispatch::CallOrigin::TargetAccount`.
+///
+/// Returns public key of the signer and the signature itself.
+pub fn ed25519_sign(target_call: &impl Encode, source_account_id: &impl Encode) -> ([u8; 32], [u8; 64]) {
```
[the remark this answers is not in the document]
svyatonik: I haven't found anything callable from `no_std`

modules/message-lane/src/benchmarking.rs
svyatonik: Yes - iirc both options (i.e. manual checks vs subtraction) gave ~30 (us?) per single message delivery.

modules/message-lane/src/lib.rs
svyatonik: Yes - the storage is used by tests, but it isn't accessible from the lane.

Issue #559, opened 2020-12-03
Provide account id to charge fees when Root|None is sending messages over the bridge.


tomusdrw:
Related: https://github.com/paritytech/parity-bridges-common/pull/558#discussion_r535384173

PR #568, opened 2020-12-09
Rialto to Millau Transfer Generator

HCastano:
Follow up to #535.

One thing that's new in this PR is support in the message generators for sending messages from the Root origin. However, they will not be sent over the bridge just yet (see discussion with Tomek below).


tomusdrw:
To send as root you would need to go through `sudo` pallet (see `sudo::sudo` call), the transaction itself will never get `Root` origin.

modules/message-lane/src/instant_payments.rs
```diff
@@ -46,10 +46,12 @@ where
 				Currency::transfer(submitter, relayer_fund_account, *fee, ExistenceRequirement::AllowDeath)
 					.map_err(Into::into)
 			}
-			Sender::Root | Sender::None => {
-				// fixme: we might want to add root account id to this struct.
-				Err("Root and None account is not allowed to send regular messages.")
+			Sender::Root => {
```
HCastano, line 49: @tomusdrw I had to change this to get Root messages to be sent across the bridge. Does this break what you had in mind when you added this check?

HCastano:
Root messages get accepted by the message lanes, but not dispatched at the moment. It's progress though 😛

modules/message-lane/src/instant_payments.rs
tomusdrw: Yes, it's not acceptable cause confirming such messages would fail (you would try to withdraw more from `relayer_fund_account` than it has - please see #559.

I'd suggest to ignore sending root messages for now.

deployments/bridges/rialto-millau/entrypoints/relay-messages-to-millau-generator-entrypoint.sh
```diff
@@ -3,29 +3,83 @@
...
-MESSAGE=${MSG_EXCHANGE_MESSAGE:-Remark}
+FERDIE_ADDR=5CiPPseXPECbkjWCa6MnjNokrgYjMqmKndv2rSnekmSK2DjL
 
-while true
-do
-	# Sleep a bit between messages
+SHARED_CMD="/home/user/substrate-relay submit-rialto-to-millau-message"
+SHARED_HOST="--rialto-host rialto-node-bob --rialto-port 9944"
+DAVE_SIGNER="--rialto-signer //Dave --millau-signer //Dave"
+ROOT_SIGNER="--rialto-signer //Alice --millau-signer //Alice"
```
svyatonik, line 18: Alice also signs Rialto-PoA headers -> Rialto transactions. Using the same account here may lead to nonces conflict and delays (alerts) in Rialto-PoA -> Rialto headers sync. Possible solutions would be either to change account used by Rialto-PoA headers relay, or change Rialto root account.

I remember Tomek had doubts about keeping Root account origin - instead, for every 'root' action we'll probably need to use separate account. So alternate solution may be to remove root transactions at all. Otoh, the use-case for root account initially was to tune bridge parameters && we may have used something like `ensure_bridged_root(origin)?;` at the target chain. Not sure if we want to keep this use-case, or not? Probably it would be easier to submit target chain transactions directly (without going through bridge) and e.g. to use pallet owner for the same purposes. It is up to you

svyatonik (approved):

deployments/bridges/rialto-millau/entrypoints/relay-messages-to-millau-generator-entrypoint.sh
tomusdrw: Yeah, I'd skip root calls for now, I have some other idea that would replace them, cause the use case is really unclear.

tomusdrw (approved):
looks good to me, except for Slava's comment about re-using Alice's account.

tomusdrw (approved):

PR #570, opened 2020-12-10
Benchmark for message delivery confirmation transaction

svyatonik:
~on top of #567~ 

Similar to #567, I'm going to introduce weight fomula later, after migrating to `trait WeightInfo`.


bin/runtime-common/src/messages_benchmarking.rs
```diff
@@ -142,3 +145,47 @@ where
...
+) -> FromBridgedChainMessagesDeliveryProof<B>
+where
+	B: MessageBridge,
+	H: Hasher,
+	R: pallet_substrate_bridge::Config,
+	<R::BridgedChain as bp_runtime::Chain>::Hash: Into<HashOf<BridgedChain<B>>>,
+	ML: Fn(LaneId) -> Vec<u8>,
+	MH: Fn(H::Out) -> <R::BridgedChain as bp_runtime::Chain>::Header,
+{
+	// prepare Bridged chain storage with inbound lane state
```
HCastano, line 163: Since out documentation is now using `source`/`target` chains we should also use that in our comments over `bridged`/`this` there possible. What do you think?

bin/runtime-common/src/messages_benchmarking.rs
```diff
@@ -142,3 +145,47 @@ where
...
+	let mut root = Default::default();
+	let mut mdb = MemoryDB::default();
+	{
+		let mut trie = TrieDBMut::<H>::new(&mut mdb, &mut root);
+		trie.insert(&storage_key, &params.inbound_lane_data.encode())
+			.map_err(|_| "TrieMut::insert has failed")
+			.expect("TrieMut::insert should not fail in benchmarks");
+	}
+
+	// generate storage proof to be delivered to This chain
```
HCastano, line 174: Same comment about `bridged`/`this`

modules/message-lane/src/benchmarking.rs
```diff
@@ -200,6 +218,111 @@ benchmarks_instance! {
 		);
 	}
 
+	// Benchmark `receive_messages_delivery_proof` extrinsic with fllowing conditions:
```
HCastano, line 221:
```suggestion
	// Benchmark `receive_messages_delivery_proof` extrinsic with following conditions:
```

modules/message-lane/src/benchmarking.rs
```diff
@@ -200,6 +218,111 @@ benchmarks_instance! {
...
+		});
+	}: receive_messages_delivery_proof(RawOrigin::Signed(relayer_id.clone()), proof)
+	verify {
+		assert_eq!(
+			T::account_balance(&relayer_id),
+			relayer_balance + MESSAGE_FEE.into(),
+		);
+	}
+
+	// Benchmark `receive_messages_delivery_proof` extrinsic with fllowing conditions:
```
HCastano, line 251:
```suggestion
	// Benchmark `receive_messages_delivery_proof` extrinsic with following conditions:
```

modules/message-lane/src/benchmarking.rs
```diff
@@ -200,6 +218,111 @@ benchmarks_instance! {
...
+		});
+	}: receive_messages_delivery_proof(RawOrigin::Signed(relayer_id.clone()), proof)
+	verify {
+		assert_eq!(
+			T::account_balance(&relayer_id),
+			relayer_balance + (MESSAGE_FEE * 2).into(),
+		);
+	}
+
+	// Benchmark `receive_messages_delivery_proof` extrinsic with fllowing conditions:
```
HCastano, line 284:
```suggestion
	// Benchmark `receive_messages_delivery_proof` extrinsic with following conditions:
```

modules/message-lane/src/benchmarking.rs
```diff
@@ -200,6 +218,111 @@ benchmarks_instance! {
...
+				relayers: vec![(1, 1, relayer_id.clone())].into_iter().collect(),
+				latest_received_nonce: 1,
+				latest_confirmed_nonce: 0,
+			}
+		});
+	}: receive_messages_delivery_proof(RawOrigin::Signed(relayer_id.clone()), proof)
+	verify {
+		assert_eq!(
+			T::account_balance(&relayer_id),
+			relayer_balance + MESSAGE_FEE.into(),
```
HCastano, line 247: So here we don't need to actually assert that `latest_confirmed_nonce = 1` since the fact that we got paid a fee is "proof" that it was confirmed?

modules/message-lane/src/benchmarking.rs
```diff
@@ -200,6 +218,111 @@ benchmarks_instance! {
...
+		);
+	}
+
+	// Benchmark `receive_messages_delivery_proof` extrinsic with fllowing conditions:
+	// * two relayers are rewarded for relaying single message each;
+	// * relayer account does not exist (in practice it needs to exist in production environment).
+	//
+	// Additional weight for paying reward to the next relayer could be computed
+	// as `weight(receive_delivery_proof_for_two_messages_by_two_relayers)
+	//   - weight(receive_delivery_proof_for_two_messages_by_single_relayer)`.
```
HCastano, line 290: Is this because we'd have to do two separate transfers instead of one transfer with twice the funds? What other overhead costs are there for paying more relayers?

modules/message-lane/src/benchmarking.rs
```diff
@@ -274,6 +397,86 @@ benchmarks_instance! {
...
 		);
 	}
+
+	// Benchmark `receive_messages_delivery_proof` extrinsic where single relayer delivers multiple messages.
+	receive_delivery_proof_for_multiple_messages_by_single_relayer {
+		// there actually should be used value of `MaxUnrewardedRelayerEntriesAtInboundLane` from the bridged
+		// chain, but we're more interested in additional weight/message than in max weight
+		let i in 1..T::MaxUnrewardedRelayerEntriesAtInboundLane::get()
+			.try_into()
+			.expect("too large MaxUnrewardedRelayerEntriesAtInboundLane");
```
HCastano, line 407:
```suggestion
			.expect("Value of MaxUnrewardedRelayerEntriesAtInboundLane is too large");
```

modules/message-lane/src/benchmarking.rs
```diff
@@ -274,6 +397,86 @@ benchmarks_instance! {
 			20,
 		);
 	}
+
+	// Benchmark `receive_messages_delivery_proof` extrinsic where single relayer delivers multiple messages.
+	receive_delivery_proof_for_multiple_messages_by_single_relayer {
+		// there actually should be used value of `MaxUnrewardedRelayerEntriesAtInboundLane` from the bridged
+		// chain, but we're more interested in additional weight/message than in max weight
```
HCastano, line 404: I'm confused with this comment. Do you mean that we should be using a "real" value from one of the runtimes, but it doesn't matter since we only care about each _additional_ message?

modules/message-lane/src/benchmarking.rs
```diff
@@ -274,6 +397,86 @@ benchmarks_instance! {
...
+		);
+	}
+
+	// Benchmark `receive_messages_delivery_proof` extrinsic where every relayer delivers single messages.
+	receive_delivery_proof_for_multiple_messages_by_multiple_relayers {
+		// there actually should be used value of `MaxUnconfirmedMessagesAtInboundLane` from the bridged
+		// chain, but we're more interested in additional weight/message than in max weight
+		let i in 1..T::MaxUnconfirmedMessagesAtInboundLane::get()
+			.try_into()
+			.expect("too large MaxUnconfirmedMessagesAtInboundLane");
```
HCastano, line 441:
```suggestion
			.expect("Value of MaxUnconfirmedMessagesAtInboundLane is too large ");
```

HCastano (approved):

modules/message-lane/src/benchmarking.rs
svyatonik: Yes, I don't think everything should be tested here - there are regular tests for that. But probably worth replacing this test with the test you've mentioned, because runtime may not pay rewards instantly (although now all runtimes are supposed to do that). Otoh probably switching from instant rewards to some other schema may require changes to weight formula, so better to fail test in this case...

modules/message-lane/src/benchmarking.rs
svyatonik: Yes. Basically weight of this tx is now `base_weight + messages_count * message_overhead + relayers_count * relayer_overhead`. So there's one test to compute that `message_overhead` (which is ~ cost of db read) and another one for computing `relayer_overhead` (which is ~ cost of transfer).

modules/message-lane/src/benchmarking.rs
svyatonik: Yes. I mean that someone probably want the weight for the worst case (when there are `MaxUnrewardedRelayerEntriesAtInboundLane` relayers rewarded in the single tx). So this comment here is to warn that this isn't necessary the value you're interested in.

bin/runtime-common/src/messages_benchmarking.rs
svyatonik: It isn't always possible to say that chain A is the source chain and chain B is the target chain. Chain A may be the source of messages, but at the same time it is the target of the confirmations. So when we talk about confirmations, A is the target and B is the source. And even more, the same message lane module instance is used to send messages to the other chain and receive messages from the other chain. 

So I think there are places where we may use source/target chain terminology, but in the code there are much more places where we should refer the other chain as a  bridged chain.

tomusdrw: >  So when we talk about confirmations, A is the target and B is the source. And even more, the same message lane module instance is used to send messages to the other chain and receive messages from the other chain.

I agree with Hernando and would prefer having consistent naming. While I understand your point, about source/target being switched, I think we should not change the frame of reference here. The entire interaction is always `source -> target` chain and the frame of reference is message delivery, confirmations are simply delivered from `target chain` to `source chain` and I don't see an issue with this. Obviously from the relayer perspective "target" might be where you send transaction, but as I said we should not switch the frame of reference, but rather keep focused on consistently naming the chain in the "message passing" frame of reference.

That said, I do think it's totally fine to use "bridged chain" when it's clear from the context and secondly message lanes use bridge/this terminology all over, so it might actually be better to change this in a separate PR instead of randomly mixing the two.

PR #571, opened 2020-12-10
Fix send_message_worst_case benchmark

### LINKED ISSUE #569: Fix send_message_worst_case benchmark
It is broken after #558

svyatonik:
closes #569

### DIFF
--- a/bin/rialto/runtime/src/lib.rs
+++ b/bin/rialto/runtime/src/lib.rs
@@ -830,7 +830,7 @@ impl_runtime_apis! {
 				}
 
 				fn prepare_message(
-					params: MessageLaneMessageParams,
+					params: MessageLaneMessageParams<Self::AccountId>,
 				) -> (millau_messages::ToMillauMessagePayload, Balance) {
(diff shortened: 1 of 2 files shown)


bin/rialto/runtime/src/lib.rs
```diff
@@ -843,7 +843,9 @@ impl_runtime_apis! {
 							* sp_std::cmp::min(factor, WORST_MESSAGE_SIZE_FACTOR),
 					};
 					let message_payload = vec![0; message_size as usize];
-					let dispatch_origin = pallet_bridge_call_dispatch::CallOrigin::SourceAccount(Default::default());
```
HCastano, line 846: Was it broken because of this?

HCastano (approved):
Looks fine, but I don't quite understand what how benchmarks were broken by #558.

If it failed to compile then we should add the `message-lane` benchmarks to the CI soon.

bin/rialto/runtime/src/lib.rs
svyatonik: Yes

PR #572, opened 2020-12-10
Add scripts for running local Rialto<->Millau header sync

HCastano:
I've intentionally left out scripts for message lanes since that adds a bit of complexity to the local setup and at that point you might as well use our Compose deployment and swap out individual components with local builds.

If you guys think that's the wrong mindset I can add scripts for message lanes.


tomusdrw (approved):
looks good enough for me :) I'm fine with leaving out the message lane scripts, I guess it's a nice exercise for the reader to run them without our help if they don't like docker-compose :joy:

PR #575, opened 2020-12-14
Limit size of message delivery transaction

### LINKED ISSUE #552: Limit maximal size of delivery transaction
follow up of #551

Now messages are limited by their cumulative weight, count (#541). In addition, they should be limited by size. Roughly - we should accumulate messages until their cumulative weight overflows 2/3 of maximal tx size on target chain.

svyatonik:
closes #552

Actually limit is for bundled messages of the delivery transaction. The limit is 1/3 of the maximal tx size of the chain. The runtime has slightly different check - it checks that single message has size <= 2/3 of the maximal size. The difference is because the more messages we have, the more proof overhead size would be.

I've also changed weight-related relay code a bit - even if single message weight/size overflows limits, the delivery transaction is still submitted. The warning is also printed (it may be invalid sometimes - i.e. when the message is actually large). But anyway it worth noticing.


primitives/kusama/src/lib.rs
```diff
@@ -122,15 +122,15 @@ sp_api::decl_runtime_apis! {
 	/// This API is implemented by runtimes that are sending messages to Kusama chain, not the
 	/// Kusama runtime itself.
 	pub trait ToKusamaOutboundLaneApi {
-		/// Returns dispatch weight of all messages in given inclusive range.
+		/// Returns dispatch weight and size of all messages in given inclusive range.
```
tomusdrw, line 125:
```suggestion
		/// Returns total dispatch weight and encoded payload size of all messages in given inclusive range.
```

tomusdrw: I'd be a bit more explicit in the comment here (and in other instances).

relays/messages-relay/src/message_lane_loop.rs
```diff
@@ -71,10 +71,21 @@ pub struct MessageDeliveryParams {
...
+	/// Maximal cumulative size of relayed messages in single delivery transaction.
+	pub max_messages_size_in_single_batch: usize,
+}
+
+/// Message weights.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct MessageWeights {
+	/// Message dispatch weight.
+	pub weight: Weight,
+	/// Message size.
```
tomusdrw, line 83:
```suggestion
	/// Message size - number of bytes in encoded payload.
```

relays/messages-relay/src/message_race_delivery.rs
```diff
@@ -369,11 +374,44 @@ impl<P: MessageLane> RaceStrategy<SourceHeaderIdOf<P>, TargetHeaderIdOf<P>, P::M
 				let to_requeue = range
 					.into_iter()
 					.skip_while(|(_, weight)| {
+						// Since we (hopefully) have some reserves in `max_messages_weight_in_single_batch`
+						// and `max_messages_size_in_single_batch`, we may still try to submit transaction
+						// with single message if message overflows these limits. The worst case would be if
```
tomusdrw, line 379: Just to clarify, the worst thing that might happen is:
1. The transaction is rejected from the pool (i.e. not even included in the block)
2. So the relayer will not really be charged fee (i.e. we don't open a griefing vector here)
3. Such thing should never really happen in practice unless runtimes are misconfigured, right? If it happens the lane is essentially clogged, unless relayers are configured incorrectly.

tomusdrw (approved):
lgtm, tiny docs grumbles.

relays/messages-relay/src/message_race_delivery.rs
svyatonik: Normally yes - afair, before accepting tx to the pool, all its signed extensions are checked by all supported runtimes. And weight and size checks are done by `CheckWeight` extension. So out-of-bounds tx shall not even reach the pool => fee would not be charged.

Such thing should not happen in practice. But in addition to runtime/relayer misconfiguration, there's another case when it might happen - if storage proof is too large. If storage trie is large enough to generate proofs that can occupy 2/3 of maximal tx size, then relay may select e.g. 2 messages for the single delivery tx. Let's say the proof would take the whole tx size (1/3 for messages and (2/3 + 1 byte) for trie nodes). In this case the lane is not actually clogged (because we may deliver messages 1-by-1), but the relayer will always select 2-messages for transaction => messages will never be delivered. This may be fixed in relayer code, but if the proof is that large, then it is actually easier to cheat source runtime, actually clogging the lane.

relays/messages-relay/src/message_lane_loop.rs
svyatonik: I'm not sure - what do you mean here? It is actually size of value stored in the messages storage map (the value that is proved later).

svyatonik: Ok, looks like I misinterpreted dash as a minus sign :) I thought you're subtracting encoded payload size from the message size

tomusdrw: Hah, yeah I just meant to clarify what "size" is. Feel free to re-phrase to make it clearer, maybe a comma would less ambiguous.

PR #580, opened 2020-12-15
Add correct Docker commands in README

HCastano:
This updates the README to use Docker commands with ones that aren't stubs. The goal here is to get somebody started with a basic one way header sync setup really quickly and easily.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ There are 3 ways to run the bridge, described below:
  - building or using Docker images for each individual component,
  - running a Docker Compose setup (recommended).
 
-### Building
+### Using the Source
 
 First you'll need to build the bridge nodes and relay. This can be done as follows:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -115,41 +115,74 @@ cargo build -p substrate-relay
...
 ```
 
 Notice that the `docker run` command will accept all the normal Substrate flags. For local
 development you should at minimum run with the `--dev` flag or else no blocks will be produced.
 
+Then we need to initialize and run the relayer:
+
+```bash
+docker run --network=host -it paritytech/substrate-relay initialize-rialto-headers-bridge-in-millau \
+	--millau-host localhost \
```
tomusdrw, line 155:
```suggestion
docker run --network=host -it \
	paritytech/substrate-relay initialize-rialto-headers-bridge-in-millau \
	--millau-host localhost \
```

README.md
```diff
@@ -115,41 +115,74 @@ cargo build -p substrate-relay
...
+	--millau-port 9945 \
+	--rialto-host localhost \
+	--rialto-port 9944 \
+	--millau-signer //Alice
+
+docker run --network=host -it paritytech/substrate-relay rialto-headers-to-millau \
+	--millau-host localhost \
```
tomusdrw, line 162:
```suggestion
docker run --network=host -it \
	paritytech/substrate-relay rialto-headers-to-millau \
	--millau-host localhost \
```

tomusdrw (approved):

tomusdrw (approved):

tomusdrw:
@General-Beck Could you take a look at the CI error? Why do we even run the pipeline if `README.md` should be in `paths-ignore`?

Issue #579, opened 2020-12-15
Add weight-related parameters to message lane calls


svyatonik:
Ideas that I've got when working on messages lane benchmarks (#567 and #570). We need to add following parameters to message lane transactions:
- [x] number of bundled messages to the [delivery transaction](https://github.com/paritytech/parity-bridges-common/blob/e31ac43b614ddc69123067f5e19c94b346b42fc9/modules/message-lane/src/lib.rs#L332). So relayer would only pay message-overhead fee for actually delivered messages. This change would also allow us to remove [MaxMessagesInDeliveryTransaction](https://github.com/paritytech/parity-bridges-common/blob/e31ac43b614ddc69123067f5e19c94b346b42fc9/modules/message-lane/src/lib.rs#L100) - it will be unnecessary then;
- [x] number of entries in [relayers](https://github.com/paritytech/parity-bridges-common/blob/e31ac43b614ddc69123067f5e19c94b346b42fc9/primitives/message-lane/src/lib.rs#L88) vec and number of messages that we'll need to read to the [confirmation transaction](https://github.com/paritytech/parity-bridges-common/blob/e31ac43b614ddc69123067f5e19c94b346b42fc9/modules/message-lane/src/lib.rs#L414). Adding these parameters would allow us to replace [MaxUnrewardedRelayerEntriesAtInboundLane](https://github.com/paritytech/parity-bridges-common/blob/e31ac43b614ddc69123067f5e19c94b346b42fc9/modules/message-lane/src/lib.rs#L88) and [MaxUnconfirmedMessagesAtInboundLane](https://github.com/paritytech/parity-bridges-common/blob/e31ac43b614ddc69123067f5e19c94b346b42fc9/modules/message-lane/src/lib.rs#L95) with actual values in weight formula. We would still need these parameters, because we can't process `relayers` vector paritally.

PR #581, opened 2020-12-16
Add messages count parameter to delivery transaction

svyatonik:
first part of #579

One thing I was thinking of is that now anyone could submit tx with some invalid data in `proof` parameter and declaring 0 messages in the tx. So it'll pay minimal fee for the tx. But since we're targeting chains where tx fee includes per-byte fee, I think it is ok. The same may be achieved with the `remark` call, which has declared constant weight e.g. on Polkadot/Kusama.

### DIFF
--- a/bin/millau/runtime/src/lib.rs
+++ b/bin/millau/runtime/src/lib.rs
@@ -313,8 +313,6 @@ parameter_types! {
 		bp_millau::MAX_UNREWARDED_RELAYER_ENTRIES_AT_INBOUND_LANE;
 	pub const MaxUnconfirmedMessagesAtInboundLane: bp_message_lane::MessageNonce =
 		bp_millau::MAX_UNCONFIRMED_MESSAGES_AT_INBOUND_LANE;
-	pub const MaxMessagesInDeliveryTransaction: bp_message_lane::MessageNonce =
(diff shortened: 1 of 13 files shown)


HCastano (approved):
Code looks good.

> One thing I was thinking of is that now anyone could submit tx with some invalid data in proof parameter and declaring 0 messages in the tx. So it'll pay minimal fee for the tx. 

Can you clarify this for me though? What's the reason that somebody would submit a transaction with a proof that indicated >0 messages but claim that it only had 0 messages? We would reject the proof eventually in this case, right?

svyatonik:
Without per-byte fee, you may store arbitrary data (of any length up to `max_extrinsic_size`) on full nodes almost for free (in our case - for `weight_to_fee(base_delivery_tx_weight)`).  [Per-byte fee](https://github.com/paritytech/substrate/blob/987a9723920217917f2708388d150add5ef52ef7/frame/transaction-payment/src/lib.rs#L226) makes (or should make) this expensive.

So I had a doubts on whether we need to alter delivery tx weight formula to charge submitter (here: relayer) for proof length. But looking at the weight of `System::remark()` call, you may see that its weight is not independent of the remark length. So it relies on 'system' per-byte fee => we should also rely on this system mechanism (since all our target chains have that per-byte fee).

PR #582, opened 2020-12-16
Use `cargo-chef` for caching dependencies in Dockerfile

### LINKED ISSUE #473: Try out cargo chef for Docker builds
Currently rebuilding docker images is a huge pain - they take really long to re-compile. We might want to try out `cargo chef` to cache build dependencies first (details: https://www.lpalmieri.com/posts/fast-rust-docker-builds/)

HCastano:
Introduces `cargo-chef` in order to improve out Docker build times. This doesn't improve clean build times (in fact, I suspect it might make them worse), but we make up for that time by caching our dependencies. I did some benchmarking with `hyperfine` and found that this does in fact provide improvements for cached builds. See my comment below for more details on that.

This PR also adds a dependency Docker image which helps cache some shared dependencies between the build stages of the main Dockerfile. While it is not published under the `paritytech` org on Docker Hub just yet, I've added a job to do that.

Closes #473.


Dockerfile
```diff
@@ -1,13 +1,25 @@
...
+WORKDIR /parity-bridges-common
+
 # show backtraces
 ENV RUST_BACKTRACE 1
-
 ENV LAST_DEPS_UPDATE 2020-06-22
 
 # install tools and dependencies
+# TODO: Check if we need all these deps (e.g jq)
 RUN set -eux; \
```
tomusdrw, line 23: We should probably publish our own image with these things, which would speed things up as well: the `cacher` and `builder` phase would be based on the same image - now we have to keep both parts exactly the same to make sure that docker cache is utilized here, otherwise prepare the base image from scratch twice.

tomusdrw (commented):
looks interesting!

Dockerfile
HCastano: Yeah, I think that a separate "base" build image is a good idea. I don't know how much time we'd save using it (since grabbing and building the dependencies is free after the first build), but it will definitely make for a more manageable Dockerfile

HCastano:
I did some benchmarking today using `hyperfine`. My goal was to try and compare how non-initial build times varied between the current Dockerfile that's on `master` and what's being proposed in this PR.

I ran all the benchmarks with the following command:

```bash
hyperfine --prepare 'echo "blep" >> foo.rs' --min-runs 2 --max-runs 5 'docker build . -t docker/millau-bridge-node --build-arg=PROJECT=millau-bridge-node'
```

I first checked out the `master` branch and ran the benchmarks on the `millau-bridge-node` and `substrate-relay` projects. After that was done I checked out this branch and did the same thing.

One thing you might find strange is the command being run by `--prepare`. That command gets run before each of the timing runs, and what it does is blow the cache for the actual build of the project (the initial dependency and tool prep is unaffected).

The table below shows the outcome of the benchmarks:

| Run | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Millau Node (Master) | 480.659 ± 12.816 | 471.596 | 489.721 | 1.00 |
| Millau Node (Chef) | 363.118 ± 2.655 | 361.241 | 364.996 | 1.32 |
| Substrate Relay (Master) | 225.246 ± 1.300 | 224.326 | 226.165 | 1.00 |
| Substrate Relay (Chef) | 224.815 ± 3.170 | 222.574 | 227.056 | 1.00 |

The gains are pretty significant from the bridge node. Interestingly enough though, there is basically no change when building the relay.

Dockerfile
```diff
@@ -1,47 +1,48 @@
...
+FROM rust as planner
+WORKDIR /parity-bridges-common
+RUN cargo install cargo-chef
+COPY . .
+RUN cargo chef prepare --recipe-path recipe.json
+
+# This second stage is where the dependencies actually get built.
+# The reason we split it from the first stage is so that the `COPY . .`
+# step doesn't blow our cache.
+FROM hcastano/bridge-deps AS cacher
```
HCastano, line 21: Currently using an image that I published since we don't have any published with the `paritytech` org just yet (will need to wait until this PR is merged)

.github/workflows/publish-deps.yml
```diff
@@ -0,0 +1,55 @@
...
+
+on:
+  push:
+    tags:
+      - v*
+    paths-ignore:
+      - 'README.md'
+      - diagrams/
+  schedule:                                         # Nightly build
+    - cron:                                         '0 0 * * *'
```
HCastano, line 11: I don't think this needs to be done nightly. Rather, we should do it weekly, but even that might be too often. It's set to nightly right now though so we can get an "official" image published quickly

deployments/BridgeDeps.Dockerfile
```diff
@@ -0,0 +1,32 @@
+# Image with dependencies required to build projects from the bridge repo.
+#
+# This image is meant to be used as a building block when building images for
+# the various components in the bridge repo, such as nodes and relayers.
+FROM ubuntu:xenial
```
adoerr, line 5: Any particular reason why to use `xenial` (16.04 LTS) and not `focal` (20.04 LTS)? `xenial` is pretty outdated.

Dockerfile
```diff
@@ -1,47 +1,48 @@
...
 
 ARG PROJECT=ethereum-poa-relay
-
 RUN cargo build --release --verbose -p ${PROJECT}
 RUN strip ./target/release/${PROJECT}
 
-FROM ubuntu:xenial
+# In this final stage we copy over the final binary and do some checks
+# to make sure that everything looks good.
+FROM ubuntu:xenial as runtime
```
adoerr, line 45: Same question here, regarding the Ubuntu version.

adoerr (approved):
LGTM, given my limited DevOps background.

deployments/BridgeDeps.Dockerfile
tomusdrw: I guess it was simply just moved, but let's maybe do an upgrade since there is a good occasion :)

HCastano: I've tried switching the base image to `focal`, but I'm having issues getting the Docker images built. For some reason Docker can't find some files 🤷 

```
Step 18/31 : RUN cargo build --release --verbose -p ${PROJECT}
 ---> Running in eba424f7d692
error: failed to read `/parity-bridges-common/bin/millau/node/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
The command '/bin/sh -c cargo build --release --verbose -p ${PROJECT}' returned a non-zero code: 101
```

If you're curious, the steps to reproduce are:
- `s/xenial/focal/g`
- Add `ENV DEBIAN_FRONTEND=noninteractive` to the top of `BridgeDeps.Dockerfile` (prevents `tzdata` from stalling the build)
- Build deps
- Build relay or node

I would say that unless there is a strong reason _for_ switching, we should keep the `xenial` base images and log an issue to upgrade them at a later date.

adoerr: We still have until April 2021. That is when standard support for `xenial` ends ;-)

tomusdrw: logged here: https://github.com/paritytech/parity-bridges-common/issues/626

.github/workflows/publish-deps.yml
tomusdrw: Logged here as well: https://github.com/paritytech/parity-bridges-common/issues/627

tomusdrw (approved):

PR #584, opened 2020-12-17
Add unrewarded relayers state param to the delivery confirmation transactions

### LINKED ISSUE #579: Add weight-related parameters to message lane calls

svyatonik:
closes #579 

Second part of #579. The weight of the call may now be computed as `base_weight + num_of_messages * message_weight + num_of_relayers * relayer_weight`

### DIFF
--- a/primitives/message-lane/src/lib.rs
+++ b/primitives/message-lane/src/lib.rs
@@ -103,13 +103,15 @@ impl<RelayerId> Default for InboundLaneData<RelayerId> {
 }
 
 /// Gist of `InboundLaneData::relayers` field used by runtime APIs.
-#[derive(Clone, Encode, Decode, RuntimeDebug, PartialEq, Eq)]
+#[derive(Clone, Default, Encode, Decode, RuntimeDebug, PartialEq, Eq)]
 pub struct UnrewardedRelayersState {
 	/// Number of entries in the `InboundLaneData::relayers` set.
 	pub unrewarded_relayer_entries: MessageNonce,
 	/// Number of messages in the oldest entry of `InboundLaneData::relayers`. This is the
 	/// minimal number of reward proofs required to push out this entry from the set.
 	pub messages_in_oldest_entry: MessageNonce,
+	/// Total number of messages in the relayers vector.
+	pub total_messages: MessageNonce,
 }
 
 /// Outbound lane data.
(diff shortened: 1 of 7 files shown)


primitives/message-lane/src/lib.rs
```diff
@@ -134,3 +136,13 @@ impl Default for OutboundLaneData {
 		}
 	}
 }
+
+/// Returns total number of messages in the `InboundLaneData::relayers` vector.
+pub fn total_unrewarded_messages<RelayerId>(
+	relayers: &VecDeque<(MessageNonce, MessageNonce, RelayerId)>,
+) -> MessageNonce {
+	match (relayers.front(), relayers.back()) {
+		(Some((begin, _, _)), Some((_, end, _))) => end.checked_sub(*begin).and_then(|d| d.checked_add(1)).unwrap_or(0),
```
HCastano, line 145: Why do we do `d.checked_add(1)` after getting the difference?

HCastano (approved):

primitives/message-lane/src/lib.rs
svyatonik: The range is inclusive, so if it is `[1; 1]`, then the number of messages is `(1 - 1) + 1 = 1`

PR #585, opened 2020-12-17
Disable sccache temporarily.

tomusdrw:


General-Beck:
@tomusdrw The problem has been resolved. Runners that are not fully configured have received the highest priority after the update.

PR #586, opened 2020-12-18
Fixes for Polkadot Integration

HCastano:
This PR updates some of the code used by our Polkadot integration branch to get it into a compiling state again. The most up to date Polkadot code is in the [`bridges-integration`](https://github.com/paritytech/polkadot/compare/bridges-integration?expand=1) branch of the main Polkadot repo.

Since the bridges repo doesn't follow the lastest `master` for Substrate our Polkadot integration branch also doesn't follow the latest `master`. Specifically, we are on the following commits for each of the repos:
- The bridges repo and Polkadot integration branch are on Substrate commit [`﻿﻿﻿9d843c3e`](https://github.com/paritytech/substrate/commit/9d843c3e1b05c0ddfa2469277f5b522f7d57f612)
- The Polkadot integration branch is based on Polkadot commit [`7a0fcb07`](https://github.com/paritytech/polkadot/commit/7a0fcb07fc50becf159fe8f61f4f65e2c4d3ee44)

With some recent changes that Slava introduced we _could_ move to the latest Substrate (I did a quick compile check), which would allow us to bump our Polkadot branch ahead further. I think that should be done in a different PR though to prevent something like #583 from slipping in.

While our Polkadot branch compiles, I haven't been able to test out any bridge functionality yet since we need some additional code in the relayer.


primitives/rialto/src/lib.rs
```diff
@@ -132,9 +110,10 @@ pub fn derive_account_from_millau_id(id: bp_runtime::SourceAccount<AccountId>) -
 	AccountIdConverter::convert(encoded_id)
 }
 
-/// Get a struct which defines the weight limits and values used during extrinsic execution.
-pub fn runtime_block_weights() -> frame_system::limits::BlockWeights {
-	frame_system::limits::BlockWeights::builder()
+frame_support::parameter_types! {
+	pub BlockLength: limits::BlockLength =
+		limits::BlockLength::max_with_normal_ratio(5 * 1024 * 1024, NORMAL_DISPATCH_RATIO);
+	pub BlockWeights: limits::BlockWeights = limits::BlockWeights::builder()
```
svyatonik, line 116: Not related, to this PR, but probably this builder may be refactored to const methods and we will be able to mark this parameter as const. Imo it'll be better to expose this struct (`BlockWeights` and `BlockLength`) as const from this crate.

svyatonik (approved):

Issue #578, opened 2020-12-15
Expose API for estimating delivery_and_dispatch_fee


svyatonik:
Right now it is not clear what to pass as [delivery_and_dispatch_fee](https://github.com/paritytech/parity-bridges-common/blob/e31ac43b614ddc69123067f5e19c94b346b42fc9/modules/message-lane/src/lib.rs#L255) to the `send_message` function. The minimal value is returned by the [estimate_message_dispatch_and_delivery_fee](https://github.com/paritytech/parity-bridges-common/blob/e31ac43b614ddc69123067f5e19c94b346b42fc9/bin/runtime-common/src/messages.rs#L225) and should be made available to the message submitter (either through runtime API or specialized RPC).

It may be a good idea to add additional fee-related params to method to compute updated fee. I.e. if source runtime has obsolete `sourceTKN / targetTKN` conversion rate and we know that it has been updated, then we may pass updated rate to the method and pay increased (passing decreased fee is impossible, unfortunately) fee to guarantee that our message will be delivered by the non-altruistic relayer (who also knows the updated rate).

PR #589, opened 2020-12-21
Decrease transfer amount

svyatonik:
Right now Millau/Rialto Dave every `RND(1s; 30s)` pays: `(100_000_000 fee + 1_000_000_000 fee + 1_000_000_000_000 transfer + 100_000_000 fee + 1_000_000_000 fee)` . Balance of Dave is `1 << 50`, which is `1125899906842624`. So this balance is enough for `1123` intervals.

I've decreased transfer amount from `1_000_000_000_000` to `1_000_000_000`. I believe fees may also be decreased (last time I was trying they've been estimated to something less than `1_000_000`), but it'll be fixed in #578, which I prefer to do after finally introducing message lane weights. Even without changing fees, with fixed amount it should be enough for `351_843` intervals, which is ~60 hours even if interval is always 1s.

### DIFF
--- a/deployments/bridges/rialto-millau/entrypoints/relay-messages-to-millau-generator-entrypoint.sh
+++ b/deployments/bridges/rialto-millau/entrypoints/relay-messages-to-millau-generator-entrypoint.sh
@@ -42,7 +42,7 @@ do
 		--fee 1000000000 \
 		--origin Target \
 		transfer \
-		--amount 1000000000000 \
+		--amount 1000000000 \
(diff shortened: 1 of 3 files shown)


svyatonik:
(extracting clippy things in separate PR)

tomusdrw (approved):

HCastano (approved):

HCastano:
> I believe fees may also be decreased (last time I was trying they've been estimated to something less than 1_000_000)

PR #590, opened 2020-12-21
Automated weights for message lane module

svyatonik:
~There's major problem with this PR - weights are generated using `wasmi`, while normally `wasmtime` is used. I hope that's the reason (and also my laptop) why there are such huge numbers in the `weights.rs`.~

~I was trying to use wasmtime, but had (and still have) problems with it. So it'll be a separate PR. This PR is still helpful, because finally I may introduce weight formulas for message-lane calls.~

~Draft until #593 is in~

With `wasmtime` it looks much better


HCastano (approved):
Cool stuff!

PR #593, opened 2020-12-21
Enable Wasmtime and Remove `secp256k1`

svyatonik:
mostly to use from benchmarks

I've also removed dependencies on `secp256k1`, - we're already using pure Rust implementation && this dependency has caused `cc` version conflict (which may have been solved without removal, but still).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16,7 +16,7 @@ version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1b6a2d3371669ab3ca9797670853d61402b03d0b4b9ebf33d677dfa720203072"
 dependencies = [
- "gimli",
+ "gimli 0.22.0",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


HCastano (approved):
> I've also removed dependencies on secp256k1, - we're already using pure Rust implementation && this dependency has caused cc version conflict (which may have been solved without removal, but still).

Just to clarify what you mean here. We're actually _adding_ `secp256k1` as a direct dependency, but removing dependencies which pull in it (since I see that you've removed `parity-crypto` and `ethereum-tx-sign`).

svyatonik:
Nope. I'm removing `parity-crypto` and `ethereum-tx-sign` (and `signing` feature of `web3`) which are using [secp256k1](https://crates.io/crates/secp256k1), which is a wrapper around native library and brings `cc` dependency. Instead, we're switching to [libsecp256k1](https://crates.io/crates/libsecp256k1) everywhere. This is pure Rust library (that may be used from `no_std`), which is also used by Substrate and so on.

HCastano:
> Nope. I'm removing `parity-crypto` and `ethereum-tx-sign` (and `signing` feature of `web3`) which are using [secp256k1](https://crates.io/crates/secp256k1), which is a wrapper around native library and brings `cc` dependency. Instead, we're switching to [libsecp256k1](https://crates.io/crates/libsecp256k1) everywhere. This is pure Rust library (that may be used from `no_std`), which is also used by Substrate and so on.

Ah gotcha! Thanks for clarifying that

PR #595, opened 2020-12-22
Fix benchmarks compilation

svyatonik:


svyatonik:
(merging as is, because it breaks master)

PR #596, opened 2020-12-22
Message lane weight formulas

svyatonik:
~on top of #590 => draft~

This PR only adds formulas - everything else (like changing affected `message_delivery_and_dispatch_fee`) would be done in follow-up PRs.

### DIFF
--- a/modules/message-lane/src/lib.rs
+++ b/modules/message-lane/src/lib.rs
@@ -26,6 +26,12 @@
 //! Once message is sent, its progress can be tracked by looking at module events.
 //! The assigned nonce is reported using `MessageAccepted` event. When message is
 //! delivered to the the bridged chain, it is reported using `MessagesDelivered` event.
+//!
+//! **IMPORTANT NOTE**: after generating weights (custom `WeighInfo` implementation) for
+//! your runtime (where this module is plugged to), please add test for these weights.
+//! The test should call the `ensure_weights_are_correct` function from this module.
+//! If this test fails with your weights, then either weights are computed incorrectly,
+//! or some benchmarks assumptions are broken for your runtime.
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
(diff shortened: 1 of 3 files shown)


modules/message-lane/src/lib.rs
```diff
@@ -136,6 +134,66 @@ pub trait Config<I = DefaultInstance>: frame_system::Config {
 	type MessageDispatch: MessageDispatch<Self::InboundMessageFee, DispatchPayload = Self::InboundPayload>;
 }
 
+/// Extended weight info.
+pub trait WeightInfoEx: WeightInfo {
```
HCastano, line 138:
```suggestion
pub trait WeightInfoExt: WeightInfo {
```

I think the convention is to use `Ext` for _Extension_ traits (`FutureExt` and `StreamExt` come to mind)

HCastano, line 138: A bit of a nitpick here, but I'd prefer if the trait was declared after all the pallet macros. I don't think this need to be one of the first things somebody reads when looking at the pallet.

modules/message-lane/src/lib.rs
```diff
@@ -525,6 +597,19 @@ impl<T: Config<I>, I: Instance> Module<T, I> {
 	}
 }
 
+/// Ensure that weights from `WeightInfo` implementation are looking correct.
+pub fn ensure_weights_are_correct<W: WeightInfoEx>() {
```
HCastano, line 601: So in the idea here that if we hit one of the `saturating` conditions in the weight calculations that we've done something wrong?

HCastano (approved):

modules/message-lane/src/lib.rs
svyatonik: Right. I was initially going to manually compare benchmarks results here, but `WeightInfoEx` is already doing the same thing (using `saturated_sub`).

modules/message-lane/src/lib.rs
svyatonik: I've extracted this trait into dedicated file

Issue #597, opened 2020-12-22
Change instant payment reward scheme to reward relayer that submit confirmation transaction


svyatonik:
Why? Original idea was that the message sender is paying for future delivery confirmation transaction (DCT). The DCT fees were a part of `message_delivery_and_dispatch_fee`. Since we don't know how much relayer will actually pay for the DCT, there were two options:
1) charge message sender only for single-message-DCT. While this is technically correct, relayer may have lost its funds if there are many unrewarded relayers entries in the inbound lane state (i.e. relayer A has delivered single message, while relayers B..Z have delivered other 1023 messages => if A would submit DCT, it'll pay fees for paying rewards for 1023 messages to B..Z, which may overflow profit from relaying that single message);
2) charge message sender only for worst-possible-DCT. This was the last plan before actual weights have been computed. Now I see that this would be too much. I.e. on Millau it'll be `WEIGHT_PER_SECOND / 2`, which I think is too much (it is a cost of sending ~2000 messages).

The most obvious solution would be:
1) message submitter only pays for its own message DCT (option 1 from above);
2) relayer that has submitted DCT that receives at least `MIN(confirmed_messages_count * single_message_confirmation_fee; total_confirmed_messages_fee)`.

So message relayer would still have a profit (normally it'll receive most of original message fee), but there may be relayers that only submit DCTs. They would receive minimal profit, though, because they're only receiving delivery confirmation fee, which they're paying anyway as a DCT cost.

PR #598, opened 2020-12-23
Use real weights to compute message delivery and dispatch fee

svyatonik:
This PR is unfinished yet && I'm opening it to illustrate #597 and its solution (#599).


svyatonik:
```
millau-to-rialto-messages-generator:  [1;38;5;8m2021-01-15 10:47:57 +0000[0m [38;5;10mINFO[0m [38;5;8mbridge[0m Sending message to Rialto. Fee: 3302170300
```

So the fee for sending relatively small remark call between our test chains is ~0.3 token (~3302170300 / 10000000000). Iiuc it'd be almost the same for DOT -> KSM transfers (0.3 DOT). It'll be decreased after we'll start using reference machine for our benchmarks. The fee does not include cost of `send_message` transaction itself.

If you feel that it is too large, please comment here.

primitives/millau/src/lib.rs
```diff
@@ -61,6 +61,20 @@ pub const MAX_UNREWARDED_RELAYER_ENTRIES_AT_INBOUND_LANE: MessageNonce = 1024;
 /// Maximal number of unconfirmed messages at inbound lane.
 pub const MAX_UNCONFIRMED_MESSAGES_AT_INBOUND_LANE: MessageNonce = 1024;
 
+/// Maximal weight of single message delivery transaction on Millau chain.
+///
+/// This value is a result of `pallet_message_lane::Module::receive_messages_proof` weight formula computation
+/// for the case when single message is delivered. The result then must be rounded up to account possible future
+/// runtime upgrades.
+pub const MAX_SINGLE_MESSAGE_DELIVERY_TX_WEIGHT: Weight = 1_000_000_000;
```
svyatonik, line 69: Actually it is `607_735_000` with current weights.

primitives/millau/src/lib.rs
```diff
@@ -61,6 +61,20 @@ pub const MAX_UNREWARDED_RELAYER_ENTRIES_AT_INBOUND_LANE: MessageNonce = 1024;
...
+
+/// Maximal weight of single message delivery confirmation transaction on Millau chain.
+///
+/// This value is a result of `pallet_message_lane::Module::receive_messages_delivery_proof` weight formula computation
+/// for the case when single message is confirmed. The result then must be rounded up to account possible future
+/// runtime upgrades.
+pub const MAX_SINGLE_MESSAGE_DELIVERY_CONFIRMATION_TX_WEIGHT: Weight = 2_000_000_000;
```
svyatonik, line 76: Actually it is `1_530_101_000` with current weights.

HCastano (approved):
> So the fee for sending relatively small remark call between our test chains is ~0.3 token (~3302170300 / 10000000000). Iiuc it'd be almost the same for DOT -> KSM transfers (0.3 DOT). It'll be decreased after we'll start using reference machine for our benchmarks. The fee does not include cost of send_message transaction itself.

> If you feel that it is too large, please comment here.

Does the 0.3 Token fee include the fees paid to the relayer? My gut feeling is that 0.3 DOT for a relatively simple token transfer seems quite high, especially if it doesn't cover the actual transaction costs. I worry that as we start building more complex applications and sending more complex messages costs will be prohibitive for regular use.

svyatonik:
Yes, `fee` includes relayer fee. My initial feeling was the same (that it is too large), but actually it is the same as cost of 2.5 KB remark on e.g. Polkadot. Which is imo fine, given that user pays for 2 transactions + one additional dispatch. Probably it is even too small - I need to double check

svyatonik:
Ok, I made some experiments and got actual values. So after last commit, message (simple remark) delivery and dispatch fee has been increased from `3302170300` to `3577172640` (i.e. from `0.33` units to `0.36` units).

Balance of relayer account on source chain has changed by `1125903359014445 - 1125899906842624 = 3452171821` (i.e. by `0.34` units).

Balance of message sender on source chain has changed by `1125896204669691 - 1125899906842624 = -3702172933` (i.e. by `-0.37` units).

Balance of relayer account on target account has changed by `1125899781841793 - 1125899906842624 = -125000831` (i.e. by `-0.01` units).

The delivery and dispatch fee has been built from:
1) delivery transaction portion - `1125001048` (`0.11`) target tokens;
2) dispatch portion - `1973000` (`0.0001973`) target tokens;
3) confirmation transaction portion - `2125001080` (`0.21`) source tokens;
4) relayer fee `0.04` source tokens;
5) the rest is send message portion.

So the only way to decrease this fee would be to change constants introduced in this PR (`MAX_SINGLE_MESSAGE_DELIVERY_TX_WEIGHT` and `MAX_SINGLE_MESSAGE_DELIVERY_CONFIRMATION_TX_WEIGHT`). They have a large reserve for future upgrades (which will never happen on our test nets). Before deploying to Polkadot/Kusama we need to do benchmarks on reference machine - I believe this would allow us to make weights maybe x2 times lesser than current even keeping some reserves (I have a 5-6 years old laptop :). For now I think it is ok, but worth additional attention when integrating with Polkadot/Kusama.

BTW - TIL that with Polkadot/Kusama large transaction weight is also affected by 'adjustable part of the fee'. It is zero for our testnets, but will be non-zero for real chains. So it'll be even larger than `0.33` (or total `0.37`).

Another option that may be explored in future is making delivery tx weight depend on message size => we may only charge submitter for its part. But I'm almost sure it won't lead to significant fee decrease.

primitives/rialto/src/lib.rs
```diff
@@ -34,6 +34,14 @@ use sp_runtime::{
 };
 use sp_std::prelude::*;
 
+/// Number of extra bytes (excluding size of storage value itself) of storage proof, built at
+/// Rialto chain. This mostly depends on number of entries (and their density) in the storage trie.
+/// Some reserve is reserved to account future chain growth.
+pub const EXTRA_STORAGE_PROOF_SIZE: u32 = 1024;
```
svyatonik, line 40: Actually this is ~800 bytes for our almost empty chains. I was pretty sure that it is will be much lesser.

primitives/millau/src/lib.rs
svyatonik: Updated actual value (previous has been missing db writes and reads cost): 1_175_165_000

primitives/millau/src/lib.rs
svyatonik: Updated actual value (previous has been missing db writes and reads cost): `1_819_511_000`

primitives/millau/src/lib.rs
```diff
@@ -43,6 +43,14 @@ use serde::{Deserialize, Serialize};
 
 pub use millau_hash::MillauHash;
 
+/// Number of extra bytes (excluding size of storage value itself) of storage proof, built at
+/// Millau chain. This mostly depends on number of entries (and their density) in the storage trie.
+/// Some reserve is reserved to account future chain growth.
+pub const EXTRA_STORAGE_PROOF_SIZE: u32 = 1024;
+
+/// Maximal size of encoded (using `Encode::encode()`) account id.
```
HCastano, line 51: Is the size here in bytes? If so, we should add the units to the comment

primitives/rialto/src/lib.rs
```diff
@@ -34,6 +34,14 @@ use sp_runtime::{
...
+
+/// Maximal size of encoded (using `Encode::encode()`) account id.
+pub const MAXIMAL_ENCODED_ACCOUNT_ID_SIZE: u32 = 32;
```
HCastano, line 43: Same comment about the units

primitives/millau/src/lib.rs
HCastano: Is there no way for us to get the consts directly from the weight calculation in `receive_messages_proof`?

If not, what's your process for manually updating these values - is it just running the runtime benchmarks? We should make sure it's documented in the code in case others need to update them.

HCastano (approved):
> 3. confirmation transaction portion - 2125001080 (0.21) source tokens;

This stands out to me. Do you know why the confirmation is so much more expensive than the delivery?

> Before deploying to Polkadot/Kusama we need to do benchmarks on reference machine - I believe this would allow us to make weights maybe x2 times lesser than current even keeping some reserves (I have a 5-6 years old laptop :). For now I think it is ok, but worth additional attention when integrating with Polkadot/Kusama.

Yeah, I guess we're merely just setting up the framework for calculating weights now and shouldn't be too worried about the numbers - the details for those will be ironed out when we run it with the reference machine.

Also, maybe it's time for a new laptop 😄?

primitives/millau/src/lib.rs
svyatonik: You can get actual values just by changing it to `0`. Thanks to [this test](https://github.com/paritytech/parity-bridges-common/pull/598/files#diff-b154f3d20c7411ee6c223dd0e365d823ad8d58afca82a5ae32375e120a4b9ffeR41) you'll get actual values. Then just round them up && put here.

svyatonik:
> This stands out to me. Do you know why the confirmation is so much more expensive than the delivery?

The main thing both transaction do is proof verification. Then delivery tx just updates single storage value + emits single event (note that it doesn't include message dispatch cost/weight). Confirmation tx pay relayer reward - this is included in this weight/cost. I suppose that's the reason.

tomusdrw:
> The delivery and dispatch fee has been built from:
1. delivery transaction portion - 1125001048 (0.11) target tokens;
2. dispatch portion - 1973000 (0.0001973) target tokens;
3. confirmation transaction portion - 2125001080 (0.21) source tokens;
4. relayer fee 0.04 source tokens;
5. the rest is send message portion.

Delivery & confirmation transactions indeed seem quite expensive, do you know what contributes to these costs the most? Is it on-chain computation (proof verification?) or just the size of the proof?
Might be good to somehow profile the execution to see where can we benefit the most from optimisations (separate issue obviously).

primitives/millau/src/lib.rs
HCastano: Thanks. I still think that we should have that written in the code somewhere though

svyatonik: Well, it was just an optimal way for me - probably not the smartest/fastest way. Other may just take weight values and compute result itself, as comment already states. Imo that's enough. Other hints/alternaties on how to do that imo better to leave for some `how-to-plug-in-message-lane-module-into-your-runtime.md` document, which we may create once we'll actually plug it in somewhere,

svyatonik:
> Delivery & confirmation transactions indeed seem quite expensive, do you know what contributes to these costs the most? Is it on-chain computation (proof verification?) or just the size of the proof?

I doubt it is size - I guess it's just verification.

> Might be good to somehow profile the execution to see where can we benefit the most from optimisations (separate issue obviously).

I'll make benchmarks in #670

Issue #561, opened 2020-12-04
Pay dispatch fee on the target chain.


tomusdrw:
Currently we pay dispatch fee on the source chain and that heavily relies on the price oracle (#540, #449) enforcing the minimal required fee.

To reduce that reliance we can consider paying the dispatch fee on the target chain. I think this should be an option though, cause it requires having some funds on the target account id, prior to any interactions.

The idea would be that on the source chain, you still pay for the delivery fee (includes the cost/weight of encoded call), but when we dispatch the call on the target chain we first try to pay the fee to the relayer's account and only dispatch if there is enough funds there.

Since the weight cost of dispatch would most likely be the largest of all other fees, this reduced the price oracle requirement.

This complicates the relayer strategy a bit and to make sure they don't really loose in case there is not enough funds on the target side to dispatch the call, the target fee the relayer pays should be reduced by unused (but declared) weight.
Hence the relayer should be profitable regardless of the fact if the target fee was payed or not, cause the source chain fee should always cover the delivery and the reward even in case of failed dispatch.

svyatonik:
While working on this, I've found a problem with this approach. So I agree that when there's insufficient funds at the target chain to pay for the dispatch, we shall return funds to the relayer (that's why I was going to solve this issue and #377 in a single PR). But then the following scenario is possible:
1) sender wants to send message `{ encoded_call: vec![], dispatch_weight: MAX_WEIGHT }` and chooses to pay dispatch fee at the target chain;
2) at the source chain sender only pays for delivery and confirmation transactions. Cost of the first transaction then only depends on the message size. The cost of the second tx is fixed. And since we are delivering empty vec, the total cost isn't super-expensive;
3) when relayer submits delivery transaction, it declares its weight. And since the sender has declared it to be `MAX_WEIGHT`, we're blocking other transactions from being included into the block;
4) when trying to dispatch the call, we find that there's no funds at the dispatch origin account. We have two options here:
4.1) refund relayer, so that it won't lose anything. But then noone is actually paying for this `MAX_WEIGHT`. So we'll get almost-full blocks at the target chain without any utility. This: (a) blocks/slows down 'mining' of other transactions at the target chain and (b) increases `pallet_transaction_payment::Config::FeeMultiplierUpdate`, which makes transactions more expensive;
[...]

PR #599, opened 2020-12-23
Reward delivery confirmation transaction submitter (aka confirmation relayer)

### LINKED ISSUE #597: Change instant payment reward scheme to reward relayer that submit confirmation transaction

svyatonik:
closes #597 (see description for answer to question "why it is done?")

I've also changed `MessageDeliveryAndDispatchPayment` a bit - instead of having fn that simply transfers value from relayers fund to relayer, new function now accepts all payments that need to be completed. So it may deduct confirmation fees from other relayer rewards.

The only caveat is case when confirmation transaction cost has changed since the moment when message has been submitted. Since we do not store different components of `message_delivery_and_dispatch_fee`, the confirmation relayer is rewarded with current cost of confirmation transaction. So if some time ago cost has been `10` and whole message fee was `100` and now the cost has increased to `1000`, then the confirmation relayer would receive the whole message fee (`100`). The message relayer won't get anything (because `100.saturating_sub(1000) == 0`). This isn't super-important imo, because confirmation transactions are designed to be submitted by message relayers soon after they have delivered messages.

Another benefit from this PR is that if there were 3 unrewarded relayers entries, of which 1st and 3rd were for the same relayer, we will make only two transfers, compared to 3 transfers before this PR.

P.S.: benchmarks and confirmation transaction cost must be changed a bit, but this conflicts with #596, so I'm going to do that in next PR


modules/message-lane/src/instant_payments.rs
```diff
@@ -55,34 +65,252 @@ where
...
+	confirmation_relayer: &AccountId,
+	relayers_rewards: RelayersRewards<AccountId, Balance>,
+	relayer_fund_account: &AccountId,
+	confirmation_fee: Balance,
+) where
+	AccountId: Debug + Default + Encode + PartialEq,
+	Balance: Debug + AtLeast32BitUnsigned + From<MessageNonce> + Copy,
+	Currency: InstantCurrency<AccountId, Balance>,
+{
+	// reward every relayer, excepting `confirmation_relayer`
```
HCastano, line 113:
```suggestion
	// reward every relayer except `confirmation_relayer`
```

primitives/message-lane/src/source_chain.rs
```diff
@@ -16,14 +16,26 @@
...
-use frame_support::Parameter;
-use sp_std::fmt::Debug;
+use frame_support::{Parameter, RuntimeDebug};
+use sp_std::{collections::btree_map::BTreeMap, fmt::Debug};
 
 /// The sender of the message on the source chain.
 pub type Sender<AccountId> = frame_system::RawOrigin<AccountId>;
 
+/// Relayers rewards, grouped by relayer account id.
+pub type RelayersRewards<AccountId, Balance> = BTreeMap<AccountId, RelayerRewards<Balance>>;
```
HCastano, line 28: It's kinda confusing that the `RelayersRewards` typedef and `RelayerRewards` struct are name _almost_ the same. Maybe we can rename `RelayerRewards` to just `Reward`?

modules/message-lane/src/instant_payments.rs
```diff
@@ -55,34 +65,252 @@ where
...
-				reward,
-			),
-			Err(error) => frame_support::debug::trace!(
-				target: "runtime",
-				"Failed to pay relayer {:?} reward {:?}: {:?}",
-				relayer,
-				reward,
-				error,
-			),
+	// we can't actually do anything here, because rewards are paid as a part of unrelated transaction
```
HCastano, line 166: What do you mean by "we can't actually do anything here"?

modules/message-lane/src/instant_payments.rs
```diff
@@ -55,34 +65,252 @@ where
...
+			*balances.entry(*source).or_default() -= value;
+			*balances.entry(*dest).or_default() += value;
+			Ok(())
 		}
 	}
+
+	#[test]
+	fn confirmation_relayer_is_rewarded_if_it_has_also_delivered_messages() {
+		let mut currency = TestCurrency::new();
+		pay_relayers_rewards(
```
HCastano, line 219: I found myself flipping back and forth between the tests and code bit. Some intermediate variables with the relayer_ids and fees for example would help in making the test a bit more legible in a "standalone" way

modules/message-lane/src/instant_payments.rs
```diff
@@ -55,34 +65,252 @@ where
...
+		confirmation_relayer: &AccountId,
+		relayers_rewards: RelayersRewards<AccountId, Currency::Balance>,
 		relayer_fund_account: &AccountId,
 	) {
-		let pay_result = Currency::transfer(
-			&relayer_fund_account,
-			relayer,
-			*reward,
-			ExistenceRequirement::AllowDeath,
+		pay_relayers_rewards(
```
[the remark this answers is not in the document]
svyatonik: This `InstantCurrency` trait is just for tests simplicity. All currency methods are static, so without `&mut self` you have to use thread locals ot something like that, which I prefer to avoid.

modules/message-lane/src/instant_payments.rs
svyatonik: We can't handle this error e.g. by failing whole transaction. So the only thing that is left is to log it.

modules/message-lane/src/instant_payments.rs
```diff
@@ -55,34 +65,252 @@ where
...
+	}
+}
+
+/// Shortened currency trait used from instant payments.
+trait InstantCurrency<AccountId, Balance> {
+	/// Transfer some free balance between accounts.
+	fn transfer(&mut self, source: &AccountId, dest: &AccountId, value: Balance) -> DispatchResult;
+}
+
+/// Currency adapter.
```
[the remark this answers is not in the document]
svyatonik: It implements `InstantCurrency` for `Currency` (see my other comment about `InstantCurrency`)

primitives/message-lane/src/source_chain.rs
svyatonik: I don't like `Reward` :) Imo using `item` and `items` for container of `item`-s is fine, even though they're almost the same. I could change e.g. `RelayersRewards` to `RelayersRewardsMap` if you feel it is better.

modules/message-lane/src/instant_payments.rs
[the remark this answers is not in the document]
tomusdrw: Can't we just blanket-impl `InstantCurrency` for any type that implements `Currency`? It seems like it's yet another abstraction and it's already quite complicated.

I do understand that the goal is to simplify writing tests, so that we don't have to set up the entire runtime, but at this point I'm really unsure if simplified test setup is worth this complexity. We already have mock runtime for message lane tests and it could simply just include balances module, which at least would make the test more end to end.

primitives/message-lane/src/source_chain.rs
```diff
@@ -16,14 +16,26 @@
...
+
+/// Single relayer rewards.
```
tomusdrw, line 30:
```suggestion
/// Single relayer reward.
```

primitives/message-lane/src/source_chain.rs
```diff
@@ -16,14 +16,26 @@
...
+#[derive(RuntimeDebug, Default)]
+pub struct RelayerRewards<Balance> {
```
tomusdrw, line 32:
```suggestion
pub struct RelayerReward<Balance> {
```

primitives/message-lane/src/source_chain.rs
```diff
@@ -16,14 +16,26 @@
...
+	/// Total rewards that are needs to be paid to the relayer.
```
tomusdrw, line 33:
```suggestion
	/// Total reward that needs to be paid to the relayer.
```

tomusdrw (commented):
lgtm, except for the `InstantCurrency` abstraction which I'm not sure is worth the extra non-test code it introduces.

modules/message-lane/src/instant_payments.rs
HCastano: Can you add that clarification as part of the comment?

primitives/message-lane/src/source_chain.rs
HCastano: I don't think adding the type to the name is a good idea either. What if we decide in the future that this should be a `Vec<Reward>` instead? Unfortunately I can't think of any good suggestions other than this:

```
pub type RelayerRewards<AccountId, Balance> = BTreeMap<AccountId, Reward<Balance>>;
```

adoerr:
I'm starting to look into #561. Would it make sense, to have this (#599) PR merged before starting to implement #561? Or in other words, do we plan to merge this PR anytime soon?

tomusdrw:
> Or in other words, do we plan to merge this PR anytime soon?

Yup! @svyatonik there are few unresolved discussions (some docs/naming grumbles) and it's not mergeable due to conflicts, but apart from that it looks pretty good to me.

svyatonik:
Fixed conflicts + removed "confusing" comment. Re renaming suggestions - I don't see any improvements over existing code. To me both options are worse than original.

modules/message-lane/src/instant_payments.rs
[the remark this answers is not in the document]
tomusdrw: @svyatonik could you comment on the above suggestions? Either blanket-impl or using actual mock runtime?

primitives/message-lane/src/source_chain.rs
tomusdrw: @svyatonik please address at least this one, "that are needs" is plain incorrect.

modules/message-lane/src/instant_payments.rs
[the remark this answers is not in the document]
svyatonik: https://github.com/paritytech/parity-bridges-common/pull/599/commits/c53794b7455af2c1d8efa3bf1814db9ac61fd458

primitives/message-lane/src/source_chain.rs
tomusdrw: I just understood that you meant "Rewards of a single relayer" not "Relayer's single rewards". It is confusing me, but given you seem to be actively opposing changing this I'd rather merge as-is.

tomusdrw (approved):

PR #603, opened 2020-12-28
`send_message` weight now depends on message size

svyatonik:
Initially I was thinking that the `weight(send_message(vec![0; max_message_size])) - weight(send_message(vec![]))` will be near zero, but apparently it is not. So I've changed weight formula of `send_message` a bit - now it depends on message size.


modules/message-lane/src/weights_ext.rs
```diff
@@ -36,6 +37,18 @@ pub fn ensure_weights_are_correct<W: WeightInfoExt>() {
...
 /// Extended weight info.
 pub trait WeightInfoExt: WeightInfo {
+	/// Returns weight of message send transaction (`send_message`).
+	fn send_message_overhead() -> Weight {
+		Self::send_minimal_message_worst_case()
+	}
+
+	/// Returns weight that needs to be accounted when message of given size is sent (`send_message`).
+	fn send_message_size_overhead(message_size: u32) -> Weight {
+		let message_size_in_kb = (1024u64 + message_size as u64) / 1024;
```
HCastano, line 47: Why do we add a kilobyte here? Shouldn't this just be `message_size as u64 / 1024`?

modules/message-lane/src/benchmarking.rs
```diff
@@ -118,7 +118,81 @@ benchmarks_instance! {
...
+	}
+
+	// Benchmark `send_message` extrinsic with the worst possible conditions:
+	// * outbound lane already has state, so it needs to be read and decoded;
+	// * relayers fund account does not exists (in practice it needs to exist in production environment);
+	// * maximal number of messages is being pruned during the call;
+	// * message size is 1KB.
+	//
+	// With single KB of message size, the weight of the call is increased (roughly) by
+	// `(send_16_kb_message_worst_case - send_1_kb_message_worst_case) / 16`.
```
HCastano, line 139:
```suggestion
	// `(send_16_kb_message_worst_case - send_1_kb_message_worst_case) / 15`.
```

Shouldn't this be 15?

modules/message-lane/src/benchmarking.rs
```diff
@@ -118,7 +118,81 @@ benchmarks_instance! {
...
+	}
+
+	// Benchmark `send_message` extrinsic with the worst possible conditions:
+	// * outbound lane already has state, so it needs to be read and decoded;
+	// * relayers fund account does not exists (in practice it needs to exist in production environment);
+	// * maximal number of messages is being pruned during the call;
+	// * message size is 16KB.
+	//
+	// With single KB of message size, the weight of the call is increased (roughly) by
+	// `(send_16_kb_message_worst_case - send_1_kb_message_worst_case) / 16`.
```
HCastano, line 176:
```suggestion
	// `(send_16_kb_message_worst_case - send_1_kb_message_worst_case) / 15`.
```

Here as well I think

HCastano (approved):

modules/message-lane/src/weights_ext.rs
tomusdrw: We want to take a `ceil` of that divison, I guess we could in theory just add `1023` to be super precise, cause currently `1024 bytes` payload will be counted as `2kb`.

tomusdrw (approved):
LGTM, unless we want to fix the KB rounding (adding 1023 instead of 1024)

svyatonik:
The idea is to map message size to 1KB intervals - i.e. `[0; 1024) => interval1, [1024, 2048) => interval2, ...`. I don't see any benefit in mapping `1024` to interval1 vs not charging payment for empty message.

Issue #161, opened 2020-07-02
Relay can't detect if node is offline


svyatonik:
Part1: It looks like after we have migrated to shared client, we also lost ability to receive transport-related RPC errors. I.e. if you start relay with both nodes online, but then stop and start some node again, relay will hang forever, because transport errors are lost somewhere in background threads.

Part2 (depends on way how we resolve part1): if stopping node would spoil `jsonrpsee::Client` forever, we'll need to create new instance. I've started implementing that before realizing that all transport errors are lost. Intermediate results are [in this branch](https://github.com/paritytech/parity-bridges-common/tree/reconnect-on-transport-error). I'll leave it there for now.

svyatonik:
related #281 #288

tomusdrw:
@svyatonik is this still an issue?

svyatonik:
Yes, the 2nd part

PR #605, opened 2020-12-29
pause relays(s) when node is syncing

### LINKED ISSUE #531: Relay shall not submit any transaction if at least one of nodes it is connected to, is out of sync
In test environment our nodes are always synced. But in reality we may have situation e.g. in messages relay when target node is syncing, so we may see large (obsolete) difference in nonces. Like: on source node we're at the block 1000/1000 and best nonce is 100. On target node we're at the block 100/1000 and best received nonce is 10. Under some additional circumstances relayer may decide to deliver these missing 900 messages, even though then may have been already delivered. The result is potentially large number of rejected transactions and loss of funds.

Some mechanism is required that will pause relay loops (headers/messages) when at least one node is out of sync. This can be relay guard (which would pause executor instead of halting process), or simply another condition in loop(s).

svyatonik:
closes #531
needs #161 to be fixed first => draft


tomusdrw:
@svyatonik since #161 is closed, is this ready to be merged now?

tomusdrw (approved):
lgtm!

relays/ethereum-client/src/client.rs
```diff
@@ -53,6 +56,23 @@ impl Client {
...
+}
+
+impl Client {
+	/// Returns true if client is connected to at least one peer and is in synced state.
+	pub async fn ensure_synced(&self) -> Result<()> {
+		match Ethereum::syncing(&self.client).await? {
+			SyncState::NotSyncing => Ok(()),
+			SyncState::Syncing(syncing) => {
+				let missing_headers = syncing.highest_block.saturating_sub(syncing.current_block);
+				if missing_headers > MAJOR_SYNC_BLOCKS.into() {
```
tomusdrw, line 68: There is probably a status call for ETH node as well, but it's not really worth implementing given we are not actively working on the PoA bridge.

svyatonik:
> @svyatonik since #161 is closed, is this ready to be merged now?

No, it still needs to be tested

svyatonik:
Testing wasn't easy (basically because you either need patched clients or block connection between several nodes, leaving connections to other nodes intact), but I managed to confirm that solution works at least for sub2sub headers sync. So it should be ready now

PR #624, opened 2021-01-06
Relayer rewards improvements

### LINKED ISSUE #559: Provide account id to charge fees when Root|None is sending messages over the bridge.

### LINKED ISSUE #618: Relayer fund account and relayer account must exist
Both accounts must have balance over Existential Deposit before we even start the bridge. Document this properly and add an Integrity Check to the pallet, to make sure it's the case.

tomusdrw:
Closes #618
Closes #559 

1. Adds documentation on account existence requirements for `InstantCurrencyPayment` fee handler.
2. Introduces `on_runtime_upgrade` which delegates to the payment handler and will panic if `relayer_fund_account_id()` does not exist, to enforce this requirement.
3. Add an option to configure root account that should be charged fees when Root/None origin is being used.

### DIFF
--- a/modules/message-lane/src/instant_payments.rs
+++ b/modules/message-lane/src/instant_payments.rs
@@ -15,24 +15,35 @@
 // along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
 
 //! Implementation of `MessageDeliveryAndDispatchPayment` trait on top of `Currency` trait.
-//! All payments are instant.
+//!
+//! The payment is first transferred to a special `relayers-fund` account and only transferred
+//! to the actual relayer in case confirmation is received.
 
 use bp_message_lane::source_chain::{MessageDeliveryAndDispatchPayment, Sender};
 use codec::Encode;
-use frame_support::traits::{Currency as CurrencyT, ExistenceRequirement};
+use frame_support::traits::{Currency as CurrencyT, ExistenceRequirement, Get};
 use sp_std::fmt::Debug;
 
-/// Instant message payments made in given currency. Until claimed, fee is stored in special
-/// 'relayers-fund' account.
-pub struct InstantCurrencyPayments<AccountId, Currency> {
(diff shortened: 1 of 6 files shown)


modules/message-lane/src/lib.rs
```diff
@@ -538,6 +546,16 @@ impl<T: Config<I>, I: Instance> Module<T, I> {
 			total_messages: total_unrewarded_messages(&relayers),
 		}
 	}
+
+	/// AccountId of the shared relayer fund account.
+	///
+	/// This account stores all the fees paid by submitters. Relayers are able to claim these
+	/// funds as at their convenience.
```
adoerr, line 553: ~~as~~

modules/message-lane/src/instant_payments.rs
```diff
@@ -15,57 +15,80 @@
...
+			// it's fine for the submitter to go below Existential Deposit and die.
+			ExistenceRequirement::AllowDeath,
+		)
+		.map_err(Into::into)
 	}
 
 	fn pay_relayer_reward(
-		_confirmation_relayer: &AccountId,
-		relayer: &AccountId,
+		_confirmation_relayer: &T::AccountId,
```
adoerr, line 81: Could this parameter be removed from `MessageDeliveryAndDispatchPayment::pay_relayer_reward()`? AFAICT it as actually never used in any implementation of the trait. Or is this a placeholder for some future improvement?

adoerr (approved):
LGTM, from my currently limited perspective. Just some typo and a question.

modules/message-lane/src/instant_payments.rs
tomusdrw: Yes, it's a placeholder, see #599. Or rather this one is a super simple implementation of the reward payment trait and we know that more sophisticated ones (which #599 implements) will need this.

HCastano (approved):

Issue #628, opened 2021-01-07
Finality Verifier Pallet


HCastano:
The current Substrate header sync pallet is susceptible to several different attacks. These attacks involve being able to write to storage indefinitely (like with the case of #454), or requiring potentially unbounded iteration in order to make progress (like in the case of #367).

To works around these issues the following is being proposed. Instead of accepting headers and finality proofs directly into the existing pallet we should instead go through a "middleware" pallet. The role of this new pallet will be to verify finality and ancestry proofs for incoming headers, and if successful write the headers to the base Substrate pallet.

This essentially flips the flow of the current pallet, in which we first import headers and later verify if they're actually part of the canonical chain (through the finality proof verification). By verifying finality first and writing headers later we can ensure that we only write "known good" headers to the base pallet.

Additionally, by keeping the base pallet as is (more or less), we are able to ensure that applications building on the bridge can continue to have access to all headers from the canonical chain .

The proposed API for the pallet is as follows:

```rust
// Ancestry proof should container headers from (`last_finalized_header`, `finalized_header`]
fn submit_finality_proof(justification: GrandpaJustification, ancestry_proof: Vec<Header>);
```

If this call is successful, the headers from the ancestry proof would be written to the base pallet.

Optionally, if the proofs don't fit into a single block, we can also have a "chunk" based API to allow relayers to submit proofs across several blocks:

```rust
// The `merkle_root` would be the root of a trie containing all the headers you plan to submit
fn signal_long_proof(deposit: T::Balance, finality_proof: GrandpaJustification, merkle_root: Vec<u8>);

// The chunk submitted here would be the leaves of the merkle trie
fn submit_chunk(ancestry_chunk: Vec<Header>);
```

After a call to `signal_long_proof()` you'd be able to call `submit_chunk()` as many times as required to complete the proof. We may want to restrict others from submitting proofs during this period, and add some timeouts in case we don't receive a chunk, or if we don't complete a proof, in a certain amount of time. In the case where we timeout, or the proof is invalid, the submitter will lose their `deposit`.

adoerr:
Could `submit_chunk()` be abused for some sort of DoS attack? Basically calling `submit_chunk()` without and prior `signal_long_proff()` forcing the pallet to repeatedly engage in some sort of  costly housekeeping activities.

HCastano:
[...]

PR #629, opened 2021-01-07
Finality Verifier Pallet

HCastano:
Implements the happy path code of #628. The assumption with the happy path code is that we'll be able to verify the given justification and ancestry proof efficiently in a single transaction.

Some things to keep in mind while going through this PR:

1. While the ancestry checker is configurable (since we hope to someday use MMRs for example), the proof type is not. I decided to hardcode the type as a `Vec` so that I could be more specific in some checks (like when checking the upper bound for the ancestry proof length). While I think this is fine for now, let me know if you guys can think of a case in which it's not.

2. I'm not entirely sure how to test the transaction rollback code. Any ideas here would be helpful.

3. The choice to make `type MaxHeadersInSingleProof` be of type `u8` was pretty arbitrary - although I expect that if we have to verify more than 255 headers we're already well past the amount of computation we can do in a single transaction.


modules/substrate/src/lib.rs
```diff
@@ -361,6 +362,38 @@ impl<T: Config> Module<T> {
...
+		Ok(())
+	}
+
+	fn import_finality_proof(header: BridgedHeader<T>, finality_proof: Vec<u8>) -> Result<(), ()> {
+		let mut verifier = verifier::Verifier {
+			storage: PalletStorage::<T>::new(),
+		};
+
+		let _ = verifier
+			.import_finality_proof(header.hash(), finality_proof.into())
```
svyatonik, line 390: Just to be sure that it is known - this fn verifies justification, but you have already did that in `submit_finality_proof`. Is it left for future PRs?

primitives/header-chain/src/lib.rs
```diff
@@ -32,8 +38,25 @@ pub mod justification;
...
+
+impl AuthoritySet {
+	/// Create a new GRANDPA Authority Set.
+	pub fn new(authorities: AuthorityList, set_id: SetId) -> Self {
+		Self { authorities, set_id }
+	}
+}
+
+/// base trait for verifying transaction inclusion proofs.
+pub trait TransactionVerifier {
```
svyatonik, line 59: Are you sure it is a good name? :) I'd imagine that `TransactionVerifier` is verifying transactions, not transaction inclusion proofs

svyatonik (approved):

modules/substrate/src/lib.rs
HCastano: Yeah I know it's being verified twice 😅

I would like to change this in the future so that we could import the headers and finality proofs directly without additional verification

primitives/header-chain/src/lib.rs
HCastano: It seemed like an okay name - but I'm open to changing it to something better. Maybe `TransactionProofVerifier`?

PR #631, opened 2021-01-08
Fix Dependency Image Publishing

HCastano:
Publishing is failing because of an empty `tag`. This uses automatic tagging to fix that.

Images are being published [here](https://hub.docker.com/r/paritytech/bridge-dependencies/tags?page=1&ordering=last_updated).

### DIFF
--- a/.github/workflows/publish-deps.yml
+++ b/.github/workflows/publish-deps.yml
@@ -26,13 +26,6 @@ jobs:
         with:
           fetch-depth:                              5
           submodules:                               recursive
-      - name:                                       Set vars
-        id:                                         vars
-        run:                                        |
-          if [[ ${GITHUB_REF} = refs/tags/* ]]
-          then
-            echo ::set-output name=TAG::${GITHUB_REF#refs/tags/}
-          fi
       - name:                                       Build and push dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-deps.yml
```diff
@@ -26,29 +26,21 @@ jobs:
...
-          then
-            echo ::set-output name=TAG::${GITHUB_REF#refs/tags/}
-          fi
       - name:                                       Build and push dependencies
         uses:                                       docker/build-push-action@v1
         with:
           username:                                 ${{ secrets.DOCKER_USER }}
           password:                                 ${{ secrets.DOCKER_PASSWORD }}
-          repository:                               paritytech/parity-bridges-common
+          repository:                               paritytech/bridge-dependencies
```
adoerr, line 34: Is this supposed to be up on Docker Hub already?

adoerr (approved):
LGTM

.github/workflows/publish-deps.yml
HCastano: The `paritytech/parity-bridges-common` image was not (I forgot to change the name). However, the `paritytech/bridge-dependencies` is since I published it to make sure that this Action was working correctly

PR #638, opened 2021-01-11
Remove redundant data from InboundLaneData

### LINKED ISSUE #616: InboundLaneData might contain redundant data
It seems that `latest_*_nonce` fields are not really required in `InboundLaneData`, cause this information can be extracted from `relayers` vector (min and max nonce from that vector).

tomusdrw:
Resolves #616 

We still need to store one of the nonces in case there is no pending `relayers` in the array.

I chose to store the last confirmed nonce (i.e. the one that target chain knows is already confirmed on the source chain) and compute last delivered nonce based on the last element of `relayers` vector.

I also decided to rename the variables a tiny bit, to distinguish them from `OutboundLane`'s parameters. Previously you had to know about the context (i.e. Inbound/Outbound) which made it a bit hard for me to reason about, now it's clear from just the variable naming.


modules/message-lane/src/inbound_lane.rs
```diff
@@ -55,29 +55,32 @@ impl<S: InboundLaneStorage> InboundLane<S> {
 	/// Receive state of the corresponding outbound lane.
 	pub fn receive_state_update(&mut self, outbound_lane_data: OutboundLaneData) -> Option<MessageNonce> {
 		let mut data = self.storage.data();
-		if outbound_lane_data.latest_received_nonce > data.latest_received_nonce {
+		let last_received_nonce = data.last_delivered_nonce();
+
+		if outbound_lane_data.latest_received_nonce > last_received_nonce {
```
adoerr, line 60: Shouldn't this be `last_received_noce` for consistency reasons?

modules/message-lane/src/inbound_lane.rs
```diff
@@ -55,29 +55,32 @@ impl<S: InboundLaneStorage> InboundLane<S> {
...
 			// this is something that should never happen if proofs are correct
 			return None;
 		}
-		if outbound_lane_data.latest_received_nonce <= data.latest_confirmed_nonce {
+		if outbound_lane_data.latest_received_nonce <= data.last_confirmed_nonce {
```
adoerr, line 64: Same here

primitives/message-lane/src/lib.rs
```diff
@@ -74,34 +74,48 @@ pub struct Message<Fee> {
...
-	/// Multiple dispatches from the same relayer one are allowed.
+	/// Multiple dispatches from the same relayer are allowed.
 	pub relayers: VecDeque<(MessageNonce, MessageNonce, RelayerId)>,
-	/// Nonce of latest message that we have received from bridged chain.
-	pub latest_received_nonce: MessageNonce,
-	/// Nonce of latest message that has been confirmed to the bridged chain.
-	pub latest_confirmed_nonce: MessageNonce,
+
+	/// Nonce of the last message which delivery to this (target) chain has been confirmed back
+	/// on the source chain.
```
adoerr, line 96: Perhaps: "Nonce of the last source chain confirmed message, which has been delivered to this (target) chain."

adoerr (approved):
Minor grumbles, other than that, LGTM.

HCastano (approved):

modules/message-lane/src/inbound_lane.rs
tomusdrw: I've changed the variable name to `last_delivered_nonce` for consistency, is that what you meant? I don't want to change the `OutboundLane` names, because of the reasons I gave in the description.

modules/message-lane/src/inbound_lane.rs
tomusdrw: That's outbound lane data, I decided to only change `InboundLaneData` to be able to distinguish variables without knowing the inbound/outbound context.

primitives/message-lane/src/lib.rs
tomusdrw: I'm not convinced it's better. In my sentence I've tried to emphasize the ordering of events as well:
1. Delivery to the target chain.
2. Delivery confirmation on the source chain.

With `source-chain-confirmed message` I feel this is lost, and it's not clear to me how that differs from `last_delivered_nonce()`.

modules/message-lane/src/inbound_lane.rs
adoerr: Yeah. As a non-native speaker I'm always a bit confused about 'latest' and 'last'. So, personally instead of 'latest', I would prefer something like 'newest' or 'current'. But that is admittedly a personal preference.

primitives/message-lane/src/lib.rs
adoerr: So how about, "Nonce of the last message that a) has been delivered to the target chain and b) for which there is a confimation on the source chain." I just felt that the original sentence reads a bit 'bumpy'.

tomusdrw: Re-phrased a bit and updated, let me know if the bumps are smaller now :)

adoerr: Smooth

PR #640, opened 2021-01-13
Move justification code to primitives crate

HCastano:
While working through #629 I realized that since I'd be checking justifications in both `pallet-substrate-bridge` and `pallet-finality-verifier` that it would be useful to share the code.

In order to work around rust-lang/cargo#2911 I've moved the justification test helpers into a new crate called `bp-test-helpers` and moved the justification tests into `header-chain/tests`. What this allows me to do is to have a fully `std` `dev-dependency` that I can use from different crates in order to verify justifications, while also allowing me to import the "main" justification code into `no_std` environments. The tests were moved into their own integration test crate in order to avoid circular dependencies between `bp-test-helpers` and `bp-header-chain`.


primitives/test-utils/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Utilities for testing runtime code.
+//!
+//! Unlike other crates in the `primitives` folder, this crate is does *not* need to compile in a
```
tomusdrw, line 19:
```suggestion
//! Unlike other crates in the `primitives` folder, this crate does *not* need to compile in a
```

tomusdrw (approved):

relays/substrate/Cargo.toml
```diff
@@ -18,6 +18,7 @@ structopt = "0.3"
 
 # Bridge dependencies
 
+bp-header-chain = { path = "../../primitives/header-chain" }
```
svyatonik, line 21: `bp-header-chain/std` is missing

svyatonik (approved):
Code changes lgtm in context of introducing separate pallet, but I have a couple of questions about where it all heads now. Why would you need this separate `finality-verifier` pallet at all (and this change particularly)? #628 says that there are problems with `pallet-substrate-bridge` && to work around these problems you're going to introduce additional pallet. What is going to happen with substrate-pallet calls? Will they still be accessible? I guess the answer is no - you can't leave this hole open, right? So I guess the current idea is to disable substrate pallet calls and expose finality-verifier pallet calls instead? But in this case - why have two pallets, one of which only provides storage and second only provides calls to update this storage?

HCastano:
> Code changes lgtm in context of introducing separate pallet, but I have a couple of questions about where it all heads now. Why would you need this separate `finality-verifier` pallet at all (and this change particularly)? #628 says that there are problems with `pallet-substrate-bridge` && to work around these problems you're going to introduce additional pallet. What is going to happen with substrate-pallet calls? Will they still be accessible? I guess the answer is no - you can't leave this hole open, right? So I guess the current idea is to disable substrate pallet calls and expose finality-verifier pallet calls instead? But in this case - why have two pallets, one of which only provides storage and second only provides calls to update this storage?

Responded to this in #628.

PR #641, opened 2021-01-13
Swap target <-> source in race logs

svyatonik:
Right now in Rialto -> Millau message relay logs I see:
```
[1;38;5;8m2021-01-13 03:02:20 +0000[0m [38;5;14mDEBUG[0m [38;5;8mbridge[0m Going to submit proof of messages in range 1..=2 to Millau::ReceivingConfirmationsDelivery node
```

It was supposed to be:
```
[1;38;5;8m2021-01-13 03:02:20 +0000[0m [38;5;14mDEBUG[0m [38;5;8mbridge[0m Going to submit proof of messages in range 1..=2 to Rialto::ReceivingConfirmationsDelivery node
```

Because proof is actually submitted to Rialto node.

### DIFF
--- a/relays/messages-relay/src/message_race_receiving.rs
+++ b/relays/messages-relay/src/message_race_receiving.rs
@@ -79,11 +79,11 @@ impl<P: MessageLane> MessageRace for ReceivingConfirmationsRace<P> {
 	type Proof = P::MessagesReceivingProof;
 
 	fn source_name() -> String {
-		format!("{}::ReceivingConfirmationsDelivery", P::SOURCE_NAME)
+		format!("{}::ReceivingConfirmationsDelivery", P::TARGET_NAME)
 	}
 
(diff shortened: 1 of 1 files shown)


tomusdrw (approved):
LGTM, but maybe we should simply swap `target -> source` usage in case of confirmations race? 

Like I think it would be cool to consistently assume that:
1. Message Delivery is `source -> target`
2. Message Delivery Confirmation is `target -> source`
3. Confirmation Delivery is `source -> target` again

I guess we currently consistently use `source` but assume a relayer PoV.

svyatonik:
Race is a programming concept, which has its own `source` and `target`. They may match message lane source/target or be different. So what happens here is exactly `we should simply swap target -> source usage in case of confirmations race` (c). The issue was that in logs we were using wrong names - e.g. `Rialto` instead of `Millau`.

PR #645, opened 2021-01-13
Add dump logs script.

tomusdrw:
This is pretty deployment-specific, so I wonder if there is a better place to put it.


HCastano (approved):
The `scripts` folder is totally fine imo. The stuff there is just a mishmash of somewhat random, but useful scripts.

If we end up with a lot of scripts specific to deployments, _then_ we should move stuff elsewhere

Issue #419, opened 2020-10-13
Prune Headers in Substrate Pallet


HCastano:
Splitting out of #369.

For storage reasons we don't want to be keeping old headers around. We should have a way to prune these headers from storage which is configurable (e.g choose how many headers you want to keep at any point in time, or how many to remove in one go). While pruning headers on the non-canonical chain is important, we should also be able to prune old headers on the canonical chain. Again, this is for storage reasons.

tomusdrw:
I didn't have time to finish this up, might be worth re-visiting now with Finality Verifier pallet being in.

PR #646, opened 2021-01-13
Headers pruning via dispatchable

tomusdrw:
Related #419 

This PR doesn't completely solve the storage issue, but allows the pallet to be cleaned up by sending calls to the pallet.

Still work in progress, since we need to:
- [ ] Make minimal number of finalized headers configurable or use time-based strategy
- [ ] Write a lot of tests.
- [ ] Add hooks (for instance for rewards or some accounting)

But opening here for visibility.

I just got an idea to couple message delivery rewards with pruning to incentivise relayers to actually clean up the pallet. Whenever the relayer wants to claim their rewards, they first need to prune some minimal number of headers. It will require a lot of tweaking and benchmarking but together with #628  we would only have canonical chain blocks to be pruned, which simplifies it a bit.


HCastano:
After the changes introduced in https://github.com/paritytech/parity-bridges-common/pull/707 we can simplify this a bit since we'll only need to prune finalized headers.

HCastano:
Given the changes with how our bridge pallet works (we now only import finality proofs instead of full header chains) the approach proposed in this PR has gone stale.

If we still want some sort of pruning mechanism we need a different approach, and thus a new PR.

PR #650, opened 2021-01-14
Improve Substrate pallet logs

svyatonik:


modules/substrate/src/lib.rs
```diff
@@ -172,15 +172,21 @@ decl_module! {
 		) -> DispatchResult {
 			ensure_operational::<T>()?;
 			let _ = ensure_signed(origin)?;
-			frame_support::debug::trace!("Got header {:?}", header);
+			let hash = header.hash();
+			frame_support::debug::trace!("Going to import header {:?}: {:?}", hash, header);
```
HCastano, line 176: I don't remember what the logs look like, what info do the `"{:?}," header` logs include?

HCastano (approved):

modules/substrate/src/lib.rs
svyatonik: For our chains, it is regular auto-derived `Debug` implementation (i.e. it shows all struct fields).

PR #652, opened 2021-01-15
Pin Rust Nightly to 2020-12-17

HCastano:
Clippy is panicking with the current nightly version that the CI is using. This rolls back our nightly to one which doesn't have this problem. I've commented about this issue [here](https://github.com/rust-lang/rust-clippy/issues/6592#issuecomment-761054533).

Note that I didn't rollback the compiler version for our Docker builds. I don't really see the point since this fix is mainly just to get the Clippy step of the CI passing.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -22,7 +22,7 @@ jobs:
         toolchain:
           - stable
           #- beta
-          - nightly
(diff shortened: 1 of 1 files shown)


HCastano:
Merging since there are some PRs waiting on this before they can be merged.

PR #653, opened 2021-01-16
Disallow Duplicate Best Headers

HCastano:
Fixes a bug in which we wrote the same header twice to the `BestHeaders` Vec.

What was happening was that we'd write the header once during import, and a second time if we finalized the header (we go through the same `write_header()` function). Since we're allowed to have multiple headers at the same height (they might be on different forks), we ended up with the same header duplicated in the `BestHeaders`.

### DIFF
--- a/modules/substrate/src/lib.rs
+++ b/modules/substrate/src/lib.rs
@@ -521,7 +521,12 @@ impl<T: Config> BridgeStorage for PalletStorage<T> {
 
 		match current_height.cmp(&best_height) {
 			Ordering::Equal => {
-				<BestHeaders<T>>::append(hash);
+				// Want to avoid duplicates in the case where we're writing a finalized header to
+				// storage which also happens to be at the best height the best height
+				let not_duplicate = !<ImportedHeaders<T>>::contains_key(hash);
+				if not_duplicate {
+					<BestHeaders<T>>::append(hash);
+				}
 			}
 			Ordering::Greater => {
 				<BestHeaders<T>>::kill();
--- a/modules/substrate/src/verifier.rs
+++ b/modules/substrate/src/verifier.rs
@@ -562,7 +562,7 @@ mod tests {
 
 			// It should be fine to import both
 			assert_ok!(verifier.import_header(best_header.hash(), best_header.clone()));
-			assert_ok!(verifier.import_header(also_best_header.hash(), also_best_header.clone()));
+			assert_ok!(verifier.import_header(also_best_header.hash(), also_best_header));
 
 			// The headers we manually imported should have been marked as the best
 			// upon writing to storage. Let's confirm that.
--- a/modules/substrate/src/verifier.rs
+++ b/modules/substrate/src/verifier.rs
@@ -24,6 +24,7 @@
 
 use crate::storage::{AuthoritySet, ImportedHeader, ScheduledChange};
 use crate::BridgeStorage;
+
 use bp_header_chain::justification::verify_justification;
 use finality_grandpa::voter_set::VoterSet;
 use sp_finality_grandpa::{ConsensusLog, GRANDPA_ENGINE_ID};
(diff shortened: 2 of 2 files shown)


HCastano:
Looks like the CI runner is out of disk space.

modules/substrate/src/lib.rs
```diff
@@ -521,7 +521,12 @@ impl<T: Config> BridgeStorage for PalletStorage<T> {
 
 		match current_height.cmp(&best_height) {
 			Ordering::Equal => {
-				<BestHeaders<T>>::append(hash);
+				// Want to avoid duplicates in the case where we're writing a finalized header to
```
svyatonik, line 524: Speaking in context of new `finality-verifier` pallet - are there going to be any optimization in this pallet? Like:
1) this function (`write_header`) is called for every finalized header, but seems that this `match` can be called only for the best finalized header;
2) iiuc you may move `<ImportedHeaders<T>>::insert(hash, header);` before this match and use something like `let not_duplicate = <ImportedHeaders<T>>::mutate(hash, |storage_header| { let not_duplicate = storage_header.is_none(); *storage_header = Some(header); not_duplicate });` instead, to avoid calling both `contains_key` and `update`

svyatonik (approved):

modules/substrate/src/lib.rs
HCastano: > 1. this function (write_header) is called for every finalized header, but seems that this match can be called only for the best finalized header;

Yes, in the case of finalized headers the best one would be the only one to enter the `match`, and it would only be able to enter the`Ordering::Equal` match arm.

> 2. iiuc you may move <ImportedHeaders<T>>::insert(hash, header); before this match and use something like let not_duplicate = <ImportedHeaders<T>>::mutate(hash, |storage_header| { let not_duplicate = storage_header.is_none(); *storage_header = Some(header); not_duplicate }); instead, to avoid calling both contains_key and update

Instead of `update` here, are you referring to `append`?

I don't quite understand the improvement with the `mutate()` version though - although I think it might have to do with me not being super familiar with the `mutate()` mechanics. From what I can tell `mutate()` gets `storage_header` by doing a call to `get()` under the hood. Is that any better than checking `contains_key`? More code from you might be helpful here :)

svyatonik: 1. My question is more about whether we'll need optimizations in this pallet at all? Are you going to update pallet storage from another pallet directly, or use existing functions from this pallet?
2. The whole

svyatonik: 2. The whole
```
		let best_height = <BestHeight<T>>::get();

		match current_height.cmp(&best_height) {
			Ordering::Equal => {
				// Want to avoid duplicates in the case where we're writing a finalized header to
				// storage which also happens to be at the best height the best height
				let not_duplicate = !<ImportedHeaders<T>>::contains_key(hash);
				if not_duplicate {
					<BestHeaders<T>>::append(hash);
				}
			}
			Ordering::Greater => {
				<BestHeaders<T>>::kill();
				<BestHeaders<T>>::append(hash);
				<BestHeight<T>>::put(current_height);
			}
			Ordering::Less => {
				// This is fine. We can still have a valid header, but it might just be on a
				// different fork and at a lower height than the "best" overall header.
			}
		}
```
block may be moved out of the loop. I agree that it has single additional storage read in case of `finalize_headers`, though :) And this read is also probably cached by `OverlayChanges`, so probably not worth it.
3. I mean `insert`, not `append`. The idea is that you're currently reading and then updating this value. You're right that the value isn't decoded in case of `contains_key`, though. So I'm also not sure which option is better

HCastano: > My question is more about whether we'll need optimizations in this pallet at all? Are you going to update pallet storage from another pallet directly, or use existing functions from this pallet?

Yes, I think we'll need to optimize storage a bit in the bridge pallet. The finality pallet at the moment is writing to the bridge pallet using existing function, like this:

```rust
// In modules/finality-verifier/src/lib.rs

pub trait Config: frame_system::Config {
	type BridgedChain: Chain;
	type HeaderChain: HeaderChain<<Self::BridgedChain as Chain>::Header>;
        ...
}

decl_module! {
    fn submit_finality_proof() {
        ...
        T::HeaderChain::import_header(header);
        ...
    }
}
```

```rust
// In modules/substrate/src/lib.rs

impl<T: Config> bp_header_chain::HeaderChain<BridgedHeader<T>> for Module<T> {
    fn import_header(header: BridgedHeader<T>) -> Result<(), ()> {
	let mut verifier = verifier::Verifier {
            storage: PalletStorage::<T>::new(),
	};

	let _ = verifier.import_header(header).map_err(|_| ())?;

	Ok(())
    }
}
```

HCastano:
I'm gonna merge this. We can address storage optimizations at a later point in time.

PR #659, opened 2021-01-18
Fixed messages count check

svyatonik:
closes https://github.com/paritytech/srlabs_findings/issues/40 

The guaratee there is not that straightforward as in https://github.com/paritytech/parity-bridges-common/pull/658 - there's no direct link betweeen maximal tx weight and maximal vector capacity. But e.g. on our testnets the max number of messages is roughly 2 * weight_per_1s / single_message_overhead ~ 6900. This doesn't include other factors - like that max extrinsic weight is only 75% of 2s or that extrinsic weight has other components. So it should be fine both for testnets + real chains. In case if you think it is too fragile, I may resurrect parts of https://github.com/paritytech/parity-bridges-common/pull/541 - it had explicit limit of messages in the delivery transaction.

### DIFF
--- a/bin/runtime-common/src/messages.rs
+++ b/bin/runtime-common/src/messages.rs
@@ -396,6 +396,10 @@ pub mod target {
 	}
 
 	/// Verify proof of Bridged -> This chain messages.
+	///
+	/// The `messages_count` argument verification (sane limits) is supposed to be made
+	/// outside of this function. This function only verifies that the proof declares exactly
+	/// `messages_count` messages.
 	pub fn verify_messages_proof<B: MessageBridge, ThisRuntime>(
 		proof: FromBridgedChainMessagesProof<B>,
 		messages_count: MessageNonce,
@@ -497,11 +501,18 @@ pub mod target {
 		let (bridged_header_hash, bridged_storage_proof, lane_id, begin, end) = proof;
 
 		// receiving proofs where end < begin is ok (if proof includes outbound lane state)
-		// => hence unwrap_or(0)
-		let messages_in_the_proof = end.checked_sub(begin).and_then(|diff| diff.checked_add(1)).unwrap_or(0);
-		if messages_in_the_proof != messages_count {
-			return Err(MessageProofError::MessagesCountMismatch);
-		}
+		let messages_in_the_proof = if let Some(nonces_difference) = end.checked_sub(begin) {
+			// let's check that the user (relayer) has passed correct `messages_count`
+			// (this bounds maximal capacity of messages vec below)
+			let messages_in_the_proof = nonces_difference.saturating_add(1);
+			if messages_in_the_proof != messages_count {
+				return Err(MessageProofError::MessagesCountMismatch);
+			}
+
+			messages_in_the_proof
+		} else {
+			0
+		};
 
 		let parser = build_parser(bridged_header_hash, bridged_storage_proof)?;
 
@@ -509,7 +520,7 @@ pub mod target {
 		// be in the proof. So any error in `read_value`, or even missing value is fatal.
 		//
 		// Mind that we allow proofs with no messages if outbound lane state is proved.
-		let mut messages = Vec::with_capacity(end.saturating_sub(begin) as _);
+		let mut messages = Vec::with_capacity(messages_in_the_proof as _);
 		for nonce in begin..=end {
 			let message_key = MessageKey { lane_id, nonce };
 			let raw_message_data = parser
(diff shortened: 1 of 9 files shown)


bin/runtime-common/src/messages.rs
```diff
@@ -1183,4 +1194,56 @@ mod tests {
...
+					}),
+				}),
+			),
+			Err(target::MessageProofError::MessagesCountMismatch),
+		);
+	}
+
+	#[test]
+	#[should_panic]
+	fn verify_messages_proof_with_parser_panic_if_too_many_messages_declared() {
```
HCastano, line 1226: I'm a little confused here. If by declaring `u64::MAX` messages a user is able to trigger the same Vec allocation as before, doesn't the problem still exist?

adoerr, line 1226: I don't think so. The problem before was that `messages_in_the_proof` would have been calculated to `0`. This allowed to bypass the triggering of the `MessagesCountMismatch` error in case `message_count` was `0` as well. But then for the actual allocation of the vec the saturated value of `end - begin`, i.e. `u64::MAX` would have been used. 

Now you could still bypass the `MessageCountMismatch` error but in this case an allocation of size `0` would be done. This prevents the unbound allocation.

svyatonik: User won't be able to declare `u64::MAX`, because this code is callable via extrinsic and `messages_count` is a part of weight formula. See description for details

svyatonik: > IDK if it is possible to just return Ok(ProvenMessages::new()) right way in case the checked_sub() overflows instead of continuing with messages_in_the_proof == 0 which in the end should result in the same thing, i.e. no proven messages.

That would be wrong, because empty proof is not allowed (`MessageProofError::Empty`). So we can't simply return `Ok(ProvenMessages::new())` - instead we just skip "messages part" and go to the "outbound lane data" part of the function.

svyatonik:
Actually there's another way to make this check explicit. Since we only work with single lane in this fn, we may use `MaxUnconfirmedMessagesAtInboundLane` as maximal value of `messages_count` instead of relying on weight formula. Will add that check now.

bin/runtime-common/src/messages.rs
adoerr: So now the unbounded allocation is gone and the condition which would have triggered the unbounded allocation is not user exposed.

adoerr:
LGTM

svyatonik:
> Actually there's another way to make this check explicit. Since we only work with single lane in this fn, we may use MaxUnconfirmedMessagesAtInboundLane as maximal value of messages_count instead of relying on weight formula. Will add that check now.

I've added explicit check of `messages_count` in the `receive_messages_proof`. So now this value will never be larger than `min(maximal-value-that-cames-from-weight-formula, MaxUnconfirmedMessagesAtInboundLane)`. And `MaxUnconfirmedMessagesAtInboundLane` needs to fit in the memory, because we already have a `VecDeque` of this length.

bin/runtime-common/src/messages.rs
```diff
@@ -1183,4 +1194,56 @@ mod tests {
...
+						latest_generated_nonce: 1,
+					}),
+				}),
+			),
+			Err(target::MessageProofError::MessagesCountMismatch),
+		);
+	}
+
+	#[test]
+	#[should_panic]
```
tomusdrw, line 1225: I'm not a fan of `should_panic` tests, had to actually run the code without this to understand that the panic is coming from `Vec::with_capacity(u64::max_value())` call (note the behavior is in theory platform specific).

TBH I'd prefer to explicitly disallow allocations over `u32::max_value()` and just return a regular error here. I know we do verify the incoming values using `MaxMessagesAtInboundLane` in practice, but I'd still be more in favor of defensive programming style here.

I'd simply change `messages_count: MessageNonce` parameter to `u32` (I'm assuming `MessageNonce: From<u32>` is implemented anyway) to make it explicit (so it doesn't even require documentation that much).

tomusdrw (commented):
The fix looks good, few suggestions though.

primitives/message-lane/src/lib.rs
```diff
@@ -158,11 +158,36 @@ impl Default for OutboundLaneData {
...
 pub fn total_unrewarded_messages<RelayerId>(
 	relayers: &VecDeque<(MessageNonce, MessageNonce, RelayerId)>,
-) -> MessageNonce {
+) -> Option<MessageNonce> {
 	match (relayers.front(), relayers.back()) {
-		(Some((begin, _, _)), Some((_, end, _))) => end.checked_sub(*begin).and_then(|d| d.checked_add(1)).unwrap_or(0),
-		_ => 0,
+		(Some((begin, _, _)), Some((_, end, _))) => {
+			if let Some(difference) = end.checked_sub(*begin) {
+				difference.checked_add(1)
```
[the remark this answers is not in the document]
svyatonik: 1. The reason is that `(begin; end)` pair means inclusive range here. And to specify empty inclusive range you shall make `end` lesser than `begin` - e.g. `10..=9`. So this is never an error to a caller.
2. `saturated_add` would return wrong result, just dumping error here without any notifications to the caller.If this function will ever be called from the place where correct result is important, this may lead to unexpected sensitive errors. Both existing calls of this function are `unwrap_or()`-ing result because it makes sense to do that there. First call - because `u64::MAX` may not be valid value of `relayers_state.total_messages` anyway. Second call just return stats && isn't used in sensitive context.

modules/message-lane/src/lib.rs
```diff
@@ -733,8 +746,11 @@ impl<T: Config<I>, I: Instance> OutboundLaneStorage for RuntimeOutboundLaneStora
 /// Verify messages proof and return proved messages with decoded payload.
 fn verify_and_decode_messages_proof<Chain: SourceHeaderChain<Fee>, Fee, DispatchPayload: Decode>(
 	proof: Chain::MessagesProof,
-	messages_count: MessageNonce,
+	messages_count: u32,
 ) -> Result<ProvedMessages<DispatchMessage<DispatchPayload, Fee>>, Chain::Error> {
+	// `receive_messages_proof` weight formula and `MaxUnconfirmedMessagesAtInboundLane` check
+	// guarantees that the `message_count` is sane and Vec<Messagae> may be allocated.
```
HCastano, line 752:
```suggestion
	// guarantees that the `message_count` is sane and Vec<Message> may be allocated.
```

HCastano (approved):
With the check against `T::MaxUnconfirmedMessagesAtInboundLane` I think we're good now

PR #662, opened 2021-01-19
Update weight related parameters in relay

svyatonik:
depends both on #598 (code) and #659 (decisions) => draft

### DIFF
--- a/relays/substrate/src/messages_lane.rs
+++ b/relays/substrate/src/messages_lane.rs
@@ -145,6 +145,15 @@ pub fn select_delivery_transaction_limits<W: pallet_message_lane::WeightInfoExt>
 		max_unconfirmed_messages_at_inbound_lane,
 	);
 
+	assert!(
+		max_number_of_messages > 0,
+		"Relay should fit at least one message in every delivery transaction",
+	);
+	assert!(
+		weight_for_messages_dispatch >= max_extrinsic_weight / 2,
+		"Relay shall be able to deliver messages with dispatch weight = max_extrinsic_weight / 2",
+	);
+
 	(max_number_of_messages, weight_for_messages_dispatch)
 }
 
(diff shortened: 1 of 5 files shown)


relays/substrate/src/messages_lane.rs
```diff
@@ -117,3 +118,53 @@ where
...
+	fn select_delivery_transaction_limits_works() {
+		assert_eq!(
+			select_delivery_transaction_limits::<RialtoToMillauMessageLaneWeights>(
+				bp_rialto::max_extrinsic_weight(),
+				bp_millau::MAX_UNREWARDED_RELAYER_ENTRIES_AT_INBOUND_LANE,
+			),
+			// We don't actually care about these values, so feel free to update them whenever test
+			// fails. The only thing to do before that is to ensure that new values looks sane: i.e. weight
+			// reserved for messages dispatch allows dispatch of non-trivial messages.
+			(1024, 866_583_333_334),
```
HCastano, line 167: How often do you expect this test to break and need to be manually fixed?

Also, what do you consider "sane" values? Based off the comment `select_delivery_transaction_limits` I _think_ that anything under 50% for dispatch wouldn't be sane - but what about the max number of messages? I'd like more clarity around this documented in the code

HCastano (approved):

relays/substrate/src/messages_lane.rs
svyatonik: > How often do you expect this test to break and need to be manually fixed?

Whenever weights or other related parameters are changed. We don't update weights often, so I guess it won't be often.

> Also, what do you consider "sane" values? Based off the comment select_delivery_transaction_limits I think that anything under 50% for dispatch wouldn't be sane - but what about the max number of messages? I'd like more clarity around this documented in the code

Number of messages should be larger than zero. If it has became much lower (no strict value - just some tendency) after some change, then it should get additional attention. E.g. when it drops from 1024 to, say, 100, it means that some significant change has been introduced.

svyatonik: I've added couple of asserts and docs

HCastano: Thanks for the clarifications!

Issue #614, opened 2021-01-05
Replay-protecting variant of CallOrigin::TargetAccount


tomusdrw:
Our current assumption with `CallOrigin::TargetAccount` is that both the private key on the source chain and the target chain is controlled by the same entity. This means that the only way to replay the message signed with the target key is to send it again with the source key.

In some scenarios it might not be desirable though, and to protect the users we should figure out how to prevent replaying such messages.

Explore possibilities:
- [ ] use source account nonce?
- [ ] explicit nonce mechanism? (`Map<source_account_id++target_account_id, Nonce>`)
- [ ] message nonce? (index in the lane?)
- [ ] recent block hash?

tomusdrw:
A combination of source account nonce + source block hash seems interesting to me as it doesn't require any extra storage entries. Block hash or mortality (#622) is required to prevent replays in case source account goes below ED (be careful cause it may happen during fee payment!) and have it's nonce re-set.

svyatonik:
more details: https://github.com/paritytech/srlabs_findings/issues/42

adoerr:
Filing some comments from @tomusdrw so they don't get forgotten

> We should add spec_version/tx_version and bridge_id even for #614 (and that's what #42 is about)

> Another angle is mortality of calls, that's something that is also partially related and worth implementing. So if we change the call to actaully be:
let call_object = { call, spec_version/tx_version, mortality }; then we should change the signing payload to bridge_id ++ source_account_id ++ call_object.

svyatonik:
irrelevant after #1376

PR #663, opened 2021-01-20
Decode call after spec_version check

### LINKED ISSUE #613: Defer decoding payload
Currently we decode the payload eagerly even if the `spec_version` does not match. This means that we might sometimes get decode error and sometimes `spec_version` mismatch error (depending on the message) after a runtime upgrade. We should consider derering `Call::decode` up until we know that everything else is correct (like `spec_version`, etc) or we may early fail decoding if we see the `spec_version` mismatch.

svyatonik:
closes #613

There's a new `type EncodedCall;` in the `pallet_call_dispatch::Config` trait. It needs to be `Encode + Decode` and these implementations must match implementations on the source chain (so that signed message matches original message). At the same time, we shall support conversions from this `EncodedCall` into `Call`. That's why there's that strange `Into<Result<<Self as Config<I>>::Call, ()>>` bound. Initially it was `TryInto<<Self as Config<I>>::Call>`, but I've ran into https://github.com/rust-lang/rust/issues/50133 . If you feel that `Into<Result<_, _>>` is too awkward, I could e.g. introduce new trait for this conversion.

### DIFF
--- a/bin/runtime-common/src/messages.rs
+++ b/bin/runtime-common/src/messages.rs
@@ -27,7 +27,7 @@ use bp_message_lane::{
 	InboundLaneData, LaneId, Message, MessageData, MessageKey, MessageNonce, OutboundLaneData,
 };
 use bp_runtime::InstanceId;
-use codec::{Compact, Decode, Encode, Input};
+use codec::{Compact, Decode, Encode};
 use frame_support::{traits::Instance, RuntimeDebug};
 use hash_db::Hasher;
 use pallet_substrate_bridge::StorageProofChecker;
@@ -302,11 +302,11 @@ pub mod target {
 	>;
 
 	/// Decoded Bridged -> This message payload.
-	pub type FromBridgedChainDecodedMessagePayload<B> = pallet_bridge_call_dispatch::MessagePayload<
+	pub type FromBridgedChainMessagePayload<B> = pallet_bridge_call_dispatch::MessagePayload<
 		AccountIdOf<BridgedChain<B>>,
 		SignerOf<ThisChain<B>>,
 		SignatureOf<ThisChain<B>>,
-		CallOf<ThisChain<B>>,
+		FromBridgedChainEncodedMessageCall<B>,
 	>;
 
 	/// Messages proof from bridged chain:
@@ -323,33 +323,21 @@ pub mod target {
 		MessageNonce,
 	);
 
-	/// Message payload for Bridged -> This messages.
-	pub struct FromBridgedChainMessagePayload<B: MessageBridge>(pub(crate) FromBridgedChainDecodedMessagePayload<B>);
-
-	impl<B: MessageBridge> From<FromBridgedChainDecodedMessagePayload<B>> for FromBridgedChainMessagePayload<B> {
-		fn from(decoded_payload: FromBridgedChainDecodedMessagePayload<B>) -> Self {
-			Self(decoded_payload)
-		}
+	/// Encoded Call of This chain as it is transferred over bridge.
+	///
+	/// Our Call is opaque (`Vec<u8>`) for Bridged chain. So it is encoded, prefixed with
+	/// vector length. Custom decode implementation here is exactly to deal with this.
+	#[derive(Decode, Encode, RuntimeDebug, PartialEq)]
+	pub struct FromBridgedChainEncodedMessageCall<B> {
+		pub(crate) encoded_call: Vec<u8>,
+		pub(crate) _marker: PhantomData<B>,
 	}
 
-	impl<B: MessageBridge> Decode for FromBridgedChainMessagePayload<B> {
-		fn decode<I: Input>(input: &mut I) -> Result<Self, codec::Error> {
-			// for bridged chain our Calls are opaque - they're encoded to Vec<u8> by submitter
-			// => skip encoded vec length here before decoding Call
-			let spec_version = pallet_bridge_call_dispatch::SpecVersion::decode(input)?;
-			let weight = frame_support::weights::Weight::decode(input)?;
-			let origin = FromBridgedChainMessageCallOrigin::<B>::decode(input)?;
-			let _skipped_length = Compact::<u32>::decode(input)?;
-			let call = CallOf::<ThisChain<B>>::decode(input)?;
-
-			Ok(FromBridgedChainMessagePayload(
-				pallet_bridge_call_dispatch::MessagePayload {
-					spec_version,
-					weight,
-					origin,
-					call,
-				},
-			))
+	impl<B: MessageBridge> From<FromBridgedChainEncodedMessageCall<B>> for Result<CallOf<ThisChain<B>>, ()> {
+		fn from(encoded_call: FromBridgedChainEncodedMessageCall<B>) -> Self {
+			let mut input = &encoded_call.encoded_call[..];
+			let _skipped_length = Compact::<u32>::decode(&mut input).map_err(drop)?;
+			CallOf::<ThisChain<B>>::decode(&mut input).map_err(drop)
 		}
 	}
 
@@ -366,22 +354,14 @@ pub mod target {
 		ThisCallDispatchInstance: frame_support::traits::Instance,
 		ThisRuntime: pallet_bridge_call_dispatch::Config<ThisCallDispatchInstance>,
 		pallet_bridge_call_dispatch::Module<ThisRuntime, ThisCallDispatchInstance>:
-			bp_message_dispatch::MessageDispatch<
-				(LaneId, MessageNonce),
-				Message = FromBridgedChainDecodedMessagePayload<B>,
-			>,
+			bp_message_dispatch::MessageDispatch<(LaneId, MessageNonce), Message = FromBridgedChainMessagePayload<B>>,
 	{
 		type DispatchPayload = FromBridgedChainMessagePayload<B>;
 
 		fn dispatch_weight(
 			message: &DispatchMessage<Self::DispatchPayload, BalanceOf<BridgedChain<B>>>,
 		) -> frame_support::weights::Weight {
-			message
-				.data
-				.payload
-				.as_ref()
-				.map(|payload| payload.0.weight)
-				.unwrap_or(0)
+			message.data.payload.as_ref().map(|payload| payload.weight).unwrap_or(0)
 		}
 
 		fn dispatch(message: DispatchMessage<Self::DispatchPayload, BalanceOf<BridgedChain<B>>>) {
@@ -389,7 +369,7 @@ pub mod target {
 				pallet_bridge_call_dispatch::Module::<ThisRuntime, ThisCallDispatchInstance>::dispatch(
 					B::INSTANCE,
 					(message.key.lane_id, message.key.nonce),
-					payload.0,
+					payload,
 				);
 			}
 		}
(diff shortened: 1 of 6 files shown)


bin/runtime-common/src/messages.rs
```diff
@@ -323,33 +323,21 @@ pub mod target {
...
-	impl<B: MessageBridge> From<FromBridgedChainDecodedMessagePayload<B>> for FromBridgedChainMessagePayload<B> {
-		fn from(decoded_payload: FromBridgedChainDecodedMessagePayload<B>) -> Self {
-			Self(decoded_payload)
-		}
+	/// Encoded Call of This chain as it is transferred over bridge.
+	///
+	/// Our Call is opaque (`Vec<u8>`) for Bridged chain. So it is encoded, prefixed with
+	/// vector length. Custom decode implementation here is exactly to deal with this.
+	#[derive(Decode, Encode, RuntimeDebug, PartialEq)]
+	pub struct FromBridgedChainEncodedMessageCall<B> {
```
tomusdrw, line 331: I'd ditch the this type here and `type EncodedCall` and rather move that logic to `call-dispatch`, like so:

```rust
#[derive(RuntimeDebug, Encode, Decode, Clone, PartialEq, Eq)]
pub struct MessagePayload<SourceChainAccountId, TargetChainAccountPublic, TargetChainSignature, Call> {
	/// Runtime specification version. We only dispatch messages that have the same
	/// runtime version. Otherwise we risk to misinterpret encoded calls.
	pub spec_version: SpecVersion,
	/// Weight of the call, declared by the message sender. If it is less than actual
	/// static weight, the call is not dispatched.
	pub weight: Weight,
	/// Call origin to be used during dispatch.
	pub origin: CallOrigin<SourceChainAccountId, TargetChainAccountPublic, TargetChainSignature>,
	/// The call itself.
	pub call: OpaqueCall<Call>,
}

#[derive(Decode, Encode)]
pub struct OpaqueCall<T>(Vec<u8>, #[codec(skip)] PhantomData<T>);

impl OpaqueCall<T: Decode> {
  pub fn decode_inner(&self) -> Result<...> {
     T::decode(&mut &*self.0)
  }
}
```

IMHO this should also be coupled with #664, so that we don't have extra `note_rejected_message`  function, but rather let the `call-dispatch` do the decoding and spec-version checking and then emit one of 4 events:
1. Spec version failed
2. Decoding Failed
3. Dispatch Failed
4. Dispatch Successful

@svyatonik what do you think?

svyatonik: I think that:
1) any parts of delivery logic (and the fact that call is 'strangely encoded' is a part of that logic) shall not be hardcoded in dispatch module. Today the call is `call.encode().encode()`, tomorrow (if we'll deal with offline storage), it'll be simply `call.encode()`;
2) if we don't have an access to the `Call` from outside (and it is what you're suggesting here), then we'll be unable to perform any filtering. Like initially iirc we were going to only dispatch balance calls;
3) you're confusing `Call` decoding (this PR) with message decoding (#614). We need to decode message before dispatch to be able to verify that the relayer has reserved enough dispatch weight for all messages.

tomusdrw: 1. IMHO The call dispatch should be the only module responsible for intepreting the payload of the message. The fact that it has to be double encoded is simply a limitation of SCALE codec we use. Even if we have offline storage, I guess it will look the same if we want to keep dispatch and message decoding separate (which I think we should do).
2. Perhaps that signifies that the filtering should be done in the `call-dispatch` pallet then? I know it complicates things a bit, but might be worth it. The `Call` type is actually known in the delivery pallet, but it's not decoded yet when the filtering is performed, we could pass the filter closure to the `call-dispatch::dispatch()` call so it's executed after we successfully decode the call.
3. No, I think we are on the same page. Message decoding should happen first - if the message can't be decoded we should simply fail the entire transaction. Decoded message will have `weight` parameter so that we can compute the total cost. Next the delivery pallet triggers dispatch (generic) for every message and that's where the decoding & spec_version checking should happen. I.e. delivery (and decoding) of messages should be successful despite `spec_version` mismatch, it's just the dispatch that should fail.

svyatonik: 1. `send_message` call of the message lane pallet accepts `Vec<u8>`, which is `call.encode()`. I agree that this first decoding should be made only by dispatch pallet. But it is not dispatch pallet who is adding that second `.encode()`. And it should not deal with that decode imo. If we'll migrate to offline storage. I didn't get your argument about keeping separating dispatch + decoding. Isn't separation is what you're suggesting to remove here (by decoding in dispatch)?
2. I think filtering should be done in runtime. Either by providing closure to the dispatch, or directly, if runtime would have decoded `Call`. I'm just saying that in your scenario `Call` is completely hidden from outside world && this won't work.
3. We can't fail entire transaction, because it'll stuck lane. We have just received proof that the other chain has sent this message (which is encoded invalid) to us - why we should fail transaction?

tomusdrw: Okay, thanks for clarifications, I see your points better now.
1. I'm opting for delivery doing `Vec<u8> -> OpaqueCall` conversion (which would be `Vec::<u8>::encode()`) on `send_message` and then symmetrically delivery pallet doing `OpaqueCall -> Vec<u8>` decoding before handing over to the dispatch pallet. I.e. to overcome the scale issue of just lazily decoding raw bytes.
2. Yes, totally agree it's in the runtime. AFAIR currently we implement `LaneMessageVerifier` and it's supposed to get decoded payload (call). I think we should keep that API, but with `call-dispatch` responsible for decoding (`Vec<u8> -> Call`) we can't do filtering at this level.
Instead I'd propose to add `BaseCallFilter`-like mechanism (this will allow us to re-use existing API) to `call-dispatch` pallet.
3. I think I'm conflating a bunch of things when saying "message" and "payload", hence confusion. But I think I see your point now, we need to attempt to decode the call to get the exact `weight` before we even start doing the dispatch. But I still think it's fine with `OpaqueCall` approach - I'd just prefer to avoid adding generic types, where we can simply add concrete types and make the process more explicit in the pallet. Similarly using `Into` for heavy stuff like `decoding` and getting a result is imho not the right way. We can be explicit that we don't `.into()` between formats, but we simply `.decode()` to overcome SCALE limitations.

I can maybe try to implement a pseudo-code of what I have in mind, perhaps there is something that's obvious to you and what I'm missing which prevents us to get to shared understanding.

svyatonik: 1 + 3. Ok, so what I got from these points,  is that all this comes from preference of explicit `Decode` + `Vec<u8>` over `EncodedCall`. I have two things to say about this:
1.1) I still do not think that `OpaqueCall` belongs to the `call-dispatch` pallet. Imo, it is about encoding + decoding and shall not be a part of this pallet => should be solved outside (on "transport" level). I haven't seen any strong arguments to not to keep it outside, other than personal preferences;
1.2) I agree that `Into` is awkward, as I already stated in description;
2. That is wrong - `LaneMessageVerifier` works on the source side. On target chain there's only call dispatch, which supposed to do filtering. Basically - before this PR you had `Call` [here](https://github.com/paritytech/parity-bridges-common/pull/663/files#diff-ded5af982e34b56ffecfe2ffeba389edefd5dc7302c45a6c1853361a7e6d749bL389) and could have just drop the `Call` there, before it reache `dispatch` method. Now you can do the same from `Into<Result<...>>`, which is I agree, is also a bit awkward.
3. So given that you agree that the `MessagePayload` should be decoded before dispatch method, does it mean that [this comment](https://github.com/paritytech/parity-bridges-common/pull/664#pullrequestreview-572186156) is now invalid?

What I can suggest is to change `EncodedCall: Into<Result<...>>;` to something like that:
```rust
enum CallError { Decode, Filter }

trait EncodedCall<Call> {
  fn decode_and_filter_call(self) -> Result<Call, DecodeOrFilterError>;
}
```

So what it changes is simply that your `OpaqueCall` would live outside of the pallet (it already lives there as `FromBridgedChainEncodedMessageCall`). The main change is removal of that `Into` conversion.

I'd even go further and introduce more traits, like:
```rust
trait EncodedMessage<Message> {
	/// Resolves #664 - fails if message decode has failed and we can't even verify spec_version, ...
	/// The pallet would emit `MessageRejected` event.
	fn decode(self) -> Result<Message, ()>;
}

trait EncodedCall<Call> {
	/// Fails if call can not be decoded.
	/// The pallet would emit `MessageCallDecodeFailed` event.
	fn decode(self) -> Result<Call, ()>;
}

trait CallFilter<Call> {
	/// Fails if call shall not be dispatched.
	/// The pallet would emit `CallRejected` event.
	fn filter(self) -> Result<Call, ()>;
}
```

But I assume that'll be too much.

tomusdrw: > I still do not think that `OpaqueCall` belongs to the call-dispatch pallet.

Agree, that was my initial proposal, cause I thought it would be the `call-dispatch` doing the decoding, it should be separate, but it shouldn't be runtime-specific either.

> So given that you agree that the MessagePayload should be decoded before dispatch method, does it mean that this comment is now invalid?

Yeah, let's continue there.

> But I assume that'll be too much.

I'd aim for introducing 0 new traits :)
1. It seems to me that `trait EncodedMessage<Mesage>` is simply `parity_scale_codec::Decode`? 
```rust
// in message delivery
T::Dispatch::dec
2. `CallFilter` is already in `frame_support::traits::Filter` and it has a bunch of useful tools to construct the filters: https://github.com/paritytech/substrate/blob/15fc239ed63c902a225cfb193ca1646a45d34931/frame/support/src/traits.rs#L44


Anyways, I feel we are going nowhere, because I'm basing most of my suggestions on gut feeling that there is too much abstraction, but in reality it's probably not possible to implement them in my suggested form without changing too much of other existing abstractions.

Let's just proceed with explicit decoding (no `Into`) and we can revist later on.

svyatonik: > It seems to me that trait EncodedMessage<Mesage> is simply parity_scale_codec::Decode?
> Let's just proceed with explicit decoding (no Into) and we can revist later on.

I think we have lost the track. The idea is that call-dispatch shall not know how the call is transmitted - it may be encoded with parity-codec, it may be `call.encode().encode()`, it may use custom encoding, or it is crafted manually at the target chain, ... This pallet imo shall just dispatch the call and emit events. Like when we have had conversation about Polkadot <-> Kusama bridge, one of ideas was to use `enum Message { TransferBalance(Source, Target) }`, or message could be just `u64`, which is a balance that must be transferred from some hardcoded account A to account B.

I'm not sure I got your idea about replacing `EncodedMessage::decode() -> Call` with `Decode::decode(&mut &input) -> Call`. Where we're going to get this `input`? It seems to me that you're just suggesting exactly what you have been suggesting initially - i.e. to change type of `pallet_bridge_call_dispatch::MessagePayload::call` to `Vec<u8>` and then just manually call `Call::decode()` on it. That's exactly what I'm stand against in this whole discussion.

> CallFilter is already in frame_support::traits::Filter

Ok, then we may reuse existing trait here.

HCastano:
@svyatonik I turned this into a draft since it looks like there's been a lot of conversation, but no new commits. Feel free to change the status back if I was wrong about this call

bin/runtime-common/src/messages.rs
tomusdrw: Let's proceed with it as-is, I initially thought it could be simplified significantly, but there are caveats that I didn't take into account (like getting the weight first).

While I don't agree with some of your reasoning for increased flexibility (i.e. supporting custom encodings, etc) it would be dumb for me to be strong on reverting the functionality if it's already here.

tomusdrw (approved):

PR #664, opened 2021-01-20
Emit dispatch message when whole message (payload) has invalid encoding

### LINKED ISSUE #612: Event for payload decoding error
Message delivery on the target chain should generate an event in case it fails to decode the payload. Currently events are not consistent, cause we do report an error in case of `spec_version` inconsistency but not if the payload fails to be decoded. We should have an event for every message in any case.

svyatonik:
closes #612

### DIFF
--- a/bin/runtime-common/src/messages.rs
+++ b/bin/runtime-common/src/messages.rs
@@ -364,7 +364,9 @@ pub mod target {
 		for FromBridgedChainMessageDispatch<B, ThisRuntime, ThisCallDispatchInstance>
 	where
 		ThisCallDispatchInstance: frame_support::traits::Instance,
-		ThisRuntime: pallet_bridge_call_dispatch::Config<ThisCallDispatchInstance>,
+		ThisRuntime: pallet_bridge_call_dispatch::Config<ThisCallDispatchInstance, MessageId = (LaneId, MessageNonce)>,
+		<ThisRuntime as pallet_bridge_call_dispatch::Config<ThisCallDispatchInstance>>::Event:
+			From<pallet_bridge_call_dispatch::RawEvent<(LaneId, MessageNonce), ThisCallDispatchInstance>>,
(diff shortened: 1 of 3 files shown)


tomusdrw (dismissed):
See: https://github.com/paritytech/parity-bridges-common/pull/663#discussion_r560933590

bin/runtime-common/src/messages.rs
```diff
@@ -385,12 +387,18 @@ pub mod target {
 		}
 
 		fn dispatch(message: DispatchMessage<Self::DispatchPayload, BalanceOf<BridgedChain<B>>>) {
+			let message_id = (message.key.lane_id, message.key.nonce);
 			if let Ok(payload) = message.data.payload {
 				pallet_bridge_call_dispatch::Module::<ThisRuntime, ThisCallDispatchInstance>::dispatch(
```
tomusdrw, line 392: I know understand why we need to decode the message before this point (getting weight info), but I still don't like the explicit `note_rejected_message` call, could we just pass the entire `message.data.payload` (instead of `payload.0`) to the `dispatch` function?

I.e. I'm proposing to change the `fn dispatch` function from the trait to take `Result<...>` instead of just the payload - this emitting the event can stay internal to `call-dispatch`.

tomusdrw (approved):

modules/call-dispatch/src/lib.rs
```diff
@@ -183,7 +185,17 @@ impl<T: Config<I>, I: Instance> MessageDispatch<T::MessageId> for Module<T, I> {
 		message.weight
 	}
 
-	fn dispatch(bridge: InstanceId, id: T::MessageId, message: Self::Message) {
+	fn dispatch(bridge: InstanceId, id: T::MessageId, message: Result<Self::Message, ()>) {
```
HCastano, line 188: Since we're not actually using the `Err` type of the `Result` we might as well just make this an `Option<Self::Message>`

HCastano (approved):

PR #667, opened 2021-01-20
Prevent potential signature reuse

adoerr:
Resolves [paritytech/srlabs_findings/#42](https://github.com/paritytech/srlabs_findings/issues/42)

Currently the signature required to submit a call with origin `CallOrigin::TargetAccount` contains only the encoded call and the submitter account id on the source chain (source_account_id). This may allow signature reuse.

In order to prevent this, we include `MessagePayload.spec_version` and the `InstanceId` of the bridge in the signature.

### DIFF
--- a/bin/millau/runtime/src/lib.rs
+++ b/bin/millau/runtime/src/lib.rs
@@ -596,12 +596,12 @@ impl_runtime_apis! {
 	}
 }
 
-/// Proof Rialto account ownership from Millau.
+/// Rialto account ownership digest from Millau.
 ///
 /// The byte vector returned by this function should be signed with a Rialto account private key.
 /// This way, the owner of `millau_account_id` on Millau proves that the Rialto account private key
 /// is also under his control.
-pub fn rialto_account_ownership_proof<Call, AccountId, SpecVersion>(
+pub fn rialto_account_ownership_digest<Call, AccountId, SpecVersion>(
 	rialto_call: Call,
 	millau_account_id: AccountId,
 	rialto_spec_version: SpecVersion,
@@ -611,7 +611,7 @@ where
 	AccountId: codec::Encode,
 	SpecVersion: codec::Encode,
 {
-	pallet_bridge_call_dispatch::account_ownership_proof(
+	pallet_bridge_call_dispatch::account_ownership_digest(
 		rialto_call,
 		millau_account_id,
 		rialto_spec_version,
--- a/bin/millau/runtime/src/lib.rs
+++ b/bin/millau/runtime/src/lib.rs
@@ -611,14 +611,12 @@ where
 	AccountId: codec::Encode,
 	SpecVersion: codec::Encode,
 {
-	let proof = pallet_bridge_call_dispatch::account_ownership_proof(
+	pallet_bridge_call_dispatch::account_ownership_proof(
 		rialto_call,
 		millau_account_id,
 		rialto_spec_version,
 		bp_runtime::MILLAU_BRIDGE_INSTANCE,
-	);
-
-	proof
+	)
 }
 
 #[cfg(test)]
--- a/bin/millau/runtime/src/lib.rs
+++ b/bin/millau/runtime/src/lib.rs
@@ -596,6 +596,31 @@ impl_runtime_apis! {
 	}
 }
 
+/// Proof account ownership on 'rialto'.
+///
+/// The byte vector returned by this function will be signed with a 'rialto' account private key.
+/// This way, the owner of `millau_account_id` on 'millau' proves that the 'rialto' account private key
+/// is also under his control.
+pub fn rialto_account_ownership_proof<Call, AccountId, SpecVersion>(
+	rialto_call: Call,
+	millau_account_id: AccountId,
+	rialto_spec_version: SpecVersion,
+) -> sp_std::vec::Vec<u8>
+where
+	Call: codec::Encode,
+	AccountId: codec::Encode,
+	SpecVersion: codec::Encode,
+{
+	let proof = pallet_bridge_call_dispatch::account_ownership_proof(
+		rialto_call,
+		millau_account_id,
+		rialto_spec_version,
+		bp_runtime::MILLAU_BRIDGE_INSTANCE,
+	);
+
+	proof
+}
+
 #[cfg(test)]
 mod tests {
 	use super::*;
--- a/bin/millau/runtime/src/lib.rs
+++ b/bin/millau/runtime/src/lib.rs
@@ -596,10 +596,10 @@ impl_runtime_apis! {
 	}
 }
 
-/// Proof account ownership on 'rialto'.
+/// Proof Rialto account ownership from Millau.
 ///
-/// The byte vector returned by this function will be signed with a 'rialto' account private key.
-/// This way, the owner of `millau_account_id` on 'millau' proves that the 'rialto' account private key
+/// The byte vector returned by this function should be signed with a Rialto account private key.
+/// This way, the owner of `millau_account_id` on Millau proves that the Rialto account private key
 /// is also under his control.
 pub fn rialto_account_ownership_proof<Call, AccountId, SpecVersion>(
 	rialto_call: Call,
--- a/bin/rialto/runtime/src/lib.rs
+++ b/bin/rialto/runtime/src/lib.rs
@@ -985,12 +985,12 @@ impl_runtime_apis! {
 	}
 }
 
-/// Proof Millau account ownership from Rialto.
+/// Millau account ownership digest from Rialto.
 ///
 /// The byte vector returned by this function should be signed with a Millau account private key.
 /// This way, the owner of `rialto_account_id` on Rialto proves that the 'millau' account private key
 /// is also under his control.
-pub fn millau_account_ownership_proof<Call, AccountId, SpecVersion>(
+pub fn millau_account_ownership_digest<Call, AccountId, SpecVersion>(
 	millau_call: Call,
 	rialto_account_id: AccountId,
 	millau_spec_version: SpecVersion,
@@ -1000,7 +1000,7 @@ where
 	AccountId: codec::Encode,
 	SpecVersion: codec::Encode,
 {
-	pallet_bridge_call_dispatch::account_ownership_proof(
+	pallet_bridge_call_dispatch::account_ownership_digest(
 		millau_call,
 		rialto_account_id,
 		millau_spec_version,
--- a/bin/rialto/runtime/src/lib.rs
+++ b/bin/rialto/runtime/src/lib.rs
@@ -1000,14 +1000,12 @@ where
 	AccountId: codec::Encode,
 	SpecVersion: codec::Encode,
 {
-	let proof = pallet_bridge_call_dispatch::account_ownership_proof(
+	pallet_bridge_call_dispatch::account_ownership_proof(
 		millau_call,
 		rialto_account_id,
 		millau_spec_version,
 		bp_runtime::RIALTO_BRIDGE_INSTANCE,
-	);
-
-	proof
+	)
 }
 
 #[cfg(test)]
--- a/bin/rialto/runtime/src/lib.rs
+++ b/bin/rialto/runtime/src/lib.rs
@@ -985,10 +985,10 @@ impl_runtime_apis! {
 	}
 }
 
-/// Proof account ownership on 'millau'.
+/// Proof Millau account ownership from Rialto.
 ///
-/// The byte vector returned by this function will be signed with a 'millau' account private key.
-/// This way, the owner of `rialto_account_id` on 'rialto' proves that the 'millau' account private key
+/// The byte vector returned by this function should be signed with a Millau account private key.
+/// This way, the owner of `rialto_account_id` on Rialto proves that the 'millau' account private key
 /// is also under his control.
 pub fn millau_account_ownership_proof<Call, AccountId, SpecVersion>(
 	millau_call: Call,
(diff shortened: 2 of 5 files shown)


HCastano:
@adoerr Can you change the name of the PR to be more descriptive? A small explanation in the body would also be useful for those that don't have access to the `srlabs_findings` repository

HCastano (changes requested):
Right now any messages sent with the Target origin fail to be dispatched due to mismatched signatures:

```
2021-01-20 20:24:30.054  TRACE tokio-runtime-worker pallet_bridge_call_dispatch:Message [114, 108, 116, 111]/([0, 0, 0, 0], 6): origin proof is invalid. Expected account: 306721211d5404bd9da88e0204360a1a9ab8b87c66c1bc2fcdd37f3c2222cc20 (5DAAnrj7...) from signature: MultiSignature::Sr25519(28b8cebf83fe1fb5655273aec2fe856a734a8404ad614188022393e1546f0c73b930749bd6dca5c815caae5e6fefbaa326c7f4e8d0da7008b3c7ada65bba5a84)
```

We need to add the `bridge_id` and `spec_version` to the [signed messages](https://github.com/paritytech/parity-bridges-common/blob/master/relays/substrate/src/main.rs#L300) being sent by the relayer.

tomusdrw:
Would be awesome to cover the CLI-runtime compatibility in a unit test (i.e. not necessarily sending the entire transaction, but maybe just verifying that the generated signature is runtime-valid).

adoerr:
Turned to draft, until I have figured out how to do the unit test.

adoerr:
@tomusdrw so my current idea would be to use `relay_millau_client` and `relay_rialto_client` `SigningParams` in order to build an origin target signature or even a `MessagePayload`. But then what? Just calling `TargetChainSignature.verify()` using the appropriate trait pram types?

tomusdrw:
> But then what? Just calling TargetChainSignature.verify() using the appropriate trait pram types?

Maybe we can extract this code to a plain function and just call this from the relayer test?
https://github.com/paritytech/parity-bridges-common/pull/667/files#diff-ff4d23aa9a4bf7c078243228b62f974fb5326ffd2037610400f6ed0718498eceR238-R246

just try to keep it as simple as possible :)

relays/substrate/src/main.rs
```diff
@@ -517,3 +526,76 @@ async fn estimate_message_delivery_and_dispatch_fee<Fee: Decode, C: Chain, P: En
...
+		let millau_account_id: bp_millau::AccountId = millau_public.into_account();
+
+		// The `CallOrigin::TargetAccount` private key ownership proof is a signature
+		// over this message.
+		let mut message = Vec::new();
+		call.encode_to(&mut message);
+		millau_account_id.encode_to(&mut message);
+		rialto_runtime::VERSION.spec_version.encode_to(&mut message);
+		bp_runtime::MILLAU_BRIDGE_INSTANCE.encode_to(&mut message);
+
```
tomusdrw, line 568: This logic is now repeated 3 times (runtime, relay and test), IMHO we should create a standalone function that is responsible either for just constructing the message or constructing & verifying the signature, so that it can be re-used between here and the runtime (assuming we don't want the relayer code to depend on the runtime directly).

The main reason for this is to make sure that if we change this in only one of the places (i.e. only in runtime, not relayer) we will get a failing test, so we prevent a situation where this get's broken in the future.

tomusdrw (changes requested):

relays/substrate/src/main.rs
adoerr: I think, I don't understand what you mean by 'runtime' here. I can see, how a separate function in the call_dispatch pallet might help. But that would not make the solution dependent on the runtime but the call_dispatch pallet. So, what am I missing here?

tomusdrw: By `runtime` I meant "the code that eventually ends up being part of the runtime", so `call-dispatch` pallet falls into that category :)
Yeah, just exposing a function in `call-dispatch` pallet should be good enough, or additionally you can have `millau_runtime` or `rialto_runtime` expose the same function but with some parameters already partially-applied (like runtime version or bridge instance).

adoerr: Next try.

deny.toml
```diff
@@ -49,7 +49,9 @@ notice = "warn"
 # output a note when they are encountered.
 ignore = [
 	# yaml-rust < clap. Not feasible to upgrade and also not possible to trigger in practice.
-	"RUSTSEC-2018-0006"
+	"RUSTSEC-2018-0006",
+   # raw-cpuid is a dep of cranelift-native. Not feasible to upgrade.
+    "RUSTSEC-2021-0013"
```
HCastano, line 54: I opened #677 for this :)

modules/call-dispatch/src/lib.rs
```diff
@@ -322,6 +321,32 @@ where
 	}
 }
 
+/// Proof account ownership on the target chain.
```
HCastano, line 324:
```suggestion
/// Prove target account ownership from the source chain.
```

modules/call-dispatch/src/lib.rs
```diff
@@ -322,6 +321,32 @@ where
...
+///
+/// The byte vector returned by this function will be signed with a target chain account
```
HCastano, line 326:
```suggestion
/// The byte vector returned by this function should be signed with a target chain account
```

bin/millau/runtime/src/lib.rs
```diff
@@ -596,6 +596,29 @@ impl_runtime_apis! {
 	}
 }
 
+/// Proof account ownership on 'rialto'.
+///
+/// The byte vector returned by this function will be signed with a 'rialto' account private key.
```
HCastano, line 601:
```suggestion
/// The byte vector returned by this function should be signed with a 'rialto' account private key.
```

bin/rialto/runtime/src/lib.rs
```diff
@@ -985,6 +985,29 @@ impl_runtime_apis! {
 	}
 }
 
+/// Proof account ownership on 'millau'.
+///
+/// The byte vector returned by this function will be signed with a 'millau' account private key.
```
HCastano, line 990:
```suggestion
/// The byte vector returned by this function should be signed with a 'millau' account private key.
```

relays/substrate/src/main.rs
```diff
@@ -517,3 +523,67 @@ async fn estimate_message_delivery_and_dispatch_fee<Fee: Decode, C: Chain, P: En
 		Decode::decode(&mut &encoded_response.0[..]).map_err(relay_substrate_client::Error::ResponseParseFailed)?;
 	Ok(decoded_response)
 }
+
+#[cfg(test)]
+mod tests {
+	use sp_core::Pair;
+	use sp_runtime::traits::{IdentifyAccount, Verify};
+
+	macro_rules! call_for_runtime {
```
HCastano, line 532: I think we can simplify this a bit. Really, all we need is:

```rust
let call = $runtime::Call::System($runtime::SystemCall::remark(vec![]))
```

At which point we might as well just hardcode the runtime in all the call sites. I don't think a macro is worth the savings for two calls :)

relays/substrate/src/main.rs
```diff
@@ -517,3 +523,67 @@ async fn estimate_message_delivery_and_dispatch_fee<Fee: Decode, C: Chain, P: En
...
+				)
+				.as_bytes()
+				.to_vec()
+			))
+		}
+	}
+
+	#[test]
+	fn millau_signature_is_valid_on_rialto() {
+		let millau_sign = relay_millau_client::SigningParams::from_suri("//Dave", None).unwrap();
```
HCastano, line 550:
```suggestion
		let millau_signer = relay_millau_client::SigningParams::from_suri("//Dave", None).unwrap();
```

nitpick

relays/substrate/src/main.rs
```diff
@@ -517,3 +523,67 @@ async fn estimate_message_delivery_and_dispatch_fee<Fee: Decode, C: Chain, P: En
...
+
+		let rialto_sign = relay_rialto_client::SigningParams::from_suri("//Dave", None).unwrap();
+		let signature = rialto_sign.signer.sign(&proof);
+
+		assert!(signature.verify(&proof[..], &rialto_sign.signer.public()));
+	}
+
+	#[test]
+	fn rialto_signature_is_valid_on_millau() {
+		let rialto_sign = relay_rialto_client::SigningParams::from_suri("//Dave", None).unwrap();
```
HCastano, line 571:
```suggestion
		let rialto_signer = relay_rialto_client::SigningParams::from_suri("//Dave", None).unwrap();
```

nitpick

HCastano (approved):
Code looks good!

@adoerr can you make sure that this actually syncs messages properly though? To do this you can build images for the nodes and relayer locally and do a quick check with the Compose deployment. You shouldn't see any errors in the `pallet_bridge_call_dispatch` logs.

bin/millau/runtime/src/lib.rs
```diff
@@ -596,6 +596,29 @@ impl_runtime_apis! {
 	}
 }
 
+/// Proof account ownership on 'rialto'.
```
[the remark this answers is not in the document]
adoerr: Done

bin/rialto/runtime/src/lib.rs
```diff
@@ -985,6 +985,29 @@ impl_runtime_apis! {
 	}
 }
 
+/// Proof account ownership on 'millau'.
```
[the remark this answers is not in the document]
adoerr: Done

modules/call-dispatch/src/lib.rs
```diff
@@ -65,7 +65,7 @@ pub enum CallOrigin<SourceChainAccountId, TargetChainAccountPublic, TargetChainS
 	///
 	/// The account can be identified by `TargetChainAccountPublic`. The proof that the
 	/// `SourceChainAccountId` controls `TargetChainAccountPublic` is the `TargetChainSignature`
-	/// over `(Call, SourceChainAccountId).encode()`.
+	/// over `(Call, SourceChainAccountId, TargetChainSpecVersion, SourceChainInstanceId).encode()`.
```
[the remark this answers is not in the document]
adoerr: Then the type should be called `BridgeId`, no? Changed.

modules/call-dispatch/src/lib.rs
```diff
@@ -322,6 +321,32 @@ where
 	}
 }
 
+/// Proof account ownership on the target chain.
+///
+/// The byte vector returned by this function will be signed with a target chain account
+/// private key. This way, the owner of `source_account_id` on the source chain proves that
+/// the target chain account private key is also under his control.
+pub fn account_ownership_proof<Call, AccountId, SpecVersion, InstanceId>(
```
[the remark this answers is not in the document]
adoerr: Done

modules/call-dispatch/src/lib.rs
adoerr: Done

bin/millau/runtime/src/lib.rs
adoerr: Done

bin/rialto/runtime/src/lib.rs
adoerr: Done

relays/substrate/src/main.rs
adoerr: Done

relays/substrate/src/main.rs
adoerr: Done

relays/substrate/src/main.rs
adoerr: Done

adoerr:
> @adoerr can you make sure that this actually syncs messages properly though? To do this you can build images for the nodes and relayer locally and do a quick check with the Compose deployment. You shouldn't see any errors in the `pallet_bridge_call_dispatch` logs.

@HCastano I don't understand this comment. Have you seen errors? I ran a local test deployment. Balance transfers got through and no message signature mismatch events appeared.

bin/millau/runtime/src/lib.rs
```diff
@@ -596,6 +596,29 @@ impl_runtime_apis! {
 	}
 }
 
+/// Proof Rialto account ownership from Millau.
+///
+/// The byte vector returned by this function should be signed with a Rialto account private key.
+/// This way, the owner of `millau_account_id` on Millau proves that the Rialto account private key
+/// is also under his control.
+pub fn rialto_account_ownership_proof<Call, AccountId, SpecVersion>(
```
tomusdrw, line 604: This technically isn't really a proof, it just returns a payload that should be signed by the target account (and the signature + payload + public address is a proof). But it's good enough.

tomusdrw (approved):

bin/millau/runtime/src/lib.rs
adoerr: How should this be named then?

tomusdrw: I'd change `call_dispatch::account_ownership_proof -> call_dispatch::target_origin_signed_payload`
and `rialto_account_ownership_proof -> rialto_origin_signed_payload`

but I don't think these names are that much better, hence I didn't even propose that.

adoerr: How about `account_ownership_digest`?

tomusdrw: works for me!

adoerr: Changed the names. Will merge, once CI is green.

PR #669, opened 2021-01-20
Migrate Finality Verifier Pallet to FRAME v2

HCastano:
After watching Guillaume's [Substrate Seminar presentation](https://www.crowdcast.io/e/substrate-seminar/30) yesterday I figured it would be cool to try and upgrade one of our pallets to the new FRAME macros. The Finality Verifier pallet seemed like a good place to start since a) it's not being audited at the moment, and b) it's super simple (for example, it doesn't even have storage items).

It took me less than an hour to do this change while following the [migration guide](https://crates.parity.io/frame_support/attr.pallet.html#upgrade-guidelines).

### DIFF
--- a/modules/finality-verifier/src/lib.rs
+++ b/modules/finality-verifier/src/lib.rs
@@ -307,10 +307,4 @@ mod tests {
 			);
 		})
 	}
-
-	#[test]
-	fn print_metadata() {
-		println!("{:#?}", TestRuntime::metadata());
-		panic!()
-	}
 }
--- a/modules/finality-verifier/src/lib.rs
+++ b/modules/finality-verifier/src/lib.rs
@@ -161,6 +161,7 @@ pub mod pallet {
 
 #[cfg(test)]
 mod tests {
+	use super::pallet::*;
 	use super::*;
 	use crate::mock::{run_test, test_header, Origin, TestRuntime};
 	use bp_test_utils::{authority_list, make_justification_for_header};
--- a/modules/finality-verifier/src/lib.rs
+++ b/modules/finality-verifier/src/lib.rs
@@ -35,59 +35,52 @@
 use bp_header_chain::{justification::verify_justification, AncestryChecker, HeaderChain};
 use bp_runtime::{Chain, HeaderOf};
 use finality_grandpa::voter_set::VoterSet;
-use frame_support::{decl_error, decl_module, decl_storage, dispatch::DispatchResult, ensure, traits::Get};
+use frame_support::{ensure, traits::Get};
 use frame_system::ensure_signed;
 use sp_runtime::traits::Header as HeaderT;
 
 #[cfg(test)]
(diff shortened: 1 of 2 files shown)


modules/finality-verifier/src/lib.rs
```diff
@@ -35,73 +35,66 @@
...
-		/// The authority set from the underlying header chain is invalid.
-		InvalidAuthoritySet,
-		/// Failed to write a header to the underlying header chain.
-		FailedToWriteHeader,
-		/// Failed to write finality proof to the underlying header chain.
-		FailedToWriteFinalityProof,
-		/// The given ancestry proof is too large to be verified in a single transaction.
-		OversizedAncestryProof,
+#[frame_support::pallet]
+pub mod pallet {
```
adoerr, line 46: I was wondering, if it is possible to use a module name other thaan `pallet`?

modules/finality-verifier/src/lib.rs
```diff
@@ -35,73 +35,66 @@
...
+		#[pallet::constant]
+		type MaxHeadersInSingleProof: Get<u8>;
 	}
-}
 
-decl_module! {
-	pub struct Module<T: Config> for enum Call where origin: T::Origin {
-		type Error = Error<T>;
+	#[pallet::pallet]
+	#[pallet::generate_store(pub(super) trait Store)]
```
adoerr, line 76: Is this even needed? `#[pallet::storage]` itself is optional and according to your own comment, we don't have storage items?

modules/finality-verifier/src/lib.rs
```diff
@@ -134,24 +126,42 @@ decl_module! {
 			with_transaction(|| {
```
adoerr, line 126: Would it be possible to use the `frame_support_procedural::transactional` attribute macro for the call?

adoerr (approved):
Apart from some questions, LGTM.

modules/finality-verifier/src/lib.rs
HCastano: Good catch, I don't think it's required

modules/finality-verifier/src/lib.rs
HCastano: It's just a normal Rust module, so yeah the name doesn't matter. I imagine the convention will be to simply stick with `mod pallet` though

modules/finality-verifier/src/lib.rs
HCastano: Just looking at the docs they look reasonably equivalent. I'm not sure if there any subtle differences though

HCastano: Looks like the [`transactional` macro is implemented using `with_transaction`](https://github.com/paritytech/substrate/blob/master/frame/support/procedural/src/transactional.rs#L23), so they're equivalent

PR #672, opened 2021-01-21
Add Unchecked Header Writes to Bridge Pallet

HCastano:
This PR adds a method to the `Verifier` in the Substrate bridge pallet which accepts finalized headers without actually checking if they are indeed finalized.

The reason why a admittedly dangerous method like this is useful is because it saves us work if we know that the incoming headers are already finalized. This is exactly the case if we're importing headers which originate from the Finality Verifier pallet.

Note that while in theory some of the storage items used by the new method could be simplified to not be fork aware, we want to (for now) piggyback off the existing pallet so we need to occasionally use a dummy fork when interacting with storage.

In the future I imagine that unchecked header import will be the only way to write to the Substrate bridge pallet, but only after doing the verification elsewhere.


modules/substrate/src/lib.rs
```diff
@@ -376,19 +376,7 @@ impl<T: Config> bp_header_chain::HeaderChain<BridgedHeader<T>> for Module<T> {
...
-
-	fn import_finality_proof(header: BridgedHeader<T>, finality_proof: Vec<u8>) -> Result<(), ()> {
-		let mut verifier = verifier::Verifier {
-			storage: PalletStorage::<T>::new(),
-		};
-
-		let _ = verifier
-			.import_finality_proof(header.hash(), finality_proof.into())
-			.map_err(|_| ())?;
+		let _ = verifier.import_header_unchecked(header).map_err(|_| ())?;
```
tomusdrw, line 379: I'm not sure about changing the existing trait. Why not add a new one that simply adds a different interface towards the same `Module`?

I.e. we keep `import_header/import_finality_proof` and have a more specific `trait FinalizedHeaderChain: HeaderChain` which only exposes a unchecked function to import these pre-verified headers.

modules/substrate/src/verifier.rs
```diff
@@ -303,6 +307,77 @@ where
 
 		Ok(())
 	}
+
+	/// Import a finalized header without checking if this is true.
```
tomusdrw, line 311: What do you think about extracting this function out of `Verifier` struct, because it (as the name and docs suggest) does not verify that much (only handles authority set hand off).
I'm thinking about something like (on the call site):
```
InjectBestFinalizedHeader { storage } ::inject(header)
```

This will also draw a clear line between the old interface (and `Verifier`) and the new "unsafe" one.

modules/substrate/src/verifier.rs
```diff
@@ -303,6 +307,77 @@ where
...
+	///
+	/// This function assumes that all the given header has already been proven to be valid and
+	/// finalized. Using this assumption it will write them to storage with minimal checks. That
+	/// means it's of great importance that this function *not* called with any headers whose
+	/// finality has not been checked, otherwise you risk bricking your bridge.
+	///
+	/// One thing this function does do for you is GRANDPA authority set handoffs. However, since it
+	/// does not do verification on the incoming header it will assume that the authority set change
+	/// signals in the digest are well formed.
+	pub fn import_header_unchecked(&mut self, header: H) -> Result<(), ImportError> {
```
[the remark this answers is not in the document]
tomusdrw: After reading the function body I think we should emphasize that this function is expecting the imported header to always be the next best finalized header as well (i.e. it has to be (direct) descendant of the previous best finalized).
Perhaps we could name it:
```
fn inject_best_finalized()
```
?

modules/substrate/src/verifier.rs
tomusdrw: Given that we call `import_header` (new one) in a loop, perhaps from the API perspective it makes more sense to have something like:
```
AppendBestFinalized { storage }::append_finalized_chain(headers: &[Header])
```

tomusdrw (commented):
Logic-wise it looks good, but I'd minimize changes to the old API and would like to start some discussion about naming.

modules/substrate/src/lib.rs
HCastano: I don't think we should be exposing both `import_header/import_finality` and `import_unchecked` because for our purposes `import_unchecked` is superior to the former methods.

Unless you can convince me that having all three methods is useful, I'm going to only work with the unchecked method 😄

modules/substrate/src/verifier.rs
HCastano: I think both of these suggestions make sense

modules/substrate/src/lib.rs
tomusdrw: I feel that redesigning (& merging) both pallets will be needed in the future, and by changing `HeaderChain`, we are risking leaving some incorrect documentation behind (or breaking some assumptions in other parts of the code).
The fact that `HeaderChain::import_header` method stays the same on the API level, but expects a completely different thing (a fully-verified, final header that builds on top of the best finalized header) worries me.
Also bear in mind, that currently the code is fully functional, and as mentioned earlier I think we should rather do a complete overhaul in the future, then partially change existing things given they are used in a different context.
The ongoing audit is another reason - the main point of introducing new pallet was to avoid touching existing code, and I'd rather keep it this way.

I know none of these arguments is actually really strong one, most of them are just based on gut feelings, but perhaps that's enough to convince you :)

HCastano: Okay so I think there's a small misunderstanding here. The `HeaderChain` trait was introduced  and implemented on `substrate_bridge_pallet::Module` as part of #629. The methods it exposes are not used by any old code (only the Finality Verifier pallet calls them). The relayer uses the `import_signed_header` and `finalize_header` dispatchables. Changing `HeaderChain` and its implementation doesn't affect any exiting functionality.

tomusdrw: >  Changing HeaderChain and its implementation doesn't affect any exiting functionality.

...most likely.

The same argument applies for the header pallet - since it's not used in any bridge we could freely change it, but again my point is to minimize changes to avoid making some weird mistakes - we shouldn't need to carefully review if the code is used or not, or if the assumptions are not getting broken by introducing this change or another.
But again, that's just my gut feeling :)

modules/substrate/src/verifier.rs
```diff
@@ -303,6 +307,77 @@ where
...
+		// Since we want to use the existing storage infrastructure we need to indicate the fork
+		// that we're on. Since we assume that everything in this function is on the same fork we'll
+		// just write everything to a dummy fork.
+		let dummy_fork_hash = H::Hash::default();
```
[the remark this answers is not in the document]
HCastano: Yes, good catch!

I've added a storage item for the hash of the header which was used to bootstrap the pallet, and all the checks are now done using that "fork". It's a little awkward, but it's a small item that gets written at genesis, and I don't think we'll need to keep it around in the future.

modules/substrate/src/lib.rs
HCastano: Take another look at the updated code and let me know if this is still bugging you

modules/substrate/src/lib.rs
```diff
@@ -371,15 +373,79 @@ impl<T: Config> bp_header_chain::HeaderChain<BridgedHeader<T>> for Module<T> {
...
+			import_header_unchecked::<_, T>(&mut storage, header);
+		}
+	}
+}
 
-		Ok(())
+/// Import a finalized header without checking if this is true.
+///
+/// This function assumes that all the given header has already been proven to be valid and
+/// finalized. Using this assumption it will write them to storage with minimal checks. That
```
tomusdrw, line 388:
```suggestion
/// finalized, and a child of current best finalized header. Using this assumption it will write them to storage with minimal checks. That
```

primitives/header-chain/src/lib.rs
```diff
@@ -75,11 +75,11 @@ pub trait HeaderChain<H> {
...
 	fn authority_set() -> AuthoritySet;
 
-	/// Write the given header to the underlying pallet storage.
+	/// Write a finalized chain of headers to the underlying pallet storage.
 	///
-	/// It is assumed that this header has been finalized.
-	#[allow(clippy::result_unit_err)]
-	fn import_header(header: H) -> Result<(), ()>;
+	/// It is assumed that each header in this chain been finalized, and that the given headers are
+	/// in order (e.g vec![header_1, header_2, ..., header_n]).
```
tomusdrw, line 81: I'd add a similar comment as in `import_header_unchecked` - the caller has to ensure that `header_1` is a child of current best finalized header or we should add this simply check to the function implementation.

modules/substrate/src/lib.rs
```diff
@@ -371,15 +373,79 @@ impl<T: Config> bp_header_chain::HeaderChain<BridgedHeader<T>> for Module<T> {
...
+/// signals in the digest are well formed.
+fn import_header_unchecked<S, T>(storage: &mut S, header: BridgedHeader<T>)
+where
+	S: BridgeStorage<Header = BridgedHeader<T>>,
+	T: Config,
+{
+	// Since we want to use the existing storage infrastructure we need to indicate the fork
+	// that we're on. We will assume that since we are using the unchecked import there are no
+	// forks, and can indicate that by using the first imported header's "fork".
+	let dummy_fork_hash = <InitialHash<T>>::get();
```
tomusdrw, line 403: Can you elaborate why this extra storage item has to be introduced? I.e. what problem exactly does it solve?

modules/finality-verifier/src/lib.rs
```diff
@@ -126,18 +126,9 @@ decl_module! {
...
-					if T::HeaderChain::import_header(header).is_err() {
-						return TransactionOutcome::Rollback(Err(<Error<T>>::FailedToWriteHeader))
-					}
-				}
-
-				TransactionOutcome::Commit(Ok(()))
-			})?;
+			// Note that this won't work if we ever change the `ancestry_proof` format to be
+			// sparse since this expects a contiguous set of finalized headers.
+			T::HeaderChain::append_finalized_chain(ancestry_proof);
```
tomusdrw, line 131: We've lost the transactional storage writes here, but we assume that the function is infallible (afaict it only panics in case of internal inconsistency)?

modules/substrate/src/lib.rs
HCastano: A wrote a little bit about it [here](https://github.com/paritytech/parity-bridges-common/pull/672#discussion_r562936774).

But the tl;dr is that I don't want to change any of the exiting storage items, so by adding this we can re-use `NextScheduledChange`.

modules/finality-verifier/src/lib.rs
HCastano: Yeah, there's really no reason why this should fail given our assumptions

primitives/header-chain/src/lib.rs
HCastano: > ... or we should add this simply check to the function implementation

Yeah, I guess the overhead of the extra storage read isn't that big a deal if it prevents people from accidentally bricking their chain

modules/substrate/src/lib.rs
```diff
@@ -151,6 +153,8 @@ decl_error! {
 		Halted,
 		/// The pallet has already been initialized.
 		AlreadyInitialized,
+		/// The given header is not a decendent of a particular header.
+		NotDecendent,
```
tomusdrw, line 157:
```suggestion
		/// The given header is not a descendent of a particular header.
		NotDescendent,
```
?

tomusdrw (approved):

modules/substrate/src/lib.rs
HCastano: Good catch, you were [close](https://grammarist.com/usage/descendant-descendent/) (and closer than I was, lol).

modules/finality-verifier/src/lib.rs
```diff
@@ -119,22 +119,10 @@ pub mod pallet {
...
-
-				if T::HeaderChain::import_finality_proof(finality_target, justification).is_err() {
-					return TransactionOutcome::Rollback(Err(<Error<T>>::FailedToWriteFinalityProof));
-				}
-
-				TransactionOutcome::Commit(Ok(()))
-			})?;
+			// Note that this won't work if we ever change the `ancestry_proof` format to be
+			// sparse since this expects a contiguous set of finalized headers.
+			let _ =
```
svyatonik, line 124: You have both `finality_target` and `ancestry_proof` in parameters. But you only `append` and `finalize` headers from `ancestry_proof`. I suppose that `finality_target` is already in the `ancestry_proof`, right? So why to have that extra `finality_target` parameter? Given that `ancestry_proof` needs to be ordered (iiuc), probably it is possible to just assume that `let finality_target = ancestry_proof.last();`?

modules/substrate/src/lib.rs
```diff
@@ -371,27 +375,93 @@ impl<T: Config> bp_header_chain::HeaderChain<BridgedHeader<T>> for Module<T> {
...
+	) -> Result<(), sp_runtime::DispatchError> {
+		let mut storage = PalletStorage::<T>::new();
+
+		let mut header_iter = headers.into_iter().peekable();
+		let first_header = header_iter.peek().ok_or(Error::<T>::NotDescendant)?;
+		let best_finalized = storage.best_finalized_header().header;
+
+		// Quick ancestry check to make sure we're not writing complete nonsense to storage
+		ensure!(
+			best_finalized.hash() == *first_header.parent_hash(),
```
svyatonik, line 389: Not for this PR, but probably you should replace `storage.best_finalized_header().header.hash()` with something like `storage.best_finalized_header_hash()` - you'll only need single read in this case, instead of two reads + decode + hash.

modules/finality-verifier/src/lib.rs
svyatonik: Also probably worth explicit mention that `ancestry_proof` must be finalized?

modules/substrate/src/lib.rs
```diff
@@ -371,27 +375,93 @@ impl<T: Config> bp_header_chain::HeaderChain<BridgedHeader<T>> for Module<T> {
...
+		// If we are unable to enact an authority set it means our storage entry for scheduled
+		// changes is missing. Best to crash since this is likely a bug.
+		let _ = storage.enact_authority_set(dummy_fork_hash).expect(ENACT_SET_PROOF);
 	}
+
+	storage.update_best_finalized(header.hash());
+
+	storage.write_header(&ImportedHeader {
+		header,
+		requires_justification: false,
```
svyatonik, line 461: Just trying to check that we are ok under all conditions - we're assuming that the header does not `requires_justification`, because we have checked justification in the finality verifier pallet. And we rely on the fact that justification of the transition header must be singed by the current authority set. So if justification is provided for the header from the next session, it will be rejected? Am I right?

So if `H` is transition header, then we will reject transaction with headers `H-2, H-1, H, H+1`, because `H+1` is signed by the next set.

modules/finality-verifier/src/lib.rs
svyatonik: > Also probably worth explicit mention that ancestry_proof must be finalized?

Ok, I see it has been discussed already, but I haven't found any checks except that one:
```rust
		ensure!(
			best_finalized.hash() == *first_header.parent_hash(),
			Error::<T>::NotDescendant,
		);
```

But it only checks first header? Or am I missing something?

svyatonik (approved):

modules/finality-verifier/src/lib.rs
HCastano: > You have both finality_target and ancestry_proof in parameters. But you only append and finalize headers from ancestry_proof. I suppose that finality_target is already in the ancestry_proof, right? So why to have that extra finality_target parameter? Given that ancestry_proof needs to be ordered (iiuc), probably it is possible to just assume that let finality_target = ancestry_proof.last();?

Yeah, you're right - we could make that assumption. Again, we would need to keep this assumption in mind if/when we do change the proof structure.

I made the change, but afterwards I realized I didn't really like ditching the `finality_target` argument. The reason is that with the change `ancestry_proof` hides a bit of information. Specifically, it's "hiding" which header the justification is meant to be targeting. Sure, we could say in the documentation that the justification should always be targeting `ancestry_proof.last()` but I'd rather keep it explicit.

If/when we do change to using a more flexible proof structure we'd likely need to re-introduce the `finality_target` anyways.

HCastano: >  > Also probably worth explicit mention that ancestry_proof must be finalized?
>
> Ok, I see it has been discussed already, but I haven't found any checks except that one:
>
>		ensure!(
>			best_finalized.hash() == *first_header.parent_hash(),
>			Error::<T>::NotDescendant,
>		);
>
> But it only checks first header? Or am I missing something?

So that check is a quick safety check to make sure that the given list of headers is not _totally_ wrong (e.g from a different fork) before we write them to storage. However, since an unchecked write function is being used it is up to the sender of the headers to be careful and first use something like `pallet-finality-verifier` before calling it.

modules/substrate/src/lib.rs
HCastano: In the finality verifier pallet we check the justification against the current authority set known to the bridge pallet. So in the scenario you're describing the authority set which signed the given justification and the current authority set would be different, and we would therefore reject the header.

modules/substrate/src/lib.rs
HCastano: Oh good point, I forgot that we had that available as a storage item

modules/finality-verifier/src/lib.rs
svyatonik: > I made the change, but afterwards I realized I didn't really like ditching the finality_target argument. The reason is that with the change ancestry_proof hides a bit of information. Specifically, it's "hiding" which header the justification is meant to be targeting. Sure, we could say in the documentation that the justification should always be targeting ancestry_proof.last() but I'd rather keep it explicit.

Hmmm. But you're then importing all headers from the ancestry proof && mark them as finalized :) So the only header that may be 'best finalized' is the `.last()`, right? I'm probably missing something - sorry :/

> So that check is a quick safety check to make sure that the given list of headers is not totally wrong (e.g from a different fork) before we write them to storage. However, since an unchecked write function is being used it is up to the sender of the headers to be careful and first use something like pallet-finality-verifier before calling it.

:_( I actually meant `ordered` here, not finalized - sorry, my hands somteimes losting connections to my mind. So what I'm asking here is whether you have check if headers in `ancestry_proof` are sorted (ASC)? Because after verifying the justification you're importing headers assuming they come in-order. What I mean is that you're overwriting `scheduled_change` in `import_unchecked_header`. So what disturbs me is that if `AncestryChecker` is able to work with unordered ancestry proof + submitter provides unordered headers in `ancestry_proof`, then we will be importing these headers out of order => there probably will be situation when we will be able to overwrite `scheduled_change`. This seems unexploitablem though, in current conditions. But still - probably worth a check that headers are orderd.

HCastano: > Hmmm. But you're then importing all headers from the ancestry proof && mark them as finalized :) So the only header that may be 'best finalized' is the .last(), right? I'm probably missing something - sorry :/

You're not missing anything here, `finality_target == ancestry_proof.last()`, and after importing `finality_target == best_finalized`. I still want to keep `finality_target` in order to be explicit.

> :_( I actually meant ordered here, not finalized - sorry, my hands sometimes losing connections to my mind.

⌨️🤪

> So what I'm asking here is whether you have check if headers in ancestry_proof are sorted (ASC)? Because after verifying the justification you're importing headers assuming they come in-order. What I mean is that you're overwriting scheduled_change in import_unchecked_header. So what disturbs me is that if AncestryChecker is able to work with unordered ancestry proof + submitter provides unordered headers in ancestry_proof, then we will be importing these headers out of order => there probably will be situation when we will be able to overwrite scheduled_change. This seems unexploitable though, in current conditions. But still - probably worth a check that headers are ordered.

We don't check if they're ordered, so your concern with the scheduled changes is totally valid. I'll need to do a bit of thinking about the best way to handle this.

HCastano: I've opened #683 to track this.

Issue #670, opened 2021-01-21
Storage proof size must be accounted in `receive_*_proof` functions weight formulas


svyatonik:
So right now both `receive_messages_proof` and `receive_messages_delivery_proof` weight formulas rely on the fact that the relayer has provided proof exactly for given messages count/single inbound lane data accordingly. But actually it may provide additional trie nodes in the proof, resulting in additional `hash()` calls that aren't accounted by the weight formula (thus maybe causing delay in block production times => slashing, ...).

The idea is to introduce benchmark that'll used to build `weight(proof-size-unit)` function && then charge relayer for `cmp::max(messages-count * weight(single-message), proof_size * weight(proof-size-unit))`.

svyatonik:
closed by #679

PR #675, opened 2021-01-25
Add proof-size related benchmarks to message lane module

svyatonik:
part of #670 - while I still have no solution for described issue, having these benchmarks may give information on how proof size affects tx weight.

This PR also fixes issues with existing benchmarks that were broken after weight formulas were introduces.


modules/message-lane/src/benchmarking.rs
```diff
@@ -28,13 +28,25 @@ use frame_system::RawOrigin;
...
 
 /// Module we're benchmarking here.
 pub struct Module<T: Config<I>, I: crate::Instance>(crate::Module<T, I>);
 
+/// Proof size requirements.
+pub enum ProofSize {
+	/// The proof is expected to be minimal.
+	Minimal,
+	/// The proof is expected to have at least given size and grow by increasing number of trie nodes
+	/// included in the prof.
```
HCastano, line 43:
```suggestion
	/// included in the proof.
```

modules/message-lane/src/benchmarking.rs
```diff
@@ -28,13 +28,25 @@ use frame_system::RawOrigin;
...
-const MESSAGE_FEE: u32 = 1_000_000;
+pub const MESSAGE_FEE: u64 = 10_000_000_000;
 
 const SEED: u32 = 0;
 
 /// Module we're benchmarking here.
 pub struct Module<T: Config<I>, I: crate::Instance>(crate::Module<T, I>);
 
+/// Proof size requirements.
+pub enum ProofSize {
```
HCastano, line 39: I don't quite understand the difference between `HasExtraNodes` and `HasLargeLeaf` - can you give me an example or something to help clarify?

modules/message-lane/src/benchmarking.rs
```diff
@@ -522,6 +532,80 @@ benchmarks_instance! {
...
+	}
+
+	// Benchmark `receive_messages_proof` extrinsic with single minimal-weight message and following conditions:
+	// * proof does not include outbound lane state proof;
+	// * inbound lane already has state, so it needs to be read and decoded;
+	// * message is successfully dispatched;
+	// * message requires all heavy checks done by dispatcher.
+	//
+	// Results of this benchmark may be used to check how large (extra) leaf in proof affect transaction performance.
+	receive_message_proofs_with_large_leaf {
```
HCastano, line 579: Just wondering, why don't you need a benchmark with`ProofSize::Minimal`?

modules/message-lane/src/lib.rs
```diff
@@ -110,7 +110,7 @@ pub trait Config<I = DefaultInstance>: frame_system::Config {
 	/// Payload type of outbound messages. This payload is dispatched on the bridged chain.
 	type OutboundPayload: Parameter + Size;
 	/// Message fee type of outbound messages. This fee is paid on this chain.
-	type OutboundMessageFee: Default + From<u32> + Parameter + SaturatingAdd + Zero;
+	type OutboundMessageFee: Default + From<u64> + PartialOrd + Parameter + SaturatingAdd + Zero;
```
HCastano, line 113: Why are we changing `From<u32>` to `From<u64>`?

HCastano (approved):
Looks alright, but since I'm unfamiliar with the storage proof stuff I'd get another pair of eyes on this before merging

modules/message-lane/src/benchmarking.rs
svyatonik: Imagine binary tree and operation `binary_tree.encode()` that avoids any optimizations and encodes nodes, not values - i.e. pointers are also included in in the encoding. So there are two kinds of grow: (1) simply add 1 large value to the tree => since we include values into encoding, encoded value will be larger (2) insert many 0-byte values => number of nodes will grow and since we put internal information in the encoding, encoded value will be larger. So `HasExtraNodes` is the latter and `HasLargeLeaf` is the former. With these benchmarks, you may see that extra-nodes way has larger impact on performance, but the difference is not that big.

modules/message-lane/src/benchmarking.rs
svyatonik: All other benchmarks are running with `Minimal` :)

modules/message-lane/src/lib.rs
svyatonik: The payment for message in some benchmarks doesn't fit `u32` :/ As I noted in the description, right now benchmarks are broken and this is a fix

tomusdrw (approved):
lgtm!

PR #679, opened 2021-01-27
Account proof size in weight formula

svyatonik:
on top of #675 (this PR includes #675)

What is missing yet:
- [x] separate document with description of module benchmarks and weights + why we choose this approach;
- [x] description + either tests or some asserts that would describe why `send_message` won't accept messages that can't be accepted by `receive_messages_proof`;
- [x] ~description of why relay will be able to relay all messages accepted by `send_message`~ there's test in relay code;
- [x] tests?
- [x] fixing issues + TODOs.

## Weights of module extrinsics

The main assumptions behind weight formulas is:
- all possible costs are paid in advance by the message submitter;
- whenever possible, relayer tries to minimize cost of its transactions. So e.g. even though sender always pays for delivering outbound lane state proof, relayer may not include it in the delivery transaction (unless message lane module on target chain requires that);
- weight formula should incentivize relayer to not to submit any redundand data in the extrinsics arguments;
- the extrinsic shall never be executing slower (i.e. has larger actual weight) than defined by the formula.

### Weight of `send_message` call

#### Related benchmarks

| Benchmark                         | Description                                            |
|-----------------------------------|--------------------------------------------------------|
| `send_minimal_message_worst_case` | Sends 0-size message with worst possible conditions    |
| `send_1_kb_message_worst_case`    | Sends 1KB-size message with worst possible conditions  |
| `send_16_kb_message_worst_case`   | Sends 16KB-size message with worst possible conditions |

#### Weight formula

The weight formula is:
```
Weight = BaseWeight + MessageSizeInKilobytes * MessageKiloByteSendWeight
```

Where:

| Component                   | How it is computed?                                                          | Description                                                                                                                                  |
|-----------------------------|------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------|
| `SendMessageOverhead`       | `send_minimal_message_worst_case`                                            | Weight of sending minimal (0 bytes) message                    |
| `MessageKiloByteSendWeight` | `(send_16_kb_message_worst_case - send_1_kb_message_worst_case)/15` | Weight of sending every additional kilobyte of the message |

### Weight of `receive_messages_proof` call

#### Related benchmarks

| Benchmark                                               | Description*                                                                                                            |
|---------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------|
| `receive_single_message_proof`                          | Receives proof of single `EXPECTED_DEFAULT_MESSAGE_LENGTH` message                                                      |
| `receive_two_messages_proof`                            | Receives proof of two identical `EXPECTED_DEFAULT_MESSAGE_LENGTH` messages                                              |
| `receive_single_message_proof_with_outbound_lane_state` | Receives proof of single `EXPECTED_DEFAULT_MESSAGE_LENGTH` message and proof of outbound lane state at the source chain |
| `receive_single_message_proof_1_kb`                     | Receives proof of single message. The proof has size of approximately 1KB**                                             |
| `receive_single_message_proof_16_kb`                    | Receives proof of single message. The proof has size of approximately 16KB**                                            |

*\* - In all benchmarks all received messages are dispatched and their dispatch cost is near to zero*

*\*\* - Trie leafs are assumed to have minimal values. The proof is derived from the minimal proof by including more trie nodes. That's because according to `receive_message_proofs_with_large_leaf` and `receive_message_proofs_with_extra_nodes` benchmarks, increasing proof by including more nodes has slightly larger impact on performance than increasing values stored in leafs*.

#### Weight formula

The weight formula is:
```
Weight = BaseWeight + OutboundStateDeliveryWeight + MessagesCount * MessageDeliveryWeight + MessagesDispatchWeight + Max(0, ActualProofSize - ExpectedProofSize) * ProofByteDeliveryWeight
```

Where:

| Component                     | How it is computed?                                                                      | Description                                                                                                                                                                                                                                                                                                                                                                                         |
|-------------------------------|------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `BaseWeight`                  | `2*receive_single_message_proof - receive_two_messages_proof`                            | Weight of receiving and parsing minimal proof                                                                                                                                                                                                                                                                                                                                                       |
| `OutboundStateDeliveryWeight` | `receive_single_message_proof_with_outbound_lane_state - receive_single_message_proof`   | Additional weight when proof includes outbound lane state                                                                                                                                                                                                                                                                                                                                           |
| `MessageDeliveryWeight`       | `receive_two_messages_proof - receive_single_message_proof`                              | Weight of of parsing and dispatching (without actual dispatch cost) of every message                                                                                                                                                                                                                                                                                                                |
| `MessagesCount`               |                                                                                          | Provided by relayer                                                                                                                                                                                                                                                                                                                                                                                 |
| `MessagesDispatchWeight`      |                                                                                          | Provided by relayer                                                                                                                                                                                                                                                                                                                                                                                 |
| `ActualProofSize`             |                                                                                          | Provided by relayer                                                                                                                                                                                                                                                                                                                                                                                 |
| `ExpectedProofSize`           | `EXPECTED_DEFAULT_MESSAGE_LENGTH * MessagesCount + EXTRA_STORAGE_PROOF_SIZE`             | Size of proof that we are expecting. This only includes `EXTRA_STORAGE_PROOF_SIZE` once, because we assume that intermediate nodes likely to be included in the proof only once. This may be wrong, but since weight of processing proof with many nodes is almost equal to processing proof with large leafs, additional cost will be covered because we're charging for extra proof bytes anyway  |
| `ProofByteDeliveryWeight`     | `(receive_single_message_proof_16_kb - receive_single_message_proof_1_kb) / (15 * 1024)` | Weight of processing every additional proof byte over `ExpectedProofSize` limit                                                                                                                                                                                                                                                                                                                     |

#### Why for every message sent using `send_message` we will be able to craft `receive_messages_proof` transaction?

We have following checks in `send_message` transaction on the source chain:
- message size should be less than or equal to `2/3` of maximal extrinsic size on the target chain;
- message dispatch weight should be less than or equal to the `1/2` of maximal extrinsic dispatch weight on the target chain.

Delivery transaction is an encoded delivery call and signed extensions. So we have `1/3` of maximal extrinsic size reserved for:
- storage proof, excluding the message itself. Currently, on our test chains, the overhead is always within `EXTRA_STORAGE_PROOF_SIZE` limits (1024 bytes);
- signed extras and other call arguments (`relayer_id: SourceChain::AccountId`, `messages_count: u32`, `dispatch_weight: u64`).

On Millau chain, maximal extrinsic size is `0.75 * 2MB`, so `1/3` is `512KB` (`524_288` bytes). This should be enough to cover these extra arguments and signed extensions.

Let's exclude message dispatch cost from single message delivery transaction weight formula:
```
Weight = BaseWeight + OutboundStateDeliveryWeight + MessageDeliveryWeight + Max(0, ActualProofSize - ExpectedProofSize) * ProofByteDeliveryWeight
```

So we have `1/2` of maximal extrinsic weight to cover these components. `BaseWeight`, `OutboundStateDeliveryWeight` and `MessageDeliveryWeight` are determined using benchmarks and are hardcoded into runtime. Adequate relayer would only include required trie nodes into the proof. So if message size would be maximal (`2/3` of `MaximalExtrinsicSize`), then the extra proof size would be `MaximalExtrinsicSize / 3 * 2 - EXPECTED_DEFAULT_MESSAGE_LENGTH`.

Both conditions are verified by `pallet_message_lane::ensure_weights_are_correct` and `pallet_message_lane::ensure_able_to_receive_messages` functions, which must be called from every runtime' tests.

### Weight of `receive_messages_delivery_proof` call

#### Related benchmarks

| Benchmark                                                   | Description                                                                              |
|-------------------------------------------------------------|------------------------------------------------------------------------------------------|
| `receive_delivery_proof_for_single_message`                 | Receives proof of single message delivery                                                |
| `receive_delivery_proof_for_two_messages_by_single_relayer` | Receives proof of two messages delivery. Both messages are delivered by the same relayer |
| `receive_delivery_proof_for_two_messages_by_two_relayers`   | Receives proof of two messages delivery. Messages are delivered by different relayers    |

#### Weight formula

The weight formula is:
```
Weight = BaseWeight + MessagesCount * MessageConfirmationWeight + RelayersCount * RelayerRewardWeight + Max(0, ActualProofSize - ExpectedProofSize) * ProofByteDeliveryWeight
```

Where:

| Component                 | How it is computed?                                                                                                   | Description                                                                                                                                                                                             |
|---------------------------|-----------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `BaseWeight`              | `2*receive_delivery_proof_for_single_message - receive_delivery_proof_for_two_messages_by_single_relayer`             | Weight of receiving and parsing minimal delivery proof                                                                                                                                                  |
| `MessageDeliveryWeight`   | `receive_delivery_proof_for_two_messages_by_single_relayer - receive_delivery_proof_for_single_message`               | Weight of confirming every additional message                                                                                                                                                           |
| `MessagesCount`           |                                                                                                                       | Provided by relayer                                                                                                                                                                                     |
| `RelayerRewardWeight`     | `receive_delivery_proof_for_two_messages_by_two_relayers - receive_delivery_proof_for_two_messages_by_single_relayer` | Weight of rewarding every additional relayer                                                                                                                                                            |
| `RelayersCount`           |                                                                                                                       | Provided by relayer                                                                                                                                                                                     |
| `ActualProofSize`         |                                                                                                                       | Provided by relayer                                                                                                                                                                                     |
| `ExpectedProofSize`       | `EXTRA_STORAGE_PROOF_SIZE`                                                                                            | Size of proof that we are expecting                                                                                                                                                                     |
| `ProofByteDeliveryWeight` | `(receive_single_message_proof_16_kb - receive_single_message_proof_1_kb) / (15 * 1024)`                              | Weight of processing every additional proof byte over `ExpectedProofSize` limit. We're using the same formula, as for message delivery, because proof mechanism is assumed to be the same in both cases |

#### Why we're always able to craft `receive_messages_delivery_proof` transaction?

There can be at most `<PeerRuntime as pallet_message_lane::Config>::MaxUnconfirmedMessagesAtInboundLane` messages and at most `<PeerRuntime as pallet_message_lane::Config>::MaxUnrewardedRelayerEntriesAtInboundLane` unrewarded relayers in the single delivery confirmation transaction.

We're checking that this transaction may be crafted in the `pallet_message_lane::ensure_able_to_receive_confirmation` function, which must be called from every runtime' tests.


modules/message-lane/README.md
```diff
@@ -0,0 +1,132 @@
+# Message Lane Module
```
HCastano, line 1: I wanna start off by saying that I really appreciate the effort you put into this document. I think it'll be super useful to anybody that comes along and wants to understand the complexities of the `message-lane` benchmarking.

However, I do think there is a lot of room for improvement in terms of the way it is written. 

1. Keeping line lengths within our specified `.editorconfig` guidelines

I'm pretty sensitive to this since I read basically everything I can inside of my terminal instead of hopping on GitHub or whatever - so I might be a little bias, but hear me out. Even using my full monitor's width there are lines I can't read without scrolling to the right. I would expect that most people that want to read this document will not be doing so from GitHub, but will instead be looking at it from their code editors (probably alongside the actual benchmarking code). Having it be legible in this way would (in my opinion) be a nice quality of life improvement.

2. Moving longer descriptions and formulas out of tables

As a way to partially mitigate (1) I would recommend that we move some content out of tables and into standalone paragraphs or sections. This would make the content more legible in general. As a concrete example, consider the description for `ExpectedProofSize`. It's a good description, but when rendered it is squished within the tiny table box making it hard to read.

3. A section on how to run, interpret, and update weights (could be done in a different PR)

Part of the key of benchmarks is that they're able to adapt given a changing environment. In order to achieved that we need a detailed, reproducible, and somewhat dummy-proof set of instructions that others can use. Even if we link to the Substrate benchmarking doc and add a few sentences I think that would be better than having nothing.

I do have to admit I haven't dove into the _content_ of the doc yet, but I will read it carefully in the coming days :).

relays/substrate-client/src/client.rs
```diff
@@ -234,14 +234,14 @@ impl<C: Chain> Client<C> {
 		instance: InstanceId,
 		lane: LaneId,
 		at_block: C::Hash,
-	) -> Result<StorageProof> {
+	) -> Result<Vec<Vec<u8>>> {
```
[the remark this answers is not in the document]
svyatonik: We need an approximate size of the proof && `StorageSize` only provides a consuming way to do that.

modules/message-lane/README.md
HCastano: @svyatonik Why'd you remove the README file and put it in the PR description? I think the code is the right place to have the explanations of how the weights are being computed. It just needed some "beautifying"

svyatonik: Sorry - I was in a hurry yesterday, so haven't left an explanation. In brief: I don't think we have a time for "beautifying" readmes right now && I do not think we have a time for arguing about whether readme is "beautiful" enough or not. So let it lay in PR description for now, until we'll have some spare time for working on beautiful readmes.

HCastano: Personally, I think this is one of those cases where the upfront work will pay dividends in the future. I'd rather have this done right now, even if time is a bit tight, than in 6 months when none of us will remember what magic went into the benchmarking.

At the very least I think this should be part of the codebase where it's easily accessible to those working on benchmarking, even if it's not quite at the cleaned up state where I think it should be (i.e the improvements from my first comment).

HCastano (approved):
Should probably get another pair of eyes on this before merging

HCastano:
If nobody else reviews this by EOD I'll just merge it

tomusdrw:
> If nobody else reviews this by EOD I'll just merge it

No, please don't. I'm reviewing the code, but it's not trivial and takes a while.

bin/runtime-common/src/messages.rs
```diff
@@ -160,7 +163,23 @@ pub mod source {
 	/// - hash of finalized header;
 	/// - storage proof of inbound lane state;
 	/// - lane id.
-	pub type FromBridgedChainMessagesDeliveryProof<B> = (HashOf<BridgedChain<B>>, StorageProof, LaneId);
+	#[derive(Clone, Decode, Encode, Eq, PartialEq, RuntimeDebug)]
+	pub struct FromBridgedChainMessagesDeliveryProof<BridgedHeaderHash> {
+		pub bridged_header_hash: BridgedHeaderHash,
```
tomusdrw, line 168:
```suggestion
		/// Hash of the bridge header the proof is for.
		pub bridged_header_hash: BridgedHeaderHash,
```

bin/runtime-common/src/messages.rs
```diff
@@ -160,7 +163,23 @@ pub mod source {
...
+		pub storage_proof: RawStorageProof,
```
tomusdrw, line 169:
```suggestion
		/// Storage trie proof generated for [`Self::bridged_header_hash`].
		pub storage_proof: RawStorageProof,
```

bin/runtime-common/src/messages.rs
```diff
@@ -160,7 +163,23 @@ pub mod source {
...
+		pub lane: LaneId,
```
tomusdrw, line 170:
```suggestion
		/// Lane id of which messages were delivered and the proof is for.
		pub lane: LaneId,
```

bin/runtime-common/src/messages.rs
```diff
@@ -332,13 +355,25 @@ pub mod target {
...
-	pub type FromBridgedChainMessagesProof<B> = (
-		HashOf<BridgedChain<B>>,
-		StorageProof,
-		LaneId,
-		MessageNonce,
-		MessageNonce,
-	);
+	#[derive(Clone, Decode, Encode, Eq, PartialEq, RuntimeDebug)]
+	pub struct FromBridgedChainMessagesProof<BridgedHeaderHash> {
+		pub bridged_header_hash: BridgedHeaderHash,
```
tomusdrw, line 360:
```suggestion
		/// Hash of the finalized bridged header the proof is for.
		pub bridged_header_hash: BridgedHeaderHash,
```

bin/runtime-common/src/messages.rs
```diff
@@ -332,13 +355,25 @@ pub mod target {
...
+		pub storage_proof: RawStorageProof,
```
tomusdrw, line 361:
```suggestion
		/// A storage trie proof of messages being delivered.
		pub storage_proof: RawStorageProof,
```

bin/runtime-common/src/messages.rs
```diff
@@ -332,13 +355,25 @@ pub mod target {
...
+		pub lane: LaneId,
+		pub nonces_start: MessageNonce,
```
tomusdrw, line 363:
```suggestion
		/// Nonce of the first message being delivered.
		pub nonces_start: MessageNonce,
```

bin/runtime-common/src/messages.rs
```diff
@@ -332,13 +355,25 @@ pub mod target {
...
+		pub nonces_end: MessageNonce,
```
tomusdrw, line 364:
```suggestion
		/// Nonce of the last message being delivered.
		pub nonces_end: MessageNonce,
```

primitives/runtime/src/lib.rs
```diff
@@ -110,3 +111,9 @@ pub trait Size {
 	/// accurate.
 	fn size_hint(&self) -> u32;
 }
+
+impl Size for usize {
```
tomusdrw, line 115: That implementation might be a bit misleading, no? The trait documentation says the function should return "approximate size of this object in bytes", not perform a `usize -> u32` conversion.

Is it only used for tests? Perhaps maybe better to have a wrapper type?

tomusdrw (approved):
lgtm! Merge at will after addressing the doc comments.

tomusdrw (approved):

PR #684, opened 2021-01-27
Bump Substrate Dependency

HCastano:
This bump of Substrate isn't to fix the `cargo-deny` issues we've been seeing, but
instead to get some new utilities from `frame_system::mocking` to improve our test setup.


HCastano:
CI failures are unrelated to the PR (`sccache` failures).

bin/millau/node/src/service.rs
```diff
@@ -69,13 +71,17 @@ pub fn new_partial(
 				AuraPair,
 			>,
 			sc_finality_grandpa::LinkHalf<Block, FullClient, FullSelectChain>,
+			Option<TelemetrySpan>,
 		),
 	>,
 	ServiceError,
 > {
+	if config.keystore_remote.is_some() {
```
adoerr, line 79: I was wondering, if we even want to support a remote keystore at all for Rialto and Millau nodes.

bin/rialto/node/src/service.rs
```diff
@@ -69,13 +71,17 @@ pub fn new_partial(
 				AuraPair,
 			>,
 			sc_finality_grandpa::LinkHalf<Block, FullClient, FullSelectChain>,
+			Option<TelemetrySpan>,
 		),
 	>,
 	ServiceError,
 > {
+	if config.keystore_remote.is_some() {
```
adoerr, line 79: Same here

adoerr (approved):
LGTM

bin/millau/node/src/service.rs
tomusdrw: Definitely not for the testnets, we might consider later on using them for production deployment.

tomusdrw (approved):
lgtm! We just need to get the build sorted.

HCastano:
There seems to be a problem related to message dispatch, not sure if it's related to this PR but I'm changing it to a draft just to be safe.

PR #686, opened 2021-01-28
Make Account Derivation More Unique

HCastano:
Closes https://github.com/paritytech/srlabs_findings/issues/45.

Since `root` and `account` aren't exactly unique prefixes it is possible that other
pallets or tools may opt to use these same prefixes when generating accounts. This could
lead to collisions and in the worst case account takeovers.

This PR makes the prefix used when deriving accounts a bit longer, reducing the chances
of collisions between us and other pallets or tools.

### DIFF
--- a/bin/millau/node/src/chain_spec.rs
+++ b/bin/millau/node/src/chain_spec.rs
@@ -185,6 +185,6 @@ fn derived_dave_account_is_as_expected() {
 	let derived: AccountId = derive_account_from_rialto_id(bp_runtime::SourceAccount::Account(dave));
 	assert_eq!(
 		derived.to_string(),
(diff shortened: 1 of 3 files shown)


tomusdrw (approved):

AsceticBear:
An off-topic question.

Is https://github.com/paritytech/srlabs_findings/issues/45 in a private repo issue? I get a 404 error when I click here.

HCastano:
> An off-topic question.
> 
> Is https://github.com/paritytech/srlabs_findings/issues/45 in a private repo issue? I get a 404 error when I click here.

@AsceticBear Hey, yes it's a private repo. We're trying to write enough of a description in our PRs so that those without access to the repo can still have enough context to follow the changes. Are there any questions you have about this change in particular?

AsceticBear:
> Are there any questions you have about this change in particular?

No, just about the repo link.

Issue #450, opened 2020-10-23
Message dispatch filtering


tomusdrw:
The dispatcher #211 should support a whilelist of calls that can actually go past the bridge. It should go beyond simply .call type, rather we might want to filter out specific parameters as well (to make sure they are in range, or to only have a limtied set of transfers initially (limited in value)).

svyatonik:
Since it is very chain-specific, it probably should be done at the runtime level, where `message-lane` <-> `call-dispatch` communication happens. I.e. in the implementation of `bp_message_lane::MessageDispatch`. So we just add additional filter there, before message is actually passed to the `pallet_bridge_call_dispatch::Module`.

tomusdrw:
I wonder if there is something we could and should do on the source chain as well? We most likely don't want to decode the entire call, but we could simply check the first byte of the payload?

tomusdrw:
This is runtime specific, so should be part of the concrete release (Kusama <> Polkadot).

adoerr:
Would it be useful to allow for dynamic filters (using `frame_support::traits::FilterStack`) as well? This could even go as far as offering a `sudo` call in order to add and remove filters.

tomusdrw:
`Filter` trait covers `FilterStack` as well (see `impl_filter_stack` macro in substrate), so the way it's implemented in #687 is good enough :)

tomusdrw:
Implemented in https://github.com/paritytech/polkadot/tree/bridges-integration-2

PR #687, opened 2021-01-29
Pre-dispatch call filter

svyatonik:
related #450 (iiuc it is more about actual filters in P <-> K)

I've made explicit filter + special event. The same (except for the event) may be achieved by calling `dispatch(Err())`.


modules/call-dispatch/src/lib.rs
```diff
@@ -172,6 +177,8 @@ decl_event!(
 		MessageDispatched(InstanceId, MessageId, DispatchResult),
 		/// We have failed to decode Call from the message.
 		MessageCallDecodeFailed(InstanceId, MessageId),
+		/// The call from the message has been rejected by the call filter.
+		MessageCallRejected(InstanceId, MessageId),
```
adoerr, line 181: Perhaps something like `MessageCallFiltered` in order to make the reason for rejecting the message more explicit?

svyatonik: I have nothing against that, except for the fact we have similar `MessageRejected` event, which rejects "filtered" message before it reaches dispatch mechanism. If you think `MessageCallRejected` it's better - just respond here, I'll change the name.

adoerr: I wasn't aware of `MessageRejected`. If we basically use 'rejected' for 'filtered by some means', then `MessageCallRejected` is better, since it is more consistent.

modules/call-dispatch/src/lib.rs
```diff
@@ -291,6 +298,18 @@ impl<T: Config<I>, I: Instance> MessageDispatch<T::MessageId> for Module<T, I> {
 			return;
 		}
 
+		// filter the call
+		if !T::CallFilter::filter(&call) {
```
tomusdrw, line 302: I wonder if it wouldn't be better to move the filtering before weight checks. I'm expecting that the filtering might be much less complicated than weight calculations and it it has a nice consequence of preventing exploiting incorrectly calculated weights (i.e. you generate excessive load by forcing the chain to calculate it even though the call is filtered out).

tomusdrw (approved):
Looks good! I've made a small suggestion to move the call filter even before weight calculations, I wonder what your thoughts on this are.

HCastano (approved):

tomusdrw (approved):
lgtm! Just the CI gods (fmt) remain to be satisfied.

PR #689, opened 2021-01-29
Allow disabling the monitoring when starting a test network.

tomusdrw:


deployments/run.sh
```diff
@@ -20,29 +20,64 @@ set -xeu
 # Since the Compose commands are using relative paths we need to `cd` into the `deployments` folder.
 cd "$( dirname "${BASH_SOURCE[0]}" )"
 
+function show_help () {
```
HCastano, line 23: Nice touch 👌

deployments/run.sh
```diff
@@ -20,29 +20,64 @@ set -xeu
...
-		;;
-	*) echo "Invalid parameter: $1 (expected poa-rialto/rialto-millau)"; exit 1;;
-esac
+SUB_COMMAND='start'
+for i in "$@"
+do
+  case $i in
+    --no-monitoring)
+      MONITORING=" -f ./monitoring/disabled.yml"
+      shift
```
HCastano, line 50: TIL about `shift`

HCastano (changes requested):
While I like this, the `--no-monitoring` option doesn't work. I run into the following error:

```
❯ ./deployments/run.sh rialto-millau --no-monitoring

...

+ docker-compose --project-directory . ... -f ./monitoring/disabled.yml up -d
Pulling grafana-dashboard (scratch:)...
ERROR: 'scratch' is a reserved name
```

The [image](https://hub.docker.com/_/scratch/) page on Docker Hub says that:

> While scratch appears in Docker’s repository on the hub, you can’t pull it, run it, or tag any image with the name scratch.

We'll need some other "dummy" image to use here.

tomusdrw:
> While I like this, the --no-monitoring option doesn't work. I run into the following error:

Apologies, I was pretty sure I tested it and it worked, but I must have mix something up. Updated to use `alpine` image instead.

HCastano (approved):
Looks good!

PR #696, opened 2021-02-01
Wire Finality Verifier Pallet Into Runtimes

HCastano:
This PR adds the Finality Verifier pallet into the Rialto and Millau runtimes.

One thing that I'd like to hear some opinions on is the `MaxHeadersInSingleProof`
constant. I've added my explanation as to how I got the 150 limit in the comments, but
maybe there's some other factors I need to take into account.


adoerr (approved):
LGTM

primitives/header-chain/src/lib.rs
```diff
@@ -110,3 +112,107 @@ impl<H, P> AncestryChecker<H, P> for () {
...
+			return false;
+		};
+
+		if let Some(last) = proof.last() {
+			if last != child {
+				return false;
+			}
+		} else {
+			return false;
+		};
```
tomusdrw, line 142:
```suggestion
		match proof.last() {
		    Some(last) if last == child => {},
		    _ => return false,
		}
```
likewise

primitives/header-chain/src/lib.rs
```diff
@@ -110,3 +112,107 @@ impl<H, P> AncestryChecker<H, P> for () {
...
+
+		// Now we actually check the proof
+		for i in (1..proof.len()).rev() {
```
tomusdrw, line 145: The `.rev()` stuff doesn't really matter here, but let's leave it out, cause it might seem more intuitive.

primitives/header-chain/src/lib.rs
```diff
@@ -110,3 +112,107 @@ impl<H, P> AncestryChecker<H, P> for () {
...
+			if proof[i - 1].hash() != *proof[i].parent_hash() {
```
tomusdrw, line 146:
```suggestion
			if &proof[i - 1].hash() != proof[i].parent_hash() {
```
No need to copy

primitives/header-chain/src/lib.rs
```diff
@@ -110,3 +112,107 @@ impl<H, P> AncestryChecker<H, P> for () {
...
+		}
+
+		// Let's make sure that the given headers are actually in the proof
+		if let Some(first) = proof.first() {
+			if first != ancestor {
+				return false;
+			}
+		} else {
+			return false;
+		};
```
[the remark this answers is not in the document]
tomusdrw: It seems they are not necessarily cached, so might be worth to do `proof.iter().map(|h| h.hash()).collect()` first and only then do the comparison.

primitives/header-chain/src/lib.rs
```diff
@@ -110,3 +112,107 @@ impl<H, P> AncestryChecker<H, P> for () {
...
 }
+
+/// A simple ancestry checker which verifies ancestry by walking every header between `child` and
+/// `ancestor`.
+pub struct LinearAncestryChecker;
+
+impl<H: HeaderT> AncestryChecker<H, Vec<H>> for LinearAncestryChecker {
+	fn are_ancestors(ancestor: &H, child: &H, proof: &Vec<H>) -> bool {
+		// You can't be your own parent
+		if ancestor == child {
```
tomusdrw, line 123: Given all other checks this can be easily optimised into:
```suggestion
		if proof.len() < 2 {
		    return false;
		}
```
which should be cheaper than exact header comparison

bin/rialto/runtime/src/lib.rs
```diff
@@ -416,6 +416,22 @@ impl pallet_substrate_bridge::Config for Runtime {
 	type BridgedChain = bp_millau::Millau;
 }
 
+parameter_types! {
+	// One hundred and fifty old headers is ~15 mins (assuming a target block time of 6s). This
+	// should give the relayer enough time to get a justification for a header.
```
tomusdrw, line 421: The justification time is not really up to the relayer. Restricting to 150 here means we assume that GRANDPA will produce at least one justification every 15 mins. I'd rather relax this and make it match the session length (GRANDPA has to provide at least one justification for each session).

bin/millau/runtime/src/lib.rs
```diff
@@ -309,6 +309,22 @@ impl pallet_substrate_bridge::Config for Runtime {
 	type BridgedChain = bp_rialto::Rialto;
 }
 
+parameter_types! {
+	// One hundred and fifty old headers is ~15 mins (assuming a target block time of 6s). This
+	// should give the relayer enough time to get a justification for a header.
+	//
+	// Since the ancestry checker we're using doesn't need to access to storage for verification
+	// this can be fairly high.
+	pub const MaxHeadersInSingleProof: u8 = 150;
```
tomusdrw, line 318: same as below - instead of setting this arbitrarily, we should make it depend on the session length. For small header sizes (or larger block times/weights) it might be totally possible to deliver a proof that spans the entire session in a single transaction.

tomusdrw (commented):
I don't like arbitrary limits for the size, I'd rather make it depend on actual session length (in blocks). Also left a few opimisation suggestions.

tomusdrw (changes requested):
Marking as request changes, cause I'd like to at least discuss the suggestions.

primitives/header-chain/src/lib.rs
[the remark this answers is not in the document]
HCastano: > It seems they are not necessarily cached, so might be worth to do proof.iter().map(|h| h.hash()).collect() first and only then do the comparison.

I don't see the advantage of this. If we have a proof where the `first` and `last` headers are fine, but the middle isn't valid we would have ended up iterating over the entire collection needlessly instead of exiting early (current behaviour). Am I missing something?

primitives/header-chain/src/lib.rs
HCastano: I knew there was a more idiomatic way to write this, but couldn't see it yesterday

HCastano:
I added you (@svyatonik) as a reviewer since I'm curious to hear about what you have to say on the `MaxHeadersInSingleProof` value

primitives/rialto/src/lib.rs
```diff
@@ -74,6 +74,24 @@ pub const MAX_SINGLE_MESSAGE_DELIVERY_TX_WEIGHT: Weight = 1_500_000_000;
 /// runtime upgrades.
 pub const MAX_SINGLE_MESSAGE_DELIVERY_CONFIRMATION_TX_WEIGHT: Weight = 2_000_000_000;
 
+/// The length of a session (how often authorities change) on Rialto measured in of number of blocks.
+pub const SESSION_LENGTH: BlockNumber = 4;
+
+/// Re-export `time_units` to make usage easier.
```
svyatonik, line 80: `time_units` aren't used here, right?

svyatonik (approved):
> I added you (@svyatonik) as a reviewer since I'm curious to hear about what you have to say on the `MaxHeadersInSingleProof` value

I think session length is fine, but probably in practice (with real chains) it'll be limited by call weight rather than by this number anyway.

tomusdrw (approved):

primitives/rialto/src/lib.rs
HCastano: No they're not. However, they were part of the old runtime code, and they could be useful in the future if we want to change how some units are defined

PR #698, opened 2021-02-02
Add finality verifier test with different validator set id.

tomusdrw:


modules/finality-verifier/src/lib.rs
```diff
@@ -203,6 +203,27 @@ mod tests {
 		})
 	}
 
+	#[test]
+	fn rejects_ancestry_with_validator_set_transition() {
```
HCastano, line 207: I think the test name could be improved a bit. Something more along the lines of the PR title:

```suggestion
	fn rejects_justification_that_skips_authority_set_transition() {
```

HCastano (approved):

Issue #690, opened 2021-01-29
Failed to Decode Calls During Dispatch


HCastano:
Recently I've been seeing the following errors in our node logs:

```
2021-01-29 20:23:18.039  TRACE tokio-runtime-worker pallet_message_lane:Received lane [0, 0, 0, 0] state update: latest_confirmed_nonce=24
2021-01-29 20:23:18.039  TRACE tokio-runtime-worker pallet_bridge_call_dispatch:Source Account: b3777b4eca64b4ad4647a1043f55275eb206d69f541be51072cb75da40670a3e (5G81vRqU...)
2021-01-29 20:23:18.039  TRACE tokio-runtime-worker pallet_bridge_call_dispatch:Failed to decode Call from message [114, 108, 116, 111]/([0, 0, 0, 0], 28)
2021-01-29 20:23:18.039  TRACE tokio-runtime-worker pallet_message_lane:Received messages: total=1, valid=1
```

```
2021-01-29 20:23:36.014  TRACE tokio-runtime-worker pallet_bridge_call_dispatch:Message [114, 108, 116, 111]/([0, 0, 0, 0], 29): origin proof is invalid. Expected account: 306721211d5404bd9da88e0204360a1a9ab8b87c66c1bc2fcdd37f3c2222cc20 (5DAAnrj7...) from signature: MultiSignature::Sr25519(42b1bc1c0dcc3eb4751adb0b362c62e41ce7ab4f4eb08770f7ebed9e175fb7137e670e97c85202e5beebc815c2047a0014b3129cd628e06e5780e8452a80cc89)
2021-01-29 20:23:36.014  TRACE tokio-runtime-worker pallet_bridge_call_dispatch:Message
[114, 108, 116, 111]/([0, 0, 0, 0], 30): origin proof is invalid. Expected account:
306721211d5404bd9da88e0204360a1a9ab8b87c66c1bc2fcdd37f3c2222cc20 (5DAAnrj7...) from
signature:
MultiSignature::Sr25519(a46134a799381f6ed6114d9ea51afc9046fa957780b563cf32732ca1b209f56f03b4957c93372a4c9e3aaf504d4d2bb6889bab46e6028726982ef0be14cc878e)
```

After a bit of investigation I found out that these errors started happening after #663
was merged (I checked #660 and saw no errors).

I tried to dig into the code a bit to see if I could fix the issue myself, but I wasn't
able to. I suspect that the messages formed by the relayer aren't being encoded
[...]

PR #699, opened 2021-02-02
Fixed call encoding in signature digest

svyatonik:
fixes issue from #690 

I've also swapped signature verification with call decoding - this removes additional requirement && seems more intuitive.

### DIFF
--- a/bin/runtime-common/src/messages.rs
+++ b/bin/runtime-common/src/messages.rs
@@ -795,6 +795,7 @@ mod tests {
 				},
 			}
 		);
+		assert_eq!(Ok(ThisChainCall::Transfer), message_on_this_chain.call.into());
 	}
 
 	const TEST_LANE_ID: &LaneId = b"test";
--- a/bin/runtime-common/src/messages.rs
+++ b/bin/runtime-common/src/messages.rs
@@ -27,7 +27,7 @@ use bp_message_lane::{
 	InboundLaneData, LaneId, Message, MessageData, MessageKey, MessageNonce, OutboundLaneData,
 };
 use bp_runtime::InstanceId;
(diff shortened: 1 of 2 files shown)


tomusdrw (approved):
lgtm! Although it looks weird that we don't need to change any tests? Maybe worth adding some runtime-level tests to make sure the entire integration is working correctly?

bin/runtime-common/src/messages.rs
```diff
@@ -352,9 +352,7 @@ pub mod target {
 
 	impl<B: MessageBridge> From<FromBridgedChainEncodedMessageCall<B>> for Result<CallOf<ThisChain<B>>, ()> {
 		fn from(encoded_call: FromBridgedChainEncodedMessageCall<B>) -> Self {
-			let mut input = &encoded_call.encoded_call[..];
-			let _skipped_length = Compact::<u32>::decode(&mut input).map_err(drop)?;
-			CallOf::<ThisChain<B>>::decode(&mut input).map_err(drop)
+			CallOf::<ThisChain<B>>::decode(&mut &encoded_call.encoded_call[..]).map_err(drop)
```
HCastano, line 355: Can you explain the fix for me? Why don't we need to decode the payload twice anymore?

HCastano (approved):

bin/runtime-common/src/messages.rs
svyatonik: `Vec` Idecoding now happens when struct itself is decoded.

svyatonik:
> lgtm! Although it looks weird that we don't need to change any tests? Maybe worth adding some runtime-level tests to make sure the entire integration is working correctly?

Yeah - I forgot to add this check to existing test.

PR #700, opened 2021-02-03
Generate large messages

### LINKED ISSUE #680: Generate and deliver large messages
Somewhat related to #679 + #500 + final message lane module calls weights.

The idea is that sometimes messages generator should generate:
1) several messages of maximal size;
2) several messages of maximal dispatch weight;
3) several (larger than `MaxUnconfirmedMessagesAtInboundLane`) messages at once.

This would help us to ensure that the relayer and both `message-lane` instances (source and target) are able to deliver messages under this conditions.

svyatonik:
closes #680

I'm not sure yet if it'll end up in our regular deployments - my laptop fan turned into helicopter when I was trying to generate, deliver and dispatch 2048 small messages in both ways. Delivery alerts were also turned red. There's also a few issues that may need to be fixed before/if this PR will be merged. So right now this PR more like TODO for myself + testing environment for my scripts.

UPD: given that ^^^, there's one thing to consider now - whether we want this in our regular deployments, or manual testing is fine? Depending on that, I I'll either update our deployment scripts in next PR, or leave it in my [local scripts](https://github.com/svyatonik/parity-bridges-common.test/blob/master/millau-to-rialto-messages-generator.sh#L54). Pros of updating scripts: we'll be able to (almost automatically) detect if anything will break with edge-case messages . Cons of having that is that the CPU load is a way bigger when large number of messages is sent, so we may see false alerts sometimtes (or we may need to increase alert interval from 5 to 10 minutes).


relays/substrate/src/cli.rs
```diff
@@ -162,12 +162,18 @@ pub enum ToRialtoMessage {
 		#[structopt(long)]
 		amount: bp_rialto::Balance,
 	},
+	/// Make an on-chain remark (comment). The size of remark is the maximal possible size, that can be
```
tomusdrw, line 165: I think we should instead take remark payload as CLI parameter.

relays/substrate/src/main.rs
```diff
@@ -524,8 +555,30 @@ async fn estimate_message_delivery_and_dispatch_fee<Fee: Decode, C: Chain, P: En
 	Ok(decoded_response)
 }
 
+fn compute_maximal_size_of_remark(maximal_source_extrinsic_size: u32, maximal_target_extrinsic_size: u32) -> u32 {
```
tomusdrw, line 558: This could be a helper CLI method.

tomusdrw (commented):
Looks good to me, I'm willing to give it a try on our test deployment, perhaps we don't have to spam the chain that often to make the load super high, but it's definitely worth having.

I'm sorry about CLI refactoring, I didn't notice this PR touches similar parts of the code :(
Regarding the CLI parameters, I'd rather keep it user-centered, so instead of having special "max size" remark type, I think it's better to let the user parameterize that.

relays/substrate/src/cli.rs
svyatonik: Maximal payload is too large to fit the command line + I don't want to make this even more complex by allowing to submit remark with content read from file or something similar. This is simple machinery, mostly built for autogenerating messages, not for sending arbitrary user-provided messages. So I've added couple of optional args - `--remark-size=<N>` and `--dispatch-weight=<N>` for providing message size+weight externally.

relays/substrate/src/main.rs
svyatonik: TBH this subcommand seems super-awkward to me :) But here it is `./bin/substrate-relay print-message-limits millau-to-rialto`

tomusdrw: Yeah, you're right indeed it feels weird. Perhaps instead we could do `--remark-size=max-allowed`?

tomusdrw (approved):
lgtm, the subcommand is indeed akward, I've made a comment how we could get rid of it, but it's your call, happy to merge either way.

tomusdrw (approved):

Issue #701, opened 2021-02-03
importedHeaders on Rialto are retrieving an empty result.


hbulgarini:
Hey! When i try to get the last importedHeader with a valid block hash on Rialto network, no results are being retrieved. 

For reproducing this issue here are the steps:

1) Connect to Rialto node.
2) Go to Chain State -> Storage -> Choose 'BridgedMillau' on 'selected state query'.
3) Choose 'bestFinalized' query and execute it.
4) Change query to  'importedHeaders' and provide as input the block hash retrieved from step 3).
5) Execute the query.

If you follow the same steps on Millau node and choose on 'BridgedRialto', the process retrieves the expected information.

HCastano:
I can reproduce this on Polkadot JS Apps.

However, it seems like the node does know about the imported headers. First I got a hash header that was known to be imported from the logs (in this case the SCALE encoded parent hash):

```
2021-02-03 17:13:48.031  TRACE tokio-runtime-worker pallet_substrate_bridge: Going to import header 0x...: Header { parent_hash: 0xd2fc8e1a7143f8ef7c4c548b9082024906dedaf844ae142dc1bda82cc77b8535bd4e0597dae854ce84443f4338cb43fcbb91da7605b4c85268aaef5645e9ac64, number: 7326, state_root: 0x..., extrinsics_root: 0x.., digest: Digest ... }
```

I then asked the node if it knew about this particular header like so:

```bash
curl -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "state_call", "params": ["MillauHeaderApi_is_known_block", "0xd2fc8e1a7143f8ef7c4c548b9082024906dedaf844ae142dc1bda82cc77b8535bd4e0597dae854ce84443f4338cb43fcbb91da7605b4c85268aaef5645e9ac64"]}' http://localhost:9933
```

And it responded with

```
{"jsonrpc":"2.0","result":"0x01","id":1}
```

This means that the issue is between the Polkadot JS Apps and node communication.

tomusdrw:
I suspect it might be because of SCALE-incompatibility. On both chains `ImportedHeader` is actually a completely different (generic) type, while we only have one custom types json. It would be best to:
1. Debug node RPC via `-lrpc=trace` and inspect the responses to make sure that the data is actually returned
2. debug polkadot.js/api internals to figure out how the parsing of this data is being done
3. Potentially prepare different custom JSON files for both networks.

[...]

PR #705, opened 2021-02-05
Add specific Rialto/Millau types

hbulgarini:
Adding specific json files for Rialto and Millau.

This need is originated in the discussion for the issue : #701


hbulgarini:
Unwanted commits. Closing PR

PR #706, opened 2021-02-05
Adding a draft of specific types for Rialto & Millau chains

hbulgarini:
There is a need of having specific json  files for defining the corresponding types of  Rialto & Millau chains.
This  was discussed and agreed on issue #701.


deployments/typesMillau.json
```diff
@@ -0,0 +1,52 @@
+{
+	"AuthoritySet": {
+		"authorities": "AuthorityList",
+		"set_id": "SetId"
+	},
+	"BridgedBlockHash": "H256",
+	"BridgedBlockHasher": "BlakeTwoAndKeccak256",
```
HCastano, line 7: Do we not need to define `BlakeTwoAndKeccak256` for Apps?

deployments/typesMillau.json
```diff
@@ -0,0 +1,52 @@
...
+	"BridgedBlockNumber": "u64",
+	"BridgedDigest": {
```
HCastano, line 9: I'm pretty curious about this. We don't have a `BridgedDigest` struct anywhere in the code, how did you figure that we needed a definition for it?

deployments/typesMillau.json
```diff
@@ -0,0 +1,52 @@
...
+			"Other": "Vec<u8>",
+			"PreRuntime": "PreRuntime",
+			"Seal": "Seal",
+			"SealV0": "SealV0"
+		}
+	},
+	"BridgedHeader": {
+		"digest": "BridgedDigest",
+		"extrinsics_root": "BridgedBlockHash",
+		"number": "Compact<BridgedBlockNumber>",
```
HCastano, line 26: Why's this need to be `Compact`?

HCastano (commented):
We should `git rm` the existing `types.json` file. 

I'd personally also rename these files to `millau-types.json` and `rialto-types.json` since that's closer to the "standard" `types.json` convention. I wouldn't wouldn't block a merge over this though

deployments/typesMillau.json
hbulgarini: @tomusdrw , any comments about this?

deployments/typesMillau.json
hbulgarini: @tomusdrw , same

deployments/typesMillau.json
hbulgarini: So when i provide this exact json file to polkadotjs , there are no errors in the console. So i assumed this was needed, but i can dive more into it.

hbulgarini:
@HCastano  i will rename the files as you recommended as soon as we confirm the definitive versions to merge ( i'm not sure if after the files are renamed the comments are lost )

deployments/typesMillau.json
tomusdrw: I think we don't even need the `Hasher` at all, given that the `BlakeTwoAndKeccak256` is for sure not defined anywhere and it doesn't complain it means it's not needed.

deployments/typesMillau.json
tomusdrw: We need to re-define the `BridgedHeader` type, because in Rust it's a generic type `generic::Header<Hash>`. The JS parser doesn't handle generic types at all, so we need separate type for every generic instance (i.e. `generic::Header<H256>` or `generic::Header<H512>`).
The types propagate further down, so since `Digest` is generic as well (and depends on hash size), we must re-define it as well to parse it correctly.

deployments/typesMillau.json
tomusdrw: Because that's part of encode/decode implementation: https://github.com/paritytech/substrate/blob/master/primitives/runtime/src/generic/header.rs#L96

I basically took the built-in `Header` type from [here](https://github.com/polkadot-js/api/blob/d2ff350db963600fb5256152ef6946856e1bbb96/packages/types/src/interfaces/runtime/definitions.ts#L72)
and redefined the parts that depend on the hash size.

tomusdrw:
@hbulgarini let's add any other types in a follow-up PR if necessary.

tomusdrw:
@hbulgarini do you mind updating to the latest (working) version of the type definition?

deployments/types-millau.json
```diff
@@ -0,0 +1,100 @@
+{
+	"Balance": "u64",
+	"Address": "AccountId",
+	"LookupSource": "AccountId",
+	"MillauBlockHash": "H512",
```
tomusdrw, line 5: As mentioned earlier already, I'd prefer to split chain-related definitions from actual overwrites. I.e.
```json
"__AllRialtoThings:": "",
"RialtoBlockHash": "H256",
"RialtoBalance": "u64",
"...": "...",
"__AllMillauThings": "",
"MillauBlockHash": "H256",
"MillauBalance": "u128",
"__ChainOverrides": "",
"Balance": "MillauBalance",
"Address": "AccountId",
...
```


Do you mind grouping things like that?

Since the top part (Rialto & Millau) should be identical for both files, the next logical step is to move it into a separate `JSON` file and add a small tool that generates `types-millau.json` and `types-rialto.json` by merging `types-common.json + types-millau.json` and `types-common.json + types-rialto.json` accordingly (that should be a separate PR though).

deployments/types-millau.json
```diff
@@ -0,0 +1,100 @@
...
+		"latest_generated_nonce": "MessageNonce",
+		"latest_received_nonce": "MessageNonce",
+		"oldest_unpruned_nonce": "MessageNonce"
+	},
+	"MessageData": {
+		"payload": "MessagePayload",
+		"fee": "Balance"
+	},
+	"MessagePayload": "Vec<u8>",
+	"OutboundMessageFee": "u64",
```
tomusdrw, line 70:
```suggestion
	"OutboundMessageFee": "Balance",
```

deployments/types-millau.json
```diff
@@ -0,0 +1,100 @@
...
+	"OutboundPayload": {
+		"call": "Vec<u8>",
+		"origin": "CallOrigin",
+		"spec_version": "u32",
```
tomusdrw, line 74:
```suggestion
		"spec_version": "SpecVersion",
```

deployments/types-millau.json
```diff
@@ -0,0 +1,100 @@
...
+		"weight": "Weight"
+	},
+	"CallOrigin": {
+		"_enum": {
+			"SourceRoot": "()",
+			"TargetAccount": "([u8;33], MultiSigner,MultiSignature)",
+			"SourceAccount": "[u8;33]"
+		}
+	},
+	"MultiSigner": {
```
tomusdrw, line 84: This is probably unnecessary, no?

deployments/types-millau.json
```diff
@@ -0,0 +1,100 @@
...
+		"_enum": {
+			"Ed25519": "H256",
+			"Sr25519": "H256",
+			"Ecdsa": "[u8;33]"
+		}
+	},
+	"MultiSignature": {
```
tomusdrw, line 91: And this as well.

deployments/types-millau.json
```diff
@@ -0,0 +1,100 @@
...
+		"_enum": {
+			"Ed25519": "H512",
+			"Sr25519": "H512",
+			"Ecdsa": "[u8;65]"
+		}
+	},
+	"Fee": "u64",
```
tomusdrw, line 98:
```suggestion
	"Fee": "Balance",
```

deployments/types-millau.json
hbulgarini: of course, i will set the chain related definitions. Building the tool was not so easy from the JS point of view because of async loading file process ( i tried already and it will take more time than the one i was ready to spend )

deployments/types-millau.json
hbulgarini: actually it is necessary for TargetAccount type: 

`
	"CallOrigin": {
		"_enum": {
			"SourceRoot": "()",
			"TargetAccount": "([u8;33], MultiSigner,MultiSignature)",
			"SourceAccount": "[u8;33]"
		}
	},
`

even if we don't use it know , it does not heart to leave it i think

deployments/types-millau.json
hbulgarini: same as above

deployments/types-millau.json
tomusdrw: You can use `jq` tool and create a bash script instead.
```
jq -s '.[0] * .[1]' file1,json file2.json
```

deployments/types-millau.json
tomusdrw: At least `MultiSignature` is a built-in type, we don't have to specify it:
https://github.com/polkadot-js/api/blob/d2ff350db963600fb5256152ef6946856e1bbb96/packages/types/src/interfaces/extrinsics/definitions.ts#L28

deployments/types-millau.json
tomusdrw: This should go away, but indeed we need to keep `MultiSigner`.

deployments/types-millau.json
```diff
@@ -0,0 +1,95 @@
...
+			"SourceRoot": "()",
+			"TargetAccount": "([u8;32], MultiSigner,MultiSignature)",
+			"SourceAccount": "[u8;32]"
+		}
+	},
+	"MultiSigner": {
+		"_enum": {
+			"Ed25519": "H256",
+			"Sr25519": "H256",
+			"Ecdsa": "[u8;32]"
```
tomusdrw, line 92: This is incorrect, ECDSA public key is 33 bytes: https://github.com/paritytech/substrate/blob/362d5e90ca14a277dada73126a4acfb6df0abba1/primitives/core/src/ecdsa.rs#L56

https://github.com/paritytech/substrate/blob/362d5e90ca14a277dada73126a4acfb6df0abba1/primitives/runtime/src/lib.rs#L242

deployments/types-rialto.json
```diff
@@ -0,0 +1,95 @@
...
+			"SourceRoot": "()",
+			"TargetAccount": "([u8;32], MultiSigner,MultiSignature)",
+			"SourceAccount": "AccountId"
+		}
+	},
+	"MultiSigner": {
+		"_enum": {
+			"Ed25519": "H256",
+			"Sr25519": "H256",
+			"Ecdsa": "[u8;32]"
```
tomusdrw, line 92: same here.

tomusdrw (changes requested):

deployments/types-millau.json
HCastano: Huh, TIL

HCastano (commented):
This looks pretty intense compared to what I had envisioned for our `types.json`, good stuff!

One thing though, I can't send "manual" transfers on the Millau chain with the Millau types. I get the following error in the console:

```
main.35f448fe.js:20 signAndSend: error: TypeError: Cannot read property 'unwrap' of undefined
    at t.project (polkadot.01.0aac97ef.js:1)
    at t._next (other.01.2ddab8b3.js:2)
    at t.next (other.01.2ddab8b3.js:2)
    at t.notifyNext (other.01.2ddab8b3.js:2)
    at t._next (other.01.2ddab8b3.js:2)
    at t.next (other.01.2ddab8b3.js:2)
    at t.notifyNext (other.01.2ddab8b3.js:2)
    at t._next (other.01.2ddab8b3.js:2)
    at t.next (other.01.2ddab8b3.js:2)
    at t.next (other.01.2ddab8b3.js:2)
```

Transfers work fine on Rialto though.

deployments/types/common.json
```diff
@@ -0,0 +1,134 @@
+{
+  "--1": "Millau Types",
```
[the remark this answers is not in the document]
tomusdrw: Nothing, it's just a JSON-comment.

deployments/types/build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+set -exu
+
+# Make sure we are in the right dir.
+cd $(dirname $(realpath $0))
+
+# Create rialto and millau types.
+jq -s '.[0] * .[1]' common.json rialto.json > ../types-rialto.json
```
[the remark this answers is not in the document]
tomusdrw: Having a common prefix places the two files next to each other when using natural sorting. There is already a lot of files in that directory, so IMHO better to keep like this.

deployments/types/common.json
```diff
@@ -0,0 +1,134 @@
...
+    }
+  },
+  "MultiSigner": {
+    "_enum": {
+      "Ed25519": "H256",
+      "Sr25519": "H256",
+      "Ecdsa": "[u8;33]"
+    }
+  },
+  "MesssageProofOf": "(Weight, FromBridgedChainMessagesProof)",
```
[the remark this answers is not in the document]
tomusdrw: Updated, I forgot to push latest version yesterday.

deployments/types/build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
tomusdrw: Done

tomusdrw:
> This looks pretty intense compared to what I had envisioned for our `types.json`, good stuff!
> 
> One thing though, I can't send "manual" transfers on the Millau chain with the Millau types. I get the following error in the console:
> 
> ```
> main.35f448fe.js:20 signAndSend: error: TypeError: Cannot read property 'unwrap' of undefined
>     at t.project (polkadot.01.0aac97ef.js:1)
>     at t._next (other.01.2ddab8b3.js:2)
>     at t.next (other.01.2ddab8b3.js:2)
>     at t.notifyNext (other.01.2ddab8b3.js:2)
>     at t._next (other.01.2ddab8b3.js:2)
>     at t.next (other.01.2ddab8b3.js:2)
>     at t.notifyNext (other.01.2ddab8b3.js:2)
>     at t._next (other.01.2ddab8b3.js:2)
>     at t.next (other.01.2ddab8b3.js:2)
>     at t.next (other.01.2ddab8b3.js:2)
> ```
> 
> Transfers work fine on Rialto though.

Millau will never fully work due to custom hasher implementation - this is currently something that can't be easily overridden using JSON file, so I'm assuming the error is related to this.

tomusdrw (approved):

deployments/types/common.json
```diff
@@ -0,0 +1,134 @@
...
+    }
+  },
+  "MultiSigner": {
+    "_enum": {
+      "Ed25519": "H256",
+      "Sr25519": "H256",
+      "Ecdsa": "[u8;33]"
+    }
+  },
+  "MesssagesProofOf": {
```
HCastano, line 115: Is there something wrong with the `s` key on your keyboard @tomusdrw 🤣 

```suggestion
  "MessagesProofOf": {
```

deployments/types/common.json
```diff
@@ -0,0 +1,134 @@
...
+	"bridged_header_hash": "BridgedBlockHash",
+	"storage_proof": "Vec<StorageProofItem>",
+	"lane": "LaneId"
+  },
+  "UnrewardedRelayersState": {
+	"unrewarded_relayer_entries": "MessageNonce",
+	"messages_in_oldest_entry": "MessageNonce",
+	"total_messages": "MessageNonce"
+  },
+  "AncestryProof": "()"
```
HCastano, line 133: We'll also need to remove this once #804 is in.

HCastano (approved):

deployments/types/common.json
tomusdrw: That's insane, I feel I fixed this dozen of times alraedy and it keeps coming back :D

tomusdrw:
FYI: #809

PR #707, opened 2021-02-05
Stop Importing Full Header Chain

HCastano:
This PR removes the need to import every single header to the base bridge pallet from the
finality verifier pallet. By doing this we are opening the doors to "sparse" proof
formats to be used as ancestry proofs. Some examples of sparse proofs include storage
proofs and MMRs.

### DIFF
--- a/modules/finality-verifier/src/lib.rs
+++ b/modules/finality-verifier/src/lib.rs
@@ -33,7 +33,6 @@
 #![allow(clippy::large_enum_variant)]
 
 use bp_header_chain::{justification::verify_justification, AncestryChecker, HeaderChain};
-use bp_runtime::Size;
 use bp_runtime::{Chain, HeaderOf};
 use finality_grandpa::voter_set::VoterSet;
 use frame_support::{dispatch::DispatchError, ensure, traits::Get};
@@ -69,7 +68,7 @@ pub mod pallet {
 		///
 		/// Will be used by the ancestry checker to verify that the header being finalized is
 		/// related to the best finalized header in storage.
-		type AncestryProof: Parameter + Size;
+		type AncestryProof: Parameter;
 
 		/// The type through which we will verify that a given header is related to the last
 		/// finalized header in our storage pallet.
@@ -110,7 +109,7 @@ pub mod pallet {
 			let _ = ensure_signed(origin)?;
 
 			if let Some(max_len) = T::MaxElementsInSingleProof::get() {
-				let proof_len = Size::size_hint(&ancestry_proof);
+				let proof_len = T::AncestryChecker::proof_size(&ancestry_proof);
 				frame_support::debug::trace!("Got an ancestry proof of length {:?}", proof_len);
 				ensure!(proof_len < max_len.as_(), <Error<T>>::OversizedAncestryProof);
 			};
--- a/modules/finality-verifier/src/lib.rs
+++ b/modules/finality-verifier/src/lib.rs
@@ -77,9 +77,10 @@ pub mod pallet {
 
 		/// The maximum length of elements we can have in a single ancestry proof.
 		///
-		/// It is an optional field since some proof types might not be concered about the length of
-		/// the proof when verifying ancestry, whereas for other verification time may grow with the
-		/// size of the proof (which if left unbounded would be bad).
+		/// It is an optional field since some ancestry proof types may require iterating through
+		/// the proof during verification. If left unbounded, this could be problematic. On the
+		/// other hand, some proof structures may not care about a maximum length, in which case
+		/// this does not need to be used.
 		#[pallet::constant]
 		type MaxElementsInSingleProof: Get<Option<u32>>;
 	}
@@ -94,14 +95,11 @@ pub mod pallet {
 	impl<T: Config> Pallet<T> {
 		/// Verify a target header is finalized according to the given finality proof.
 		///
-		/// Will use the underlying storage pallet to fetch information about the current
+		/// It will use the underlying storage pallet to fetch information about the current
 		/// authorities and best finalized header in order to verify that the header is finalized.
 		///
-		/// If successful in verification, it will write the header as well as its ancestors (from
-		/// the given `ancestry_proof`) to the underlying storage pallet.
-		///
-		/// Note that the expected format for `ancestry_proof` is a continguous list of finalized
-		/// headers containing (current_best_finalized_header, finality_target]
+		/// If successful in verification, it will write the target header to the underlying storage
+		/// pallet.
 		#[pallet::weight(0)]
 		pub fn submit_finality_proof(
 			origin: OriginFor<T>,
@@ -157,7 +155,7 @@ pub mod pallet {
 		InvalidAuthoritySet,
 		/// Failed to write a header to the underlying header chain.
 		FailedToWriteHeader,
-		/// The given ancestry proof is too large to be verified in a single transaction.
+		/// The given ancestry proof is too large to be verified efficiently.
 		OversizedAncestryProof,
 	}
 }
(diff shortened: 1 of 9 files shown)


modules/finality-verifier/src/lib.rs
```diff
@@ -135,17 +137,8 @@ pub mod pallet {
...
-				frame_support::debug::error!("Failed to append finalized header chain.");
-				<Error<T>>::FailedToWriteHeader
-			})?;
-
-			frame_support::debug::info!(
-				"Succesfully imported finalized header chain for target header {:?}!",
-				finality_target
-			);
+			T::HeaderChain::append_header(finality_target);
+			frame_support::debug::info!("Succesfully imported finalized header with hash {:?}!", hash);
```
svyatonik, line 141: It seems to that you're not only "opening the doors to "sparse" proof" here, but also closing the doors to "full" proof here, right? Simply because ancestry proof is always ignored here && you're only "importing" finality target. So unless we'll submit `submit_finality_proof` for every header (which requires GRANDPA justification for every header, which we can't provide), the pallet will always miss some headers. This breaks assumption (at least I was 100% sure in this) that we were going to deploy initial version of bridge with all-headers-in-the-pallet (+historical pruning) && then, eventually. migrate to MMRs (or whatever). Could you confirm that this is an intentional change?

modules/finality-verifier/src/lib.rs
```diff
@@ -87,45 +95,39 @@ pub mod pallet {
...
-			ancestry_proof: Vec<BridgedHeader<T>>,
+			ancestry_proof: T::AncestryProof,
 		) -> DispatchResultWithPostInfo {
 			let _ = ensure_signed(origin)?;
 
-			frame_support::debug::trace!("Going to try and finalize header {:?}", finality_target);
-			frame_support::debug::trace!("Got ancestry proof of length {}", ancestry_proof.len());
+			if let Some(max_len) = T::MaxElementsInSingleProof::get() {
+				let proof_len = Size::size_hint(&ancestry_proof);
+				frame_support::debug::trace!("Got an ancestry proof of length {:?}", proof_len);
```
svyatonik, line 114: For lengths there's `frame_support::traits::Len` - probably it's better to use it here?

modules/finality-verifier/src/lib.rs
```diff
@@ -64,17 +65,24 @@ pub mod pallet {
...
-		/// unbounded iteration when verifying proofs.
+		/// The maximum length of elements we can have in a single ancestry proof.
+		///
+		/// It is an optional field since some ancestry proof types may require iterating through
+		/// the proof during verification. If left unbounded, this could be problematic. On the
+		/// other hand, some proof structures may not care about a maximum length, in which case
+		/// this does not need to be used.
 		#[pallet::constant]
-		type MaxHeadersInSingleProof: Get<<Self::BridgedChain as Chain>::BlockNumber>;
+		type MaxElementsInSingleProof: Get<Option<u32>>;
```
tomusdrw, line 85: IMHO if we were to create a proper abstraction this should be function from the `AncestryChecker` trait and it should simply receive the `ancestry_proof` parameter and it will know the type, so can easily verify whatever is needed (without `Size::size_hint` hacks).

modules/finality-verifier/src/lib.rs
tomusdrw: Indeed, initially I was more inclined to have full header chain, but it seems it creates more problems than it really solves.

Especially given that:
1. We don't need full header chain for our `MessageLanes` implementations
2. Migrating to MMR would remove full header chain anyway.
3. Pruning is a bit more problematic with full header chain (more data to remove)

HCastano: > It seems to that you're not only "opening the doors to "sparse" proof" here, but also closing the doors to "full" proof here, right?

Yes and no. So _technically_ you can still provide "full" proofs, but now the pallet won't import all the headers from that proof to the base bridge pallet (for the reasons Tomek stated above).

modules/finality-verifier/src/lib.rs
HCastano: I don't think that limiting the number of elements in the proof should be part of the trait's responsibility. It should be done as part of wiring up the runtime.

I have changed the `AncestryChecker` trait to provide a method for easily getting the proof size, let me know if that works for you.

tomusdrw: It's better than `size_hint` now, but I still don't think it's the right way. Since the `proof` is already abstract the notion of `proof_size` is completely ambiguous.

What `proof_size` does `frame_system` storage proof or MMR proof have? Is it the number of proof nodes or maybe total encoded bytes length, or maybe even it's always 1 cause it contains only one header, given that `Vec<Header>` `proof_size` is defined as the length of the vector.

Delegation to the abstract `AncestryChecker` will be required for weight computation anyway, so IMHO either we do a proper abstraction, where we don't make assumptions about what properties the "ancestry proof" has (i.e. `proof_size`) or we avoid introducing this abstraction at all, and let's simply take `Vec<Header>` for now and verify by calling into a regular function not a trait (my preference actually).

I feel this is all moot anyway, since the code we talk about here should be removed ASAP and replaced with storage proof verification (as described in #628)

HCastano: > It's better than size_hint now, but I still don't think it's the right way. Since the proof is already abstract the notion of proof_size is completely ambiguous.

While I figured it would be up to the implementer of whatever type of ancestry checker to decide how they want to size their proof, I agree with you it's not a great property to add directly to the ancestry checker.

> Delegation to the abstract AncestryChecker will be required for weight computation anyway, so IMHO either we do a proper abstraction, where we don't make assumptions about what properties the "ancestry proof" has (i.e. proof_size) or we avoid introducing this abstraction at all, and let's simply take Vec<Header> for now and verify by calling into a regular function not a trait (my preference actually).

I like the flexibility given by the trait, so I'm going to stick with (roughly) the original - maybe removing the size checks since their meaning is ambiguous (🤠).

> I feel this is all moot anyway, since the code we talk about here should be removed ASAP and replaced with storage proof verification (as described in #628)

Do you mind expanding on this point here? I envisioned that the same pallet code could be used. The only difference would be that in the runtime we would need an implementation of `AncestryChecker` that used storage proofs.

tomusdrw: > maybe removing the size checks since their meaning is ambiguous

That's a step backwards and will cause problems when calculating weights, but okay, let's deal with it when we implement weights.

> Do you mind expanding on this point here? I envisioned that the same pallet code could be used. 

Yes, the same pallet will be used.

> The only difference would be that in the runtime we would need an implementation of AncestryChecker that used storage proofs.

No, that's not the only difference. The differences are:
1. Different `AncestryChecker` implementation
2. Different proof type (hence I suggested an associated type on `AncestryChecker` trait)
3. Different limiting mechanism, specific to `AncestryChecker` (hence suggested to simply move the check to the trait)
4. Different weight mechanism, specific to `AncestryChecker` (that would come next, but it would simply be another trait method on the trait).

TBH Given that we only plan to have one `AncestryChecker` anyway, introducing the trait is pre-mature generalisation (is that a thing even :D), hence the entire discussion.

That said, let's just merge it and start making some progress instead of discussing.

tomusdrw (approved):

modules/finality-verifier/src/lib.rs
HCastano: > TBH Given that we only plan to have one AncestryChecker anyway, introducing the trait is pre-mature generalisation (is that a thing even :D), hence the entire discussion.

When I wrote this up I thought we'd have at least three `AncestryChecker` implementations, which is why it's traitified :)

> That said, let's just merge it and start making some progress instead of discussing.

👌

PR #714, opened 2021-02-09
Rate Limiting Pallet

HCastano:
This PR adds a pallet which is meant to provide rate limiting capabilities. The idea
behind it is that we will only allow a certain number of requests (in our case `Call`s)
over a pre-defined window of time (some number of blocks). If we get any requests once
the limit of the rolling window has been saturated they are denied.

As for the implementation, I've gone with a sliding window implementation. This is
because it is pretty simple and reasonably accurate despite being an approximation. I
considered a fixed window approach which would've been simpler, but it has problems at
the window boundaries (people can wait for a window reset and stampede the pallet at that
point in time). See [here](https://nlogn.in/design-a-scalable-rate-limiting-algorithm-system-design/) for some brief explanations of different rate-limiting algorithms.

I initially broke this out into a standalone pallet to avoid the "noise" of working
either in the Substrate or Finality pallets, but I realized that having it standalone
could be benefitial. I imagine (with a bit of work) it could be used sort of like the
Proxy or Sudo pallets are used today - they wrap a `Call`, and ensure that some
invariant(s) holds before dispatching the inner `Call`.

For today though, the main goal of this is to help address https://github.com/paritytech/srlabs_findings/issues/47.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4381,6 +4381,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "pallet-rate-limiter"
+version = "0.1.0"
+dependencies = [
+ "bp-test-utils",
+ "frame-support",
+ "frame-system",
+ "parity-scale-codec 1.3.6",
+ "serde",
+ "sp-io",
+ "sp-runtime",
+ "sp-std",
+]
+
+[[package]]
 name = "pallet-session"
 version = "2.0.1"
 source = "git+https://github.com/paritytech/substrate.git?branch=master#0b0d124d5f9be89f614f2be8e9da038fcb9f540e"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ members = [
 	"modules/ethereum-contract/builtin",
 	"modules/finality-verifier",
 	"modules/currency-exchange",
+	"modules/rate-limiter",
 	"relays/ethereum",
 	"relays/substrate",
 ]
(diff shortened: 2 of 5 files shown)


tomusdrw (changes requested):
I feel this is an overkill for a very simple mitigation that could be implemented directly in the `FinalityVerifier` pallet.

We basically need this:
```rust
decl_storage! {
   ImportCounter: u32;
}

...
fn on_initialize() {
   ImportCounter::mutate(|c| c.saturating_sub(1)); 
   // If it's really needed we could configure the decrement here
   // if we expect weirdly paced chains (like one is twice as fast as
   // the other and we want all finalized headers for some weird reason)).
}
fn submit_finality_proof(
...
)  {
    let counter = ImportCounter::get() + 1;
    if counter > T::MaxImportRate::get() {
       return Err(...)
    }
    ImportCounter::put(counter);
}
```

It's simple and AFAICT get's the job done. We really don't need so much configurability here and a pallet with it's own set of dispatchables (which will really cripple the UX).

HCastano:
> It's simple and AFAICT get's the job done. 

Sure, it works, but I'm not sure its ideal. I think the problem we'll run into is that in the long run we'll always be at the limit of `MaxImportRate` and be essentially be limited to only importing one header per block (or whatever config constant we choose since you mention that).

> We really don't need so much configurability here and a +360 lines pallet with it's own set of dispatchables (which will really cripple the UX).

I don't think this is totally fair, since the meat of it is ~100 LoC. The rest is FRAME boiler plate and tests.

> I feel this is an overkill for a very simple mitigation that could be implemented directly in the `FinalityVerifier` pallet.

To quote myself: "I initially broke this out into a standalone pallet to avoid the "noise" of working
either in the Substrate or Finality pallets...", but yes maybe it's not the right solution for right now. I'll go back to the drawing board.

svyatonik:
As far as I can tell, it is superseded by #720

Issue #560, opened 2020-12-04
Fee top-ups for messages stuck in a lane.


tomusdrw:
Currently we heavily rely on on-chain price oracle (#540, #449) to make sure messages that don't pay enough fee (on the source chain) to cover the delivery & dispatch cost (on the target chain) don't get into the queue.

However it may still happen (in case of large price fluctuations) that we will end up with messages that are below profitability point from relayers perspective.

We assume that this can be relayed by either altruistic relayers (see #546) or if the cumulative reward in the lane is high enough (also #546).

It's also assumed that users can "unclog" the lane by putting high-paying messages inside, however we also have a fixed limit of unconfirmed messages in the lane, so at some point it might not be possible to add more rewards.

Hence it would be good to be able to top-up arbitrary message in the lane - i.e. increase the reward for relaying the message, without changing the content.

svyatonik:
> It's also assumed that users can "unclog" the lane by putting high-paying messages inside, however we also have a fixed limit of unconfirmed messages in the lane, so at some point it might not be possible to add more rewards.

This doesn't block us from sending message - it blocks relayer from relaying these messages to the target chain. So you still can 'unblock' the lane by sending empty message and non-zero fee.

tomusdrw:
> This doesn't block us from sending message - it blocks relayer from relaying these messages to the target chain.

Hmm, so on the source chain we don't limit number of pending messages in the lane? I think from the security perspective we should to prevent arbitrary state growth from user-generated data (i.e. the one-time cost doesn't really cover storing the message ad infinitum).

svyatonik:
I don't think that is relevant to the number of messages, right? I mean - we should not be storing even single message for inifinity? 

This could be solved e.g. by replacing instant payment mechanism with some schema where message fee is decaying (burnt) with every block until the fee reaches absolute minimum (e.g. the value returned by `estimate_message_dispatch_and_delivery_fee`). Then  the message payload is replaced with `vec![]`. So relayers will be trying to relay messages asap, because this way they may get maximal reward. At the same time, the fee can't be decayed to the value less than minimal relay cost => this won't stuck the line.

tomusdrw:
> I don't think that is relevant to the number of messages, right? I mean - we should not be storing even single message for inifinity?

[...]

PR #715, opened 2021-02-10
Limit number of pending messages at outbound lane

svyatonik:
from discussion inside #560 
Decay strategy may be implemented after migrating to offchain storage - then the most of changes will be in the relay code.

This PR introduces two additional checks in our default `LaneMessageVerifier`. The checks are:
1) messages that are being sent through the disabled lane are immediately rejected. For testnets I've only enabled lane `[0, 0, 0, 0]` - we don't use any other lanes here. But may change to whitelist all lanes, if you're doing some other tests there. For P+K I suggest to enable single lane initially. Then, in some later upgrades - enable all (or more lanes);
2) if there are too many pending (aka undelivered) messages at outbound lane, message is immediately rejected. For testnets there's no limit - I'm often submit a lot of messages. For P+K I suggest this being some `N * BridgedChain::MaxUnconfirmedMessagesAtInboundLane` (this constant may even come from storage for easier upgrades). This would also require support in ML clients (like in our UI) - we should get confirmation if we're close to limit + it shall warn user that his message may be rejected. And we should not allow to submit any messages if we're already at limit.

Why it is made in `LaneMessageVerifier`, not in the pallet itself? Because these checks may be lane-specific - i.e. for some kind of sudo calls we may have a separate lane without any limits. For apps that are sending large messages (which may be controlled indide the `LaneMessageVerifier` again), we may have lower limit and so on. So even if our `LaneMessageVerifier` implementation currently assumes that all lanes are identical, I feel that this may change in the future && I prefer to have this code outside of the pallet.

### DIFF
--- a/bin/runtime-common/src/messages.rs
+++ b/bin/runtime-common/src/messages.rs
@@ -44,7 +44,7 @@ pub trait MessageBridge {
 	const RELAYER_FEE_PERCENT: u32;
 
 	/// This chain in context of message bridge.
-	type ThisChain: ChainWithMessageLanes;
+	type ThisChain: ThisChainWithMessageLanes;
 	/// Bridged chain in context of message bridge.
 	type BridgedChain: ChainWithMessageLanes;
 
@@ -102,6 +102,17 @@ pub trait ChainWithMessageLanes {
 
 	/// Instance of the message-lane pallet.
 	type MessageLaneInstance: Instance;
+}
+
+/// This chain that has `message-lane` and `call-dispatch` modules.
+pub trait ThisChainWithMessageLanes: ChainWithMessageLanes {
+	/// Are we accepting any messages to the given lane?
+	fn is_outbound_lane_enabled(lane: &LaneId) -> bool;
+
+	/// Maximal number of pending (not yet delivered) messages at this chain.
+	///
+	/// Any messages over this limit, will be rejected.
+	fn maximal_pending_messages_at_outbound_lane() -> MessageNonce;
 }
 
 pub(crate) type ThisChain<B> = <B as MessageBridge>::ThisChain;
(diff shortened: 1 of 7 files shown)


bin/runtime-common/src/messages.rs
```diff
@@ -104,6 +104,17 @@ pub trait ChainWithMessageLanes {
...
 
+/// This chain that has `message-lane` and `call-dispatch` modules.
+pub trait ThisChainWithMessageLanes: ChainWithMessageLanes {
+	/// Are we accepting any messages to the given lane?
+	fn is_outbound_lane_enabled(lane: &LaneId) -> bool;
+
+	/// Maximal number of pending (not yet delivered) messages at this chain.
+	///
+	/// Any messages over this limit, will be rejected.
+	fn maximal_pending_messages_at_outbound_lane() -> MessageNonce;
```
tomusdrw, line 115: > Because these checks may be lane-specific - i.e. for some kind of sudo calls we may have a separate lane without any limits.

Given what you said, shouldn't the method then get `lane: &LaneId` parameter?

tomusdrw (commented):
lgtm, although this now creates the `top-up` problem mentioned in #560. Basically if you discover that the price oracle is behind, you spam the lane with underpriced messages and because of the limit the only way to unclog it is to run an altruistic relayer.
I feel doing that on-chain (topping up) should be a bit simpler than setting up the relayer.

svyatonik:
> lgtm, although this now creates the top-up problem mentioned in #560. Basically if you discover that the price oracle is behind, you spam the lane with underpriced messages and because of the limit the only way to unclog it is to run an altruistic relayer.

Yes, you're right - let's add the call to the pallet for increasing fee. I've changed the #560 milestone to P+K

bin/runtime-common/src/messages.rs
svyatonik: No, I has commented specifically that our implementation of `MessageLaneVerifier` assumes that all lanes are identical. There are other checks (i.e. some lanes may accept messages with zero fee) that are not lane-specific in our implementation. So making this one lane-specific makes no sense

tomusdrw: Ah, I didn't notice that the check is actually being done in some verifier, so we can skip the check (or alter it) in case we don't want that.

Although, now that I think about it, perhaps it's better to use `frame_support::traits::Filter<LaneId>` for enablement (given there is some tooling around it) and just harcode the max pending messages in the verifier implementation (as a const)?

svyatonik: > Although, now that I think about it, perhaps it's better to use frame_support::traits::Filter<LaneId> for enablement (given there is some tooling around it) and just harcode the max pending messages in the verifier implementation (as a const)?

This could be changed, but I feel that it'll just bring more code lines without any notable benefits. You may still use whatever you feel is better in the implementation - `Filter`s, `Get`s, constants or whatever.

tomusdrw (approved):

tomusdrw:
Looks good modulo the `top-up` requirement, which should be covered in #560

PR #716, opened 2021-02-10
Add a fuzzer for storage proofs

viniul:
CC @tomusdrw 

This PR adds a fuzzer that tests whether storage proofs work correctly. 
The fuzzer:
1. Takes a vector of key-value pairs
2. Insert them into storage and creates a proof
3. Check if the proofs worked as expected

The fuzzer is based on the respective unit test for the storage proofs.


adoerr:
Perhaps we should add a top-level directory `fuzz` and put all fuzzing related stuff there

adoerr:
Perhaps we should add a top-level directory fuzz and put all fuzzing related stuff there

adoerr (commented):
Perhaps we should add a top-level directory fuzz and put all fuzzing related stuff there

adoerr (approved):
LGTM

adoerr (approved):
LGTM

tomusdrw:
Merging since `cargo-deny` is unrelated.

fewensa:
@tomusdrw the fuzzer is required? because now in windows `honggfuzz` only support WSL, not everyone use it. So I think set fuzz to not default member?

https://doc.rust-lang.org/cargo/reference/workspaces.html#package-selection
like this

```toml
[workspace]
resolver = "2"

members = [
	"bin/*/node",
	"bin/*/runtime",
	"fuzz/*",
	"modules/*",
	"primitives/*",
	"relays/*",
]
default-members = [
	"bin/*/node",
	"bin/*/runtime",
	"modules/*",
	"primitives/*",
	"relays/*",
]
```

or use exclude

```toml
[workspace]
resolver = "2"

members = [
	"bin/*/node",
	"bin/*/runtime",
	"fuzz/*",
	"modules/*",
	"primitives/*",
	"relays/*",
]
exclude  = [
	"fuzz/storage-proof",
]
```

this way help build simply in Windows, when add this, if want build storage-proof, use `cargo build -p storage-proof-fuzzer`

tomusdrw:
@fewensa we currently neither support nor test on Windows at all. I'd rather keep the fuzzer as the default member to make sure it's properly maintained. Feel free to fork and change the members as you suggested.

PR #718, opened 2021-02-11
Increase message fee call

### LINKED ISSUE #560: Fee top-ups for messages stuck in a lane.

svyatonik:
closes #560

This also fixes broken benchmarks (as a result of recent #715)


modules/message-lane/src/weights.rs
```diff
@@ -72,100 +73,105 @@ pub trait WeightInfo {
...
 			.saturating_add(T::DbWeight::get().writes(12 as Weight))
 	}
 	fn send_16_kb_message_worst_case() -> Weight {
-		(194_483_000 as Weight)
+		(214_721_000 as Weight)
 			.saturating_add(T::DbWeight::get().reads(4 as Weight))
 			.saturating_add(T::DbWeight::get().writes(12 as Weight))
 	}
+	fn increase_message_fee() -> Weight {
+		(8_395_221_000 as Weight)
```
svyatonik, line 91: The fee is quite large because we're reading the maximal possible message payload (`Vec<u8>`) in the benchmark. We can't actually predict this size without reading => can't use user-provided params.

modules/message-lane/src/lib.rs
```diff
@@ -348,6 +352,59 @@ decl_module! {
...
+			// if someone tries to pay for already-delivered message, we're rejecting this intention
+			// (otherwise this additional fee will be locked forever in relayers fund)
+			//
+			// if someone tries to pay for not-yet-sent message, we're rejeting this intention, or
+			// we're risking to have mess in the storage
+			let lane = outbound_lane::<T, I>(lane_id);
+			ensure!(
+				nonce > lane.data().latest_received_nonce && nonce <= lane.data().latest_generated_nonce,
+				Error::<T, I>::InactiveMessage
+			);
```
HCastano, line 372: I'd be tempted to split this into two different errors, just to ensure that the caller knows precisely what went wrong

modules/message-lane/src/benchmarking.rs
```diff
@@ -81,6 +81,10 @@ pub struct MessageDeliveryProofParams<ThisChainAccountId> {
 
 /// Trait that must be implemented by runtime.
 pub trait Config<I: Instance>: crate::Config<I> {
+	/// Lane id to use in benchmarks.
+	fn bench_lane_id() -> LaneId {
+		Default::default()
+	}
```
HCastano, line 87: Why did this need to get moved into the `Config` trait? It looked fine as a standalone function

HCastano (approved):

modules/message-lane/src/benchmarking.rs
svyatonik: We're blocking all lanes except default in our runtimes + will be blocking all lines except *some-lane* in P+K. So it's controlled by runtime

PR #720, opened 2021-02-11
Finality Pallet Rate Limiter

HCastano:
This PR uses Tomek's suggestion from [here](https://github.com/paritytech/parity-bridges-common/pull/714#pullrequestreview-587490435) and provides a simplified approach
to rate limiting for the finality verifier pallet.

The number I chose for `MaxRequests` is pretty arbirary. I figured that it wouldn't be
the worst thing in the world if we were wrong in choosing this number, since:

- Chances are we're not going to import a lot of finality proofs in a short period of
  time, so 50 seemed like a high enough limit if there is a burst of activity
- The pallet would be throttled to one request per block after we hit the limit, limiting
  any damage in case it's too low

The real test will be to see how it performs in practice and adjust the value
accordingly.

Closes https://github.com/paritytech/srlabs_findings/issues/47.

### DIFF
--- a/modules/finality-verifier/src/lib.rs
+++ b/modules/finality-verifier/src/lib.rs
@@ -89,9 +89,9 @@ pub mod pallet {
 		fn on_initialize(_n: T::BlockNumber) -> frame_support::weights::Weight {
 			<RequestCount<T>>::mutate(|count| *count = count.saturating_sub(1));
 
-			(0 as Weight)
-				.saturating_add(T::DbWeight::get().reads(1 as Weight))
-				.saturating_add(T::DbWeight::get().writes(1 as Weight))
+			(0_u64)
+				.saturating_add(T::DbWeight::get().reads(1))
+				.saturating_add(T::DbWeight::get().writes(1))
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


modules/finality-verifier/src/lib.rs
```diff
@@ -72,13 +72,28 @@ pub mod pallet {
 		/// The type through which we will verify that a given header is related to the last
 		/// finalized header in our storage pallet.
 		type AncestryChecker: AncestryChecker<<Self::BridgedChain as Chain>::Header, Self::AncestryProof>;
+
+		/// The upper bound on the number of requests allowed by the pallet.
+		///
+		/// Once this bound is reached the pallet will not allow any dispatchables to be called
+		/// until the request count has decreased.
+		#[pallet::constant]
+		type MaxRequests: Get<u32>;
```
adoerr, line 81: Since this is a configuration param, perhaps MaxRequestsPerBlock or even MaxSubmittedProofsPerBlock might be more accurate.

adoerr (approved):
Reason for this limiter is that we assume, that a source chain can be fully compromised. In that case, I think we should also have some sort of a deny-list for source chains. This deny-list would block a source chain under suspicion from any interaction with the bridge. This deny-list would be populated and cleared by a bridge admin.

modules/finality-verifier/src/lib.rs
tomusdrw: It's actually not "per block", it's more like `MaxBurstRequests`, cause after this number is reached, as Hernando noted, we throttle the import to one per block.

tomusdrw (approved):
<3 Thanks a lot, looks good to me.

PR #722, opened 2021-02-12
Workaround for SS58Prefix issues in the UI.

tomusdrw:
Having SS58Prefix set to number greater than `64` create an incompatibility between [UI](https://github.com/polkadot-js/common/blob/master/packages/util-crypto/src/address/encode.ts#L26) and [Rust](https://github.com/paritytech/substrate/blob/9f6e97ebdf2d2c86da837e3d6879a07bfb11e22c/primitives/core/src/crypto.rs#L242) code.

To temporarily overcome the issue until it's solved, I've change the SS58Prefix to 42 on both chains.

### DIFF
--- a/bin/millau/runtime/src/lib.rs
+++ b/bin/millau/runtime/src/lib.rs
@@ -148,7 +148,7 @@ parameter_types! {
 		read: 60_000_000, // ~0.06 ms = ~60 µs
 		write: 200_000_000, // ~0.2 ms = 200 µs
 	};
-	pub const SS58Prefix: u8 = 86;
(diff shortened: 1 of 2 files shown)


tomusdrw:
Okay, seems it's just because our substrate dependency is out of date: https://github.com/paritytech/substrate/pull/8039

HCastano:
So should we close this and bump Substrate instead?

tomusdrw:
Bumping substrate is non-trivial, let's merge and revert after it's done.

svyatonik:
From #723 : "Supersedes #722", so I'm closing it

PR #727, opened 2021-02-15
Remove lower limit of messages weight

### LINKED ISSUE #721: Consider removing lower bound for message dispatch weight
It seems that this [weight' lower limit](https://github.com/paritytech/parity-bridges-common/blob/76d0224010138e0795da0ceb10d7847f4c36e11c/bin/rialto/runtime/src/millau_messages.rs#L102) may be redundant now - it has appeared before we have started charging sender for message size [here](https://github.com/paritytech/parity-bridges-common/blob/76d0224010138e0795da0ceb10d7847f4c36e11c/bin/runtime-common/src/messages.rs#L142). It was there to make sure that e.g. `system::remark()` call coming through the bridge will never be cheaper than simple `system::remark()`call on the target chain. TLDR: this function is intended to only work with dispatch weights, but lower limit seems to come from transaction payments.

This all needs to be re-checked, though.

svyatonik:
closes #721 

It may still be verified by other bridges - if e.g. bridge only needs to accept transfer calls, it may verify that weight is >= hardcoded weight of transfer call. But for our (testnets) bridge we don't need any limits like that, so this check is invalid. For example, we were unable to deliver `sudo::set_key()` with declared weight set to zero, even though dispatch weight of this call is actually zero. So the sender must have declared non-zero weight and pay extra cost just to overcome this limit.

### DIFF
--- a/bin/millau/runtime/src/rialto_messages.rs
+++ b/bin/millau/runtime/src/rialto_messages.rs
@@ -95,17 +95,16 @@ impl MessageBridge for WithRialtoMessageBridge {
 		bp_rialto::max_extrinsic_size()
 	}
 
-	fn weight_limits_of_message_on_bridged_chain(message_payload: &[u8]) -> RangeInclusive<Weight> {
+	fn weight_limits_of_message_on_bridged_chain(_message_payload: &[u8]) -> RangeInclusive<Weight> {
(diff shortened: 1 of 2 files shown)


tomusdrw (approved):
sounds reasonable

PR #728, opened 2021-02-15
Message lane pallet parameters + updatable conversion rate

### LINKED ISSUE #540: Centralised price oracle module.
To convert balances between chains, we need some price oracle (see #449).
Since in initial DOT<>KSM deployment it's probably fine to overpay, we can have centralised version of that price oracle for instance that could be used here:
https://github.com/svyatonik/polkadot/blob/20a43e0f24aab1d31bd983ba1462f4d049cd2093/runtime/kusama/src/polkadot_messages.rs#L117

The idea would be to allow either `root`, `bridge admin` or `bridged chain root` to update the price ratio.

svyatonik:
closes #540 

This PR introduces slightly different approach to what's described in the original issue. Why:
1) when I was integrating our bridge pallets into Polkadot/Kusama runtimes I've realized that max number of modules is [`256`](https://github.com/paritytech/parity-scale-codec/blob/master/derive/src/encode.rs#L211) (or maybe I miss something?). And we're going to occupy `3` of them (I even had intention to merge `call-dispatch` and `message-lane`, but we need a separate set of events for that). Conversion-rate-pallet would be the `4th` occupied index, which is imo too much;
2) having this pallet may lead to false sense that we're guaranteeing that it'll always has actual conversion rate. Simply because it is announced too loud (as a separate pallet). For example it may be used by offchain app developers. But that isn't actually true - e.g. if we'll halt our bridge, I believe we won't guarantee that this value is actual.

So I've decided to make something that is called "message lane pallet parameter". The set of parameters is defined by runtime. Runtime also defines the method of updating these parameters. The pallet itself guarantees that only root or the pallet owner is able to update these parameters. For illustration, I've added conversion rate parameter to our test runtimes. But I don't expect it'll be ever changed actually.

As a final note: I believe that there may be other caveats that I don't see && separate-pallet approach may be better than message-lane-parameters approach (probably someone else is already relying on this pallet?). Feel free to announce it here - I can close this PR && open another one with separate pallet. But right now it seems better option to me.

### DIFF
--- a/modules/message-lane/src/lib.rs
+++ b/modules/message-lane/src/lib.rs
@@ -288,7 +288,7 @@ decl_module! {
 		///
 		/// May only be called either by root, or by `ModuleOwner`.
 		///
-		/// The weight is: single read for permissions check +
+		/// The weight is: single read for permissions check + 2 writes for parameter value and event.
 		#[weight = (T::DbWeight::get().reads_writes(1, 2), DispatchClass::Operational)]
 		pub fn update_pallet_parameter(origin, parameter: T::Parameter) {
 			ensure_owner_or_root::<T, I>(origin)?;
--- a/modules/message-lane/src/lib.rs
+++ b/modules/message-lane/src/lib.rs
@@ -84,7 +84,7 @@ pub trait Config<I = DefaultInstance>: frame_system::Config {
 	type Event: From<Event<Self, I>> + Into<<Self as frame_system::Config>::Event>;
 	/// Benchmarks results from runtime we're plugged into.
 	type WeightInfo: WeightInfoExt;
-	/// Pallet parameteri that is opaque to the pallet itself, but may be used by the runtime
+	/// Pallet parameter that is opaque to the pallet itself, but may be used by the runtime
 	/// for integrating the pallet.
 	///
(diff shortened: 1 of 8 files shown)


modules/message-lane/src/lib.rs
```diff
@@ -84,6 +84,12 @@ pub trait Config<I = DefaultInstance>: frame_system::Config {
 	type Event: From<Event<Self, I>> + Into<<Self as frame_system::Config>::Event>;
 	/// Benchmarks results from runtime we're plugged into.
 	type WeightInfo: WeightInfoExt;
+	/// Pallet parameteri that is opaque to the pallet itself, but may be used by the runtime
```
tomusdrw, line 87:
```suggestion
	/// Pallet parameter that is opaque to the pallet itself, but may be used by the runtime
```

tomusdrw (approved):
Looks reasonable. Initially I thought that perhaps this whole machinery is not needed, since governance can just change arbitrary storage entries, but your approach makes it easy for the bridge owner to do that as well.

I feel that we could have a separate owner of the price oracle (to decouple "bridge ownership" and "updating price oracle feed") - simply for security measures, but I think this can be easily done as a runtime upgrade.

PR #732, opened 2021-02-16
Fix updated clippy grumbles

svyatonik:
clippy has stopped crashing on our codebase after I've updated it today, but it has couple of new warnings

### DIFF
--- a/bin/millau/runtime/src/lib.rs
+++ b/bin/millau/runtime/src/lib.rs
@@ -360,6 +360,7 @@ impl pallet_message_lane::Config for Runtime {
 	type MessageDispatch = crate::rialto_messages::FromRialtoMessageDispatch;
 }
 
+#[allow(clippy::from_over_into)]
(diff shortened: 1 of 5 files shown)


svyatonik:
Actually this clippy warning looks wrong - removing this "readundant slicing" leads to wrong code:
```
warning: redundant slicing of the whole range
   --> modules/ethereum-contract/builtin/src/lib.rs:103:47
    |
103 |                 .and_then(|log| ConsensusLog::decode(&mut &log[..]).ok())
    |                                                           ^^^^^^^^ help: use the original slice instead: `log`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_slicing
```

I don't want to change code much or introduce `allow`s. So better to wait until it'll be fixed

tomusdrw:
Perhaps explicit deref is needed? `&mut &*log`?

PR #733, opened 2021-02-16
Fix updated clippy grumbles

svyatonik:
this is reincarnation of #732 (force-pushed => can't reopen that other PR)
I've also reverted pinning of nightly version here

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -22,7 +22,7 @@ jobs:
         toolchain:
           - stable
           #- beta
-          - nightly-2020-12-17
+          - nightly
     runs-on:                       self-hosted
     container:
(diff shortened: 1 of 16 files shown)


tomusdrw (approved):

adoerr (commented):
Running `clippy` locally gives me the following error:

image

Any ideas?

svyatonik:
`rustup update` shall help - see #652

adoerr:
Nope. Ran `rustup update` already earlier this morning/

svyatonik:
Most probably you're using stable clippy version then, which we are not using.

adoerr:
Yep, that is it. Bit of a shame, since stable `clippy` do work well for me so far.

PR #736, opened 2021-02-17
Message lane integration documentation

svyatonik:
Things to do yet:
- [x] `bridge-runtime-common` documentation;
- [x] `pallet-bridge-call-dispatch` documentation;
- [x] cross-links between READMEs;
- [x] rereading it once again;
- [x] 'beautifying' all READMEs.


HCastano (approved):
I didn't "beautify" the tables, since imo the info there shouldn't actually be in tables - but that's something to address in the future since the actual content is fine.

Docs look good, thanks for the hard work on this Slava!

PR #739, opened 2021-02-17
Change SS58Prefixes

tomusdrw:
Resurrecting #722  cause it seems that simply bumping substrate does not solve the issue. I didn't have time to investigate further, but for the sake of development speed let's just change the SS58 prefixes for now.

CC @hbulgarini


adoerr (approved):
LGTM. Both prefixes are not allocated ATM according to [ss58-registry](https://github.com/paritytech/substrate/blob/master/ss58-registry.json)

PR #740, opened 2021-02-18
Disable self-hosted runner and sccache

General-Beck:


General-Beck:
Check and test [stable](https://github.com/paritytech/parity-bridges-common/runs/1923471847?check_suite_focus=true) 42 min
Check and test [nightly](https://github.com/paritytech/parity-bridges-common/runs/1923471857?check_suite_focus=true) 38 min
Check and cache stable with [cache](https://github.com/paritytech/parity-bridges-common/runs/1923730853?check_suite_focus=true) 37 min
Check and test nightly with [cache ](https://github.com/paritytech/parity-bridges-common/runs/1923730873?check_suite_focus=true) 25 min
Clippy first [check](https://github.com/paritytech/parity-bridges-common/runs/1923471919?check_suite_focus=true) 25 min
Clippy with [cache](https://github.com/paritytech/parity-bridges-common/runs/1923730715?check_suite_focus=true) 9 min

.github/workflows/rust.yml
```diff
@@ -102,6 +73,34 @@ jobs:
           args:                    --manifest-path ./bin/millau/node/Cargo.toml --no-default-features --features runtime-benchmarks --verbose
 
 ## Build Stage
+  build:
+    name:                          Build
```
tomusdrw, line 77: AFAICT we do `--release` build, but I don't think anyone is using the binaries, so we could cut down the time a bit by doing debug build instead.

tomusdrw (approved):
Awesome!

PR #743, opened 2021-02-18
updating paths to binaries

General-Beck:


svyatonik (approved):
Not sure where our artifacts are used :) But the fix seems ok to me

tomusdrw (approved):
Yeah, we don't really use the artifacts, but looks good.

PR #746, opened 2021-02-19
Send message doc

adoerr:


docs/send-message.md
```diff
@@ -0,0 +1,131 @@
...
+		--millau-host=127.0.0.1 \
+		--millau-port=10946 \
+		--millau-signer=//Dave \
+		--rialto-signer=//Dave \
+		--lane=00000000 \
+		--origin Target \
+		remark
+```
+Messages are basically regular transactions. That means, they have to be signed. In order
+to send a message, you have to control an account private key on both, the source and
```
svyatonik, line 117: That is only partially correct :) You only need private key on target chain when `--origin Target` specified (you need to sign that digest you've recently changed). With other origins, private key on target chain isn't required.

I guess there's an issue with relay cli now, though - it always requires target chain key. This probably comes from the moment when `--origin` param has been added. Anyway - we should fix it, until then we may leave this description as-is :)

docs/send-message.md
```diff
@@ -0,0 +1,131 @@
...
+the target chain. Those accounts are specified using the `--millau-signer` and `--rialto-signer`
+arguments in the example above. 
+
+Message delivery and dispatch requires a fee to be payed. In the example above, we have not
```
svyatonik, line 121:
```suggestion
Message delivery and dispatch requires a fee to be paid. In the example above, we have not 
```

svyatonik (approved):

PR #754, opened 2021-02-22
Set weird substrate global state for custom SS58 prefixes

tomusdrw:
Fixes https://github.com/polkadot-js/api/issues/3188#issuecomment-783347268

CC @hbulgarini


adoerr (approved):
LGTM

Issue #39, opened 2020-03-19
PoA -> Substrate: detect and report validator misbehavior when headers are submitted using unsigned transactions


svyatonik:
imagine #34 in unsigned transactions context - we can't penalize validator for spamming tx pool and blocks with good bad' headers AND validator can just do not relay these blocks to other PoA peers => it'll be never reported as malicious using builtin PoA mechanisms. So we need to detect such validators in bridge and report them back to PoA chain.

The easiest fix (suggested by @tomusdrw ): (c) "...should basically just make sure that every header seen on the bridge is gossiped in the network and that should be enough."

svyatonik:
Update: it would be good to use the same mechanism with signed transactions. Most problems with malicious validators will arise when unsigned transactions are used, but if we'll have that mechanism, then we would we simply ignore misbehavior when signed transactions are used?

tomusdrw:
> Update: it would be good to use the same mechanism with signed transaction

Yes, I think in fact it should be simply a separate component, that just listens to on-chain state (no matter how it is changed (signed / unsigned)) and then makes sure the headers are propagated.

tomusdrw:
Note that it should be done not only for headers, but also for finality proofs (justifications).

tomusdrw:
Re-labeling as Substrate<->Substrate bridge feature.

serban300:
Not sure what would be the best solution, but started implementing some primitives. This way it will also be easier to merge them to the `polkadot-staging` branch. And then we'll put all the pieces together.

PR #755, opened 2021-02-22
Use No-Op Ancestry Checker

HCastano:
This removes the use of a "real" ancestry checker when importing finalized headers.

The reason why we're removing this is because it doesn't really add much security to the
bridge. Consider the following: if a malicious actor is performing a long-range attack
on the bridge (which it can't protected against), ensuring that the forged headers being
submitted to the bridge are ancestors of each other is trivial.

The most important checks we need to do when importing a header are:
    - Verifying that the given justification is valid for the given header
    - Ensuring that the header was signed off by the correct authority set


svyatonik (commented):
Imo we need at least to protect against accidental finalization reverts - i.e. when we have finalized header 100 and then some relayer just submits correct finality proof of the header 50, causing havoc. I don't think there's protection against that in the code right now?

Also - since we don't have all headers now, what's the plan with scheduled changes? I mean - that's absolutely ok for block `N` to schedule changes to happen at block `N+X`. And we only need justification for the block which enacts changes (i.e. for block `N+X`, where `X` is normally zero, but let's consider the case when it's not). So there's no requirement in having justification for block `N` and if we will provide justification for block `N+X`, then we'll never know that the change should have happened at this block. Sorry - I may be missing something, tbh I'm lost in all those lat changes.

adoerr:
My understanding is that the GRANDPA [prove_finality()](https://github.com/paritytech/substrate/blob/904495bc14eff5171fca5912247e772948439411/client/finality-grandpa/rpc/src/lib.rs#L85-L92) RPC could be used to address those issues mentioned. The finality proofs returned by that RPC call also do account for authority set changes.

adoerr (commented):
My understanding is that the GRANDPA [prove_finality()](https://github.com/paritytech/substrate/blob/904495bc14eff5171fca5912247e772948439411/client/finality-grandpa/rpc/src/lib.rs#L85-L92) RPC could be used to address those issues mentioned. The finality proofs returned by that RPC call also do account for authority set changes.

svyatonik:
> My understanding is that the GRANDPA prove_finality() RPC could be used to address those issues mentioned. The finality proofs returned by that RPC call also do account for authority set changes.

RPC (offchain) cannot be used to address any onchain issues - it can be used to prepare data for the onchain transaction (or other alternatives may be used). Also - seems that now [it is](https://github.com/paritytech/substrate/blob/904495bc14eff5171fca5912247e772948439411/client/finality-grandpa/src/finality_proof.rs#L236) simple justification + ancestry proof and ancestry proof is exactly what's removed in this PR :)

adoerr:
> RPC (offchain) cannot be used to address any onchain issues -

I don't understand that. If that is generally true, then how can the (offchian) message relayer deal with (onchain) message delivery proofs, for example.

svyatonik:
So for justification we have signatures and authorities set. If signatures match authorities set, then we accept this "proof". For messages in message-lanes we have storage proof and storage root. If we successfully verify storage proof against storage root (which we take from pre-imported finalized header), then we accept this proof.

Here any relayer may just prepare arbitrary headers with anything he wants (e.g. switching to other authorities set that he controls in `X` blocks) and since we don't/can't verify this ancestry, we may only blindly trust that relayer provides correct information. This slightly differs from what has been before - we were verifying that there's a 'hash-path' from best finalized to new finalized header. And we have signatures for this new finalized header. So we only trusted authorities in this case (this doesn't protect us from long-range attacks, though). Compare this to new scheme (again - probably it has been already discussed somewhere && I just don't know the solution) where we trust both authorities and relayer.

adoerr:
> Here any relayer may just prepare arbitrary headers with anything he wants
Ok, so let's assume the relayer comes up with a set of arbitrary headers and claims that those headers have been finalized in block number 101.

Now we use `prove_finality()` and ask for the finality proof for block number 101. The justifications contained in the finality proof returned by `prove_finality()` would not match with what is provided by the relayer, right? What am I missing here?

tomusdrw (changes requested):
> when we have finalized header 100 and then some relayer just submits correct finality proof of the header 50, causing havoc.

That's a very good point and indeed it has to be addressed. I initially thought, that this is something that `HeaderChain` verifies, but since we use `append_header` method, which basically skips all the checks it's not the case.

So to summarize everything:
1. We should only import headers which have numbers greater than current best finalized header.
2. We have to know latest authority set.
2. We should only import headers which contain valid justifications signed by latest authority set.
3. If the imported header contains scheduled changes, we mark enactment block number as: `SCHEDULED + DELAY`.
4. If the next imported header with block number `N` is below `SCHEDULED + DELAY` we use current authority set.
5. If the next imported header with block number `N` is above `SCHEDULED + DELAY` we perform set transition (swap `latest <> pending`, clear `pending`)

GRANDPA is guaranteed to produce justifications for every block that schedules the change and afaict that's good enough for us.

The security is guaranteed by having the relayer work in a timely fashion (i.e. to avoid going beyond bonding period and session change) and #39 is now very important to implement to make sure we get some protection (not long-range) against validators collusion.

tomusdrw (changes requested):
> when we have finalized header 100 and then some relayer just submits correct finality proof of the header 50, causing havoc.

That's a very good point and indeed it has to be addressed. I initially thought, that this is something that `HeaderChain` verifies, but since we use `append_header` method, which basically skips all the checks it's not the case.

So to summarize everything:
1. We should only import headers which have numbers greater than current best finalized header.
2. We have to know latest authority set.
2. We should only import headers which contain valid justifications signed by latest authority set.
3. If the imported header contains scheduled changes, we mark enactment block number as: `SCHEDULED + DELAY`.
4. If the next imported header with block number `N` is below `SCHEDULED + DELAY` we use current authority set.
5. If the next imported header with block number `N` is above `SCHEDULED + DELAY` we perform set transition (swap `latest <> pending`, clear `pending`)

GRANDPA is guaranteed to produce justifications for every block that schedules the change and afaict that's good enough for us.

The security is guaranteed by having the relayer work in a timely fashion (i.e. to avoid going beyond bonding period and session change) and #39 is now very important to implement to make sure we get some protection (not long-range) against validators collusion.

svyatonik:
> > Here any relayer may just prepare arbitrary headers with anything he wants
> 
> Ok, so let's assume the relayer comes up with a set of arbitrary headers and claims that those headers have been finalized in block number 101.
> 
> Now we use `prove_finality()` and ask for the finality proof for block number 101. The justifications contained in the finality proof returned by `prove_finality()` would not match with what is provided by the relayer, right? What am I missing here?

Who are "we" in "we use `prove_finality()`"? :) Relayer that we can't trust? Runtime that can't access RPC? Again - this RPC method seems to return ancestry + justification. And we're going to remove ancestry check in this PR.

HCastano:
Since it wasn't totally clear from Tomek's comment, I want to clarify that removing the
ancestry checks isn't _wrong_ per say. However, there is a small check we need to add in
order to make this PR safer.

First off, I want to re-iterate that we don't need ancestry checks if for a given header
we:
    - Have a valid justification
    - Can verify that it was signed-off by the correct authority set

If a malicious actor can produce a valid justification, and have the correct authority
set sign off on a header then we're screwed anyways. Adding the requirement that the
current header has to be an ancestor of the last finalized header would be a trivial
addition step for this malicious actor.

As for the tweak needed: We need to ensure that submitted header has a higher number than
the last imported header. This prevents the bridge from time travelling as Slava pointed
out.

Note: If there exists a header with a valid justification at a lower height than the best
finalized header we have then that means there was a hard-fork on the source chain. In
this case the target chain's governance would need to step in to correct the bridge.

modules/substrate/src/lib.rs
```diff
@@ -375,8 +380,18 @@ impl<T: Config> bp_header_chain::HeaderChain<BridgedHeader<T>, sp_runtime::Dispa
...
-	fn append_header(header: BridgedHeader<T>) {
+	fn append_header(header: BridgedHeader<T>) -> Result<(), sp_runtime::DispatchError> {
+		// We do a quick check here to ensure that our header chain is making progress and isn't
+		// "travelling back in time" (which would be indicative of something bad, e.g a hard-fork).
+		//
+		// Note: It is valid for GRANDPA to finalize the same header twice as long as it is
+		// finalized in different rounds. This is why we don't check for a strictly greater-than
+		// relationship here.
+		let best_finalized = PalletStorage::<T>::new().best_finalized_header().header;
+		ensure!(best_finalized.number() <= header.number(), <Error<T>>::ConflictingFork);
```
svyatonik, line 391: Imo we shouldn't care about what happens in GRANDPA. The pallet should care whether transaction is improving its state. If transaction brings the same header and (the same or other) justification again, why should we go further - we'll end up with the same state (but with block that has been executed slower than it may be)?

svyatonik: Also two points:
1) malicious relayer may just submit the same header+justification over and over again to increase `RequestCount` counter value, thus blocking valid transactions to be mined and new headers to be finalized;
2) iiuc the second import of the header that schedules+enacts authorities set change will end up at [this line](https://github.com/paritytech/parity-bridges-common/blob/master/modules/substrate/src/lib.rs#L409). So we'll increase the `set_id` twice => the pallet would stuck.

modules/finality-verifier/src/lib.rs
```diff
@@ -141,7 +141,7 @@ pub mod pallet {
 				<Error<T>>::InvalidAncestryProof
 			);
 
-			T::HeaderChain::append_header(finality_target);
+			let _ = T::HeaderChain::append_header(finality_target)?;
```
svyatonik, line 144: I've probably missed some substrate updates, but last time I was checking, returning error from extrinsic **after** it  has made some storage changes wasn't considered a good practice (because storage changes aren't reverted in this case). So `RequestCount` will be increased despite execution failure. This may allow malicious relayer to submit invalid headers over and over again, blocking pallet from progressing.

modules/substrate/src/lib.rs
tomusdrw: I second Slava on this. Importing the same header twice really doesn't bring any new information. Let's be strict here.

modules/finality-verifier/src/lib.rs
HCastano: Yeah, this is a problem whose scope is out side of this PR: https://github.com/paritytech/srlabs_findings/issues/47#issuecomment-779096609.

modules/substrate/src/lib.rs
HCastano: > Imo we shouldn't care about what happens in GRANDPA. The pallet should care whether transaction is improving its state. If transaction brings the same header and (the same or other) justification again, why should we go further - we'll end up with the same state (but with block that has been executed slower than it may be)?

So _technically_ it's not the same state since GRANDPA has made progress by finalizing the same block in different rounds. However, from the perspective of the bridge you're right, we shouldn't care about this detail since we want "real" progress.

> iiuc the second import of the header that schedules+enacts authorities set change will end up at this line. So we'll increase the set_id twice => the pallet would stuck.

That's a good point.  In the future we'd ideally be getting the `set_id` directly from the digest, so this could be avoided. However, in its current form this doesn't work.

modules/finality-verifier/src/lib.rs
HCastano: > Actaully Slava has a good point here, it seems it was missed in previous reviews but a failed extrinsic (returning an error) MUST NOT alter the storage. It's recommended to either use transactional storage here or move storage modifications after all verifications are done.

I was aware that it was writing to storage despite having failed (specifically having failed to verify justifications, since that used to be the only exit point to the extrinsic). What I had in mind was that _any request_ that was made should count towards the request count. 

> Regarding the audit issue itself, since we don't import ancestry it doesn't matter how many blocks get finalized (indirectly). We should be limitting how fast the on-chain storage grows.

You make a good point here. The real goal here isn't to limit the number of requests, but instead to limit the _storage_ growth. In our case storage doesn't grow if somebody sends invalid data, so it doesn't matter if we don't keep track of that request. 

So I think it was my fault since I lost sight of the true goal of the rate limiter.

I will address this in a different PR since it's distinct from the ancestry checker changes in this PR.

svyatonik (approved):

modules/finality-verifier/src/lib.rs
HCastano: I've opened #765.

tomusdrw (approved):

Issue #747, opened 2021-02-19
Use self-hosted runner for Docker image build.


tomusdrw:
Seems that building Docker images is quite heavy (especially on disk space), we might consider using self-hosted runner just for this one job - it's being run only nightly anyway, so it shouldn't be a big deal.

Failing build: https://github.com/paritytech/parity-bridges-common/runs/1931930586?check_suite_focus=true

TriplEight:
also https://github.com/paritytech/parity-bridges-common/actions/runs/587642086

TriplEight:
closed with https://github.com/paritytech/parity-bridges-common/pull/766

PR #758, opened 2021-02-23
Generate large messages from deployment scripts

### LINKED ISSUE #749: Relay messages that have maximal size + maximal dispatch weight at the same time
Thanks to #680 and #700 we now may relay largest possible and max dispatch weight messages. Another (hope final) edge case is maximal-size message that declares maximal possible dispatch weight. Are we able to relay that? I suggest to complete this before P+K release.

svyatonik:
closes #749 
iiuc this won't work until we'll have new docker images with updated relay, so it is blocked by #747 => draft


deployments/bridges/rialto-millau/entrypoints/relay-messages-to-millau-generator-entrypoint.sh
```diff
@@ -58,4 +64,49 @@ do
 		transfer \
 		--amount 1000000000 \
 		--recipient $FERDIE_ADDR
+
+	# every other hour we're sending 3 large (size, weight, size+weight) messages
+	if [ $SECONDS -ge $LARGE_MESSAGES_TIME ]; then
+		LARGE_MESSAGES_TIME=$((SECONDS + 7200))
```
HCastano, line 70: I know it's not any moonmath, but having:

```bash
TWO_HOURS=7200
```

Would be nice imo

HCastano (approved):
This is neat, I like this. 

Curious to see if our network will heel over because of it, haha

deployments/bridges/rialto-millau/entrypoints/relay-messages-to-millau-generator-entrypoint.sh
tomusdrw: The variable is an overkill, if we want to make it more explicit I'd rather go with `SECONDS + 2 * 60 * 60`, but it's fine as-is for me.

tomusdrw (approved):
lgtm

PR #761, opened 2021-02-24
Relay Millau && Rialto headers using (future) finality verifier API

### LINKED ISSUE #703: Relay Proofs to Finality Verifier Pallet
Since the Finality Verifier pallet has now been added to the Millau and Rialto runtimes
we should start testing that it actually works by writing some relayer code that calls
the [`pallet_finality_verifier::submit_finality_proof()`](https://github.com/paritytech/parity-bridges-common/blob/master/modules/finality-verifier/src/lib.rs#L99)
extrinsic. To do this we need three things: a header, justification, and finality proof.

We already subscribe to justifications using the `grandpa_subscribeJustifications` RPC
[here](https://github.com/paritytech/parity-bridges-common/blob/master/relays/substrate-client/src/client.rs#L259). The stream of justifications are used in the
[`headers_maintain`](https://github.com/paritytech/parity-bridges-common/blob/master/relays/substrate/src/headers_maintain.rs) pipeline. We can probably leverage some similar code to get the 
`finality_target` and `justification` required by `pallet-finality-verifier`.

As for the ancestry proof, we keep a few queues of the headers
[here](https://github.com/paritytech/parity-bridges-common/blob/master/relays/headers-relay/src/headers.rs). I think we can add a queue to keep track of the headers between
`[last_finalized, finality_target]`.

With this in hand I think what we need to do next is add a method to
[`SubstrateHeaderSyncPipeline`](https://github.com/paritytech/parity-bridges-common/blob/master/relays/substrate/src/headers_pipeline.rs#L36) that actually generates a transaction with a call to
`submit_finality_proof()`.

Since @svyatonik is the expert when it comes to the relayer code, do you think this
sounds reasonable, or are there other places in the code that need to be updated to
accomodate this?

svyatonik:
closes #703

I've removed all obsolete sub2sub headers sync that was using `headers-relay`. `headers-relay` is still used by sub2eth and eth2sub - we may reconsider this in future and remove this crate at all. So the new `finality-relay` crate is built around syncing finality proofs (justifications for Substrate). Most of the stuff from `headers-relay` isn't actually necessary now && new relay loop is more like exchange loop (not fully async) - there aren't any states && internal buffers (there's one actually, but it's only for faster sync).

New crate needs some more tests - I'll add them here, if review process will be slow. Or in a follow-up PR, if it'll be reviewed fast.

UPD: we may also remove "Blocks in state" widget from sub2sub headers sync dashboard. It'll have no data anyway. Will be done in follow-up PRs


HCastano:
Would close #703.

HCastano (commented):
Haven't been able to get through this whole PR yet. I'll try and finish up by tomorrow, but I think Monday is more likely.

By the way, have you tried syncing Rialto and Millau with this yet?

primitives/header-chain/src/lib.rs
```diff
@@ -138,6 +138,22 @@ impl<H: HeaderT> AncestryChecker<H, Vec<H>> for LinearAncestryChecker {
 	}
 }
 
+/// Find header digest that schedules next GRANDPA authorities set.
+pub fn find_grandpa_authorities_scheduled_change<H: HeaderT>(
```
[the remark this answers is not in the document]
svyatonik: I think it's required - it lays within something that is called `bp_header_chain`, which doesn't mention GRANDPA. There may be (and are) other changes in header digest.

relays/finality-relay/src/finality_loop.rs
```diff
@@ -0,0 +1,558 @@
...
+	/// Timeout before we treat our transactions as lost.
+	pub stall_timeout: Duration,
+}
+
+/// Source client used in finality synchronization loop.
+#[async_trait]
+pub trait SourceClient<P: FinalitySyncPipeline>: RelayClient {
+	/// Stream of new finality proofs. The stream is allowed to miss proofs for some
+	/// headers, even if those headers are mandatory.
+	type FinalityProofsStream: Stream<Item = P::FinalityProof>;
```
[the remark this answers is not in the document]
svyatonik: It's stream that has multiple proofs, not a single proof.

relays/finality-relay/src/finality_loop.rs
```diff
@@ -0,0 +1,558 @@
...
+
+	/// Get best finalized block number.
+	async fn best_finalized_block_number(&self) -> Result<P::Number, Self::Error>;
```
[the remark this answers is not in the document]
svyatonik: Connection lost, decode failed, versions mismatch, ...

relays/finality-relay/src/finality_loop.rs
```diff
@@ -0,0 +1,558 @@
...
+	/// to the target chain => bridge applications will run faster, but pallet storage may explode
+	/// (but if pruning is there, then it's fine).
+	pub tick: Duration,
+	/// Number of finality proofs to keep in internal buffer between loop wakeups.
+	///
+	/// While in "major syncing" state, we still read finality proofs from the stream. They're stored
+	/// in the internal buffer between loop wakeups. When we're close to the tip of the chain, we may
+	/// met finality delays if headers are not finalized frequently. So instead of waiting for next
+	/// finality proof to appear in the stream, we may use existing proof from that buffer.
+	pub recent_finality_proofs_limit: usize,
```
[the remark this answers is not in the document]
svyatonik: It limits buffer length && `usize` is the type that is used for lengths.

relays/finality-relay/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+	/// Finality proof type.
+	type FinalityProof: FinalityProof<Self::Number>;
+}
+
+/// Header that we're receiving from source node.
+pub trait SourceHeader<Number>: Clone + Debug + PartialEq + Send + Sync {
+	/// Returns number of header.
+	fn number(&self) -> Number;
+	/// Returns true if this header needs to be submitted to target node.
+	fn is_mandatory(&self) -> bool;
```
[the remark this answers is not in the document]
svyatonik: It's up to implementer. If we'll ever use this pallet for poa2sub, then it'll be the header where special contract event is emitted (so we'll be looking at the bloom && returning `true` if it has some bits set).

relays/finality-relay/src/finality_loop_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+struct TestFinalityProof(Option<TestNumber>);
+
+impl FinalityProof<TestNumber> for TestFinalityProof {
+	fn target_header_number(&self) -> Option<TestNumber> {
+		self.0
+	}
+}
+
+#[derive(Debug, Clone, Default)]
+struct ClientsData {
```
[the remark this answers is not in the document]
svyatonik: It holds both clients data

primitives/header-chain/src/lib.rs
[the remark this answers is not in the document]
HCastano: The only other change (that I'm aware of) is a force change.

I prefer the shorter name, but I won't you fight over this :)

relays/finality-relay/src/finality_loop.rs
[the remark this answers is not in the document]
HCastano: I don't quite have a good counterexample here, but from my point of view having an `<Object>Stream` implies that we're going to have multiple `<Object>`s coming through - so the extra "s" is unnecessary.

relays/finality-relay/src/finality_loop.rs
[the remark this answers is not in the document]
HCastano: Do you mind adding that to the comment so implementors know this?

relays/finality-relay/src/lib.rs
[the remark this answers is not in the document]
HCastano: Do you mind adding an example or two to the comment?

relays/finality-relay/src/finality_loop.rs
```diff
@@ -0,0 +1,563 @@
...
+				.submit_finality_proof(header, justification)
+				.await
+				.map_err(Error::Target)?;
+			Ok(Some(new_transaction))
+		}
+		None => Ok(None),
+	}
+}
+
+async fn select_header_to_submit<P, SC, TC>(
```
HCastano, line 354: This is a little dense. I think it could benefit from being split into a couple of different functions + a high level description of what's going on here

relays/finality-relay/src/finality_loop_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+	async fn submit_finality_proof(&self, header: TestSourceHeader, proof: TestFinalityProof) -> Result<(), TestError> {
+		let mut data = self.data.lock();
+		(self.on_method_call)(&mut *data);
+		data.target_best_block_number = header.number();
+		data.target_headers.push((header, proof));
+		Ok(())
+	}
+}
+
+fn run_sync_loop(state_function: impl Fn(&mut ClientsData) -> bool + Send + Sync + 'static) -> ClientsData {
```
HCastano, line 168: I don't understand how this test setup function works, can you ELI5 it for me?

relays/finality-relay/src/finality_loop_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+		// header#9 has persistent finality proof, but it isn't mandatory => it is submitted, because
+		//   there are no more persistent finality proofs
+		//
+		// once this ^^^ is done, we generate more blocks && read proof for blocks 12, 14 and 16 from the stream
+		// but we only submit proof for 16
+		//
+		// proof for block 15 is ignored - we haven't managed to decode it
+		if data.target_best_block_number == 9 {
+			data.source_best_block_number = 17;
+			data.source_headers.insert(11, (TestSourceHeader(false, 12), None));
```
HCastano, line 231: Why are we inserting `TestSourceHeader(false, 12)` at height 11, and then inserting the same `TestSourceHeader` at height 12 right after?

relays/finality-relay/src/finality_loop_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+	assert_eq!(
+		prune_unjustified_headers::<TestFinalitySyncPipeline>(10, &mut unjustified_headers),
+		Some(TestSourceHeader(false, 10)),
+	);
+	assert_eq!(
+		&original_unjustified_headers.clone().make_contiguous()[1..],
+		unjustified_headers.make_contiguous(),
+	);
+
+	// when there's no header in the collection
```
HCastano, line 280:
```suggestion
	// when the header doesn't exist in the collection
```

relays/finality-relay/src/finality_loop_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+		prune_unjustified_headers::<TestFinalitySyncPipeline>(19, &mut unjustified_headers),
+		Some(TestSourceHeader(false, 19)),
+	);
+
+	assert_eq!(
+		&original_unjustified_headers.clone().make_contiguous()[5..],
+		unjustified_headers.make_contiguous(),
+	);
+
+	// when post-last entry is pruned
```
HCastano, line 303:
```suggestion
	// when we try and prune past last entry
```

relays/finality-relay/src/finality_loop_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+		&original_unjustified_headers.clone().make_contiguous()[1..],
+		unjustified_headers.make_contiguous(),
+	);
+
+	// when last entry is pruned
+	let mut unjustified_headers = original_unjustified_headers.clone();
+	assert_eq!(
+		prune_unjustified_headers::<TestFinalitySyncPipeline>(19, &mut unjustified_headers),
+		Some(TestSourceHeader(false, 19)),
+	);
```
HCastano, line 296: Is there anything special about the last entry? Just trying to figure out why you have a test for it explicitly

relays/finality-relay/src/finality_loop_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+		recent_finality_proofs.make_contiguous(),
+	);
+
+	// when there are no proof for justified header in the deque
+	let mut recent_finality_proofs = original_recent_finality_proofs.clone();
+	prune_recent_finality_proofs::<TestFinalitySyncPipeline>(11, &mut recent_finality_proofs, 1024);
+	assert_eq!(
+		&original_recent_finality_proofs.clone().make_contiguous()[1..],
+		recent_finality_proofs.make_contiguous(),
+	);
```
HCastano, line 341: It's not clear to me what happens here - does the first header (since its the most recent compared with the given header) get pruned?

relays/finality-relay/src/finality_loop.rs
```diff
@@ -0,0 +1,563 @@
...
+		recent_finality_proofs,
+		recent_finality_proofs_limit,
+		|(header_number, _)| *header_number,
+	);
+}
+
+fn prune_ordered_deque<T, Number: relay_utils::BlockNumberBase>(
+	header_number: Number,
+	ordered_deque: &mut VecDeque<T>,
+	maximal_deque_size: usize,
```
HCastano, line 503: Maybe something like `pruned_deque_size` would be more clear here

relays/finality-relay/src/finality_loop.rs
```diff
@@ -0,0 +1,563 @@
...
+					qed",
+			);
+			*ordered_deque = updated_deque;
+			Some(extracted_element)
+		}
+		Err(position) => {
+			*ordered_deque = ordered_deque.split_off(position);
+			None
+		}
+	};
```
HCastano, line 528: Since the header found here is only used by `prune_unjustified_headers()` I would move this code there. Then, since the splitting code is pretty short, I'd just inline that in `prune_recent_finality_proofs()` and `prune_unjustified_headers()`

relays/finality-relay/src/finality_loop.rs
```diff
@@ -0,0 +1,558 @@
...
+					None
+				},
+				exit_signal.clone(),
+			)
+		},
+	);
+}
+
+/// Unjustified headers container. Ordered by header number.
+pub(crate) type UnjustifiedHeaders<P> = VecDeque<<P as FinalitySyncPipeline>::Header>;
```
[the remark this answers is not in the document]
HCastano: The space makes it more legible

HCastano (commented):
I still haven't been able to look at everything yet

relays/finality-relay/src/finality_loop.rs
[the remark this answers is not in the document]
adoerr: Reading the comment, isn't that more like a `buffered_proofs_limit`?

relays/finality-relay/src/finality_loop.rs
```diff
@@ -0,0 +1,563 @@
...
+	/// Number of finality proofs to keep in internal buffer between loop wakeups.
+	///
+	/// While in "major syncing" state, we still read finality proofs from the stream. They're stored
+	/// in the internal buffer between loop wakeups. When we're close to the tip of the chain, we may
+	/// meet finality delays if headers are not finalized frequently. So instead of waiting for next
+	/// finality proof to appear in the stream, we may use existing proof from that buffer.
+	pub recent_finality_proofs_limit: usize,
+	/// Number of recent finality proofs to keep.
+	///
+	/// When we're in "major syncing" state, we still read finality proofs from the stream and
```
adoerr, line 62: Same here w.r.t. 'major syncing'.

relays/finality-relay/src/finality_loop.rs
```diff
@@ -0,0 +1,563 @@
...
+pub trait SourceClient<P: FinalitySyncPipeline>: RelayClient {
+	/// Stream of new finality proofs. The stream is allowed to miss proofs for some
+	/// headers, even if those headers are mandatory.
+	type FinalityProofsStream: Stream<Item = P::FinalityProof>;
+
+	/// Get best finalized block number.
+	async fn best_finalized_block_number(&self) -> Result<P::Number, Self::Error>;
+
+	/// Get canonical header and its finality proof by number.
+	async fn header_and_finality_proof_by_number(
```
adoerr, line 78: `canonical_finality_proof()`. The fact that the selection criteria is a header number, should be clear from the signature. Same is true for the result, i.e. that we do get a (Header, FinalityProof) tuple.

adoerr (commented):
Couple of comments. In general, I **really** would prefer having PRs like this split up into smaller ones.

primitives/header-chain/src/lib.rs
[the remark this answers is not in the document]
svyatonik: > The only other change (that I'm aware of) is a force change.

There's also at least BABE and Aura, which have their own consensus changes with digests. Don't know about other consensus algorithms, but pretty sure that every will have something like that (if they're sealing blocks with signatures).

> The header_chain module is clearly tied to GRANDPA by using `sp_finality_grandpa`

Looking at crate dependencies/module usages to understand what the function does isn't a super-optimal workflow imo. When you're exposing public function, it should be clear what it does just by looking at its path+name. Otherwise you'll need all those extra steps (looking into deps, usages) to understand what's happening. And since header-chain is neutral name, it'll be confusing to provide something that returns anything other than grandpa-authorities-scheduled-change imo. This name may be shortened if we'll change crate name to something grandpa-specific.

relays/finality-relay/src/finality_loop.rs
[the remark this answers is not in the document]
svyatonik: Googling e.g. "notifications stream" gives me 23700 results, while googling "notification stream" only gives 13500 results. So it's a common practice to use plural in the stream type. I'd imagine `<Object>Stream` is linked to single object - e.g. to return events that happen with this object (I'd use longer name, like `<Object>EventsStream` though).

relays/finality-relay/src/finality_loop.rs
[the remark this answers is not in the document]
svyatonik: This documentation won't be helpful - why should I suggest implementer to return "connection lost" error if e.g. he has no connections and reads finality proofs from file? There's an error which implementer defines - he is able to define subset of errors that he expects to return and there are traits that we require from these errors (`Debug`) - imo that's enough.

relays/finality-relay/src/finality_loop.rs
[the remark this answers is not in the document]
svyatonik: > `PreliminaryHeaders`

Let me explain my logic - probably it'll help. Justification is used in Substrate to denote some data, associated with header/block that may be interpreted by consensus algorithm (GRANDPA in finality proof context) to prove that the block/header is valid for this algorithm. So unjustified here means headers without this associated data - we're waiting for this data to select single header from this container and submit it to the target node.

If you still think that `unjustified` doesn't fit here - please ping me again && let's change it. But I'd prefer something like `HeadersBuffer` then :)

relays/finality-relay/src/lib.rs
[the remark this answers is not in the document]
svyatonik: I don't want to include documentation about specific implementation details into generic trait. The crate is generic && don't know anything about GRANDPA, eth contracts, blooms and whatever.

relays/finality-relay/src/finality_loop.rs
svyatonik: Yeah - I have already splitted it for tests in other branch. Since this PR has already received too many comments, I don't want to merge these two branches now. Will do in follow-up PR

relays/finality-relay/src/finality_loop_tests.rs
svyatonik: It just prepares hardcoded clients data (see `clients_data` initialization for details) and runs the loop until `state_function` returns `true`.

relays/finality-relay/src/finality_loop_tests.rs
svyatonik: That's a typo

relays/finality-relay/src/finality_loop_tests.rs
svyatonik: The implementation uses `binary_search_by_key` and `split_off` - there's a lot of ways you could screw up with last element when using this function

relays/finality-relay/src/finality_loop_tests.rs
svyatonik: Yes

relays/finality-relay/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+/// Abstract finality proof that is justifying block finality.
+pub trait FinalityProof<Number>: Clone + Send + Sync + Debug {
```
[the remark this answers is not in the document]
svyatonik: If we'll ever need to `verify` proof in the relay, it'll be here :) `FinalityTarget` is clearly wrong - what do you mean by target?

relays/finality-relay/src/finality_loop.rs
```diff
@@ -0,0 +1,563 @@
...
+	/// This parameter may be used to limit transactions rate. Increase the value && you'll get
+	/// rare updates => sparse headers => potential slow down of bridge applications, but pallet storage
+	/// won't be super large. Decrease the value to near `source_block_time` and you'll get
+	/// transaction for (almost) every block of the source chain => all source headers will be known
+	/// to the target chain => bridge applications will run faster, but pallet storage may explode
+	/// (but if pruning is there, then it's fine).
+	pub tick: Duration,
+	/// Number of finality proofs to keep in internal buffer between loop wakeups.
+	///
+	/// While in "major syncing" state, we still read finality proofs from the stream. They're stored
```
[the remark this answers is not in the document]
svyatonik: It's a well-known term in the blockchain. Basically - if your client is missing many blocks/headers, then it may change its behavior to avoid wasting CPU/memory on stuff that isn't required to reach the tip of the chain. Like it may reject all incoming transactions, block announcements, blocks requests and etc. Then when number of unsynced headers falls below some limit, client may switch to syncing/importing state where it starts behaving normally again.

relays/substrate/src/finality_pipeline.rs
```diff
@@ -0,0 +1,130 @@
...
+use finality_relay::{FinalitySyncParams, FinalitySyncPipeline};
+use relay_substrate_client::{
+	finality_source::{FinalitySource, Justification},
+	BlockNumberOf, Chain, Client, Error as SubstrateError, HashOf, SyncHeader,
+};
+use relay_utils::BlockNumberBase;
+use std::{fmt::Debug, marker::PhantomData, time::Duration};
+
+/// Default synchronization loop timeout.
+const STALL_TIMEOUT: Duration = Duration::from_secs(120);
```
HCastano, line 32:
```suggestion
const STALL_TIMEOUT: Duration = Duration::from_secs(120);

```

relays/substrate/src/finality_target.rs
```diff
@@ -0,0 +1,91 @@
...
+// Parity Bridges Common is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Substrate client as Substrate finality proof target. The chain we connect to should have
+//! runtime that implements `<BridgedChainName>HeaderApi` to allow bridging with
```
HCastano, line 18: FYI, I've changed this to `<BridgedChainName>FinalityApi` in my pallet refactor PR

relays/substrate/src/millau_headers_to_rialto.rs
```diff
@@ -17,70 +17,52 @@
...
+pub(crate) type MillauFinalityToRialto = SubstrateFinalityToSubstrate<Millau, Rialto, RialtoSigningParams>;
 
 #[async_trait]
-impl SubstrateHeadersSyncPipeline for MillauHeadersToRialto {
-	const BEST_BLOCK_METHOD: &'static str = BEST_MILLAU_BLOCKS_METHOD;
-	const FINALIZED_BLOCK_METHOD: &'static str = FINALIZED_MILLAU_BLOCK_METHOD;
-	const IS_KNOWN_BLOCK_METHOD: &'static str = IS_KNOWN_MILLAU_BLOCK_METHOD;
-	const INCOMPLETE_HEADERS_METHOD: &'static str = INCOMPLETE_MILLAU_HEADERS_METHOD;
+impl SubstrateFinalitySyncPipeline for MillauFinalityToRialto {
+	const BEST_FINALIZED_SOURCE_HEADER_ID_AT_TARGET: &'static str = bp_millau::FINALIZED_MILLAU_BLOCK_METHOD;
```
HCastano, line 35: Another heads up that this will need to change once my refactor PR goes in

relays/substrate/src/rialto_headers_to_millau.rs
```diff
@@ -14,72 +14,55 @@
...
+pub(crate) type RialtoFinalityToMillau = SubstrateFinalityToSubstrate<Rialto, Millau, MillauSigningParams>;
 
 #[async_trait]
-impl SubstrateHeadersSyncPipeline for RialtoHeadersToMillau {
-	const BEST_BLOCK_METHOD: &'static str = BEST_RIALTO_BLOCKS_METHOD;
-	const FINALIZED_BLOCK_METHOD: &'static str = FINALIZED_RIALTO_BLOCK_METHOD;
-	const IS_KNOWN_BLOCK_METHOD: &'static str = IS_KNOWN_RIALTO_BLOCK_METHOD;
-	const INCOMPLETE_HEADERS_METHOD: &'static str = INCOMPLETE_RIALTO_HEADERS_METHOD;
+impl SubstrateFinalitySyncPipeline for RialtoFinalityToMillau {
+	const BEST_FINALIZED_SOURCE_HEADER_ID_AT_TARGET: &'static str = bp_rialto::FINALIZED_RIALTO_BLOCK_METHOD;
```
HCastano, line 35: Same thing here

relays/substrate/src/finality_target.rs
```diff
@@ -0,0 +1,91 @@
...
+	P::Number: Decode,
+	P::Hash: Decode,
+	P: SubstrateFinalitySyncPipeline,
+{
+	async fn best_finalized_source_block_number(&self) -> Result<P::Number, SubstrateError> {
+		// we can't continue to relay finality if target node is out of sync, because
+		// it may have already received (some of) headers that we're going to relay
+		self.client.ensure_synced().await?;
+
+		Ok(crate::messages_source::read_client_state::<C, P::Hash, P::Number>(
```
HCastano, line 75: `read_client_state` seems generally useful, I don't think it should be part of `messages_source`. You don't need to change it in this PR if it's too annoying to do though

relays/substrate/src/finality_target.rs
```diff
@@ -0,0 +1,91 @@
...
+		.best_finalized_peer_at_best_self
+		.0)
+	}
+
+	async fn submit_finality_proof(&self, header: P::Header, proof: P::FinalityProof) -> Result<(), SubstrateError> {
+		self.pipeline
+			.make_submit_finality_proof_transaction(header, proof)
+			.and_then(|tx| self.client.submit_extrinsic(Bytes(tx.encode())))
+			.await
+			.map(drop)
```
HCastano, line 89:
```suggestion
			.map(|_| ())
```

nit: I find this easier to read, but don't worry if you don't want to change it

relays/substrate/src/millau_headers_to_rialto.rs
```diff
@@ -17,70 +17,52 @@
 //! Millau-to-Rialto headers sync entrypoint.
```
HCastano, line 17:
```suggestion
//! Millau-to-Rialto finality sync entrypoint.
```

Although I wouldn't say that header sync is wrong either. It's just confusing during this transition time. What do you think?

relays/substrate/src/rialto_headers_to_millau.rs
```diff
@@ -14,72 +14,55 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
 
-//! Rialto-to-Millau headers sync entrypoint.
+//! Millau-to-Rialto headers sync entrypoint.
```
HCastano, line 17:
```suggestion
//! Rialto-to-Millau finality sync entrypoint.
```

relays/substrate/src/rialto_headers_to_millau.rs
```diff
@@ -14,72 +14,55 @@
...
-		let transaction = Millau::sign_transaction(&self.target_client, &self.target_sign.signer, nonce, call);
-		Ok(transaction)
-	}
-
-	async fn make_complete_header_transaction(
+	async fn make_submit_finality_proof_transaction(
 		&self,
-		id: RialtoHeaderId,
-		completion: Justification,
+		header: relay_rialto_client::SyncHeader,
```
HCastano, line 41: You import and typedef this in `millau_headers_to_rialto.rs`. I'd be consistent (doesn't matter which way you choose though)

relays/substrate/src/rialto_headers_to_millau.rs
```diff
@@ -14,72 +14,55 @@
...
+			(),
+		)
+		.into();
 		let transaction = Millau::sign_transaction(&self.target_client, &self.target_sign.signer, nonce, call);
 		Ok(transaction)
 	}
 }
 
-/// Run Rialto-to-Millau headers sync.
+/// Run Millau-to-Rialto finality sync.
```
HCastano, line 57:
```suggestion
/// Run Rialto-to-Millau finality sync.
```

relays/finality-relay/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "finality-relay"
+version = "0.1.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+license = "GPL-3.0-or-later WITH Classpath-exception-2.0"
+description = "Finality proofs relay"
```
HCastano, line 7:
```suggestion
description = "Relay finality proofs from a source Substrate chain to a target Substrate chain."
```

relays/finality-relay/src/finality_loop.rs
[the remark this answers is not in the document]
adoerr: Thanks for the explanation! I know this under the term 'fast-sync' or our own one, 'warp-sync'.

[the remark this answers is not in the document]
svyatonik: Warp-sync is slightly another type of sync, where you're allowed to skip some headers/blocks and jump directly to some future header/block/state. So it requires special protocol, probably special data (like that snapshot on eth), etc. Major sync is just a local (your own) client state, where it decides that is too far beyond the tip && should probably concentrate on importing headers/blocks using regular import pipeline.

relays/finality-relay/Cargo.toml
svyatonik: This is clearly wrong - this crate knows nothing about Substrate

relays/substrate/src/millau_headers_to_rialto.rs
svyatonik: Why would this need to be changed? Have you change the constant value there or introduced another constant? If there are two constants now, why would we need both?

HCastano: I've introduced a constant called `BEST_FINALIZED_MILLAU_HEADER_METHOD`. I didn't remove the exiting constants so things wouldn't end up insta-breaking.

After my PR is merged I'll clean up the unused constants as a followup, and in that I'll probably be the one to update the methods used by the relayer as well.

relays/finality-relay/Cargo.toml
HCastano: Yeah, I guess it doesn't know anything about it - but as humans we know, and it could be helpful if we publish these crates at some point.

svyatonik: As a human, I'm saying that this part: "from a source Substrate chain to a target Substrate chain" is clearly wrong.

relays/substrate/src/millau_headers_to_rialto.rs
svyatonik: Why would they insta-break? Do you have something missing from your crate?

HCastano (approved):

relays/finality-relay/src/finality_loop.rs
```diff
@@ -0,0 +1,558 @@
...
+			(false, None) => {
+				unjustified_headers.push(header);
+			}
+		}
+
+		header_number = header_number + One::one();
+	}
+
+	// see if we can improve finality by using recent finality proofs
+	if !unjustified_headers.is_empty() && !recent_finality_proofs.is_empty() {
```
tomusdrw, line 408: Shouldn't this step be only performed if the lower loop didn't select any header? Right now it seems that we are trying to get recent proof, perform some pruning, but later we are overwriting this again (possibly couple of time in `loop` iterations).

Since you've mentioned that it's changed in a follow up PR with tests, I'm fine to revisit later.

relays/finality-relay/src/finality_loop.rs
```diff
@@ -0,0 +1,558 @@
...
+		};
+		let finality_proof_target_header_number = match finality_proof.target_header_number() {
+			Some(target_header_number) => target_header_number,
+			None => {
+				continue;
+			}
+		};
+
+		let justified_header =
+			prune_unjustified_headers::<P>(finality_proof_target_header_number, &mut unjustified_headers);
```
tomusdrw, line 452: Why are we doing pruning in a loop? Wouldn't it be better to first run the loop and select the "best" finality proof to submit and only then prune everything once?

relays/finality-relay/src/finality_loop.rs
```diff
@@ -0,0 +1,558 @@
...
+	prune_recent_finality_proofs::<P>(
+		oldest_finality_proof_to_keep,
+		recent_finality_proofs,
+		sync_params.recent_finality_proofs_limit,
+	);
+
+	Ok(selected_finality_proof)
+}
+
+pub(crate) fn prune_unjustified_headers<P: FinalitySyncPipeline>(
```
tomusdrw, line 477:
```suggestion
/// Remove headers from `unjustified_headers` collection with number lower or equal than `justified_header_number`.
///
/// Returns the header that matches `justified_header_number` (if any).
pub(crate) fn prune_unjustified_headers<P: FinalitySyncPipeline>(
```

relays/finality-relay/src/finality_loop.rs
```diff
@@ -0,0 +1,558 @@
...
+
+fn print_sync_progress<P: FinalitySyncPipeline>(
+	progress_context: (Instant, Option<P::Number>),
+	best_number_at_source: P::Number,
+	best_number_at_target: P::Number,
+) -> (Instant, Option<P::Number>) {
+	let (prev_time, prev_best_number_at_target) = progress_context;
+	let now = Instant::now();
+
+	let need_update = now - prev_time > Duration::from_secs(10)
```
tomusdrw, line 540:
```suggestion
	let need_update = prev_time.elapsed() > Duration::from_secs(10)
```
?

relays/finality-relay/src/lib.rs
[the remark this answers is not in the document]
tomusdrw: Currently `finality-relay` crate does not depend on any substrate components, while this could be simplified at this point (and maybe didn't need to be generic in the first place) it will require substantial effort which I'd rather consider spending after we run a test deployment.

relays/substrate/src/finality_target.rs
tomusdrw: I'm sorry @svyatonik, I've actually reviewed the PRs in the incorrect order - probably the refactor should go after this to avoid these conflicts.

tomusdrw (approved):
In general looks good to me, few clarifications might be good, but other than that let's merge it ASAP. I'm sorry @svyatonik I took so long to do a review for this one.

relays/finality-relay/src/finality_loop.rs
```diff
@@ -0,0 +1,558 @@
...
+
+		// find last proof from intersection
+		let selected_finality_proof_index = recent_finality_proofs
+			.binary_search_by_key(intersection.end(), |(number, _)| *number)
+			.unwrap_or_else(|index| index.saturating_sub(1));
+		let (selected_header_number, finality_proof) = &recent_finality_proofs[selected_finality_proof_index];
+		if intersection.contains(selected_header_number) {
+			// now remove all obsolete headers and extract selected header
+			let selected_header = prune_unjustified_headers::<P>(*selected_header_number, &mut unjustified_headers)
+				.expect("unjustified_headers contain all headers from intersection; qed");
```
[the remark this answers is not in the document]
svyatonik: `unjustified_headers` can't have gaps - every `header` from the interval is read and pushed into this `vec`. And we only remove headers from this `vec` by splitting it.

relays/finality-relay/src/finality_loop.rs
```diff
@@ -0,0 +1,558 @@
...
+
+		// we have two ranges => find intersection
+		let intersection_begin = std::cmp::max(unjustified_range_begin, buffered_range_begin);
+		let intersection_end = std::cmp::min(unjustified_range_end, buffered_range_end);
+		let intersection = intersection_begin..=intersection_end;
+
+		// find last proof from intersection
+		let selected_finality_proof_index = recent_finality_proofs
+			.binary_search_by_key(intersection.end(), |(number, _)| *number)
+			.unwrap_or_else(|index| index.saturating_sub(1));
```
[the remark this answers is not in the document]
svyatonik: `intersection` may be non-empty, but there may be no entries in the `recent_finality_proofs` in this `intersection`. So this won't work.

relays/finality-relay/src/finality_loop.rs
svyatonik: Yes, it may be done - I'll probably do that in next PR. But please keep in mind that normally we'll either never reach this point with more than 1-2 headers in `unfinalized_headers`, or will reach it only once in a relayer lifetime. It is when relayer just have started and there are many unsynced headers from the last GRANDPA session.

relays/finality-relay/src/finality_loop.rs
svyatonik: We'll need `now` anyways - it is a part of `progress_context`. So I'm just reusing the same `now` value.

PR #762, opened 2021-02-24
Add inspect command to take a look at extrinsics.

tomusdrw:
CC @hbulgarini

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1839,7 +1839,7 @@ checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 [[package]]
 name = "fork-tree"
 version = "3.0.0"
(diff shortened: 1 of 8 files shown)


adoerr (commented):
Running `cargo test` locally for this PR, the [should_fail_on_weight_mismatch()](https://github.com/paritytech/parity-bridges-common/blob/345e84a2146b56628e9888c9f5e129cb40e868a9/modules/call-dispatch/src/lib.rs#L592-L615) test always fails unless I change `assert_eq!` to `assert_ne!`.

adoerr:
The weight returned from `get_dispatch_info()` is `1279000` expected is `1973000`.

adoerr:
`WeightInfo` for `remark` has been changed with the [Remove suicide from frame_system weights](https://github.com/paritytech/substrate/commit/ad8c585d1c1704b534a92c49e0700080c9d8e13d) PR.

I think it is a bit unfortunate to do this kind of change without a bit of a heads-up for everyone.

PR #763, opened 2021-02-25
Add default for some options.

tomusdrw:
Add some sensible defaults to make it easier to use the CLI to send transactions.


adoerr (approved):
LGTM

PR #765, opened 2021-02-25
Stop counting invalid requests towards rate limit

HCastano:
This PR stops counting all the calls made to `submit_finality_proof()` as requests.
Instead it will only count calls which _actually_ write to storage as requests.

The reason why we don't want to count invalid calls as a request is that these calls
do not end up writing to storage. The goal of the rate limiter isn't to limit requests
per say, but instead to limit the rate of state growth.


adoerr (approved):
LGTM

svyatonik (approved):

tomusdrw (approved):

PR #766, opened 2021-02-25
CI: self-host docker build

TriplEight:
Changes building and publishing containers to a self-hosted runner.
Please raise an issue to inspect your dockerfiles, it's maybe something over-optimised is going on there.
Also FYI, these self-hosted runners are subject to change soon, they'll have to be re-deployed.

### DIFF
--- a/.github/workflows/publish-deps.yml
+++ b/.github/workflows/publish-deps.yml
@@ -15,7 +15,7 @@ jobs:
 ## Publish to Docker hub
   publish:
     name:                                           Publishing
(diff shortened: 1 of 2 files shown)


HCastano:
>  Please raise an issue to inspect your dockerfiles, it's maybe something over-optimised is going on there.

Can you elaborate a bit on this?

HCastano (approved):
New images have been published on Docker Hub, so we're good here. Thanks!

PR #770, opened 2021-02-26
Bump Substrate and Deps

adoerr:
Bump Substrate in order to get [Fix transactions not being propagated to authorities](https://github.com/paritytech/substrate/commit/ad626b00990e57f7d23dee212d3b78e758ae88c8) fixed.

### DIFF
--- a/relays/ethereum/src/main.rs
+++ b/relays/ethereum/src/main.rs
@@ -243,6 +243,8 @@ fn ethereum_deploy_contract_params(matches: &clap::ArgMatches) -> Result<Ethereu
 	let eth_contract_code = parse_hex_argument(matches, "eth-contract-code")?.unwrap_or_else(|| {
 		hex::decode(include_str!("../res/substrate-bridge-bytecode.hex")).expect("code is hardcoded, thus valid; qed")
 	});
+
+	#[allow(clippy::manual_map)]
 	let sub_initial_authorities_set_id = match matches.value_of("sub-authorities-set-id") {
 		Some(sub_initial_authorities_set_id) => Some(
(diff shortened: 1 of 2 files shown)


adoerr:
Had to add a couple of `#[allow(clippy::manual_map)]`.

This check seems to be quite new and still has issues. See [manual_map recommends noncompilable code involving moved values](lhttps://github.com/rust-lang/rust-clippy/issues/6797).

adoerr:
WTF, was all green, requested the reviews and then it turned read

adoerr:
PR has been tested with a local test net and looked good, AFAICT.

HCastano:
I tested the builds locally and they're fine. 

There's probably an issue with the CI runners, @TriplEight, would you be able to take a look at this on Monday?

relays/ethereum/src/main.rs
```diff
@@ -243,6 +243,8 @@ fn ethereum_deploy_contract_params(matches: &clap::ArgMatches) -> Result<Ethereu
 	let eth_contract_code = parse_hex_argument(matches, "eth-contract-code")?.unwrap_or_else(|| {
 		hex::decode(include_str!("../res/substrate-bridge-bytecode.hex")).expect("code is hardcoded, thus valid; qed")
 	});
+
+	#[allow(clippy::manual_map)]
```
HCastano, line 247: Can you open an issue so that we can address this once the lint is working correctly?

HCastano (approved):

relays/ethereum/src/main.rs
adoerr: #773

PR #771, opened 2021-02-26
Clean Up Substrate Bridge Pallet

HCastano:
Since #761 is coming along, I figured that now was a good time to start getting rid of
bridge code that wouldn't be used anymore. In particular, I'm getting rid of all the code
related to importing and finalizing headers in favour of using the scheme introduced in
#628.

Eventually I would like to have `pallet-substrate-bridge` and `pallet-finality-verifier`
be a single pallet. I think it'll take a couple of steps to get there though:

1. Remove old, unused code
2. Migrate `pallet-substrate-bridge` to use FRAME v2 macros
3. Merge pallets

This PR does (1), while (2) and (3) will be done in follow up PRs in order to make
reviews easier.

If this PR is approved before #761 we should wait to merge it, since at the moment it'll
insta-break our testnet.


modules/substrate/src/fork_tests.rs
```diff
@@ -1,515 +0,0 @@
-// Copyright 2020 Parity Technologies (UK) Ltd.
```
HCastano, line 1: I'm sad to see these go, I thought they were pretty great

tomusdrw:
I'd wait with merging this until we get Rococo<>Westend bridge out.

HCastano:
> I'd wait with merging this until we get Rococo<>Westend bridge out.

That means we also shouldn't merge #761 since that removes support for the old header sync.

Maybe we should think about having an integration branch of Rococo<>Westend, but continue development on `master`?

tomusdrw:
> That means we also shouldn't merge #761 since that removes support for the old header sync.

Rococo<>Westend should not use the old header sync, but the new one, so I think #761 is definitely something we have to merge before we deploy.

> Maybe we should think about having an integration branch of Rococo<>Westend, but continue development on master?

How would that work? Frankly, I don't see any reason to do any other kind of development before Rococo<>Westend is out. Like the goal is to have Rococo<>Westend be a test deployment before Kusama<>Polkadot, so it has to include all the features we want there.

svyatonik:
> How would that work? Frankly, I don't see any reason to do any other kind of development before Rococo<>Westend is out. Like the goal is to have Rococo<>Westend be a test deployment before Kusama<>Polkadot, so it has to include all the features we want there.

If that's the goal (R+W bridge to be exactly what we're going to deploy on K+P), then we should definitely clean up and merge two pallets. We were going to stabilize storage before deploying the bridge, right? Not doing this cleanup + merge will lead to redundant + obsolete (with the new approach) entries in the storage.

HCastano:
Closing in favour of #783.

PR #772, opened 2021-02-26
Disable cache in CI builds.

tomusdrw:
Builds are randomly running out of disk space, and the cache seems to be around 2.5G. I'd rather have predictable, but slower builds than fast and randomly failing, so trying out this.

Closes: https://github.com/paritytech/parity-bridges-common/issues/773

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -42,8 +42,6 @@ jobs:
         run:                       rustup toolchain add $NIGHTLY
       - name:                      Add WASM Utilities
         run:                       rustup target add wasm32-unknown-unknown --toolchain $NIGHTLY
-      - name:                      Rust Cache
-        uses:                      Swatinem/rust-cache@v1.2.0
       - name:                      Checking rust-${{ matrix.toolchain }}
         uses:                      actions-rs/cargo@master
         with:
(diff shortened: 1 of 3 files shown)


HCastano:
Do you know what the difference between build times with and without the caching is?

For PRs relatively quick feedback is pretty important, as opposed to building Docker images for example. If we're running out of space we should consider using beefier self-hosted runners and periodically cleaning them up.

HCastano:
Hmm, I wonder if this fixes the CI failures in #770 🤔

tomusdrw:
> Do you know what the difference between build times with and without the caching is?

Yes, it's 5 minutes in the most optimistic case where we compile the same thing twice (no changes):
https://github.com/paritytech/parity-bridges-common/pull/740#issuecomment-781008727

> For PRs relatively quick feedback is pretty important,

Agree, but right now you get random, unpredictable feedback which is even worse than slow and stable. I'd rather wait a little bit more, but receive reasonable feedback than have to inspect & restart the job every 40 minutes.

Our CI team plans to re-write to gitlab, but honestly I think we can probably just do a bunch of smart things to optimize the time other than relying on blind caching (which pretty much never works). Some examples that might be worth testing:
- Disable unused features (we really can get away without RocksDB compilation)
- Use pre-compiled RocksDB (it takes significant amount of build time and it's simply useless), we could make RocksDB part of the docker image (or just disable it).
- Split building runtime WASM blob - we can build the node with native runtime and build the WASM blob as a separate job.
- Split relayer & node compilation (instead of using `cargo --all`) - they don't share that much.

adoerr:
Resolves #773

adoerr (approved):
PR is named in a somewhat deceiving way ;-) Does a bit more than it announces.

LGTM

PR #774, opened 2021-02-28
Fix finality_verifier test

adoerr:
Somehow, we missed one.


adoerr:
Fix with #772 already

PR #779, opened 2021-03-01
Overhaul releasing process

TriplEight:
-> Due to the long image builds we have to use the self-hosted runner.
-> A self-hosted runner can't `buildah`
Have to update to the newer version of the deprecated official docker action. 
This is, in fact, not bad. Now the action uses BuildX.
-> BuildX has an ugly [bug](https://github.com/docker/buildx/issues/101) 

Overhaul of publishing
- remove multistage build as it is a CI recursion
- upgrade to `ubuntu:20.04` base
- CI: move to use `buildah` actions
- chore

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -1,11 +1,7 @@
 name:                              Publish images to Docker hub
 
 on:
-  pull_request:
   push:
-    branches:
-      - master
-      - 3x8_use_buildah
     tags:
       - v*
     paths-ignore:
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -1,24 +1,20 @@
-name:                                               Publish images to Docker hub
+name:                              Publish images to Docker hub
 
 on:
-  pull_request:
   push:
-    branches:
-      - master
-      - 3x8_use_buildah
     tags:
       - v*
     paths-ignore:
       - '**/README.md'
       - diagrams/*
       - docs/*
-  schedule:                                         # Nightly build
-    - cron:                                         '0 1 * * *'
+  schedule:                        # Nightly build
+    - cron:                        '0 1 * * *'
 
 jobs:
-## Publish to Docker hub
+                                   ## Publish to Docker hub
   publish:
-    name:                                           Publishing
+    name:                          Publishing
     strategy:
       matrix:
         project:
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -59,8 +75,6 @@ USER user
 
 WORKDIR /home/user
 
-ARG PROJECT=ethereum-poa-relay
```
tomusdrw, line 62: This breaks shared base cache between different `PROJECTs`.

TriplEight: even if I put it on [L50](https://github.com/paritytech/parity-bridges-common/pull/779/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R50)?

tomusdrw: Yeah, the point is that commands between `50-62` do not depend on `PROJECT` so they can be re-used. It's not much (like only `libssl-dev/curl` installation that might take some time), but it's always something (when you build locally).

Maybe just move all the `ARGs` and `LABEL` to this line instead? I know it's nice to have all parameters at the top, but it does affect docker caching ability.

TriplEight:
Got these images building sending them to the public runners with `buildah` action.

Caching is not helpful if it gets removed every day, invalidated every day with the nightly rust update, executes every time on a fresh VM, etc. Also, there was a lot of duplicated processes and esp when launched 4 times in parallel it doesn't make anything faster.

Still, they take a surprisingly long time to build (the last image gets built in 46min), normally it's 3.5 min for each without any cache.

I'd suggest an optimization - produce the binaries separately (it's faster and probs you'll want the binaries somewhere anyway) in CI and on the images stage just put them in the containers.

TriplEight:
Oh btw I've updated the images and now they run off `ubuntu:20.04` please check if they are working properly.

TriplEight:
The next step is I'll steal your CI dockerfile to build it nightly with the other images in #paritytech/scripts.

TriplEight:
Regarding
> Besides, --package and other flags are not yet working in cargo-chef.
I lied, just found that they've [addressed](https://github.com/LukeMathWalker/cargo-chef/issues/21#event-4385117897) my issue.

TriplEight:
please check if these images work correctly before merging:
`paritytech/millau-bridge-node:ubuntu`
`paritytech/ethereum-poa-relay:ubuntu`
`paritytech/rialto-bridge-node:ubuntu`
`paritytech/substrate-relay:ubuntu`
`paritytech/bridge-dependencies:latest`

tomusdrw:
> please check if these images work correctly before merging:

We will learn that on our testnet deployment :)

tomusdrw (approved):

deployments/BridgeDeps.Dockerfile
```diff
@@ -2,9 +2,10 @@
 #
 # This image is meant to be used as a building block when building images for
 # the various components in the bridge repo, such as nodes and relayers.
-FROM ubuntu:xenial
+FROM ubuntu:20.04
 
 ENV LAST_DEPS_UPDATE 2020-12-21
```
HCastano, line 7: We should bump this and the one on line 21. 

The idea behind these lines is that we can use them to break the cache and update dependencies if needed.

Dockerfile
```diff
@@ -8,51 +8,33 @@
 #
 # See the `deployments/README.md` for all the available `PROJECT` values.
 
-# This first stage prepares our dependencies to be built by `cargo-chef`.
-FROM rust as planner
+FROM docker.io/paritytech/bridge-dependencies as builder
```
HCastano, line 11: Is it not good enough to just do `paritytech/bridge-dependencies`?

Dockerfile
```diff
@@ -8,51 +8,33 @@
...
-COPY --from=cacher $CARGO_HOME $CARGO_HOME
 
 ARG PROJECT=ethereum-poa-relay
 RUN cargo build --release --verbose -p ${PROJECT}
 RUN strip ./target/release/${PROJECT}
 
 # In this final stage we copy over the final binary and do some checks
 # to make sure that everything looks good.
-FROM ubuntu:xenial as runtime
+FROM docker.io/ubuntu:20.04 as runtime
```
HCastano, line 22: Same question here, why do we need to specify `docker.io`?

Dockerfile
```diff
@@ -8,51 +8,33 @@
...
 	apt-get update && \
-	apt-get install -y libssl-dev curl
-
-RUN groupadd -g 1000 user \
-  && useradd -u 1000 -g user -s /bin/sh -m user
+	apt-get install -y --no-install-recommends \
+		libssl-dev curl && \
+	groupadd -g 1000 user && \
+	useradd -u 1000 -g user -s /bin/sh -m user && \
+# apt clean up
```
HCastano, line 34:
```suggestion

	# apt clean up
```

.github/workflows/publish-docker.yml
```diff
@@ -31,46 +31,57 @@ jobs:
...
           fi
-      - name:                                       Build and push ${{ matrix.project }}
-        uses:                                       docker/build-push-action@v1
+          TAGS="${VERSION} sha-${GITHUB_SHA::8} latest"
+          echo ::set-output name=TAGS::${VERSION}
+          echo ::set-output name=TAGS::${TAGS}
+          echo ::set-output name=DATE::$(date +%d-%m-%Y)
+
+      - name:                                       Build image for ${{ matrix.project }}
+        uses:                                       redhat-actions/buildah-build@v2.2
```
HCastano, line 63: Why did you choose `buildah` over `build-push-action`?

HCastano (approved):
Looks good!

HCastano:
Pinging @TriplEight since I've got some questions here

HCastano:
I fixed some of my nitpicks since I wanted this to get used on tonight's testnet deployment.

Dockerfile
TriplEight: it's better to have the full names when building with something that's not docker, we use buildah.

PR #781, opened 2021-03-01
Derive bridged accounts

### LINKED ISSUE #607: CLI to generate derived addresses
Add a helper CLI that can be fed with source chain address and generates the derived address of the target chain (i.e. the address that will be used during dispatch of messages sent over the bridge, see #518).

### LINKED ISSUE #769: Relayer CLI should accept addresses with any SS58 prefix
Currently we rely on default parsing method of `AccountId` (which comes from `serde::Deserialize` I believe) and this means that all addresses have to be SS58 encode with prefix=42 (due to decoding relying on some global state).

We should have a custom parser that is able to decode any SS58 encoding and detect the prefix.

tomusdrw:
Closes #607 
Closes #769


relays/substrate/src/cli.rs
```diff
@@ -63,6 +64,8 @@ pub enum Command {
 	EncodeMessagePayload(EncodeMessagePayload),
 	/// Estimate Delivery and Dispatch Fee required for message submission to message lane.
 	EstimateFee(EstimateFee),
+	/// Given Source chain AccountId, derive a corresponding AccountId on the target chain.
```
HCastano, line 67:
```suggestion
	/// Given a source chain `AccountId`, derive the corresponding `AccountId` for the target chain.
```

relays/substrate/src/cli.rs
```diff
@@ -268,6 +271,20 @@ pub enum EstimateFee {
 	},
 }
 
+/// Given Source chain AccountId, derive a corresponding AccountId on the target chain.
```
HCastano, line 274: Same suggestion here

HCastano (approved):

relays/substrate/src/cli.rs
tomusdrw: 😅

PR #782, opened 2021-03-01
Stop CI from failing with new security advisories

HCastano:
Most of the time that `cargo-deny` complains about a new security advisory there's
nothing we can do because the broken crate is way too deep in our dependency tree.
This PR continues to run the advisory check, but makes sure that the CI doesn't fail
when it does encounter a new security advisory.

### DIFF
--- a/.github/workflows/deny.yml
+++ b/.github/workflows/deny.yml
@@ -16,6 +16,13 @@ on:
 jobs:
   cargo-deny:
     runs-on:            ubuntu-latest
+    strategy:
+      matrix:
+        checks:
+          - advisories
+          - bans licenses sources
+    # Prevent sudden announcement of a new advisory from failing CI:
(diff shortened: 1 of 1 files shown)


adoerr (approved):
LGTM

HCastano:
So I can make the "advisory" step here not required, but have the others (bans, licenses, sources) required.

We'd need to live with the little red X beside the step for a bit depending on the fix (a valid fix being ignoring the advisory in the `deny.yml` config) - but at least it wouldn't block the CI pipeline randomly.

What do you guys think?

adoerr:
I think, the `Cargo deny` step failing is actually useful. This is a kind of notification to update `deny.yml`. 

And not having CI failing because of this is a big plus. IMO CI is currently more of an impediment than anything else.

Issue #760, opened 2021-02-24
Assume Scheduled Change Delay is Zero


HCastano:
When checking headers for scheduled authority set changes in the import pipeline of the
bridge pallet we have some small checks which use the `delay` field of [GRANDPA's
`ScheduledChange` struct](https://github.com/paritytech/substrate/blob/master/primitives/finality-grandpa/src/lib.rs#L82-L87).

I recently learned that it is safe to assume that as long as GRANDPA has not stalled the
`delay` will [always be zero](https://github.com/paritytech/substrate/blob/master/frame/grandpa/src/lib.rs#L611-L615).

We can use this fact to check for a non-zero delay in any header which schedules changes
and reject such headers. We want to reject these headers since they indicate that either:
a relayer is submitting invalid headers, or a forced change has occured. The former
should not be accepted, and the latter can only be remedied by the chain's governance
anyways (since the pallet wouldn't be able to verify finality for any future headers).

svyatonik:
I also suggest to do two things:
1) optionally - remove all delay traces from our pallet code - like we don't need scheduled changes anymore. This will simplify code + we'll get rid of extra storage entries;
2) also reject headers with `ConsensusLog::ForcedChange` digest - we can't follow the finalized chain after such digest anyway + we don't need all headers.

tomusdrw:
@HCastano is this already done?

HCastano:
@tomusdrw yep, this should've been closed with #783.

PR #783, opened 2021-03-02
Unify Substrate Bridge Pallet with Finality Verifier

HCastano:
This is an alternative approach to what was started in #771. In a sense it basically just
jumps to step 3 from that list.

What I've done is taken all the relevant parts from the existing bridge pallet and
re-wrote them in FRAME v2 within the finality verifier pallet. For the most part I've
avoided changing logic around on both sides - with the exception that I did make the
assumptions outlined in #760 since that simplifies the pallet a bit.

The reason why this approach is being tried is to make the review process easier and
faster.

Some things I still need to do in this PR:

- [x] Move Runtime APIs
- [x] Fix outstanding TODOs I've left in the code

Notice that no code from the bridge pallet was removed in this PR. I plan to remove
unused code after this is merged in order to avoid making this PR too noisy.

### DIFF
--- a/modules/finality-verifier/src/lib.rs
+++ b/modules/finality-verifier/src/lib.rs
@@ -778,7 +778,7 @@ mod tests {
 	}
 
 	#[test]
-	fn disallows_imports_once_limit_is_hit_in_single_block() {
+	fn rate_limiter_disallows_imports_once_limit_is_hit_in_single_block() {
 		run_test(|| {
 			initialize_substrate_bridge();
 			assert_ok!(submit_finality_proof(1, 2));
@@ -788,7 +788,7 @@ mod tests {
 	}
 
 	#[test]
-	fn invalid_requests_do_not_count_towards_request_count() {
+	fn rate_limiter_invalid_requests_do_not_count_towards_request_count() {
 		run_test(|| {
 			let submit_invalid_request = || {
 				let child = test_header(1);
@@ -820,7 +820,7 @@ mod tests {
 	}
 
 	#[test]
-	fn allows_request_after_new_block_has_started() {
+	fn rate_limiter_allows_request_after_new_block_has_started() {
 		run_test(|| {
 			initialize_substrate_bridge();
 			assert_ok!(submit_finality_proof(1, 2));
@@ -832,7 +832,7 @@ mod tests {
 	}
 
 	#[test]
-	fn disallows_imports_once_limit_is_hit_across_different_blocks() {
+	fn rate_limiter_disallows_imports_once_limit_is_hit_across_different_blocks() {
 		run_test(|| {
 			initialize_substrate_bridge();
 			assert_ok!(submit_finality_proof(1, 2));
@@ -845,7 +845,7 @@ mod tests {
 	}
 
 	#[test]
-	fn allows_max_requests_after_long_time_with_no_activity() {
+	fn rate_limiter_allows_max_requests_after_long_time_with_no_activity() {
 		run_test(|| {
 			initialize_substrate_bridge();
 			assert_ok!(submit_finality_proof(1, 2));
--- a/modules/finality-verifier/src/lib.rs
+++ b/modules/finality-verifier/src/lib.rs
@@ -348,10 +348,11 @@ pub mod pallet {
 	/// means it's of great importance that this function *not* called with any headers whose
 	/// finality has not been checked, otherwise you risk bricking your bridge.
 	pub(crate) fn import_header<T: Config>(header: BridgedHeader<T>) -> Result<(), sp_runtime::DispatchError> {
-		let best_finalized = <ImportedHeaders<T>>::get(<BestFinalized<T>>::get()).expect(
-			"In order to reach this point the bridge must have been initialized. Therefore
-			`ImportedHeaders` must contain an entry for `BestFinalized`.",
-		);
+		const BEST_FINALIZED_PROOF: &str = "In order to reach this point the bridge must have been
+			initialized. Afterwards, every time `BestFinalized` is updated `ImportedHeaders` is
+			also updated. Therefore `ImportedHeaders` must contain an entry for `BestFinalized`.";
+
+		let best_finalized = <ImportedHeaders<T>>::get(<BestFinalized<T>>::get()).expect(BEST_FINALIZED_PROOF);
 
 		// We do a quick check here to ensure that our header chain is making progress and isn't
 		// "travelling back in time" (which would be indicative of something bad, e.g a hard-fork).
--- a/modules/finality-verifier/src/lib.rs
+++ b/modules/finality-verifier/src/lib.rs
@@ -37,10 +37,10 @@ use bp_runtime::{BlockNumberOf, Chain, HashOf, HasherOf, HeaderOf};
 use codec::{Decode, Encode};
 use finality_grandpa::voter_set::VoterSet;
 use frame_support::{dispatch::DispatchError, ensure};
-use frame_system::ensure_signed;
+use frame_system::{ensure_signed, RawOrigin};
 #[cfg(feature = "std")]
 use serde::{Deserialize, Serialize};
-use sp_runtime::traits::{Header as HeaderT, Zero};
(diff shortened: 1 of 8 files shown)


modules/finality-verifier/src/lib.rs
```diff
@@ -175,32 +322,207 @@ pub mod pallet {
...
+				Default::default(),
+				Default::default(),
+				Default::default(),
+				Default::default(),
+			)
+		})
+	}
+
+	/// Check if a particular header is known to the bridge pallet.
+	pub fn is_known_header(hash: BridgedBlockHash<T>) -> bool {
```
HCastano, line 444: I don't think this is super useful anymore for syncing since it looks like relayers only care about the best finalized headers. Do you think this is helpful to keep @svyatonik?

HCastano:
As part of the review process I would take a look at the tests in the Substrate pallet
(including those in the `verifier.rs` module) and see if there are any tests from there
which we should also have in the merged finality-verifier pallet.

primitives/millau/src/lib.rs
```diff
@@ -281,6 +286,13 @@ sp_api::decl_runtime_apis! {
 		fn is_finalized_block(hash: Hash) -> bool;
 	}
 
+	pub trait MillauFinalityApi {
```
tomusdrw, line 289: Missing docs?

modules/finality-verifier/src/lib.rs
```diff
@@ -142,13 +154,86 @@ pub mod pallet {
 				<Error<T>>::InvalidAncestryProof
 			);
 
-			let _ = T::HeaderChain::append_header(finality_target)?;
+			let _ = T::HeaderChain::append_header(finality_target.clone())?;
```
tomusdrw, line 157: Why do we still need the `HeaderChain`?

modules/finality-verifier/src/lib.rs
```diff
@@ -142,13 +154,86 @@ pub mod pallet {
...
 			frame_support::debug::info!("Succesfully imported finalized header with hash {:?}!", hash);
```
tomusdrw, line 158: This is only partially correct now, since `import_header` might fail. We may also end up with header being appended to `HeaderChain`, but not imported locally.

modules/finality-verifier/src/lib.rs
```diff
@@ -175,32 +322,207 @@ pub mod pallet {
...
+	/// updating the current authority set accordingly.
+	///
+	/// Note: This function assumes that the given header has already been proven to be valid and
+	/// finalized. Using this assumption it will write them to storage with minimal checks. That
+	/// means it's of great importance that this function *not* called with any headers whose
+	/// finality has not been checked, otherwise you risk bricking your bridge.
+	pub(crate) fn import_header<T: Config>(header: BridgedHeader<T>) -> Result<(), sp_runtime::DispatchError> {
+		let best_finalized = <ImportedHeaders<T>>::get(<BestFinalized<T>>::get()).expect(
+			"In order to reach this point the bridge must have been initialized. Therefore
+			`ImportedHeaders` must contain an entry for `BestFinalized`.",
```
tomusdrw, line 352: This is only one part of the proof. When we initialize bridge indeed we set `BestFinalized` and also `ImportedHeaders`, but we should include in the proof a statement that every time `BestFinalized` is being set `ImportedHeaders` is updated as well.

modules/finality-verifier/src/lib.rs
```diff
@@ -175,32 +322,207 @@ pub mod pallet {
...
+			<Error<T>>::UnsupportedScheduledChange
+		);
+
+		if let Some(change) = super::find_scheduled_change(&header) {
+			// GRANDPA only includes a `delay` for forced changes, so this isn't valid.
+			ensure!(change.delay == Zero::zero(), <Error<T>>::UnsupportedScheduledChange);
+
+			let next_authorities = bp_header_chain::AuthoritySet {
+				authorities: change.next_authorities,
+				set_id: <CurrentAuthoritySet<T>>::get().set_id + 1,
```
tomusdrw, line 371: This should be coming from `change` actually, maybe worth adding a `TODO` so that we don't forget to update after @andresilva adds set id to the digest.

modules/finality-verifier/src/lib.rs
```diff
@@ -175,32 +322,207 @@ pub mod pallet {
...
+	/// finality has not been checked, otherwise you risk bricking your bridge.
+	pub(crate) fn import_header<T: Config>(header: BridgedHeader<T>) -> Result<(), sp_runtime::DispatchError> {
+		let best_finalized = <ImportedHeaders<T>>::get(<BestFinalized<T>>::get()).expect(
+			"In order to reach this point the bridge must have been initialized. Therefore
+			`ImportedHeaders` must contain an entry for `BestFinalized`.",
+		);
+
+		// We do a quick check here to ensure that our header chain is making progress and isn't
+		// "travelling back in time" (which would be indicative of something bad, e.g a hard-fork).
+		ensure!(best_finalized.number() < header.number(), <Error<T>>::ConflictingFork);
```
tomusdrw, line 357: I wonder if it makes more sense to verify these simple requirements before verifying the justification (which might be quite heavy).

modules/finality-verifier/src/lib.rs
```diff
@@ -175,32 +322,207 @@ pub mod pallet {
...
+				authorities: change.next_authorities,
+				set_id: <CurrentAuthoritySet<T>>::get().set_id + 1,
+			};
+
+			// Since our header schedules a change and we know the delay is 0, it must also enact
+			// the change.
+			<CurrentAuthoritySet<T>>::put(next_authorities);
+		};
+
+		<BestFinalized<T>>::put(header.hash());
```
tomusdrw, line 379: Nit: Hash could be cached to avoid computing it multiple times.

modules/finality-verifier/src/lib.rs
```diff
@@ -142,13 +154,86 @@ pub mod pallet {
...
+			frame_support::debug::warn!("Stopping pallet operations.");
+
+			Ok(().into())
+		}
+
+		/// Resume all pallet operations. May be called even if pallet is halted.
+		///
+		/// May only be called either by root, or by `ModuleOwner`.
+		#[pallet::weight((T::DbWeight::get().reads_writes(1, 1), DispatchClass::Operational))]
+		pub fn resume_operations(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
```
tomusdrw, line 230: Do you think it make sense to merge the two into `set_operational(origin, is_operational: bool)` for code reduction?

tomusdrw (approved):
lgtm modulo tiny grumbles

modules/finality-verifier/src/lib.rs
HCastano: We don't. I wanted to avoid removing this until later PRs

modules/finality-verifier/src/lib.rs
HCastano: Yep, I know. I can open an issue for it

modules/finality-verifier/src/lib.rs
HCastano: Yeah maybe. I wanted to keep as much of the "new" code in `import_header` as opposed to adding it into `submit_finality_proof` for this PR, but maybe for this small stuff it's not a big deal

modules/finality-verifier/src/lib.rs
HCastano: Yeah, it makes sense. However, we'd also have to change the code in the message lane pallet and I don't want to touch that code as part of this PR.

Can do in a follow up.

modules/finality-verifier/src/lib.rs
HCastano: Opened #788.

modules/finality-verifier/src/lib.rs
tomusdrw: Fair enough. I was actually thinking that you'll start a completely fresh pallet instead of re-using `finality-verifier`, but so be it :)

adoerr (approved):
Personally, I would always prefer a rewrite, given the chance to do so.

LGTM

PR #786, opened 2021-03-02
Update Substrate Node build script

HCastano:
It looks like the Node Template switched from using `vergen` to
`substrate-build-script-utils` almost a year ago. I guess we missed the memo, haha.


tomusdrw (approved):
Nice catch!

PR #787, opened 2021-03-03
Migrate to jsonrpsee v2

### LINKED ISSUE #661: Consider migration to jsonrpsee v2
Right now we're using our [own fork](https://github.com/svyatonik/jsonrpsee/commits/shared-client-in-rpc-api) of jsonrpsee. Looks like most (or even all) of issues we had, are fixed in [v2](https://github.com/paritytech/jsonrpsee/pull/180). Probably we may drop or fork in favor of 'official' v2?

### LINKED ISSUE #741: Relay is losing connection to Substrate node when transaction is submitted
There's some connection-related issue with tonight deployments. Following pattern repeats in millau-to-rialto messages relay logs:
```
[1;38;5;8m2021-02-18 09:00:22 +0000[0m [38;5;14mDEBUG[0m [38;5;8mbridge[0m Received nonces from Rialto::ReceivingConfirmationsDelivery: SourceClientNonces { new_nonces: 978..=977, confirmed_nonce: None }
[1;38;5;8m2021-02-18 09:00:22 +0000[0m [38;5;14mDEBUG[0m [38;5;8mbridge[0m Asking Rialto::ReceivingConfirmationsDelivery to prove nonces in range 866..=977 at block HeaderId(2388, 0xf23b8a49a96c5f05629491fa2e571cab99140e05bb8419a3b59f12e96b567f4b)
[1;38;5;8m2021-02-18 09:00:22 +0000[0m [38;5;14mDEBUG[0m [38;5;8mbridge[0m Received best nonces from Millau::ReceivingConfirmationsDelivery: TargetClientNonces { latest_nonce: 865, nonces_data: () }
...
[1;38;5;8m2021-02-18 09:00:22 +0000[0m [38;5;14mDEBUG[0m [38;5;8mbridge[0m Received proof for nonces in range 866..=977 from Rialto::ReceivingConfirmationsDelivery
[1;38;5;8m2021-02-18 09:00:22 +0000[0m [38;5;14mDEBUG[0m [38;5;8mbridge[0m Going to submit proof of messages in range 866..=977 to Millau::ReceivingConfirmationsDelivery node
[1;38;5;8m2021-02-18 09:00:22 +0000[0m [1;38;5;9mERROR[0m [38;5;8mbridge[0m Error submitting proof Millau::ReceivingConfirmationsDelivery: Request(TransportError(Custom { kind: Other, error: "background task closed" })). Going to restart
[1;38;5;8m2021-02-18 09:00:32 +0000[0m [38;5;14mDEBUG[0m [38;5;8mbridge[0m Restarting relay loop
...
```

So it seems that everything works fine - we are receiving data from both Rialto and Millau nodes. Then, suddenly, when proof is submitted to the Millau node, something breaks in jsonrpsee ("background task closed") and we restart the loop. After restart, everything repeats again. There's nothing strange in `millau-node-bob` logs when this happens. We have no `jsonrpsee` + `soketto` logs enabled for relayer, so there's nothing to work with :/

The "solution" is:
1) #661 
2) enable transport-related logs for message relays.

This must be done before P+K release

UPD: just trying to log how often this issue happens
1st reproduce: 18.02
2nd reproduce: 04.03

svyatonik:
closes #661
closes #741 (presumably)

Thanks to @niklasad1 for making this possible. I'm opening this PR early, because: otherwise this branch will be lost && I really think that we should do this upgrade before P+K (I don't want to maintain separate branch anymore + hope it'll resolve #741).

Since there are too many pending breaking changes in relay + Substrate pallet, I believe we shall merge this PR only after Millau<->Rialto bridge will be running smoothly for at least 1 night.

Known issues:
- [x] ~Rialto-PoA <-> Rialto is not working because of [this panic](https://gist.github.com/svyatonik/2ac86f41fb5c7f3cb79dab6fde3fb801). The panic is related to the fact that we're connecting to eth nodes using http client and to the substrate node using ws client (related: https://github.com/paritytech/jsonrpsee/issues/200, https://github.com/paritytech/jsonrpsee/pull/204, https://github.com/paritytech/jsonrpsee/pull/208, https://github.com/paritytech/jsonrpsee/issues/175#issuecomment-778600467).~ Eth-client is now also using ws

Some more manual tests are required:
- [x] That transport errors are propagated
- [x] That RPC clients are able to reconnect after connection error


niklasad1:
Ok, I never updated the `WsClient`.
Should be quick and there are logs in `jsonrpsee v2`  that you could use if you run it #741 to debug it.

svyatonik:
We don't have any logs, unfortunately. Also we don't have any jsonrpsee (and related libs) logs enabled right now - I'll re-add this again in near future. Worth to say that we're using jsonrpsee for a long time already and only seen this issue once, so don't expect any logs soon. To me, it looks like there's an issue with some particular payload or payload size (possibly in how fragmentation is implemented in soketto?). But I haven't been looking so deep yet

niklasad1:
> We don't have any logs, unfortunately. Also we don't have any jsonrpsee (and related libs) logs enabled right now - I'll re-add this again in near future. Worth to say that we're using jsonrpsee for a long time already and only seen this issue once, so don't expect any logs soon. To me, it looks like there's an issue with some particular payload or payload size (possibly in how fragmentation is implemented in soketto?). But I haven't been looking so deep yet

Ok, the default max payload size is currently in 10MB `jsonrpsee` FWIW. I recommend to run wireshark on it then to investigate it further but this gives us motivation to fix https://github.com/paritytech/jsonrpsee/issues/188 :)

svyatonik:
UPD: we've had #741 again this morning. So since we now have a possible fix for this issue (see https://github.com/paritytech/parity-bridges-common/issues/741#issuecomment-789590396), I'm going to complete+deploy this PR asap.

relays/ethereum-client/Cargo.toml
```diff
@@ -10,7 +10,9 @@ bp-eth-poa = { path = "../../primitives/ethereum-poa" }
 codec = { package = "parity-scale-codec", version = "2.0.0" }
 headers-relay = { path = "../headers-relay" }
 hex-literal = "0.3"
-jsonrpsee = { git = "https://github.com/svyatonik/jsonrpsee.git", branch = "shared-client-in-rpc-api", default-features = false, features = ["http"] }
+jsonrpsee-ws-client = { git = "https://github.com/paritytech/jsonrpsee" }
```
niklasad1, line 13: published alpha releases on crates.io that you can use https://crates.io/crates/jsonrpsee-ws-client
```suggestion
jsonrpsee-ws-client = "0.2.0-alpha"
```

relays/substrate-client/src/client.rs
```diff
@@ -36,6 +34,7 @@ use sp_version::RuntimeVersion;
 use std::ops::RangeInclusive;
 
 const SUB_API_GRANDPA_AUTHORITIES: &str = "GrandpaApi_grandpa_authorities";
+const MAX_SUBSCRIPTION_CAPACITY: usize = 4096;
```
tomusdrw, line 37: What does this control exactly? Is it the maximal number of open subscriptions or pending notifications in the sbuscription?

relays/substrate/src/headers_maintain.rs
```diff
@@ -153,6 +159,11 @@ where
 		})
 		.await;
 
+		// if justifications subscription has been dropped, resubscribe
+		if justifications.stream.is_none() {
```
tomusdrw, line 163: Is there a place in code that should be setting `justifications.stream = None`? I couldn't find anything, which means that the stream can be `None` only initially, no?

tomusdrw (approved):
lgtm! Asked for few clarifications though.

relays/substrate-client/src/client.rs
niklasad1: Pending notifications, https://github.com/paritytech/jsonrpsee/blob/master/types/src/client.rs#L66-#L70

relays/substrate/src/headers_maintain.rs
niklasad1: only returns None when the stream has been terminated.

svyatonik: > Is there a place in code that should be setting justifications.stream = None

It may have been set to `None` if first subscription has failed. But it was not enough ofc - I've now added reconnect when subscription has been dropped after initiation.

Cargo.lock
```diff
@@ -1,5 +1,7 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
+version = 3
```
HCastano, line 3: Was this added manually, or did Cargo do it?

Also, I can't find any info on `v2` vs `v3`, can someone point me to the right links?

deployments/bridges/poa-rialto/entrypoints/poa-exchange-tx-generator-entrypoint.sh
```diff
@@ -13,7 +13,7 @@ RELAY_BINARY_PATH=${RELAY_BINARY_PATH:-./ethereum-poa-relay}
 # Ethereum node host
 ETH_HOST=${ETH_HOST:-poa-node-arthur}
 # Ethereum node port
```
HCastano, line 15:
```suggestion
# Ethereum node websocket port
```

HCastano (approved):
Looks good!

relays/substrate/src/headers_maintain.rs
niklasad1: @svyatonik sorry I forgot to mention that an annoyance is https://github.com/paritytech/jsonrpsee/issues/220

Basically if you unsubscribe to a subscription the client won't register the request and if the server responds with the request ID it will most likely fail, as `invalid requestID` (which I think `jsonrpc server pubsub` does). Fail means closing the connection and it has to restarted :(

The reason for this is that `jsonrpsee ws server` didn't include request ID in responses for subscription from the beginning we will fix it soon.

svyatonik: @niklasad1 Thanks! :) So if I understand correctly, until that fix we'll be disconnecting from the server if there are too many unread notifications? That's ok - iiuc this disconnect will be reported as `Error::Internal(_)` when we'll try to use connection next time && we're now able to process it properly.

niklasad1: yeah almost,

1. The subscription capacity gets full because of too many unread notifications
2. If you then drop the entire subscription as I think you do
3. client send `unsubscribe_to_some_method, id: x`
4. server ACKs `unsubscribe_to_some_method` the request with `id: x` 
5. client doesn't know what `id: x`  is and closes the connection because of `Error::InvalidRequestId`.

A super hacky-workaround is to keep the all subscriptions but this means that server will send out needless messages that are not processed or make the buffer huge for now, I don't think you will run into this but you might ^^

I can ping you when the fix is merged but I haven't tested against `jsonrpc ws server` FWIW.

svyatonik: Yes, please ping me :) I'll be able to test if it works with substrate (jsonrpc) as well

niklasad1: @svyatonik 

Published 0.2.0-alpha.2, it has some breaking changes renaming some fields.
Also I think it will fix the compilation warnings generated the proc macros.

svyatonik: Thank you - will try to check it soon :)

Cargo.lock
niklasad1: @HCastano it was probably my fault but it is https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#changed

It's already on `nightly` so I think I triggered that ^^

relays/substrate/src/headers_maintain.rs
svyatonik: @niklasad1 I've opened #828 - could confirm that there's no warnings + connection is not lost when we unsubscribe. Thank you! :)

PR #790, opened 2021-03-03
Adding message relayer scripts, reformating send message scripts

hbulgarini:
- Adding header relayer script from rialto to millau.
- Adding both directions message relayer scripts.
- Updating and creating new send message scripts.
- Updated port numbers and accounts.


README.md
```diff
@@ -91,17 +94,18 @@ the `relays` which are used to pass messages between chains.
...
 ## Running the Bridge
 
 To run the Bridge you need to be able to connect the bridge relay node to the RPC interface of nodes
 on each side of the bridge (source and target chain).
 
 There are 3 ways to run the bridge, described below:
- - building & running from source,
- - building or using Docker images for each individual component,
- - running a Docker Compose setup (recommended).
+
```
HCastano, line 105:
```suggestion
```

My personal preference, but no big deal

deployments/local-scripts/relay-messages-millau-to-rialto-entrypoint.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
HCastano, line 1: Can you remove the "entrypoint" from the file names? That's a Docker convention that's used, but not relevant to these shell scripts.

Also, a small description of what this script is for would be nice. For example, look at `relay-rialto-to-millau.sh`.

deployments/local-scripts/relay-messages-rialto-to-millau-entrypoint.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
HCastano, line 1: Same comments here

deployments/local-scripts/relay-messages-millau-to-rialto-entrypoint.sh
```diff
@@ -0,0 +1,13 @@
...
+set -xeu
+
+RUST_LOG=bridge=debug \
+./target/debug/substrate-relay relay-messages millau-to-rialto \
+	--lane 00000000 \
+	--millau-host localhost \
+	--millau-port 9945 \
+	--millau-signer //Eve \
```
HCastano, line 9: The other scripts that are part of `local-scripts` don't funds for `Eve` since they run as `--dev` nodes, so I don't think this will work. Can you either a) check if this works or b) use a different account

deployments/local-scripts/relay-messages-rialto-to-millau-entrypoint.sh
```diff
@@ -0,0 +1,13 @@
...
+set -xeu
+
+RUST_LOG=bridge=debug \
+./target/debug/substrate-relay relay-messages rialto-to-millau \
+	--lane 00000000 \
+	--rialto-host localhost \
+	--rialto-port 9944 \
+	--rialto-signer //Ferdie \
```
HCastano, line 9: Same comment about accounts here

scripts/send-message-from-millau-rialto.sh
```diff
@@ -11,9 +11,9 @@ case "$1" in
 		RUST_LOG=runtime=trace,substrate-relay=trace,bridge=trace \
 		./target/debug/substrate-relay send-message millau-to-rialto \
 			--millau-host localhost \
-			--millau-port 20044 \
-			--millau-signer //Dave \
-			--rialto-signer //Dave \
+			--millau-port 9944 \
```
HCastano, line 14: So this changes the use case a little bit, and we should make a note of that in the documentation.

The ports originally used here match our Docker deployment. That meant you could use this script for testing a message which this after spinning up a network using `docker-compose`.

This changes that, and it's fine as long as we also mention what the requirements for running a full message sync network are in the documentation. That's to say: you need nodes + header relays + message relays running _before_ this can even work.

We also do want to keep the docs related to Docker since it's pretty useful for some quick testing.

scripts/send-message-from-rialto-millau.sh
```diff
@@ -0,0 +1,35 @@
...
+		./target/debug/substrate-relay send-message rialto-to-millau \
+			--rialto-host localhost \
+			--rialto-port 9944 \
+			--millau-signer //Alice \
+			--rialto-signer //Bob \
+			--lane 00000000 \
+			--origin Target \
+			transfer \
+			--amount 100000000000000 \
+			--recipient 5DZvVvd1udr61vL7Xks17TFQ4fi9NiagYLaBobnbPCP14ewA \
```
HCastano, line 32: iirc this used to be an address you could see on the UI. Given the recent address encoding changes I think this will need to be changed in both `send-message` scripts

scripts/send-message-from-rialto-millau.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
HCastano, line 1: Another approach to take here is to expand the existing `send-message` script to take a network argument.

Usage would be something like: `./send-message.sh remark --to rialto`

README.md
```diff
@@ -125,6 +129,9 @@ both Substrate chains it must be run last.
 ./deployments/local-scripts/run-rialto-bridge-node.sh
 ./deployments/local-scripts/run-millau-bridge-node.sh
 ./deployments/local-scripts/relay-millau-to-rialto.sh
+./deployments/local-scripts/relay-rialto-to-millau.sh
+./deployments/local-scripts/relay-messages-millau-to-rialto-entrypoint.sh
+./deployments/local-scripts/relay-messages-millau-to-rialto-entrypoint.sh
```
HCastano, line 134: I'd prefer if you didn't add this here.  This section specifically talks about a simple one way _header_ relay sync. This adds extra complexity, and will also not match the example shown in the Docker section below.

It does look like Andreas added a section on sending messages further down in the README, so maybe this is something that could be added there?

adoerr (commented):
I'll leave this review to @HCastano

scripts/send-message-from-millau-rialto.sh
hbulgarini: i see the local scripts as an alternative version of docker compose in case the user does not want to use docker or whatever. We could say in the documentation that this is the alternative way of the docker compose but with same ( or similar connection details) as port numbers for example.

README.md
hbulgarini: well... from the UI point of view, the simple one way header relay sync is not enough since the transactions/remark  wouldn't work because it needs the message relayers. So based on this could we consider the relay messages as a part of the minimum set up to run the environment?

deployments/local-scripts/relay-messages-millau-to-rialto-entrypoint.sh
hbulgarini: Should i set Alice or Bob ?

scripts/send-message-from-millau-rialto.sh
tomusdrw: Let's use an environment variable in both scripts and default to either docker (my preference) or local setup.
Then in case you want to override you could simply
```
export MILLAU_PORT=9944
export RIALTO_PORT=9955
./local_scripts/...
```
or even create an `.env` file.

scripts/send-message-from-rialto-millau.sh
tomusdrw: Actually after #781 CLI works with addresses encode with any SS58 prefix, it will only display a warning in case it doesn't match the network.

deployments/local-scripts/relay-messages-millau-to-rialto-entrypoint.sh
HCastano: I would do `Bob` since `Alice` is used by the header sync scripts

README.md
HCastano: For running the UI yeah I agree the setup is not enough. However, the goal of the main README is not to get somebody started on building a UI on top of the bridge, it's to get them quickly running a barebones bridge. For anything more involved we have more detailed documentation elsewhere, such as the messaging docs or the deployment README.

scripts/send-message-from-millau-rialto.sh
HCastano: > Let's use an environment variable in both scripts and default to either docker (my preference) or local setup.
> Then in case you want to override you could simply
> 
> ```
> export MILLAU_PORT=9944
> export RIALTO_PORT=9955
> ./local_scripts/...
> ```
> 
> or even create an `.env` file.

Yeah, something like this could work. Docker is also my preference (if you haven't already noticed) since getting to the point where you can even send a message requires _at minimum_ six different processes running.

README.md
hbulgarini: ok, fair enough. I will include this in specific UI README then.

tomusdrw (approved):
lgtm! But AFAICT there is no README for how to run these local scripts, which is probably fine and could be done in a follow up PR.

hbulgarini:
> lgtm! But AFAICT there is no README for how to run these local scripts, which is probably fine and could be done in a follow up PR.

based on the discussion we had here with Hernando : https://github.com/paritytech/parity-bridges-common/pull/790#discussion_r586799572

i was thinking to put the instructions on the UI README ... anyway in that UI README the first option will be to link instructions for using docker first. @tomusdrw

deployments/local-scripts/relay-messages-millau-to-rialto.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+# A script for relaying Millau messages to the Rialto chain.
+#
+# Will not work unless both the Rialto and Millau are running (see `run-rialto-node.sh`
+# and `run-millau-node.sh).
```
HCastano, line 5: It's not just Rialto and Millau, we also need the header relays running :)

deployments/local-scripts/relay-messages-rialto-to-millau.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+# A script for relaying Rialto messages to the Millau chain.
+#
+# Will not work unless both the Rialto and Millau are running (see `run-rialto-node.sh`
+# and `run-millau-node.sh).
```
HCastano, line 5: Same comment here

deployments/local-scripts/run-millau-node.sh
```diff
@@ -1,8 +1,11 @@
 #!/bin/bash
 
 # Run a development instance of the Millau Substrate bridge node.
+# To override the default port just export MILLAU_PORT=9945
```
HCastano, line 4:
```suggestion
To override the default websockets port just export MILLAU_WS_PORT=9945
```

deployments/local-scripts/run-millau-node.sh
```diff
@@ -1,8 +1,11 @@
...
+
+MILLAU_PORT="${MILLAU_PORT:-9945}"
```
HCastano, line 6:
```suggestion
MILLAU_PORT="${MILLAU_WS_PORT:-9945}"
```

deployments/local-scripts/run-rialto-node.sh
```diff
@@ -1,8 +1,11 @@
 #!/bin/bash
 
 # Run a development instance of the Rialto Substrate bridge node.
+# To override the default port just export RIALTO_PORT=9944
```
HCastano, line 4: Same thing about having `WS`

deployments/local-scripts/relay-messages-rialto-to-millau.sh
```diff
@@ -0,0 +1,20 @@
...
+set -xeu
+
+MILLAU_PORT="${MILLAU_PORT:-9945}"
+RIALTO_PORT="${RIALTO_PORT:-9944}"
```
HCastano, line 9: I think it would be more clear from a `.env` file point of view to call these `NETWORK_WS_PORT`

scripts/send-message-from-rialto-millau.sh
HCastano: This would be nice, but can be done in a follow up

HCastano (approved):

HCastano:
@hbulgarini I forgot that it would auto merge after my approval, can you update the environment variable names and documentation in a new PR 😅

PR #791, opened 2021-03-04
Create CODE_OF_CONDUCT.md

BobinMathew:
issue #789 resolved 😀

PR #792, opened 2021-03-04
Stop running CI when Markdown files are changed

HCastano:
While making a small formatting fix for #791 I noticed that the CI was running when in
reality there's no reason for it to. This PR makes is so that no checks are run when the
only files being changed in a PR are Markdown files.


tomusdrw (approved):

HCastano:
This really doesn't need two reviews, I'm gonna go ahead and merge it.

PR #793, opened 2021-03-04
Move Storage Parser from Bridge Pallet

HCastano:
This PR continues the migration of certain components from `pallet-substrate-bridge` into
`pallet-finality-verifier` (which begun in #783).

In this particular case the storage parser was moved to `finality-verifier`. This
required moving the storage checker around, and then updating some existing code to call
the storage checker and parser from their new homes.


tomusdrw (approved):
Looks good to me, although it's not a direction I was expecting this to go. I hoped that we will get a single PR where we will be actually able to review how all things are structured and how it all fits together.
Now we are reviewing a series of move around PRs and I expect later we will see a removal PR - but since we are re-using old crates it's really not easy to see if EVERYTHING that is supposed to be removed is removed. So we will end up with re-reviewing the final outcome (after removal) anyway and we might be left with some unneeded abstractions inherited from the previous approach.

HCastano:
> Looks good to me, although it's not a direction I was expecting this to go. I hoped that we will get a single PR where we will be actually able to review how all things are structured and how it all fits together.
> Now we are reviewing a series of move around PRs and I expect later we will see a removal PR - but since we are re-using old crates it's really not easy to see if EVERYTHING that is supposed to be removed is removed. So we will end up with re-reviewing the final outcome (after removal) anyway and we might be left with some unneeded abstractions inherited from the previous approach.

I thought it would be a single PR adding code, and another one removing code. When I started this particular PR in fact I was removing code and then realized some other things (related to the storage parser) stopped compiling 😄

I don't see how a "new pallet" would've changed anything though. At the end of the day we still have to remain compatible with other layers of the stack (message lanes, relayers) and a full blown re-write would've made that more painful. There's something to be said for "Kaizen", or continuous improvement. I think the priority now is not ensuring that we don't have unused abstractions, but just making sure that we have a product that works throughout the entire stack. Once we've done that then we can worry about everything else.

PR #795, opened 2021-03-05
Changed delivery and dispatch fee computation methods

### LINKED ISSUE #748: Clean up associated types and methods in`MessageBridge` and `ChainWithMessageLanes` traits
While writing documentation about our ML-integration helpers, I've found couple of things to clean-up:
1) we need to implement `ChanWithMessageLanes` for both This and Bridged chain. There's associated `Call` type in this trait. Obviously, on This chain we never know the `Call` type, so `()` is always used. There's `ThisChainWithMessageLanes` trait, which must be implemented for This Chain. So idea is to move associated `Call` type from base `ChainWithMessageLanes` to `ThisChainWithMessageLanes` trait;
2) there are couple of methods in the `MessageBridge` trait that are not about the bridge itself (like conversion rate, which is realted to both chains), but about specific chain. For example - `fn MessageBridge::maximal_extrinsic_size_on_target_chain()`. It may be moved to `BridgedChainWithMessageLanes` trait.

This only touches glue code, so may be done after initial release.

svyatonik:
closes #748 

With current fee computation scheme, it is almost impossible to compute fee properly for chains where we can't convert between weights and fees with simple `as _`. This was expected, but until yesterday I thought it'll be easy with some hacks.

The main problem is that previously to the weight that was used in the `bridge_runtime_common::messages::*` functions was meant to represent a sum of dispatch weight (which we define with that `#[weight(...)]` attribute) and 'extra weight' (`TotalWeight = DispatchWeight + ExtraWeight`). This extra weight represents base extrinsic weight and per-byte fee (and possibly adjusted fee on chains that have it enabled). After getting this `TotalWeight` we may have computed transaction fee by simply calling `WeightToFee::calc(...)`. But in complex chains (non test chains) the fee computation schema is more complex and there's no that `ExtraWeight`. Instead, both per-byte fee and adjusted fee are something that we may call `ExtraFee`. And converting from that `ExtraFee` to `ExtraWeight` (and current set of functions assumed that conversion) isn't easy (or isn't even possible).

So I've changed a set of methods a bit - now we have `MessageLaneTransaction` instead of simply `Weight`. And this transaction is used when fee computation formula is invoked. So now we have clear border between weights and fees.


relays/substrate/src/main.rs
```diff
@@ -934,4 +934,40 @@ mod tests {
 		);
 		assert!(Rialto::verify_message(&payload).is_err());
 	}
+
+	#[test]
+	fn rialto_tx_extra_bytes_constant_is_correct() {
```
tomusdrw, line 939:
```suggestion
	fn rialto_tx_extra_bytes_constant_is_correct() {
```

:heart: Thanks for this, in the instant I saw `103` hardcoded I started writing a comment that we need a test to know if it breaks :)

tomusdrw (approved):
lgtm!

HCastano (approved):

PR #797, opened 2021-03-05
bridges: fix random_seed

tomusdrw:
Backport of https://github.com/paritytech/polkadot/pull/2504#issuecomment-791599920

Running
```bash
git subtree push --prefix=bridges bridges td-polkadot
````
ends up with 2-commit history (squash commit from polkadot repo & Andre's commit) this doesn't really share any commits with `master`, so I ended up just cherry-picking Andre's commit on top of `master`. Since after `subtree push` it was already part of the repo it was as easy as doing regular cherry-pick (no diffs/patches or anything)

### DIFF
--- a/bin/millau/runtime/src/lib.rs
+++ b/bin/millau/runtime/src/lib.rs
@@ -458,7 +458,7 @@ impl_runtime_apis! {
 		}
 
 		fn random_seed() -> <Block as BlockT>::Hash {
-			RandomnessCollectiveFlip::random_seed().0
+			RandomnessCollectiveFlip::random_seed().0.into()
 		}
 	}
 
--- a/bin/millau/runtime/src/lib.rs
+++ b/bin/millau/runtime/src/lib.rs
@@ -458,7 +458,7 @@ impl_runtime_apis! {
 		}
 
 		fn random_seed() -> <Block as BlockT>::Hash {
-			RandomnessCollectiveFlip::random_seed()
+			RandomnessCollectiveFlip::random_seed().0
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


adoerr (approved):
LGTM

HCastano (approved):
Code looks good, but I have some questions.

> git subtree push --prefix=bridges bridges td-polkadot

Since I'm still getting familiar with the `git subtree` workflow, can you let me know if this is correct: Basically what's happening here is that in the Polkadot repo you pushed Andres changes to a branch `td-polkadot` on a remote you have called `bridges` (which I guess points to `https://github.com/paritytech/parity-bridges-common`). The changes are those found in the `bridges` folder which is where the bridge subtree is.

> Since after subtree push it was already part of the repo it was as easy as doing regular cherry-pick (no diffs/patches or anything)

Can you expand on this point a bit?

tomusdrw:
> Basically what's happening here is that in the Polkadot repo you pushed Andres changes to a branch td-polkadot on a remote you have called bridges (which I guess points to https://github.com/paritytech/parity-bridges-common). The changes are those found in the bridges folder which is where the bridge subtree is.

Yes. Basically `git` is filtering all changes done on the `polkadot` repo to isolate ones that affect `bridges` sub-directory. Then it only picks commits that touch this sub-dir and creates a completely separate git history, that get's pushed to `bridges/td-polkadot` branch.,

Since that history, from Polkadot's PoV is just 2 commits (the squashed changes of bridges repo and Andre's changes) that's why we end up with a branch that doesn't share any history with our `master`, that's why I ended up cherry-picking from that branch. I could probably also rebase and resolve conflicts favouring `master` in case of conflicts (the code would be identical anyway) and that should end up with the same results.


Since Andre's PR was a companion PR, I've added `.into()` at the very end, so that the changes are compatible even before the [upstream PR](https://github.com/paritytech/substrate/pull/8180) to substrate is merged :)

Issue #360, opened 2020-09-18
Make Substrate Pallet Instantiable 


HCastano:
We want a runtime to be able to bridge to many different Substrate chains. To do that we'll need to make it [instantiable](https://substrate.dev/recipes/instantiable.html). We may also need to add items to the configuration trait to account for differences in Substrate chains, such as different number, hash, and header types.

PR #798, opened 2021-03-07
Make Substrate Pallet Instantiable

MaciejBaj:
Hello, I'm making the bridge pallet instantiable per the requirements of my project to have multiple bridges connected and found the issue https://github.com/paritytech/parity-bridges-common/issues/360 which is open and covers exactly that. 

WIP - need to fix tests.

### DIFF
--- a/modules/substrate/src/lib.rs
+++ b/modules/substrate/src/lib.rs
@@ -35,7 +35,7 @@ use crate::storage::ImportedHeader;
 use bp_header_chain::AuthoritySet;
 use bp_runtime::{BlockNumberOf, Chain, HashOf, HasherOf, HeaderOf};
 use frame_support::{
-	decl_error, decl_module, decl_storage, dispatch::DispatchResult, ensure, traits::Get, weights::DispatchClass,
+	decl_error, decl_module, decl_storage, dispatch::DispatchResult, ensure, traits::Get, weights::DispatchClass
 };
 use frame_system::{ensure_signed, RawOrigin};
 use sp_runtime::traits::Header as HeaderT;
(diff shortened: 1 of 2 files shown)


HCastano:
@MaciejBaj Hey! What are you using the `pallet-substrate-bridge` for right now? We were thinking of removing it in favour of `pallet-finality-verifier` now that https://github.com/paritytech/parity-bridges-common/pull/783 has landed.

MaciejBaj:
> @MaciejBaj Hey! What are you using the `pallet-substrate-bridge` for right now? We were thinking of removing it in favour of `pallet-finality-verifier` now that #783 has landed.

 I see `pallet-finality-verifier` is the new `substrate-bridge` now and nicely migrated to FRAME v2. I'm happy to make `pallet-finality-verifier` instantiable instead.

A side question: in the new `pallet-finality-verifier` is there a way to omit the finality source check and to simply forward + validate headers? e.g. when running a bridge that connects with a light client there is no "grandpa_subscribeJustifications" i could stream that data from.

MaciejBaj:
thanks for ur reply. 

> Since you only have a light client though I don't think you'll be able to submit anything to pallet-finality-verifier. Since you seemed to have been using pallet-substrate-bridge before, did you ever submit finality proofs via pallet_substrate_bridge::finalize_header(), or were you more concerned with just importing headers?

my main concern is indeed in just importing headers. base functionality won't need finality checks for all connected bridges.

if the `substrate-bridge` pallet goes there won't be the `verifier` module and `finality-verifier` doesn't track the `BestHeight`, what would be the way to simply `import_headers`?

tomusdrw:
> if the substrate-bridge pallet goes there won't be the verifier module and finality-verifier doesn't track the BestHeight, what would be the way to simply import_headers?

tomusdrw:
> if the substrate-bridge pallet goes there won't be the verifier module and finality-verifier doesn't track the BestHeight, what would be the way to simply import_headers?

The main reason for removing the `substrate-bridge` pallet is the fact that we were not verifying almost anything about the imported headers - this allowed growing arbitrary forks and submitting any kind of headers one would like. This is an obvious DDoS vector and unbounded state growth of your chain. That's why we switched to only import headers with finality proofs.
In theory you could verify BABE signature to make sure that the headers are legit, but you'd still have a problem of potentially tracking multiple forks. That's not something we plan to support in this repository any time soon though.

MaciejBaj:
got it, thank you @HCastano and @tomusdrw for taking the time elaborating on this.

HCastano:
No problem!

I'm gonna close this. @MaciejBaj, feel free to make a PR to make `pallet-finality-verifier` instantiable

svyatonik:
@MaciejBaj I've opened #825 for that, if you still need it.

PR #802, opened 2021-03-08
Use globs for workspace members.

tomusdrw:


Cargo.toml
```diff
@@ -1,17 +1,11 @@
...
-	"modules/substrate",
-	"modules/ethereum",
-	"modules/ethereum-contract/builtin",
-	"modules/finality-verifier",
-	"modules/currency-exchange",
-	"relays/ethereum",
-	"relays/substrate",
+	"bin/*/node",
+	"bin/*/runtime",
+	"bin/runtime-common",
```
HCastano, line 6: I don't think we need this since it's used by some other crates

HCastano (approved):

PR #804, opened 2021-03-08
Clean Finality Verifier Pallet

HCastano:
This PR cleans up some unused types from `pallet-finality-verifier` and moves a
couple of things into helper functions.

With these changes I think `pallet-finality-verifier` is at stage where it can be
reviewed as "standalone".


tomusdrw:
If it passes CI, I'm willing to merge this as-is. As I've mentioned couple of times already, reviewing removal PRs is really difficult, so I'd rather just remove everything and then review the end goal - i.e. complete, streamlined finality-verifier pallet.

modules/finality-verifier/src/lib.rs
```diff
@@ -16,27 +16,25 @@
...
+//! Once verified, finalized headers are stored in the pallet, thereby creating a sparse
+//! header chain. This sparse header chain can be used as a source of truth for other
+//! higher-level applications.
 //!
-//! The underlying pallet used for storage is assumed to be a pallet which tracks headers and
-//! GRANDPA authority set changes. This information is used during the verification of GRANDPA
-//! finality proofs.
+//! Since this pallet only tracks finalized headers it does not deal with forks. If a
+//! fork does occur on the bridged chain there is nothing the pallet can do to handle
+//! this - governance intervention will be required.
```
tomusdrw, line 28:
```suggestion
//! Since this pallet only tracks finalized headers it does not deal with forks. 
//! Forks can only occur in case GRANDPA Validator Set on the bridged chain
//! is either colluding or there is a sever bug causing equivocation. Such events are
//! outside of the scope of this pallet. Shall the fork occur on the bridged chain governance
//! intervention will be required to re-initialize the bridge and track the right fork.
```

modules/finality-verifier/src/lib.rs
```diff
@@ -16,27 +16,25 @@
...
-//! whose finality has been verified will be written to the underlying pallet.
+//! This pallet is an on-chain GRANDPA light client for Substrate based chains.
 //!
-//! By verifying the finality of headers before writing them to storage we prevent DoS vectors in
-//! which unfinalized headers get written to storage even if they don't have a chance of being
-//! finalized in the future (such as in the case where a different fork gets finalized).
+//! This pallet achieves this by trustlessly verifying GRANDPA finality proofs on-chain.
+//! Once verified, finalized headers are stored in the pallet, thereby creating a sparse
+//! header chain. This sparse header chain can be used as a source of truth for other
+//! higher-level applications.
```
tomusdrw, line 24:
```suggestion
//! higher-level applications.
//!
//! The pallet is responsible for tracking GRANDPA Validator Set hand-offs. We only import headers with
//! justifications signed by the current validator set we know of. The header is inspected for `ScheduledChanges`
//! digest item, which is then used to update next validator set.
```

tomusdrw (approved):
lgtm! Added some suggestion to the documentation of the pallet.

PR #806, opened 2021-03-09
Add Transaction Payment Custom RPC

tomusdrw:
This allows retrieving `weight` of a `Call` over RPC and is required by the UI.
CC @hbulgarini


hbulgarini (approved):
I can confirm from the JS API perspective that this change is letting the API to read the weight of the transaction correctly.

svyatonik (approved):

PR #811, opened 2021-03-09
Log runtime logs from `pallet-finality-verifier`

HCastano:
In our testnet we're now using `pallet-finality-verifier` instead of
`pallet-substrate-bridge`. This updates our deployment logging to match that.


HCastano:
Merging this so we can have these logs in tonight's deployment.

PR #812, opened 2021-03-10
Add mising MessageFeeData type.

tomusdrw:


deployments/types/common.json
```diff
@@ -130,5 +130,9 @@
 	"messages_in_oldest_entry": "MessageNonce",
 	"total_messages": "MessageNonce"
   },
-  "AncestryProof": "()"
+  "AncestryProof": "()",
+  "MessageFeeData": {
+	"lane_id": "LaneId",
```
hbulgarini, line 135: not a blocker at all but it would be nice to have it indented

hbulgarini (approved):
lgtm!

deployments/types/common.json
tomusdrw: Interesting :D I've re-formatted the entire file, cause it wasn't using the right indentation settings from `.editorconfig`. Should be better now.

PR #808, opened 2021-03-09
Add Westend & Rococo primitives

tomusdrw:
Adding primitive crates for Westend & Rococo.
Extracted all polkadot-like chains primitive types into a single crate.

Issue #451, opened 2020-10-23
Sync Kusama to Millau


tomusdrw:
Introduce unidirectional sync between Polkadot / Kusama to Millau to make sure we can actually sync it.

Come up with a more detailed plan how to test bridges.

tomusdrw:
I've renamed the issue, since the rest of the "testing plan" is being realised right now:
1. Running test bridge with Kusama
2. High level documentation (#565)
3. Internal & External audits
4. Deployment details

svyatonik:
We now have this Westend (more or less Kusama) -> Millau bridge. Imo adding another one won't be super-useful.

Issue #438, opened 2020-10-23
Integrate with Polkadot & Kusama runtimes


tomusdrw:
Figure out if forking the repo is required (simlar to #218). Instead of bridging directly with the relay chain, maybe it's a better idea to deploy a parachain - this depends on the parachain availability.

HCastano:
I think we should look into integrating as a parachain. I'm not sure if the Polkadot council will approve of a bridge being added to the relay chain since it could potentially introduce "bloat".

tomusdrw:
A tentative integration in https://github.com/svyatonik/polkadot/commits/bridges-integration

tomusdrw:
We need to make sure to tune weight-related constants for the polkadot / kusama integration.

svyatonik:
https://github.com/paritytech/polkadot/tree/bridges-integration-with-council-support

svyatonik:
We no longer plan to deploy our pallets at Kusama/Polkadot directly, so this issue is no longer actual

PR #813, opened 2021-03-10
Reorganize relay code to make it easy to add new networks.

tomusdrw:
Currently the relayer is designed around `rialto-millau` bridge. While it is possible to add support for more networks, we would end up growing (already large) common files like `cli.rs` or `main.rs`. Recently I've added a bunch of other sub-commands, which made the problem even worse.

This PR is an attempt to split bridge-specific commands into a sub-module and introduce a bit more hierarchical structure (opposed to current flat one) to make it easier to add (partial or full) support for new networks.

While in the future we could add a bunch of macros to make it super easy to add more networks or build bridge-specific binaries, I think the mid-term solution is to simply extend existing binary, but in a way that will make future refactors easier.

I've tried to keep changes to minimum. The new code is mainly to delegate logic from abstract `main.rs` to bridge-specific `millau_rialto/cli.rs` file.

Related #808 #451 #438


relays/substrate/src/rialto_millau/cli.rs
```diff
@@ -0,0 +1,423 @@
...
+
+// Parity Bridges Common is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Deal with CLI args of substrate-to-substrate relay.
```
HCastano, line 17:
```suggestion
//! Deal with CLI args of Rialto <> Millau relay.
```

HCastano (approved):
I like this new re-org 👍

PR #814, opened 2021-03-10
Move CI from GitHub Actions to GitLab

TriplEight:
# New CI.

### Nightly test pipeline
Is triggered by a CI image rebuild at 6 AM CET.
It gets a new Rust nightly and runs all types of stable and nightly tests.
Doesn't run builds.
Looks like this:
image
(but upstream will be the scripts project)

### Nightly build pipeline
Is triggered by a schedule at 1 AM CET.
It runs all the regular tests and then builds and publishing.
Looks like this:
image

### New tag pipeline
Is triggered by a push of a new git tag.
Similar to the Nightly build pipeline.

### Master pipeline
Is triggered by a new commit in the `master` branch.
Looks like this:
image

### PR pipeline
Is triggered by a new commit in any PR.
For now similar to the Master pipeline.

### Web pipeline
Is triggered by [Run Pipeline](https://gitlab.parity.io/parity/parity-bridges-common/-/pipelines) button.
Depends on which branch is set there, generally similar to PR or Master pipeline.


.gitlab-ci.yml
```diff
@@ -0,0 +1,83 @@
...
+  stage:                           check
+  <<:                              *docker-env
+  script:
+    - echo "cargo check --verbose --all-features"
+
+check-wasm:
+  stage:                           check
+  <<:                              *docker-env
+  script:
+    - echo "cargo check --verbose --no-default-features --target wasm32-unknown-unknown"
```
tomusdrw, line 62: This is not really needed, we don't care about WASM.

.gitlab-ci.yml
```diff
@@ -0,0 +1,83 @@
...
+    - echo "cargo build --verbose --all-features --release"
+
+build-wasm:
+  stage:                           test
+  <<:                              *docker-env
+  needs:
+    - job:                         check-wasm
+      artifacts:                   false
+  script:
+    - echo "cargo build --verbose --no-default-features --release --target wasm32-unknown-unknown"
```
tomusdrw, line 82: same here.

tomusdrw (commented):
Nice! We still need `fmt`, `clippy` and `cargo check` and docker publish (images nightly) though.

HCastano:
Any updates here @TriplEight?

TriplEight:
New pipeline [example](https://gitlab.parity.io/parity/parity-bridges-common/-/pipelines/138854).
Please highlight when do you want which checks, whether should all of them run on PRs, on tags, on master and on schedules?
Do you want `fmt` on stable, because that's how it was in GHA.
Do you want artifacts from both stable and nightly build jobs?
When do you want to publish an image?
Which jobs you want to be non-blocking?

HCastano:
> New pipeline example.

👌

> Please highlight when do you want which checks, whether should all of them run on PRs, on tags, on master and on schedules?

So I don't think we need to run the `nightly` compiler checks on every PR. If possible I think we should run those overnight alongside the image/artifact build.

As for the timing we'll want to run all the checks (except the `nightly` ones) on every single PR, and it would also be good to run them on `master` when a new PR is merged. 

Tags are a little special, and we haven't been using them yet tbh. However, we should run the full CI suite + image/artifact jobs when a new tag is pushed.

> Do you want fmt on stable, because that's how it was in GHA?

Yes, please run `RustFmt` on stable

> Do you want artifacts from both stable and nightly build jobs?

No, only artifacts built with the `stable` compiler should be published.

> When do you want to publish an image?

Images should be built from the `master` branch nightly (midnight UTC). Binaries should also be published alongside this image

> Which jobs you want to be non-blocking?

We don't really care if `cargo-deny` fails for security vuls since most of the time we can't do anything about them. However, we do care if it fails with a license check. If that logic is too hard, just mark it as failed and we can manually check on the reasoning and deal with it accordingly.

Everything else, including Clippy and the Spellcheck, should be required

Thanks for looking at this again @TriplEight

TriplEight:
Thanks

> So I don't think we need to run the `nightly` compiler checks on every PR. If possible I think we should run those overnight alongside the image/artifact build.

Then since you already have `nightly` rust installed in the CI image, I'll run the `nightly` tests on every CI image (hence `nightly` Rust update). You will see these tests in the same Gitlab pipelines.

Other questions about CI:
- it's contradictive: either you want **all** nightly-rust stuff running on schedules or `clippy` which is also nightly on all PR commits
- why do you even need `nightly` `build` when you have both `nightly` `test` and `check`
- can I optimize and run `build` together with `test` but making them both `--release`?

HCastano:
>  it's contradictive: either you want **all** nightly-rust stuff running on schedules or `clippy` which is also nightly on all PR commits

I wouldn't quite say that. Just because we have Clippy running on every PR doesn't mean we need to have the rest of the `nightly` dependent checks running on every PR. We need the `nightly` compiler for the `stable` PR builds anyways because of WASM

>  why do you even need `nightly` `build` when you have both `nightly` `test` and `check`

So I think we should run nightly builds using the `nightly` compiler which do both `build` and `test`. We want to do this to be warned of any potential breakages coming our way (or to report them to the compiler team)

>  can I optimize and run `build` together with `test` but making them both `--release`?

What do you mean by "run them together"?

TriplEight:
> What do you mean by "run them together"?

just to reuse the cache within the same job. Given the same flags, it's often quite fast (sometimes seconds) to run `cargo build x` and then `cargo test x`. And then, if tests are good, the binary from build will be passed further.

TriplEight:
I've executed a [bench](https://gitlab.parity.io/TriplEight/parity-bridges-common/-/pipelines/140370/builds): build + test in one job instead of both in separate jobs, no cache.
Here, `cargo test` runs with `--release` in both cases:
image

Here a separate test job runs without `--release`
image

## Results

In case of build pipeline it's almost free to execute test together with build, but we do not intend to build often, so I'll stick to having them separately and `cargo test` without the `--release` flag, this will be the fastest.

TriplEight:
This is ready to be merged.
I'll take care of publishing binaries to s3 in the next PR, can't do it without CI merged to master.

TriplEight:
Before the merge, I'll switch the required tests by github to the ones that are coming from gitlab

.gitlab-ci.yml
```diff
@@ -0,0 +1,274 @@
...
+    - echo "$Docker_Hub_Pass_Parity" |
+        buildah login --username "$Docker_Hub_User_Parity" --password-stdin docker.io
+    - buildah info
+    - buildah push --format=v2s2 "${IMAGE_NAME}:${VERSION}"
+    - buildah push --format=v2s2 "${IMAGE_NAME}:sha-${CI_COMMIT_SHORT_SHA}"
+    - buildah push --format=v2s2 "${IMAGE_NAME}:latest"
+  after_script:
+    - env REGISTRY_AUTH_FILE= buildah logout "$IMAGE_NAME"
+
+rialto-bridge-node:
```
tomusdrw, line 258: Don't we need to update `docker-compose.yml`?

tomusdrw (approved):
Awesome, thanks!

.config/spellcheck.toml
```diff
@@ -2,6 +2,8 @@
 lang = "en_US"
 search_dirs = ["."]
 extra_dictionaries = ["lingua.dic"]
+skip_os_lookups = true
+use_builtin = true
```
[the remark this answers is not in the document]
TriplEight: I've contacted spellcheck's maintainer on why it didn't use the built-in dictionary and they told me this way to activate it.

.gitlab-ci.yml
```diff
@@ -0,0 +1,274 @@
...
+  rules:
+    - if: $CI_COMMIT_TAG
+    - if: $CI_COMMIT_BRANCH
+
+variables:                         &default-vars
+  GIT_STRATEGY:                    fetch
+  GIT_DEPTH:                       100
+  CARGO_INCREMENTAL:               0
+  ARCH:                            "x86_64"
+  CI_IMAGE:                        "paritytech/bridges-ci:production"
```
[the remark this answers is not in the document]
TriplEight: [This](https://github.com/paritytech/scripts/blob/master/.gitlab-ci.yml#L375-L384) is how the image updates.
image
1. the new CI image is built at 6 AM nightly and pushed with `staging` tag.
2. `staging` is tested against `bridges` CI on master and gets a vuln scan
3. if green, it gets into production

in this case it's failed because Gitlab CI is not in master yet.

.gitlab-ci.yml
TriplEight: I need to know more about it

HCastano:
I didn't get a chance to review this!

modules/grandpa/src/lib.rs
```diff
@@ -693,7 +693,7 @@ mod tests {
 				CurrentAuthoritySet::<TestRuntime>::get().authorities,
 				init_data.authority_list
 			);
-			assert_eq!(IsHalted::<TestRuntime>::get(), false);
+			assert!(!IsHalted::<TestRuntime>::get());
```
HCastano, line 696: Tbh I'm not a fan of this Clippy suggestion, the other was easier to read (no double !!), but 🤷‍♂️

ci.Dockerfile
```diff
@@ -0,0 +1,53 @@
+# This file is a "runtime" part from a builder-pattern in Dockerfile, it's used in CI.
```
HCastano, line 1: To clarify, what's going on here is that we're taking the binaries built by the CI pipeline and putting them inside a Dockerfile, whose image we then publish?

Where does this leave our top level Dockerfile then?

HCastano (commented):
In general the pipeline looks good, I still have some questions though

.gitlab-ci.yml
```diff
@@ -0,0 +1,274 @@
...
+    - time cargo build --release --verbose --workspace
+  after_script:
+    # Prepare artifacts
+    - mkdir -p ./artifacts
+    - strip ./target/release/rialto-bridge-node
+    - mv -v ./target/release/rialto-bridge-node ./artifacts/
+    - strip ./target/release/millau-bridge-node
+    - mv -v ./target/release/millau-bridge-node ./artifacts/
+    - strip ./target/release/ethereum-poa-relay
+    - mv -v ./target/release/ethereum-poa-relay ./artifacts/
```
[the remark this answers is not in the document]
TriplEight: It's built anyway with `--workspace`. Do you mean we shouldn't use this artifact for now? Ain't it used on one of the Dockerfiles?

.gitlab-ci.yml
```diff
@@ -0,0 +1,274 @@
...
+      fi
+    - echo "Effective tags = ${VERSION} sha-${CI_COMMIT_SHORT_SHA} latest"
+  script:
+    - test "${Docker_Hub_User_Parity}" -a "${Docker_Hub_Pass_Parity}" ||
+        ( echo "no docker credentials provided"; exit 1 )
+    - cd ./artifacts
+    - buildah bud
+        --format=docker
+        --build-arg VCS_REF="${CI_COMMIT_SHORT_SHA}"
+        --build-arg HEALTH="${HEALTH}"
```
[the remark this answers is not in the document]
TriplEight: Oh missed it here, thanks!

Dockerfile
```diff
@@ -8,14 +8,14 @@
...
+FROM paritytech/bridges-ci:latest as builder
 WORKDIR /parity-bridges-common
 
 COPY . .
 
 ARG PROJECT=ethereum-poa-relay
-RUN cargo build --release --verbose -p ${PROJECT}
-RUN strip ./target/release/${PROJECT}
+RUN cargo build --release --verbose -p ${PROJECT} && \
+    strip ./target/release/${PROJECT}
```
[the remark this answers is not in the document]
TriplEight: every `RUN ...` introduces a new layer in the image. It's good when you're debugging builds, but adds additional weight to the resulting image.

README.md
```diff
@@ -49,7 +49,7 @@ and external processes called relayers. A bridge chain is one that is able to fo
 of a foreign chain independently. For example, consider the case below where we want to bridge two
 Substrate based chains.
 
-```
+```bash
```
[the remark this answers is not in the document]
TriplEight: ah ok, was satisfying the md linter xD

ci.Dockerfile
TriplEight: Yeah, exactly.
Top-level `./Dockerfile` is a CI on its own with that docker builder pattern. It's not beneficial to use such a pattern within another CI system as it won't reuse cache normally.
However, it's great for local usage.
I couldn't come up with the idea on how to inherit one from another to stay DRY, there's a `--target` in `docker build` which can choose the exact step of the builder pattern to start from, but
- the unavoidable `COPY` instruction will trigger the initial step anyway, hence it won't help
- it's not yet working in buldah man_shrugging

modules/grandpa/src/lib.rs
tomusdrw: It gives much better debugging info if it fails (shows the entire expression instead of `expected true, got false`).

.gitlab-ci.yml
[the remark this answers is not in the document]
tomusdrw: Let's keep them until they are part of the codebase I don't think it's going to affect build times that much.

.gitlab-ci.yml
tomusdrw: I thought that the docker image names might have changed and the docker-compose file is fetching images from dockerhub, but I guess there were no changes.

.gitlab-ci.yml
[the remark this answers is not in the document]
TriplEight: as long as we run `cargo build --workspace` we're simply building everything

ci.Dockerfile
HCastano: I think we'll need to note down somewhere that our published images don't match that of the `Dockerfile`. I'm afraid it could cause some confusion if people use different images without realizing

Issue #391, opened 2020-10-01
Use Automated Benchmarking for Runtimes


HCastano:
As previously mentioned in #78, Substrate has gotten a feature where we're able to automatically generate weights for pallets based on the current runtime configuration. We should use this feature to benchmark the Rialto and Millau runtimes (and their included pallets).

For further reading see [here](https://github.com/paritytech/substrate/tree/master/frame/benchmarking).

Related: #78, #390 

**Cumulus repo:**
- `bridge-hub-kusama` - `type Weigher = ` according to `bridge-hub-rococo -> BridgeHubXcmWeight` 
- https://github.com/paritytech/cumulus/blob/bridge-hub-rococo-wococo/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/weights/xcm/mod.rs#L216-L223
```
fn export_message(_: &NetworkId, _: &Junctions, _: &Xcm<()>) -> XCMWeight {
		log::error!(
			target: crate::LOG_TARGET,
			"TODO: Calling weight: export_message -> triggers unpaid_execution -> need fix here!"
		);
		// TODO:check-parameter - add correct weight also add it to the polkadot gav-xcm-v3
		XcmGeneric::<Runtime>::unpaid_execution().ref_time()
	}
```

tomusdrw:
See #815

tomusdrw:
I believe this is handled automatically in the Polkadot repository while releasing, but leaving it here just to make sure.

svyatonik:
I'll move it to K<>P milestone, as it is not super-relevant for testnets and it is better to implement it with next `bridges` subtree update

svyatonik:
Done in https://github.com/paritytech/polkadot/tree/bridges-integration-with-council-support . Will close this issue once we have this branch merged

svyatonik:
Benchmarks for RBH/WBH are ready and weights are gemnerated by bench bot

Issue #390, opened 2020-10-01
Add Benchmarking to Millau Runtime


HCastano:
The Millau runtime does not have any benchmarking capabilities at the moment. We need to set that up in a similar fashion as we have done with the Rialto runtime

tomusdrw:
Is it required for any feature? Closing, since I think weights are covered in #391 and I don't see a reason just to have benchmarking for Millau, just for the sake of having it :)

HCastano:
> Is it required for any feature? Closing, since I think weights are covered in #391 and I don't see a reason just to have benchmarking for Millau, just for the sake of having it :)

No, I don't think this is required anymore. The main thing for benchmarking is that we have a good benchmark framework for each pallet, and that we know they compile with at least one runtime (in our case that's Rialto). We can then automate benchmarking with our "production" runtimes (e.g Kusama).

PR #815, opened 2021-03-10
Weights for `pallet-bridge-grandpa`

HCastano:
This PR sets up our benchmarking framework for `pallet-bridge-grandpa`. There is still some future work to be done, like running benches with more authorities, but in general I think this is at a point where it can be merged and we can improve upon it in follow up PRs. 

I have a description of my early findings in [this comment](https://github.com/paritytech/parity-bridges-common/pull/815#issuecomment-807801419). I would recommend having this alongside `benchmarking.rs` while reviewing this PR.


modules/grandpa/src/benchmarking.rs
```diff
@@ -0,0 +1,210 @@
...
+
+		<CurrentAuthoritySet<T, I>>::put(&authority_set);
+
+	}: {
+		let authority_set = <CurrentAuthoritySet<T, I>>::get();
+		<CurrentAuthoritySet<T, I>>::put(&authority_set);
+	}
+}
+
+#[cfg(test)]
```
svyatonik, line 186: Iiuc benchmarks tests won't run on CI. Do you think we still need it?

svyatonik: (also seems that the tests are actually `assert(true)`)

modules/grandpa/src/benchmarking.rs
```diff
@@ -0,0 +1,210 @@
...
+		// Need to play with the types here to get this to compile...
+		// assert_eq!(<BestFinalized<mock::TestRuntime>>::get(), header.hash());
+		// assert!(<ImportedHeaders<mock::TestRuntime>>::contains_key(header.hash()));
+	}
+
+	// What we want to check here is the effect of many pre-commits on justification verification.
+	// We do this by creating many forks, whose head will be used as a signed pre-commit in the
+	// final justification.
+	submit_finality_proof_on_many_forks {
+		let n in 1..2;
```
svyatonik, line 91: (this probably also needs to be increased)

modules/grandpa/src/benchmarking.rs
```diff
@@ -0,0 +1,210 @@
...
+
+		let justification = make_justification_for_header(params).encode();
+
+	}: submit_finality_proof(RawOrigin::Signed(caller), header, justification)
+	verify {
+		assert!(true)
+	}
+
+
+	// Here we want to find out what the overheader of looking for an enacting an authority set is.
```
svyatonik, line 124: This comment seems to be lost

modules/grandpa/src/benchmarking.rs
```diff
@@ -0,0 +1,210 @@
...
+	// We need some way for the benchmarks to use headers in a "generic" way. However, since we do
+	// use a real runtime we need a way for the runtime to tell us what the concrete type is.
+	fn bridged_header(num: BridgedBlockNumber<Self, I>) -> BridgedHeader<Self, I>;
+}
+
+benchmarks_instance_pallet! {
+	// What we want to check here is the effect of vote ancestries on justification verification
+	// time. We will do this by varying the number of ancestors our finality target has.
+	submit_finality_proof_on_single_fork {
+		let n in 1..10;
```
svyatonik, line 53: The difference between 1 and 10 headers is 9 hash operations and BTreeMap inserts, right? So probably you'll want more headers to see actual difference. Probably you'll even get conflicting results here (e.g. time of bench with depth=10 will be lesser than time of bench with depth=1 e.g. because of context switches or similar). So if the point of this benchmark is to build fn(depth) -> time, then it makes sense to make it much larger.

modules/grandpa/src/lib.rs
```diff
@@ -511,6 +514,8 @@ pub(crate) fn find_forced_change<H: HeaderT>(
 }
 
 /// (Re)initialize bridge with given header for using it in external benchmarks.
+// TODO: Will probably need to change this a bit for the finality verifier benchmarks, while keeping
```
svyatonik, line 517: Do you still need this TODO? Which btw references pallet as finality verifier :)

svyatonik (approved):
Some benchmarks seems to be too detailed to me - i.e. you have `read_write_authority_sets` benchmark, which, just measures the time of decoding+encoding the authorities set. But my guess is that there are other parameters that depend on authorities set size and which affect extrinsic weight significantly - e.g. signatures verification. So imo it is better to build function which shows how authorities set size affects the whole `submit_finality_proof` extrinsic execution time (which will also indirectly include your `read_write_authority_sets`).

modules/grandpa/src/benchmarking.rs
HCastano: > Iiuc benchmarks tests won't run on CI. Do you think we still need it?

We should be running them on the CI, even if it's just periodically. I think that'll be a change for a different PR though.

I've found the tests to be pretty useful even if they're kinda "dumb" because they've helped me figure out if I've been setting up my benchmarks correctly (e.g are my justifications actually valid)

modules/grandpa/src/lib.rs
HCastano: I'm still mulling this over, I'll get back to it eventually :P

bin/rialto/runtime/src/lib.rs
```diff
@@ -1005,7 +1007,24 @@ impl_runtime_apis! {
...
-			add_benchmark!(params, batches, pallet_bridge_eth_poa, BridgeKovan);
+			use pallet_bridge_grandpa::BridgedHeader;
+			use pallet_bridge_grandpa::benchmarking::{
+				Config as BridgeGrandpaConfig
+			};
+
+			impl BridgeGrandpaConfig<()> for Runtime {
+				fn bridged_header(num: bp_millau::BlockNumber) -> BridgedHeader<Self, ()> {
+					use sp_runtime::traits::Header;
+					bp_millau::Header::new(
```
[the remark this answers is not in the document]
HCastano: There were problems with types iirc, but I'll look into your suggestion

HCastano:
Even though this is still a WIP I want to share some of my findings.

To produce the graphs I ran the following command:

```bash
./target/release/rialto-bridge-node benchmark --chain dev --execution wasm --extrinsic "*" --pallet pallet_bridge_grandpa --wasm-execution compiled --steps 50 --repeat 20 --raw > bench-output.txt
```

And then dumped the appropriate parts of the output file into [here](https://www.shawntabrizi.com/substrate-graph-benchmarks/) which did the plotting.

## `submit_finality_proof` Benchmarks

Below are graphs for the benchmarks I have set up at the moment. Note that I have three benchmarks related to `submit_finality_proof`. 
1. The "actual" benchmark
2. Pre-Commits vs Verification Time
3. Vote Ancestries vs Verification Time

### Actual 
The "actual" benchmark takes into account both variables that I'm interested in (number of pre-commits, and vote ancestries). 

submit_finality_vote_ancestries

submit_finality_precommits

It looks pretty linear overall, but I'm a bit suspicious of the second graph since my standalone benchmark for how pre-commits affect verification time suggests a superlinear relationship (see next section for that).

Something worth nothing here is how "heavy" this call is in the (somewhat) worst case scenario - nearly 200ms or roughly ~10% of the time we have to process extrinsics. I didn't run these benchmarks on a laptop either, I ran them on a fairly beefy server. We'll need to keep an eye on this.

### Pre-Commits
The second is just for seeing how the number of pre-commits affects justification verification time.

submit_finality_many_forks

One important thing to note here is that the graph looks every so slightly superlinear. While I suspected that the GRANDPA code wouldn't be linear, I thought it would be more clear with `u8::MAX` pre-commits. I'll need to investigate this further with a higher number of pre-commits, but probably in a different PR since it'll involve changes to the testing utilities.

### Vote Ancestries
The third is just for seeing how the number of vote ancestries affects justification verification time.

submit_finality_single_fork

It's a linear relationship which is good to see.

## "Micro" Benchmarks

Lastly I have some "micro" benchmarks. While I may not keep these around long term, I was curious to see the effects of some operations that happened in the middle of `submit_finality_proof`. Namely, I wanted to see the impact of searching through `Digests` in a header, and seeing the effects of larger authority sets when we enact changes (since we write the new set to storage).

### Digests 

It turns out that looking for digests is super quick and basically negligible in the context of `submit_finality_proof`. 

digests

### Enacting Authority Sets

As for enacting authority set changes, it's not negligible, but the impact is somewhat small. 

authority_sets

Something I've been thinking about related to authority set enactments is refunding some of the weight if a header *doesn't* enact an authority set change, since that's one less storage write we have to do.

andresilva:
> One important thing to note here is that the graph looks every so slightly superlinear. While I suspected that the GRANDPA code wouldn't be linear, I thought it would be more clear with u8::MAX pre-commits. I'll need to investigate this further with a higher number of pre-commits, but probably in a different PR since it'll involve changes to the testing utilities.

If I understood correctly, this benchmark is essentially the same as the one just before it but with `s: 2` instead of `s: 50`. I suppose if you kept increasing `p` over 250 it would become "more linear". Overall the benchmarks LGTM and the results are in-line with my expectations.

modules/grandpa/src/lib.rs
```diff
@@ -107,7 +133,10 @@ pub mod pallet {
 		///
 		/// If successful in verification, it will write the target header to the underlying storage
 		/// pallet.
-		#[pallet::weight(0)]
+		#[pallet::weight(T::WeightInfo::submit_finality_proof(
+			T::BridgedSessionLength::get().as_() as u32,
+			T::BridgedValidatorCount::get() as u32,
```
tomusdrw, line 138: I think we can be much more accurate with the validator count and get rid of this hardcoded parameter.
Either we can read it from the storage (I don't know the exact rules about weights, but it's probably discouraged due to extra state read) or we can have the submitter provide the current validator set length for more accurate weight calculation and we simply compare that with an actual validator set.

Happy to work on this under separate issue though as a weight cost optimisation, since it requires a change in the relayer code as well.

modules/grandpa/src/lib.rs
```diff
@@ -134,16 +163,18 @@ pub mod pallet {
...
-			try_enact_authority_change::<T, I>(&finality_target)?;
+			let _enacted = try_enact_authority_change::<T, I>(&finality_target, set_id)?;
 			<BestFinalized<T, I>>::put(hash);
 			<ImportedHeaders<T, I>>::insert(hash, finality_target);
 			<RequestCount<T, I>>::mutate(|count| *count += 1);
 
 			log::info!(target: "runtime::bridge-grandpa", "Succesfully imported finalized header with hash {:?}!", hash);
 
-			Ok(().into())
+			Ok(Some(T::WeightInfo::submit_finality_proof(votes as u32, precommits as u32)).into())
```
tomusdrw, line 177: I think this is incorrect and may result in an even higher weight than declared (BTW, what happens in such case?).

In the benchmarks we have two parameters: `forks` and `depth` which correspond to number of pre-commits and length of forks the pre-commits are on - i.e. the total number of headers in justification will be `headers = forks * depth`. Here, afaiu, `votes` indicates the total number of headers in the justification, so the weight we calculate is going to be for a worst case, where we have `votes * precommits` number of headers.

primitives/chain-westend/src/lib.rs
```diff
@@ -57,6 +57,9 @@ pub const FROM_WESTEND_LATEST_CONFIRMED_NONCE_METHOD: &str = "FromWestendInbound
 /// Name of the `FromWestendInboundLaneApi::unrewarded_relayers_state` runtime method.
 pub const FROM_WESTEND_UNREWARDED_RELAYERS_STATE: &str = "FromWestendInboundLaneApi_unrewarded_relayers_state";
 
+/// The length of a session (how often authorities change) on Westend measured in of number of blocks.
+pub const SESSION_LENGTH: BlockNumber = 10 * time_units::MINUTES;
```
tomusdrw, line 61: Note that afair, this doesn't necessarily need to be accurate - the session may finish earlier, so it's rather a max/desired session length.

primitives/header-chain/src/justification.rs
```diff
@@ -59,7 +59,7 @@ pub fn verify_justification<Header: HeaderT>(
 	authorities_set_id: SetId,
 	authorities_set: &VoterSet<AuthorityId>,
 	raw_justification: &[u8],
-) -> Result<(), Error>
+) -> Result<(usize, usize), Error>
```
tomusdrw, line 62: Why not return a `GrandpaJustification` instead and let the caller read values? Two `usize` variables might be confusing.

modules/grandpa/src/lib.rs
```diff
@@ -82,6 +90,24 @@ pub mod pallet {
...
+		#[pallet::constant]
+		type BridgedSessionLength: Get<BridgedBlockNumber<Self, I>>;
+
+		/// The number of validators on the bridged chain.
+		///
+		/// The pallet uses this to bound justification verification since justifications may
+		/// contain up to `BridgedValidatorCount` number of signed `pre-commit` messages which need to
+		/// be verified.
+		#[pallet::constant]
+		type BridgedValidatorCount: Get<u64>;
```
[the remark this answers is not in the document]
tomusdrw: Why isn't it `u32` to prevent casting?

tomusdrw (commented):
Looks good, but I think we can easily be more accurate with `validator count`.
Another possible optimisation would be to have the required number of votes be passed as a param as well.

bin/millau/runtime/src/lib.rs
```diff
@@ -311,18 +311,36 @@ parameter_types! {
...
 pub type RialtoGrandpaInstance = ();
 impl pallet_bridge_grandpa::Config for Runtime {
 	type BridgedChain = bp_rialto::Rialto;
 	type MaxRequests = MaxRequests;
+	type BridgedSessionLength = RialtoSessionLength;
+	type BridgedValidatorCount = RialtoValidatorCount;
+
+	// NOTE: We should technically be using weights generated for the Millau runtime, but since we
+	// don't have those we'll just use the Rialto runtime ones.
+	type WeightInfo = pallet_bridge_grandpa::weights::RialtoWeight<Runtime>;
```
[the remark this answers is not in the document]
HCastano: I left it as a note instead of a TODO because I remembered [this issue](https://github.com/paritytech/parity-bridges-common/issues/390) related to the Millau benches. However, reading the comments again I realize that this should fall under a TODO for #391 instead.

Will update.

primitives/chain-westend/src/lib.rs
HCastano: Not saying that I don't believe you, but can you point to any docs/code where this is the case?

modules/grandpa/src/lib.rs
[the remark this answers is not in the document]
HCastano: > This is quite misleading right now, cause the validator count can be easily changed on the bridged chain and we should really be able to support not significant changes without updating the bridge setup.
>
> Like in case we set this to close to actual values we risk stalling the bridge, so it needs at least 2x buffer.

Good point, I hadn't thought about that

modules/grandpa/src/lib.rs
HCastano: > Either we can read it from the storage (I don't know the exact rules about weights, but it's probably discouraged due to extra state read) 

Yeah, we shouldn't be reading from storage in order to calculate weights.

> Happy to work on this under separate issue though as a weight cost optimisation, since it requires a change in the relayer code as well.

👍

modules/grandpa/src/lib.rs
HCastano: I don't quite follow your math here. The benchmarks are set up to account for the effects of two things: how long it takes to check pre-commits (`p`), and how long it takes to check vote ancestries (`s`). They are used in the weight formula as follows:

```rust
(0 as Weight)
	.saturating_add((3_248_661_000 as Weight).saturating_mul(s as Weight))
	.saturating_add((776_552_000 as Weight).saturating_mul(p as Weight))
```

As you can see these two things can be thought of as independent of one another.

Maybe can you clarify what you mean by "total number of headers in the justification"? Because I don't see how we can get a higher than declared weight by using `votes`.

> BTW, what happens in such case?

I don't think the benchmarks are doing to charge your account more than the declared weight, so I guess it would open up a DoS vector for the chain.

svyatonik (approved):
So when parameter type would be changed to the decoded justification, we'll remove both `MaxBridgedSessionLength` and `MaxBridgedValidatorCount`, right?

modules/grandpa/src/benchmarking.rs
```diff
@@ -0,0 +1,264 @@
...
+
+	}: {
+		crate::find_scheduled_change(&header)
+	}
+
+	// What we want to check here is how long it takes to read and write the authority set tracked
+	// by the pallet as the number of authorities grows.
+	read_write_authority_sets {
+		// The current max target number of validators on Polkadot/Kusama
+		let n in 1..1000;
```
svyatonik, line 219: I don't quite get where you have got values:
```rust
	pub const RialtoValidatorCount: u32 = 255;
	pub const WestendValidatorCount: u32 = 255;
```
then :) And what's we're going to do with Kusama/Polkadot/Westend/Rococo benchmarks?

On Rialto max validators count is 5. Where did you get that value for Westend? I thought that params for all polka-like chains are similar?

primitives/chain-westend/src/lib.rs
tomusdrw: Session is controlled via [`should_session_end`](https://github.com/paritytech/substrate/blob/3472a656d23a8bb0850a074b13deac2a91b2e6e3/frame/session/src/lib.rs#L139) and it can implement arbitrary logic really.

If you look at [BABE implementation](https://github.com/paritytech/substrate/blob/3472a656d23a8bb0850a074b13deac2a91b2e6e3/frame/babe/src/lib.rs#L475), there are some edge cases in case of missed slots, where the session ends a bit ealier than compared to uniform periodic session.

modules/grandpa/src/lib.rs
tomusdrw: >  how long it takes to check vote ancestries (s). 

That's not accurate. `s` is passed as `depth` to `make_justification_for_header` function. Take a look at the code that [generates justifications here](https://github.com/paritytech/parity-bridges-common/blob/a937b074f76a36868e358d222c133318e6087cb5/primitives/test-utils/src/lib.rs#L106).
For every `fork` you create a chain of `depth` blocks and add all of them to `votes_ancestries`. Hence `votes_ancestries ~ forks * depth`.

So the weight calculation is based on the worst case and the interpretation of `s` and `p` should be "maximal length of EACH fork" and "number of forks" accordingly.

Hence if you pass `votes` coming from actual justification you end up over estimating the weight quite significantly - I think you should rather be be passing "average length of EACH fork" calculated as `p = pre_commits.len(); s = (votes.len() + p - 1) / p` to get realistic results.

HCastano:
> So when parameter type would be changed to the decoded justification, we'll remove both `MaxBridgedSessionLength` and `MaxBridgedValidatorCount`, right?

No, I plan to keep them but use them differently. We'll instead do something like:

```rust
ensure!(justification.commit.precommits.len() <= T::MaxBridgedValidatorCount)
```

We may need to change the names/meaning of the bounds a bit, but in general they'll remain

modules/grandpa/src/benchmarking.rs
HCastano: Right now I'm just using the max value allowed for this configuration option for all networks (see #846). Since the max value is above the actual validator number for all supported networks this is fine. The downside is that calls will end up being more expensive initially.

This might change a bit with the changes in #847 so I'll need to revisit this.

modules/grandpa/src/lib.rs
HCastano: Okay I think I see what you mean. Basically, our weight formula is a function of `s` and `p`:

```rust
fn weight(s, p) -> Weight;
```

During the refund, we call `weight(votes, precommits)`, which can be thought of as `weight(s * p, p)`, which is clearly an overestimate of the declared weight.

Since what we really want is `weight(s, p)` you're suggesting that we take the average in order to isolate `s` from `s * p`. 

This makes sense to me now, thanks for spotting this bug!

svyatonik:
> but in general they'll remain

This comment isn't related to this PR - just my thoughts. Imo they aren't required anymore. The only reason to keep them is to add more restrictions? Or is there something else? You'll have indirect by-weight restriction anyway.

modules/grandpa/src/benchmarking.rs
svyatonik: Yes, but you're saying in the comment that for Polkadot/Kusama "max target number of validators" is 1000. So iiuc on Westend it is also 1000. Am I right here? Where did you got this thousand? We'e verifying justifications from Westend => we may overflow this 255 value for Westend justifications => end up with underpaying?

It is fine to keep it as is if you'll just remove these constants in next PRs.

tomusdrw:
> We may need to change the names/meaning of the bounds a bit, but in general they'll remain

Why? I think they are completely unnecessary and will result in excessive, imprecise weight charging. What's the point if we can get exact values right from the parameter that users passes?

modules/grandpa/src/benchmarking.rs
HCastano: > Yes, but you're saying in the comment that for Polkadot/Kusama "max target number of validators" is 1000. So iiuc on Westend it is also 1000.

I don't think so actually. Since there's no incentive to run a validator on Westend nobody is doing it. From what I can tell there are only like 10 validators on Westend, with no "goal" of it reaching 1000.

> We'e verifying justifications from Westend => we may overflow this 255 value for Westend justifications => end up with underpaying?
>
> It is fine to keep it as is if you'll just remove these constants in next PRs.

If Westend had more than 255 validators, yes this would be a concern. However, it doesn't. At the same time I do plan to raise these limits (if we keep them) so that they are above the current target validator set size of 1000.

svyatonik:
> So the reason why I'm hesitant to remove these checks is because there are places in verify_justification which can loop in an unbounded fashion.

Loops are now (not now - but after expanded justification PR) indirectly bounded by the weight formula, because weight formula will depend on both number of forks and precommits, right? So it shouldn't be a problem.

> What I'm afraid of is that somebody is able to craft a valid justifications which consumes all the weight in a block. Full blocks are then propagated to the network and all the network participants are busy churning away just doing this single thing and nothing else. Is this an irrational fear to have?

Your assert will be called **after** transaction will be included into the block. And it doesn't matter whether the justification is valid or not - in any case it'll be included into the block. And weight computed with weight formula will be deducted from the total available block weight, blocking other transactions from being 'mined'. So what happens with this decoded-justification-argument is that we give relayer full control over tx weight.

I guess your concern is valid, but relayer would have to pay for this. AFAIU weights are here to protect chain from this kind of attacks. So basically you could achieve the same by, say, submitting 1000 remark transactions at every block (or some other potentially heavy transactions that occupy either block size or block weight) - there's only tx payment who prevents you from doing that.

tomusdrw:
#847 just landed (sorry @HCastano :)), are you planning to update the code to use the concrete values as part of this PR or separately (I'm honestly fine with either).

HCastano:
> #847 just landed (sorry @HCastano :)), are you planning to update the code to use the concrete values as part of this PR or separately (I'm honestly fine with either).

I think it's best to merge this and fix stuff in a follow up since this PR. It would make the reviewers lives easier

tomusdrw:
> I think it's best to merge this and fix stuff in a follow up since this PR. It would make the reviewers lives easier

Please create an issue or draft PR, so that it doesn't get forgotten :)

PR #818, opened 2021-03-11
Add Parameter type.

tomusdrw:
Draft to confirm it uses correct types.

### DIFF
--- a/deployments/types-millau.json
+++ b/deployments/types-millau.json
@@ -137,9 +137,15 @@
   },
   "Fee": "MillauBalance",
   "Balance": "MillauBalance",
+  "Hash": "MillauBlockHash",
   "BlockHash": "MillauBlockHash",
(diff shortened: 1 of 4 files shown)


svyatonik:
1) the change works fine on Rialto && divisor is `1_000_000_000_000_000_000` as expected (so to update rate to `2.0`, you should pass `2_000_000_000_000_000_000`). Since we have decided to go with updates-via-UI, this should be documented somewhere;
2) I was pretty sure that `CompactAs` [here](https://github.com/paritytech/substrate/blob/master/primitives/arithmetic/src/fixed_point.rs#L342) means that all `Fixed*` types use compact encoding. But apparently not - with type changed to `Compact<U128>`, extrinsic fails with parameter decoding error.

svyatonik (approved):

PR #822, opened 2021-03-12
Added some ML-related constants for polkadot-like chains

svyatonik:
This PR assumes that we can't reference `polkadot-runtime-common` crate.

There are actually only 4 constants: `EXTRA_STORAGE_PROOF_SIZE`, `MAXIMAL_ENCODED_ACCOUNT_ID_SIZE`, `max_extrinsic_weight` and `max_extrinsic_size`. Other stuff is here just to compute last two 'constants'.

From critical/non-critical point of view, constants comprise two groups:
1) `max_extrinsic_weight` and `max_extrinsic_size` - if these values will change (decrease) drastically, message lanes may stuck;
2) `EXTRA_STORAGE_PROOF_SIZE` and `MAXIMAL_ENCODED_ACCOUNT_ID_SIZE` - if these constants increase, then ML-transactions submitter may pay lesser than expected. Also - if `EXTRA_STORAGE_PROOF_SIZE` will drastically increase, then message lanes may also stuck, but this is highly unlikely.

From administrative/monitoring point of view, constants comprise two groups:
1) `MAXIMAL_ENCODED_ACCOUNT_ID_SIZE`, `max_extrinsic_weight` and `max_extrinsic_size` may be changed only during runtime upgrade, so we may just have some kind of pre-upgrade document of what to track (remember that we'll need to recompile and restart relayers anyway, so any upgrade is already painful);
2) `EXTRA_STORAGE_PROOF_SIZE` may only be tracked by (manually or automatically) running some software that points to given chain. Then we may have alerts that will say us that some particular value needs to be updated. Please also note that even though we have 'ML-parameters', which may be updated more often than runtime upgrade happens, I don't think that it worth making this constant a 'ML-parameter'. Reasons: (a) I don't think it'll break barrier too often and (b) even if it'll break barrier, it won't be super critical for us (as opposed to e.g. conversion rate).

There are also two other 'constants' missing from this PR - `TX_EXTRA_BYTES` (needs a test => needs a relayer) and `WeightToFee` (still testing it).

### DIFF
--- a/primitives/polkadot-core/src/lib.rs
+++ b/primitives/polkadot-core/src/lib.rs
@@ -65,31 +65,31 @@ pub const MAXIMAL_ENCODED_ACCOUNT_ID_SIZE: u32 = 32;
 
 /// All Polkadot-like chains allow normal extrinsics to fill block up to 75%.
 ///
-/// This is a copy-paste from `polkadot-runtime-common` crate.
+/// This is a copy-paste from the Polkadot repo's `polkadot-runtime-common` crate.
 const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(75);
 
 /// All Polkadot-like chains allow 2 seconds of compute with a 6 second average block time.
 ///
-/// This is a copy-paste from `polkadot-runtime-common` crate.
+/// This is a copy-paste from the Polkadot repo's `polkadot-runtime-common` crate.
 pub const MAXIMUM_BLOCK_WEIGHT: Weight = 2 * WEIGHT_PER_SECOND;
 
 /// All Polkadot-like chains assume that an on-initialize consumes 1% of the weight on average,
 /// hence a single extrinsic will not be allowed to consume more than `AvailableBlockRatio - 1%`.
 ///
-/// This is a copy-paste from `polkadot-runtime-common` crate.
+/// This is a copy-paste from the Polkadot repo's `polkadot-runtime-common` crate.
 pub const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(1);
 
 parameter_types! {
 	/// All Polkadot-like chains have maximal block size set to 5MB.
 	///
-	/// This is a copy-paste from `polkadot-runtime-common` crate.
+	/// This is a copy-paste from the Polkadot repo's `polkadot-runtime-common` crate.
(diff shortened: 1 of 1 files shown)


primitives/polkadot-core/src/lib.rs
```diff
@@ -18,18 +18,111 @@
...
+pub const EXTRA_STORAGE_PROOF_SIZE: u32 = 1024;
+
+/// Maximal size (in bytes) of encoded (using `Encode::encode()`) account id.
+///
+/// All polkadot-like chains are using same crypto.
+pub const MAXIMAL_ENCODED_ACCOUNT_ID_SIZE: u32 = 32;
+
+/// All Polkadot-like chains allow normal extrinsics to fill block up to 75%.
+///
+/// This is a copy-paste from `polkadot-runtime-common` crate.
```
HCastano, line 68:
```suggestion
/// This is a copy-paste from the Polkadot repo's `polkadot-runtime-common` crate.
```

I'd add this for all the consts which refer to `polkadot-runtime-common` just to make it clear that we're referring to something that's not part of the bridges repo

primitives/polkadot-core/src/lib.rs
```diff
@@ -18,18 +18,111 @@
...
+/// Polkadot-like chain. This mostly depends on number of entries in the storage trie.
+/// Some reserve is reserved to account future chain growth.
+///
+/// To compute this value, we've synced Kusama chain blocks [0; 6545733] to see if there were
+/// any significant changes of the storage proof size (NO):
+///
+/// - at block 3072 the storage proof size overhead was 579 bytes;
+/// - at block 2479616 it was 578 bytes;
+/// - at block 4118528 it was 711 bytes;
+/// - at block 6540800 it was 779 bytes.
```
HCastano, line 51: Nice work with this. If we ever want to check and update these in the future, will we basically have to do what you did and sync and node and check the storage overhead manually? 

Is this a check that could potentially be built into relayers?

HCastano (approved):

primitives/polkadot-core/src/lib.rs
svyatonik: Yes. Either relayers, or some other tool that will generate warnings for us.

tomusdrw (approved):
lgtm! Also created https://github.com/paritytech/parity-bridges-common/issues/826

PR #823, opened 2021-03-12
Make Test Helpers `no_std` Compatible

HCastano:
I've been struggling to get #815 compiling. Today I realized that my problem was my use
of `bp-test-utils` to create valid, but dummy, justifications in the "setup" phase of the
benchmarks via `bp-test-utils::make_justification_for_header()`. The problem here is that
`bp-test-utils` is a `std` only crate, and benchmarks need to be able to compile in a
`no_std` environment.

The reason that `bp-test-utils` is `std` only is `sp-keyring`, which is used while
signing pre-commit messages. Right now I'm trying to replace `sp-keyring` with
`sp-application-crypto` since that can compile in `no_std`. Currently running into the
following error while running tests:

```
"No `keystore` associated for the current context!"
```

Another idea I had for solving this was trying to make `sp-keyring` compile to `no_std`
by changing a few data structure and hiding some functionalily behind feature flags, but
I'm not sure if that's a good rabbit hole to go down.

Any ideas for how to solve this would be helpful!

### DIFF
--- a/primitives/test-utils/src/lib.rs
+++ b/primitives/test-utils/src/lib.rs
@@ -51,13 +51,11 @@ pub fn make_justification_for_header<H: HeaderT>(
 	// I'm using the same header for all the voters since it doesn't matter as long
 	// as they all vote on blocks _ahead_ of the one we're interested in finalizing
 	for (id, _weight) in authorities.iter() {
-		let signer = extract_keyring(&id);
-		let precommit = signed_precommit::<H>(
-			signer,
-			(precommit_header.hash(), *precommit_header.number()),
-			round,
-			set_id,
-		);
+		// use sp_runtime::RuntimeAppPublic;
+		// let signed_payload = id.sign(&vec![1u8]);
+
+		// let signer = extract_keyring(&id);
+		let precommit = signed_precommit::<H>(id, (precommit_header.hash(), *precommit_header.number()), round, set_id);
 		precommits.push(precommit);
 		votes_ancestries.push(precommit_header.clone());
 	}
@@ -74,22 +72,26 @@ pub fn make_justification_for_header<H: HeaderT>(
 }
 
 fn signed_precommit<H: HeaderT>(
-	signer: Ed25519Keyring,
+	signer: &AuthorityId,
 	target: (H::Hash, H::Number),
 	round: u64,
 	set_id: SetId,
 ) -> finality_grandpa::SignedPrecommit<H::Hash, H::Number, AuthoritySignature, AuthorityId> {
+	use sp_runtime::RuntimeAppPublic;
+
 	let precommit = finality_grandpa::Precommit {
 		target_hash: target.0,
 		target_number: target.1,
 	};
 	let encoded =
 		sp_finality_grandpa::localized_payload(round, set_id, &finality_grandpa::Message::Precommit(precommit.clone()));
-	let signature = signer.sign(&encoded[..]).into();
+
+	let signature = signer.sign(&encoded).expect("TODO"); // .into();
+
 	finality_grandpa::SignedPrecommit {
 		precommit,
 		signature,
-		id: signer.public().into(),
+		id: signer.clone(), // signer.public().into(),
 	}
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -671,6 +671,7 @@ dependencies = [
  "serde",
  "sp-core",
  "sp-finality-grandpa",
+ "sp-io",
  "sp-runtime",
  "sp-std",
 ]
@@ -788,6 +789,7 @@ version = "0.1.0"
 dependencies = [
  "bp-header-chain",
  "finality-grandpa 0.14.0",
+ "sp-application-crypto",
  "sp-finality-grandpa",
  "sp-keyring",
  "sp-runtime",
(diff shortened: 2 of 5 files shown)


svyatonik:
I can't say if it'll be possible to enable keystore in benchmarks environment, but alternatively you may just use some no_std-enabled crate that is able to generate ed25519 signatures (I did that for PoA pallet). Or just pregenerate (and hardcode) some headers+justifications, if you don't need many.

tomusdrw:
Using application crypto in WASM context (i.e. within `TestExecution`) will make host calls instead of performing crypto operations in WASM, hence you need to register `KeystoreExt` (see [example](https://github.com/paritytech/substrate/blob/aae186113c17386f54b087416c0dc41451217ba4/bin/node/executor/tests/submit_transaction.rs#L122))

Making `sp-keyring` to compile for `no_std` is IMHO not a good idea, I think we should rather follow what Slava suggested, i.e.:
1. Either using no_std-compatible crypto libraries directly.
2. or hardcoding the signatures

HCastano:
Thanks for the suggestions guys. I'm still a bit stuck though.

> Or just pregenerate (and hardcode) some headers+justifications, if you don't need many.

This won't work because of the planned usage in benchmarking. I want to see how the number of authorities (and thus the number of signed pre-commits and vote ancestries) affects justification verification time. For this I want to create a good number of justifications and see how it scales.

> alternatively you may just use some no_std-enabled crate that is able to generate ed25519 signatures (I did that for PoA pallet)

I tried using `ed25519_dalek` today but I ran into issues with the Public Key and Signature types. During the [pre-commit generation](https://github.com/paritytech/parity-bridges-common/blob/master/primitives/test-utils/src/lib.rs#L76) step I end up with `SignedPrecommit<..., ed25519_dalek::Signature, ed25519_dalek::PublicKey>`. This is problematic because the `GrandpaJustification` type we use uses [application crypto specific signatures and public keys](https://github.com/paritytech/parity-bridges-common/blob/master/primitives/header-chain/src/justification.rs#L138), and I can't change these to the `ed25519_dalek` equivalents since those don't implement things like `codec::{Encode, Decode}`. Ideally I'd like some way to wrap/convert the dalek types into the application crypto specific types so I can return the correct type of `SignedPrecommit`.

For the Public Key I can do something like:

```rust
let alice = ed25519_dalek::Keypair::from_bytes(&[1; 32]).unwrap();
// ... use `alice` for signing ...
use sp_application_crypto::Public;
let id = sp_finality_grandpa::AuthorityId::from_slice(&alice.public.to_bytes());
// Use `id` in `SignedPrecommit`
```

But I haven't been able to find a way to do that with the Signature field.

tomusdrw:
I'm pretty sure that all crypto primitives implement `UncheckedInto` - so if you have a fixed array you can `unchecked_into()` that into desired `Public/Private/Signature`.
Also looking at the app crypto implementation [`Signature` implements `TryFrom<Vec<u8>>`](https://github.com/paritytech/substrate/blob/aae186113c17386f54b087416c0dc41451217ba4/primitives/application-crypto/src/lib.rs#L474).

In case there are no other options, you can always simply convert from `ed25519_dalek` into raw `Vec<u8>` and then `Decode` the app-crypto from that.

svyatonik (approved):

primitives/test-utils/src/lib.rs
```diff
@@ -132,20 +179,25 @@ pub fn voter_set() -> VoterSet<AuthorityId> {
...
+	Keyring::Bob
 }
 
-/// Get the Public key of the Charlie test account.
-pub fn charlie() -> AuthorityId {
-	Ed25519Keyring::Charlie.public().into()
+/// Convenience function to get a handle to the Charlie test account.
+pub fn charlie() -> Keyring {
+	Keyring::Charlie
 }
```
tomusdrw, line 203: Is it that much of a convenience? You can actually do `use Keyring::*` and simply use `Alice` instead of `alice()`.

primitives/test-utils/src/lib.rs
```diff
@@ -118,11 +123,53 @@ pub fn header_id<H: HeaderT>(index: u8) -> (H::Hash, H::Number) {
...
+	}
+}
+
+impl From<Keyring> for AuthorityId {
+	fn from(k: Keyring) -> Self {
+		AuthorityId::from_slice(&k.public().to_bytes())
+	}
+}
+
+impl From<&Keyring> for AuthorityId {
```
tomusdrw, line 169: `Keyring` is a Copy type so probably all conversions work anyway without this impl, don't they?

primitives/test-utils/src/lib.rs
```diff
@@ -118,11 +123,53 @@ pub fn header_id<H: HeaderT>(index: u8) -> (H::Hash, H::Number) {
...
+	fn from(k: &Keyring) -> Self {
+		AuthorityId::from_slice(&k.public().to_bytes())
```
tomusdrw, line 171:
```suggestion
		(*k).into()
```
In case this is actually needed (See other comment).

tomusdrw (approved):
lgtm!

primitives/test-utils/src/lib.rs
HCastano: I kept it more to reduce the changes in the existing code, since `alice()` and friends were already being used. Using the variants directly doesn't sound bad though

primitives/test-utils/src/lib.rs
HCastano: Yeah good point

Issue #819, opened 2021-03-11
Sync Rococo/Westend to Millau/Rialto


tomusdrw:
Add relayer for Rococo/Westend >Millau/Rialto and deploy it as part of the test network.

The relayer should use public Rococo/Westend node and perform initialisation starting from the current best block followed by one-way sync of headers.

svyatonik:
Westend -> Millau sync will start tomorrow

Issue #788, opened 2021-03-03
Use SetId Field from ConsensusLog


HCastano:
Sometime in the future GRANDPA is going to start including the `set_id` it expects in the
`ScheduledChange` digests it issues. We should use this field instead of manually
computing the expected `set_id` ourselves.

With this change we'll also be able to check that `set_id` is actually increasing
monotonically, ensuring that authority set handoffs are happening correctly.

adoerr:
You have an issue or code permalink where this GRANDPA change could be tracked? This is useful for BEEFY as well.

HCastano:
> You have an issue or code permalink where this GRANDPA change could be tracked? This is useful for BEEFY as well.

@andresilva do you have an issue made, or is it just in your head at the moment?

tomusdrw:
> I am thinking of changing this to ForcedChange(SetId, N, ScheduledChange<N>). In which case, for forced changes we'd unconditionally apply the set_id provided in the digest rather than doing + 1. We'd still keep the old variant as OldForcedChange(N, ScheduledChange<N>) since we need this for backwards compatibility.

Sounds like a good idea to me. Indeed if the `set_id` is part of the signed payload in justification this should be fine.

PR #824, opened 2021-03-13
Sync Westend to Millau

svyatonik:
related: #819 (need to add this into our test deployments in a separate PR)

~Right now it doesn't work, because of Millau state (it had forced changes && there's some issue with that), so probably I'll need a way how to receive current GRANDPA set id before marking it ready for review. There's also couple of things that needs to be done.~

The main change (apart from supporting instances in `finality-verifier` pallet, which I have extracted in a separate PR) here is the way how we initialize bridge pallet. Until recently we have had support for scheduled GRANDPA changes in the bridge pallet, so this operation (initialization) on non-test chains had to be a manual (and very dangerous) process - we had to select finalized header and guarantee that there's no active scheduled change that will be enacted at one of its descendants. But now we know that (currently and in the future) regular changes are always enacted at the block where they're scheduled. So we don't need to do header selection manually - we may take any finalized header && start syncing with it. Since we are not going to sync any chains from the genesis block, it now makes sense to always start with best finalized header => I've removed some cli parameters.

The only problem that is left is authorities set id - we can call `GrandpaApi::grandpa_authorities()` to get authorities ids, but there's no similar call for the set id. There's `pallet_grandpa::Current`, but it returns wrong value on Westend (it'll be fixed, but we can't wait this fix). Another option is to get GRANDPA round state using RPC call, but there are two problems: (1) if there has been change recently, it may return `set_id + 1` and we can't verify it and (2) by some reason fields of structs returned by this RPC are private, so we either need to deal with raw JSON, or copy all structs to our code. So I chose another option - we're getting justification from the stream and then trying to guess `set_id` with brutefoce - it is quite fast (5s to guess 3395 value on Westend).

All other changes are trivial - copypasting client code && similar.

### DIFF
--- a/relays/substrate/src/headers_initialize.rs
+++ b/relays/substrate/src/headers_initialize.rs
@@ -26,6 +26,7 @@ use pallet_finality_verifier::InitializationData;
 use relay_substrate_client::{Chain, Client};
 use sp_core::Bytes;
 use sp_finality_grandpa::{AuthorityList as GrandpaAuthoritiesSet, SetId as GrandpaAuthoritiesSetId};
+use sp_runtime::traits::Header as HeaderT;
 
 /// Submit headers-bridge initialization transaction.
 pub async fn initialize<SourceChain: Chain, TargetChain: Chain>(
@@ -34,6 +35,7 @@ pub async fn initialize<SourceChain: Chain, TargetChain: Chain>(
 	raw_initial_header: Option<Bytes>,
 	raw_initial_authorities_set: Option<Bytes>,
 	initial_authorities_set_id: Option<GrandpaAuthoritiesSetId>,
+	authorities_set_id_key: Vec<u8>,
 	prepare_initialize_transaction: impl FnOnce(InitializationData<SourceChain::Header>) -> Result<Bytes, String>,
 ) {
 	let result = do_initialize(
@@ -42,6 +44,7 @@ pub async fn initialize<SourceChain: Chain, TargetChain: Chain>(
 		raw_initial_header,
 		raw_initial_authorities_set,
 		initial_authorities_set_id,
+		authorities_set_id_key,
 		prepare_initialize_transaction,
 	)
 	.await;
@@ -71,6 +74,7 @@ async fn do_initialize<SourceChain: Chain, TargetChain: Chain>(
 	raw_initial_header: Option<Bytes>,
 	raw_initial_authorities_set: Option<Bytes>,
 	initial_authorities_set_id: Option<GrandpaAuthoritiesSetId>,
+	authorities_set_id_key: Vec<u8>,
 	prepare_initialize_transaction: impl FnOnce(InitializationData<SourceChain::Header>) -> Result<Bytes, String>,
 ) -> Result<TargetChain::Hash, String> {
 	let initialization_data = prepare_initialization_data(
@@ -78,8 +82,17 @@ async fn do_initialize<SourceChain: Chain, TargetChain: Chain>(
 		raw_initial_header,
 		raw_initial_authorities_set,
 		initial_authorities_set_id,
+		authorities_set_id_key,
 	)
 	.await?;
+
+	log::info!(
+		target: "bridge",
+		"Trying to initialize {}-headers bridge. Initialization data: {:?}",
+		SourceChain::NAME,
+		initialization_data,
+	);
+
 	let initialization_tx = prepare_initialize_transaction(initialization_data)?;
 	let initialization_tx_hash = target_client
 		.submit_extrinsic(initialization_tx)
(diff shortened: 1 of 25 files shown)


relays/substrate/src/headers_initialize.rs
```diff
@@ -88,50 +82,134 @@ async fn do_initialize<SourceChain: Chain, TargetChain: Chain>(
...
-	let initial_authorities_set =
-		GrandpaAuthoritiesSet::decode(&mut &raw_initial_authorities_set[..]).map_err(|err| {
-			format!(
-				"Failed to decode {} initial authorities set: {:?}",
-				SourceChain::NAME,
-				err
-			)
-		})?;
+	// In ideal world we just need to get best finalized header and then to read GRANDPA authorities
+	// set (`pallet_grandpa::CurrentSetId` + `GrandpaApi::grandpa_authorities()`) at this header.
```
tomusdrw, line 90: This should also be solved by https://github.com/paritytech/parity-bridges-common/issues/788

relays/substrate/src/headers_initialize.rs
```diff
@@ -88,50 +82,134 @@ async fn do_initialize<SourceChain: Chain, TargetChain: Chain>(
...
+			initial_authorities_set_id,
+			&authorities_for_verification,
+			&justification.0,
+		)
+		.is_ok();
+		if is_valid_set_id {
+			break;
+		}
+
+		initial_authorities_set_id += 1;
```
tomusdrw, line 161: Should it have some upper bound or fail instead of wrapping (which will probably take forever anyway)?

relays/substrate/src/rialto_millau/cli.rs
```diff
@@ -47,6 +47,17 @@ pub enum RelayHeaders {
 		#[structopt(flatten)]
 		prometheus_params: PrometheusParams,
 	},
+	/// Relay Westend headers to Millau.
+	WestendToMillau {
```
tomusdrw, line 51: I was imagining this will be part of a separate `westend_millau` module, but I can refactor this later on.

tomusdrw (approved):
LGTM! We still need deployment as mentioned in the description.

relays/substrate/src/headers_initialize.rs
svyatonik: Not actually.  The pallet needs to be initialized with set id - otherwise you won't be able to verify most justifications. You can try to find `set_id` by inspecting initial header' ancestors. But reading (possibly many) headers using our public RPC nodes isn't a best idea - I was trying to sync Westend from the genesis on Friday && relay was unable to even submit single transaction - it was able to read ~7000 blocks in 20 minutes, then connection has been lost && it has restarted. So ideally we'll be reading `CurrentSetId`, even if `set_id` will be added to the digest.

relays/substrate/src/headers_initialize.rs
svyatonik: I was writing that the upper bound is too different for different chains => you can't have any hardcoded value. But actually there can't be more changes than blocks => `initial_block_number` may be used as an upper bound

tomusdrw (approved):

relays/substrate/src/headers_initialize.rs
```diff
@@ -88,50 +83,153 @@ async fn do_initialize<SourceChain: Chain, TargetChain: Chain>(
...
+		.await
+		.map_err(|err| format!("Failed to subscribe to {} justifications: {:?}", SourceChain::NAME, err))?;
+
+	// Read next justification - the header that it finalizes will be used as initial header.
+	let justification = justifications.next().await.ok_or_else(|| {
+		format!(
+			"Failed to read {} justification from the stream: stream has ended unexpectedly",
+			SourceChain::NAME,
+		)
+	})?;
```
HCastano, line 107: Would it make sense to have a retry mechanism here?

relays/substrate/src/headers_initialize.rs
```diff
@@ -88,50 +83,153 @@ async fn do_initialize<SourceChain: Chain, TargetChain: Chain>(
...
+	let initial_authorities_set = source_authorities_set(&source_client, initial_header_hash).await?;
+	log::trace!(target: "bridge", "Selected {} initial authorities set: {:?}",
+		SourceChain::NAME,
+		initial_authorities_set,
+	);
+
+	// If initial header changes the GRANDPA authorities set, then we need previous authorities
+	// to verify justification.
+	let mut authorities_for_verification = initial_authorities_set.clone();
+	let schedules_change = find_grandpa_authorities_scheduled_change(&initial_header).is_some();
```
HCastano, line 130: Do you think it would be worth it to check for forced changes?

If the initial header schedules forced changes with a delay of `N` blocks the finality pallet has no way to knowing that header `initial + N` will enact a new authority set change and I think we'll get stuck

relays/substrate/src/headers_initialize.rs
```diff
@@ -88,50 +83,153 @@ async fn do_initialize<SourceChain: Chain, TargetChain: Chain>(
...
+
+		initial_authorities_set_id += 1;
+		min_possible_block_number += One::one();
+		if min_possible_block_number > initial_header_number {
+			// there can't be more authorities set changes than headers => if we have reached `initial_block_number`
+			// and still have not found correct value of `initial_authorities_set_id`, then something
+			// else is broken => fail
+			return Err(format!(
+				"Failed to guess initial {} GRANDPA authorities set id: checked all\
+			possible ids in range [0; {}]",
```
HCastano, line 177:
```suggestion
				possible ids in range [0; {}]",
```

HCastano (approved):

relays/substrate/src/headers_initialize.rs
svyatonik: It's not about what this PR is about - if we'll ever change initialization approach from "just submit transaction" to "check that pallet is not initialized, submit transaction and wait until it's mined", then retry will make sense. This particular error can't happen in practice, because we need to miss 4096 justifications and we're reading just after subscription.

PR #825, opened 2021-03-13
Make finality verifier pallet instantiable

### LINKED ISSUE #360: Make Substrate Pallet Instantiable

svyatonik:
closes #360 

I needed that for #824

### DIFF
--- a/modules/finality-verifier/src/lib.rs
+++ b/modules/finality-verifier/src/lib.rs
@@ -55,13 +55,13 @@ mod mock;
 pub use pallet::*;
 
 /// Block number of the bridged chain.
-pub type BridgedBlockNumber<T> = BlockNumberOf<<T as Config>::BridgedChain>;
+pub type BridgedBlockNumber<T, I> = BlockNumberOf<<T as Config<I>>::BridgedChain>;
 /// Block hash of the bridged chain.
-pub type BridgedBlockHash<T> = HashOf<<T as Config>::BridgedChain>;
+pub type BridgedBlockHash<T, I> = HashOf<<T as Config<I>>::BridgedChain>;
 /// Hasher of the bridged chain.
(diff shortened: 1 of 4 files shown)


modules/finality-verifier/src/lib.rs
```diff
@@ -70,7 +70,7 @@ pub mod pallet {
 	use frame_system::pallet_prelude::*;
 
 	#[pallet::config]
-	pub trait Config: frame_system::Config {
+	pub trait Config<I: 'static = ()>: frame_system::Config {
```
MaciejBaj, line 73: is `I: 'static = ()` equal `I: DefaultInstance`?

svyatonik: I guess so :) From what I've learned from this new pallet declaration macros docs, now 'instance' should only be `'static`. Previously, with `decl_module! {}`, it had to implement `Instance` trait. I've been copypasting this from the `balances` pallet, which atm is the only pallet that uses this new pallet declaration syntax and supports multiple instances. There, `()` is used. Tbh, I'm not even sure if `DefaultInstance` is still being generated by the `#[pallet]` :)

MaciejBaj: cool, thanks a lot!

tomusdrw (approved):

modules/finality-verifier/src/lib.rs
HCastano: So according to Gui `DefaultInstance` is not being supported anymore, and `()` should be used as the "default".

HCastano (approved):

Issue #817, opened 2021-03-11
Headers + messages relay


svyatonik:
Since we now have finality verifier pallet that doesn't require all source headers to be imported, there's a possible optimization for relayers. It is to spawn headers relay only when messages relay has something to deliver. This has two main pros:
1) we'll need to submit less transactions, meaning less storage entries at target chains, less pruning calls, etc;
2) we won't be spending too much funds on non-required transactions.

Otoh, standalone headers and messages relay may still be required (for other bridge apps or in other incentivization schemas). So I see this merged relay as optional mode for messages relay.

PR #828, opened 2021-03-16
Update jsonrpsee reference

svyatonik:
related: https://github.com/paritytech/parity-bridges-common/pull/787#discussion_r590299503 and #817


tomusdrw (approved):

niklasad1:
@svyatonik 

I hope to release plan to release another version today but blocked on https://github.com/paritytech/jsonrpsee/pull/243 we'll see how it goes:
- It will introduce a new error variant, that you can use to figure out if the background task has been terminated
- Makes the async methods cancel-safe
- Makes the `WsClient` !Clone

PR #830, opened 2021-03-16
Add multiplier parameter to transaction_payment function

svyatonik:
This change is required for all Polkadot-like chains

### DIFF
--- a/bin/millau/runtime/src/rialto_messages.rs
+++ b/bin/millau/runtime/src/rialto_messages.rs
@@ -37,7 +37,7 @@ use sp_std::{convert::TryFrom, ops::RangeInclusive};
 
 parameter_types! {
 	/// Rialto to Millau conversion rate. Initially we treat both tokens as equal.
-	storage RialtoToMillauConversionRate: FixedU128 = 1.into();
+	storage RialtoToMillauConversionRate: FixedU128 = FixedU128::one();
 }
 
 /// Storage key of the Millau -> Rialto message in the runtime storage.
@@ -146,9 +146,10 @@ impl messages::ThisChainWithMessageLanes for Millau {
 
 	fn transaction_payment(transaction: MessageLaneTransaction<Weight>) -> bp_millau::Balance {
 		// in our testnets, both per-byte fee and weight-to-fee are 1:1
-		messages::transaction_payment_without_multiplier(
+		messages::transaction_payment(
 			bp_millau::BlockWeights::get().get(DispatchClass::Normal).base_extrinsic,
 			1,
+			FixedU128::zero(),
 			|weight| weight as _,
 			transaction,
 		)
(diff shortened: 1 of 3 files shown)


bin/runtime-common/src/messages.rs
```diff
@@ -160,9 +166,8 @@ pub fn transaction_payment_without_multiplier<Balance: AtLeast32BitUnsigned>(
 	let len_fee = per_byte_fee.saturating_mul(Balance::from(transaction.size));
 
 	// the adjustable part of the fee
-	//
-	// here we assume that the fee multiplier is zero, so this part is also always zero
-	let adjusted_weight_fee = Balance::zero();
+	let unadjusted_weight_fee = weight_to_fee(transaction.dispatch_weight);
+	let adjusted_weight_fee = multiplier.saturating_mul_int(unadjusted_weight_fee);
```
tomusdrw, line 170: Can you add a test for this function or add one that set's non-zero multiplier?

bin/millau/runtime/src/rialto_messages.rs
```diff
@@ -146,9 +146,10 @@ impl messages::ThisChainWithMessageLanes for Millau {
 
 	fn transaction_payment(transaction: MessageLaneTransaction<Weight>) -> bp_millau::Balance {
 		// in our testnets, both per-byte fee and weight-to-fee are 1:1
-		messages::transaction_payment_without_multiplier(
+		messages::transaction_payment(
 			bp_millau::BlockWeights::get().get(DispatchClass::Normal).base_extrinsic,
 			1,
+			FixedU128::zero(),
```
tomusdrw, line 152: Is there a way how we can test non-zero values on rialto<>millau? What does the multiplier depend on? Is it block utilisation?

svyatonik: > Is there a way how we can test non-zero values on rialto<>millau?

No

> What does the multiplier depend on? Is it block utilisation?

This is defined by `pallet_transaction_payment::FeeMultiplierUpdate`. In P+K if block has weight > 25% of its maximal weight, then this multiplier grows. If less than 25%, then it drops (to some minimal value).

bin/runtime-common/src/messages.rs
```diff
@@ -1401,4 +1406,38 @@ mod tests {
...
+				100,
+				10,
+				FixedU128::zero(),
+				|weight| weight,
+				MessageLaneTransaction {
+					size: 50,
+					dispatch_weight: 777
+				},
+			),
+			100 + 50 * 10,
```
HCastano, line 1423: nit: Having something like `base_weight + size * per_byte_fee` here would be nice

HCastano (approved):

bin/runtime-common/src/messages.rs
svyatonik: Then it'll be duplicate of the `transaction_payment` itself :) What it does is simply gives names to these numbers

PR #831, opened 2021-03-17
 Westend->Millau headers relay in deployments

svyatonik:
I've found that we're not able to run multiple bridges at once. Is that something that hasn't been added because we didn't need that (but actually we did - why we're not running both bridges at test-deployments?) or there are some other concerns?


deployments/bridges/westend-millau/docker-compose.yml
```diff
@@ -0,0 +1,40 @@
...
+    volumes:
+      - ./bridges/westend-millau/dashboard/prometheus/prometheus.yml:/etc/prometheus/prometheus-westend-millau.yml
+    depends_on:
+      - relay-headers-westend-to-millau
+
+networks:
+  deployments_external:
+    driver: bridge
+  deployments_default:
+    external: true
```
HCastano, line 40: Why do we need to specify networking options for this Compose deployment but not the others?

bin/millau/node/src/chain_spec.rs
```diff
@@ -176,6 +176,13 @@ fn testnet_genesis(
 				.map(|x| (x.0.clone(), x.0.clone(), session_keys(x.1.clone(), x.2.clone())))
 				.collect::<Vec<_>>(),
 		}),
+		pallet_finality_verifier_Instance1: Some(BridgeWestendFinalityVerifierConfig {
+			// for our deployments to avoid multiple same-nonces transactions:
+			// //Alice is already used to initialize Rialto<->Millau bridge
+			// => let's use //George to initialize Westend->Millau bridge
```
HCastano, line 182:
```suggestion
			// `//Alice` is already used to initialize Rialto<->Millau bridge
			// => let's use `//George` to initialize Westend->Millau bridge
```

deployments/bridges/westend-millau/entrypoints/relay-headers-westend-to-millau-entrypoint.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+set -xeu
+
+sleep 3
+curl -v http://millau-node-alice:9933/health
```
HCastano, line 5:
```suggestion
curl -v http://millau-node-alice:9933/health
curl -v https://westend-rpc.polkadot.io/health
```

deployments/run.sh
```diff
@@ -28,9 +28,13 @@ function show_help () {
   echo "Usage:"
   echo "  ./run.sh poa-rialto [stop|update]          Run PoA <> Rialto Networks & Bridge"
   echo "  ./run.sh rialto-millau [stop|update]       Run Rialto <> Millau Networks & Bridge"
+  echo "  ./run.sh westend-millau [stop|update]      Run Westend -> Millau Networks & Bridge"
   echo " "
   echo "Options:"
   echo "  --no-monitoring                            Disable monitoring"
+  echo " "
+  echo "You can start multiple bridges at once by passing several bridge names:"
+  echo "  ./run.sh poa-rialto rialto-millau westend-millau [stop|update]"
```
HCastano, line 37: Can you add this to the comments at the top of the script as well?

HCastano (commented):
Currently trying to sync Westend -> Millau, but it seems to be taking a long time. Gonna leave it running and see how it goes before approving

deployments/bridges/poa-rialto/docker-compose.yml
```diff
@@ -79,12 +81,14 @@ services:
...
-      - ./bridges/poa-rialto/dashboard/prometheus/:/etc/prometheus/
+      - ./bridges/poa-rialto/dashboard/prometheus/prometheus.yml:/etc/prometheus/prometheus-poa-rialto.yml
     depends_on: *all-nodes
 
   grafana-dashboard:
     volumes:
-       - ./bridges/poa-rialto/dashboard/grafana/provisioning/:/etc/grafana/provisioning/
+       - ./bridges/poa-rialto/dashboard/grafana/relay-poa-to-rialto-exchange-dashboard.json:/etc/grafana/provisioning/dashboards/relay-poa-to-rialto-exchange-dashboard.json
+       - ./bridges/poa-rialto/dashboard/grafana/relay-poa-to-rialto-headers-dashboard.json:/etc/grafana/provisioning/dashboards/relay-poa-to-rialto-headers-dashboard.json
+       - ./bridges/poa-rialto/dashboard/grafana/relay-rialto-to-poa-headers-dashboard.json:/etc/grafana/provisioning/dashboards/relay-rialto-to-poa-headers-dashboard.json
```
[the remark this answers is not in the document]
svyatonik: Yes. When you run multiple bridges, the folder is populated by several compose files

.gitignore
```diff
@@ -5,6 +5,7 @@
 hfuzz_target
 hfuzz_workspace
 **/Cargo.lock
+deployments/monitoring/grafana/provisioning/dashboards/*.json
```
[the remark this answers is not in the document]
svyatonik: https://github.com/paritytech/parity-bridges-common/pull/831#discussion_r596087706 . The files are physically copied to the monitoring/grafana/provisioning/dashboards/ folder

deployments/bridges/rialto-millau/dashboard/prometheus/prometheus.yml
```diff
@@ -1,23 +1,5 @@
-scrape_configs:
-  # The job name is added as a label `job=<job_name>` to any timeseries scraped from this config.
-  - job_name: 'millau_to_rialto_headers_relay_node'
-
-    # Override the global default and scrape targets from this job every 15 seconds.
-    scrape_interval: 15s
-    static_configs:
-      - targets: ['relay-headers-millau-to-rialto:9616']
-
-  - job_name: 'rialto_to_millau_headers_relay_node'
```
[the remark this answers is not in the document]
svyatonik: This file can't contain jobs - it only can contain targets. There's now a single job defined in prometheus.yml from monitoring

deployments/bridges/westend-millau/docker-compose.yml
svyatonik: Because it needs access to external node (so needs access to both defaul network and bridged network)

deployments/bridges/rialto-millau/dashboard/prometheus/prometheus.yml
```diff
@@ -1,23 +1,5 @@
-scrape_configs:
-  # The job name is added as a label `job=<job_name>` to any timeseries scraped from this config.
-  - job_name: 'millau_to_rialto_headers_relay_node'
-
-    # Override the global default and scrape targets from this job every 15 seconds.
-    scrape_interval: 15s
```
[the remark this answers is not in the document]
svyatonik: https://github.com/paritytech/parity-bridges-common/pull/831/files#diff-2e8699b5f79644c305be1283d1c00c267d12adaa39e1c17cc6ca8ff594c61de3R2

deployments/bridges/westend-millau/entrypoints/relay-headers-westend-to-millau-entrypoint.sh
svyatonik: This command leaves errors in logs. So even though it returns health of the node, there are some lines with "error: " inside

svyatonik:
> Update: I've had it running for an hour or two and it still hasn't synced a single header. I'll leave it running overnight and see if it just take a really long time to sync

That's probably because you're using images from the hub?

deployments/bridges/poa-rialto/docker-compose.yml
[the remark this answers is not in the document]
tomusdrw: AFAIU the `volumes` definitions get's merged, so that we need multiple sources, but listing all the files here is sub-optimal. Can we have a (sub-) folder here instead?

.gitignore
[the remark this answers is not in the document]
tomusdrw: Why do we mount the entire folder though? Can we only add a sub-folder to make sure that there are no local modifications done by the container? Or maybe there is a way to mount read-only?

deployments/bridges/rialto-millau/docker-compose.yml
```diff
@@ -85,9 +87,12 @@ services:
...
-       - ./bridges/rialto-millau/dashboard/grafana/provisioning/:/etc/grafana/provisioning/
+       - ./bridges/rialto-millau/dashboard/grafana/relay-millau-to-rialto-headers-dashboard.json:/etc/grafana/provisioning/dashboards/relay-millau-to-rialto-headers-dashboard.json
+       - ./bridges/rialto-millau/dashboard/grafana/relay-millau-to-rialto-messages-dashboard.json:/etc/grafana/provisioning/dashboards/relay-millau-to-rialto-messages-dashboard.json
+       - ./bridges/rialto-millau/dashboard/grafana/relay-rialto-to-millau-headers-dashboard.json:/etc/grafana/provisioning/dashboards/relay-rialto-to-millau-headers-dashboard.json
+       - ./bridges/rialto-millau/dashboard/grafana/relay-rialto-to-millau-messages-dashboard.json:/etc/grafana/provisioning/dashboards/relay-rialto-to-millau-messages-dashboard.json
 
   prometheus-metrics:
     volumes:
-      - ./bridges/rialto-millau/dashboard/prometheus/:/etc/prometheus/
+      - ./bridges/rialto-millau/dashboard/prometheus/prometheus.yml:/etc/prometheus/prometheus-rialto-millau.yml
```
tomusdrw, line 97:
```suggestion
      - ./bridges/rialto-millau/dashboard/prometheus/prometheus.yml:/etc/prometheus/rialto-millau.yml
```
Is `prometheus-` prefix adding any value here?

deployments/bridges/westend-millau/dashboard/grafana/relay-westend-to-millau-headers-dashboard.json
```diff
@@ -0,0 +1,694 @@
+{
```
tomusdrw, line 1: I'm assuming this is just a copy of existing dashboard of other networks, right?

deployments/bridges/westend-millau/docker-compose.yml
tomusdrw: I'm not convinced this is actually needed at all. What it does it makes all containers part of a single docker host-bridged network and we want isolation.
Are you sure it's not an issue with your local setup: https://docs.docker.com/network/bridge/#enable-forwarding-from-docker-containers-to-the-outside-world ?

deployments/bridges/westend-millau/entrypoints/relay-headers-westend-to-millau-entrypoint.sh
tomusdrw: Can you elaborate on the errors?  It would be good to quick feedback from `curl` if `westend-rpc.polkadot.io` cannot be connected to before even running the `init-bridge` command.

deployments/bridges/rialto-millau/docker-compose.yml
tomusdrw: EDIT: I see we glob-import all `prometheus-*.yml` files, so maybe better to use `job-rialto-millau.yml` to avoid confusion with the main `prometheus.yml` configuration file?

deployments/run.sh
```diff
@@ -48,17 +52,28 @@ do
     --no-monitoring)
       MONITORING=" -f ./monitoring/disabled.yml"
       shift
+      continue
       ;;
     poa-rialto)
-      BRIDGE=$i
+      BRIDGES+=($i)
       NETWORKS+=${RIALTO}
+      RIALTO=''
```
tomusdrw, line 60: What are these network-specific variables about?

relays/substrate/src/rialto_millau/mod.rs
```diff
@@ -115,6 +115,9 @@ async fn run_init_bridge(command: cli::InitBridge) -> Result<(), String> {
 				.next_account_index(millau_sign.signer.public().into())
 				.await?;
 
+			// at Westend -> Millau initialization we're not using sudo, because otherwise our deployments
+			// may fail, because we need to initialize both Rialto -> Millau and Westend -> Millau bridge.
+			// => since there's single possible sudo account, one of transaction may fail with duplicate nonce error
```
tomusdrw, line 120: Ugh, that sounds a bit like a nasty hack. Maybe let's set up an owner for all netwoks and never use `sudo` (sounds reasonable to me, cause it's one runtime requirement less). Please add an issue for this if it makes sense for you too.

tomusdrw (commented):
lgtm! Couple of minor things.

.gitignore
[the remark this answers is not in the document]
svyatonik: > Why do we mount the entire folder though? Can we only add a sub-folder to make sure that there are no local modifications done by the container?

I don't get your questions. If we don't mount 'the entire folder' (/etc/grafana/provisioning/dashboards), then where we're going to mount subfolders? And where from we need to read dashboards from?

> Or maybe there is a way to mount read-only?

Readonly mount fails, because we then try to populate this folder with dashboards.

deployments/bridges/westend-millau/docker-compose.yml
svyatonik: > it makes all containers part of a single docker host-bridged network and we want isolation

Why do you think so? I see that this particular container will have access two networks, of which one is bridged. We can't have isolated network and reach external node, can we?

deployments/bridges/westend-millau/entrypoints/relay-headers-westend-to-millau-entrypoint.sh
svyatonik:
```
+ curl -v https://westend-rpc.polkadot.io/health
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying 104.21.60.44:443...
* TCP_NODELAY set
* Connected to westend-rpc.polkadot.io (104.21.60.44) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* error setting certificate verify locations:
  CAfile: /etc/ssl/certs/ca-certificates.crt
  CApath: /etc/ssl/certs
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
* Closing connection 0
curl: (77) error setting certificate verify locations:
  CAfile: /etc/ssl/certs/ca-certificates.crt
  CApath: /etc/ssl/certs
{"isSyncing":false,"peers":4,"shouldHavePeers":true}
```

deployments/bridges/rialto-millau/docker-compose.yml
svyatonik: It isn't job, it's a set of targets. There's single job now

deployments/bridges/westend-millau/dashboard/grafana/relay-westend-to-millau-headers-dashboard.json
svyatonik: That's correct

deployments/run.sh
svyatonik: They're defined before the loop. This 'nullifying' is needed to avoid duplicating the same (Millau|Rialto) nodes

relays/substrate/src/rialto_millau/mod.rs
svyatonik: No, I don't see this particular code as a hack. The issue is how we start bridges - i.e. simultaneously, but I don't see a way on how to improve that now.

deployments/bridges/westend-millau/docker-compose.yml
svyatonik: Actually, it seems to work for me, because deployments_default is also using bridge driver. But then what do you mean by "isolation"?

tomusdrw: Right, I meant isolation from the host machine, but perhaps I'm confusing `bridge` and `host` drivers. 
However you've changed `nginx-proxy` network as well, and I've tried this one to stay isolated from containers that don't need it, hence all containers ended up in both networks, I guess we can revert this back.

deployments/bridges/westend-millau/entrypoints/relay-headers-westend-to-millau-entrypoint.sh
tomusdrw: Seems there is no CA cerificates? [Similar issue here](https://github.com/curl/curl/issues/6090). Perahps we should either:
1. `update-ca-certificates` in the docker image (or install some missing package)
2. Disable SSL checking (`--insecure`?) cause without CA certificates it doesn't really work anyway I guess.

deployments/bridges/rialto-millau/docker-compose.yml
tomusdrw: Then a `target-` prefix maybe? My point is to avoid confusion about using `prometheus.yml`-level options in that file (cause as you've already mentioned to Hernando they are not valid in this context).

deployments/bridges/westend-millau/entrypoints/relay-headers-westend-to-millau-entrypoint.sh
svyatonik: I don't mind dealing with it in a separate issue - it isn't about Westend -> Millau bridge.

deployments/bridges/westend-millau/docker-compose.yml
svyatonik: Yes, I've changed nginx-proxy as well - it's already reverted after removal of that extra network.

relays/substrate/src/rialto_millau/mod.rs
HCastano: I don't see this as a hack either. We're just deciding to use a different owner than Sudo in this particular case (because of nonce issues)

tomusdrw: We had to put at least 5 comments to justify why we are not using `sudo` here (and when we set the owner).
So what I'm proposing is that instead of special casing, maybe let's never use sudo and just assign owners to specific bridges? Then we don't have to justify that. Is there a compelling reason to keep `sudo` for other bridge initialisations?

tomusdrw (approved):

relays/substrate/src/rialto_millau/mod.rs
HCastano: Well the main problem here isn't the use of `Sudo` per say, but the use of `Alice`, who happens to be `Sudo` for all our local/development networks.

I agree though, it could make sense to have different non-Sudo accounts set as owners instead.

HCastano (approved):

PR #833, opened 2021-03-18
on-demand headers relay

svyatonik:
related #817 (will need to use in deployments to close it)

There's a caveat with this approach - since we're serving lanes in both directions (messages are delivered from Rialto to Millau and from Millau to Rialto), it means that sometimes there'll be two active headers relays in, say, Rialto->Millau direction. The more 'busy' messages relay is, the more chance for two relays to be active. Since we're sending a lot of messages, the chance is quite big. I assume that at least initially on P<->K and W<->R we won't see too much messages => the chance won't be that big.

Two active headers relays means that we may actually spent more funds than when single relay is active (what we have now). So I suggest to run on-demand relays on real chains initially (where transactions have real cost), but if we'll see many messages, then it worth switching back to separate relays. On test deployments it makes sense to use on-demand relays for Millau <-> Rialto headers and use regular relay for Westend -> Millau. So we'll have both relays 'covered with integration tests'.


relays/substrate/src/rialto_millau/millau_messages_to_rialto.rs
```diff
@@ -131,11 +131,30 @@ pub async fn run(
 	rialto_client: RialtoClient,
 	rialto_sign: RialtoSigningParams,
 	lane_id: LaneId,
+	enable_headers_relay: bool,
```
HCastano, line 134:
```suggestion
	enable_on_demand_relay: bool,
```

relays/substrate/src/rialto_millau/millau_messages_to_rialto.rs
```diff
@@ -131,11 +131,30 @@ pub async fn run(
...
 
+	let rialto_to_millau_headers_relay = if enable_headers_relay {
+		Some(crate::rialto_millau::rialto_headers_to_millau::on_demand(
+			rialto_client.clone(),
+			millau_client.clone(),
+			millau_sign.clone(),
+		))
+	} else {
+		None
+	};
```
HCastano, line 148: Sharing since I just learned about this yesterday: [`bool::then`](https://doc.rust-lang.org/std/primitive.bool.html#method.then) was introduced in Rust `v1.50.0`, so this could be rewritten as:

```rust
 let rialto_to_millau_headers_relay = enable_headers_relay.then(|| {
     crate::rialto_millau::rialto_headers_to_millau::on_demand(
         rialto_client.clone(),
         millau_client.clone(),
         millau_sign.clone(),
     )
 });
```

I'm still unsure as to how I feel about it though, since it's quite "functional"

relays/substrate/src/rialto_millau/rialto_messages_to_millau.rs
```diff
@@ -131,11 +131,30 @@ pub async fn run(
 	millau_client: MillauClient,
 	millau_sign: MillauSigningParams,
 	lane_id: LaneId,
+	enable_headers_relay: bool,
```
HCastano, line 134:
```suggestion
	enable_on_demand_relay: bool,
```

relays/clients/substrate/src/client.rs
```diff
@@ -52,6 +53,14 @@ pub struct Client<C: Chain> {
...
 	/// Genesis block hash.
 	genesis_hash: C::Hash,
+	/// If several tasks are preparing their transactions simultaneously, they may get the same
+	/// response (nonce) from the `next_account_index` call. So one of transactions will be rejected
+	/// from the pool. This lock is here to prevent situations like that.
+	///
+	/// Note that it's just lock, not the `HashMap<AccountId, Index>`. This is because querying index
+	/// does not necessarily mean that transaction with this nonce will be submitted. If we fail to
+	/// submit it, we may end up with incorrect (future) nonce in the map => infinite relay loop.
+	next_account_index_lock: Arc<Mutex<()>>,
```
HCastano, line 63: Would this happen if we have both the on-demand relay and the message relay trying to submit transactions at the same time (e.g one submitting a new header and one submitting a new message from the same account?)

HCastano (approved):
Nice optimization 👍

relays/clients/substrate/src/client.rs
svyatonik: Yes - there will be a collision, so this lock is actually useless. This needs to be fixed

svyatonik:
Actually, I got an another idea on weekend - we may even go further and start two messages relays (rialto -> millau and millau -> rialto) and two on-demand headers relays in single process. So we'll always have at most 1 (our) headers relay running in single direction. I'll move this PR back to draft to deal with this and fix issue with nonce race

svyatonik:
I've changed a scope of this PR because it overlaps with other yet-to-be-open metrics-related PR:
1) this PR changes nothing in existing relays, so there's a dead code that will be used in follow-up PRs;
2) there's now atomic `fn submit_signed_extrinsic()` function in the Substrate client, so all nonces conflicts should gone;
3) `relays/bin-substrate/src/on_demand_headers_relay.rs` is not used anywhere yet - I'll add a composite relay after dealing with overlapping metrics. Could remove changes from this PR if you think it should be a part of that future PR

relays/client-substrate/src/client.rs
```diff
@@ -52,6 +53,15 @@ pub struct Client<C: Chain> {
 	client: RpcClient,
 	/// Genesis block hash.
 	genesis_hash: C::Hash,
+	/// If several tasks are submitting their transactions simultaneously using `sumbit_signed_extrinsic`
```
HCastano, line 56:
```suggestion
	/// If several tasks are submitting their transactions simultaneously using `submit_signed_extrinsic`
```

relays/client-substrate/src/client.rs
```diff
@@ -52,6 +53,15 @@ pub struct Client<C: Chain> {
...
+	/// method, they may get the same transaction nonce. So one of transactions will be rejected
+	/// from the pool. This lock is here to prevent situations like that.
+	///
+	/// Note that it's just lock, not the `HashMap<AccountId, Index>`. This is because querying index
+	/// does not necessarily mean that transaction with this nonce will be included into the block. If
+	/// it'll be rejected or dropped, we may end up with incorrect (future) nonce in the map => infinite
+	/// relay loop.
```
HCastano, line 63: This is an implementation detail, so I don't think this should be part of the public doc comments. We should either turn it into "regular" comments, or just remove it

relays/bin-substrate/src/rialto_millau/mod.rs
```diff
@@ -111,29 +114,31 @@ async fn run_init_bridge(command: cli::InitBridge) -> Result<(), String> {
...
+						millau_runtime::Runtime,
+						millau_runtime::WestendGrandpaInstance,
+					>::initialize(initialization_data);
+
+					Bytes(
+						Millau::sign_transaction(
+							*millau_client.genesis_hash(),
+							&millau_sign.signer,
+							transaction_nonce,
+							millau_runtime::SudoCall::sudo(Box::new(initialize_call.into())).into(),
```
HCastano, line 136: Doesn't using `Sudo` here invalidate the big comment block above?

HCastano (commented):
I'm a bit confused as to what the scope of this new PR is.

If the "on-demand" code is dead code right now, I'd prefer to push that into a different PR where we can clearly see the new code in the correct usage context. That leaves the remaining changes in this PR, which are the "atomic" version of `submit_signed_extrinsic` and also new initialization logic?

svyatonik:
> I'm a bit confused as to what the scope of this new PR is.
> 
> If the "on-demand" code is dead code right now, I'd prefer to push that into a different PR where we can clearly see the new code in the correct usage context. That leaves the remaining changes in this PR, which are the "atomic" version of `submit_signed_extrinsic` and also new initialization logic?

Ok, I've removed on_demand_headers_relay.rs from relayer. The scope is atomic submit and ability to start on-demand headers relay from messages relay.

relays/client-substrate/src/client.rs
```diff
@@ -192,15 +204,35 @@ impl<C: Chain> Client<C> {
...
 		Ok(tx_hash)
 	}
 
+	/// Submit an extrinsic signed by given account.
+	///
+	/// All calls of this method are synchronized, so there can't be more than one active
+	/// `submit_extrinsic_by()` call. This guarantees that no nonces collision may happen
+	/// if all client instances are clones of the same initial `Client`.
+	///
+	/// Note: The extrinsic returned by `prepare_extrinsic` does not need be SCALE encoded beforehand.
```
HCastano, line 222: It looks like we stopped SCALE encoding the payload at this layer a while ago, so this comment (and the one for `submit_unsigned_extrinsic` are now outdated.

We should instead have the comment note that we *do* expect the transaction (or result from `prepare_extrinsic` to be SCALE encoded.

HCastano (approved):

PR #834, opened 2021-03-18
Rename pallet-message-lane into pallet-bridge-messages

svyatonik:
(this seems to be common convention after https://github.com/paritytech/substrate/pull/8128)

Another (not super-important) thing is that our pallets have a bit different naming convention. We're going to include `pallet-finality-verifier`, `pallet-bridge-call-dispatch` and `pallet-message-lane` into P+K/W+R runtimes. Do we probably need to add that `bridge-` thing to all pallets? Or remove it from call dispatch pallet? This seems a bit sloppy now.

If we're going to add `bridge-` thing to all pallets names, then it worth doing it in this PR for ML pallet.


tomusdrw:
What about other pallets though (re: `runtime` prefix)? Let's maybe unify logging everywhere not just for ML.


Regarding naming I don't know if adding `bridge-` prefix actually solves the issue. Before the deployment I'd rather re-think the entire naming and maybe add a custom prefix as well:
1. `pallet-bridge-grandpa`
2. `pallet-bridge-messages`
3. `pallet-bridge-dispatch`

would be my preference, but happy to discuss that in a dedicated issue.

HCastano:
> What about other pallets though (re: runtime prefix)? Let's maybe unify logging everywhere not just for ML.

I agree here, we might as well change the logging targets in the other pallets. I think something like `runtime::bridge::<target>` would be ideal.

> Regarding naming I don't know if adding bridge- prefix actually solves the issue. Before the deployment I'd rather re-think the entire naming and maybe add a custom prefix as well:
>
>    pallet-bridge-grandpa
>    pallet-bridge-messages
>    pallet-bridge-dispatch
>
> would be my preference, but happy to discuss that in a dedicated issue.

Something like this seems reasonable as well, but yeah this is for a different issue/PR.

svyatonik:
It makes sense to me to rename pallet in this PR, because suffix in logs (at least now) depends on the pallet name - it is "runtime::<pallet-name-without-pallet-prefix>". So rename-PR will touch at least exactly the same code lines that this PR is touching. It also makes sense to me to rename other pallets in other PR - as I've already said logging suffix and pallet name better to be changed in single PR, so I do not want to rename half of repository in this PR. This PR is only about ML pallet.

So if we're ok with `pallet-bridge-(grandpa|messages|dispatch)`, then I'm going to rename ML to `pallet-bridge-messages` and the logging target would be "runtime::bridge_messages".

modules/messages/src/lib.rs
```diff
@@ -1469,29 +1479,38 @@ mod tests {
 	fn storage_message_key_computed_properly() {
 		// If this test fails, then something has been changed in module storage that is breaking all
 		// previously crafted messages proofs.
+		let storage_key = storage_keys::message_key::<TestRuntime, DefaultInstance>(&*b"test", 42).0;
 		assert_eq!(
-			storage_keys::message_key::<TestRuntime, DefaultInstance>(&*b"test", 42).0,
-			hex!("87f1ffe31b52878f09495ca7482df1a48a395e6242c6813b196ca31ed0547ea79446af0e09063bd4a7874aef8a997cec746573742a00000000000000").to_vec(),
+			storage_key,
+			hex!("dd16c784ebd3390a9bc0357c7511ed018a395e6242c6813b196ca31ed0547ea79446af0e09063bd4a7874aef8a997cec746573742a00000000000000").to_vec(),
```
HCastano, line 1485: Why did this change?

HCastano (approved):

modules/messages/src/lib.rs
svyatonik: Afaik the storage key is computed as `twox(pallet-name ++ storage-key-name)` => then pallet name has been changed

tomusdrw (approved):

PR #835, opened 2021-03-20
Tunable Justification Generator

HCastano:
This PR updates the code we have for generating mock GRANDPA justifications in order to
make it tunable. What I mean by that is that we are now able to indicate how many
pre-commits and vote ancestries we want our justification to contain.

This is especially useful in the context of benchmarking (see #815) since we want to be
able to see how verification scales given different factors.

### DIFF
--- a/primitives/test-utils/src/lib.rs
+++ b/primitives/test-utils/src/lib.rs
@@ -31,9 +31,41 @@ use sp_std::prelude::*;
 pub const TEST_GRANDPA_ROUND: u64 = 1;
 pub const TEST_GRANDPA_SET_ID: SetId = 1;
 
+/// Configuration parameters when generating test GRANDPA justifications.
+pub struct JustificationGeneratorParams<H> {
+	/// The header which we want to finalize.
+	pub header: H,
+	/// The GRANDPA round number for the current authority set.
+	pub round: u64,
+	/// The current authority set ID.
+	pub set_id: SetId,
+	/// The current GRANDPA authority set.
+	pub authorities: Vec<(Keyring, AuthorityWeight)>,
+	/// The number of headers included in our justification's vote ancestries.
+	pub depth: u32,
+	/// The number of forks, and thus the number of pre-commits in our justification.
+	pub forks: u32,
+}
+
+impl<H: HeaderT> Default for JustificationGeneratorParams<H> {
+	fn default() -> Self {
+		Self {
+			header: test_header(One::one()),
+			round: TEST_GRANDPA_ROUND,
+			set_id: TEST_GRANDPA_SET_ID,
+			authorities: keyring(),
+			depth: 2,
+			forks: 1,
+		}
+	}
+}
+
 /// Make a valid GRANDPA justification with sensible defaults
 pub fn make_default_justification<H: HeaderT>(header: &H) -> GrandpaJustification<H> {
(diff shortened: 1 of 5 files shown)


svyatonik (approved):

primitives/test-utils/src/lib.rs
```diff
@@ -31,29 +31,58 @@ use sp_std::prelude::*;
...
+///
+/// Note: This needs at least three authorities or else the verifier will complain about
+/// being given an invalid commit.
 pub fn make_justification_for_header<H: HeaderT>(
 	header: &H,
 	round: u64,
 	set_id: SetId,
 	authorities: &[(Keyring, AuthorityWeight)],
+	depth: u32,
+	forks: u32,
```
tomusdrw, line 54: There is already 6 parameters and 3 of them are plain numbers, I think it deserves a parameters structs now. Understanding what 2 and 1 in `&keyring(), 2, 1)` is supposed to mean will quickly fade away :)

primitives/test-utils/src/lib.rs
```diff
@@ -31,29 +31,58 @@ use sp_std::prelude::*;
...
 	let mut votes_ancestries = vec![];
 
-	// We want to make sure that the header included in the vote ancestries
-	// is actually related to our target header
-	let mut precommit_header = test_header::<H>(target_number + One::one());
-	precommit_header.set_parent_hash(target_hash);
+	assert!(
+		forks as usize <= authorities.len(),
+		"If we have more forks than authorities we can't create valid pre-commits for all the forks."
+	);
```
tomusdrw, line 63: Similarly we might want to disallow `depth == 0`, no?

primitives/test-utils/src/lib.rs
```diff
@@ -67,6 +96,29 @@ pub fn make_justification_for_header<H: HeaderT>(
...
+		let parent = &headers[(i - 1) as usize];
+		let (hash, num) = (parent.hash(), *parent.number());
+
+		let mut header = test_header::<H>(num + One::one());
+		header.set_parent_hash(hash);
+
+		// Modifying the digest so headers at the same height but in different forks have different
+		// hashes
+		let digest = header.digest_mut();
+		*digest = sp_runtime::Digest {
```
tomusdrw, line 112: We could use authority id as well (would be a bit more like babe :)).

primitives/test-utils/src/lib.rs
```diff
@@ -67,6 +96,29 @@ pub fn make_justification_for_header<H: HeaderT>(
...
+			logs: vec![sp_runtime::DigestItem::Other(vec![fork_id])],
```
tomusdrw, line 113: Can't we simply push another item instead of overwriting the entire `Digest`?
```
header.digest_mut().logs.push(DigestItem::Other(vec![fork_id]))
```
?

tomusdrw, line 113: Can't we simply push another item instead of overwriting the entire `Digest`?
```
header.digest_mut().logs.push(DigestItem::Other(vec![fork_id]));
```
?

tomusdrw (approved):
lgtm! Few tiny grumbles

primitives/test-utils/src/lib.rs
HCastano: Haha, we could, but I don't really see the benefits in this case. We need the number of forks anyways, might as well use that since it's easy

PR #837, opened 2021-03-22
Flatten back the structure

tomusdrw:
It seems that nested glob patterns for member inclusion doesn't really work that well with `exclude` key.
`exclude` seems to be ignoring the entire subtree instead of a single path without `Cargo.toml` file.

This PR partially reverts #813 and makes the structure a bit flatter to avoid using `exclude` for members, I've still added a prefix to folder names so that the crates are nicely sorted.


HCastano (commented):
Lol at the amount of conflicting files.

Anyways, the new flattened structure looks fine, but I want to clarify the motivation. So from what I understand we were using the `exclude` key to make sure that Cargo didn't try and build "empty" (i.e without a `Cargo.toml`) folders at the top level of subtrees (so the `relays/clients` folder for example). However, by excluding that folder we've ended up excluding all the crates nested within, which we don't want.

Now you're flattening everything to avoid using `exclude` and go back to correctly including everything in the Cargo workspace, right?

tomusdrw:
> Now you're flattening everything to avoid using exclude and go back to correctly including everything in the Cargo workspace, right?

Exactly. You can already see the consequences - neither `cargo fmt` nor `cargo test` was running correctly for these directories, hence I have failures here. Will fix them tomorrow.

svyatonik (approved):
Alternatively you could just avoid using wildcards in the `Cargo.toml` and leave folders in their current state.

PR #838, opened 2021-03-22
Rename Finality Verifier and Call Dispatch Pallets

HCastano:
This is a follow up to #834. The main changes here are that `pallet-finality-verifier` is
now `pallet-bridge-grandpa` and `pallet-bridge-call-dispatch` is now
`pallet-bridge-dispatch`. This was basically a game of `fd ... --exec sed ...` with
different ways to mention/spell the pallet names.

I've also updated the logging targets to follow the `runtime::bridge-*` convention.

CC @hbulgarini


svyatonik (approved):
1. There's "Substrate Finality Verifier Pallet" line that may need to be fixed in the top-level comment of `modules/grandpa/src/lib.rs`
2. Dispatch pallet is still referenced as "call dispatch" or "call-dispatch" from within several places (documentation). But since it is actually dispatching calls, this is ok imo

README.md
```diff
@@ -86,7 +86,7 @@ the `relays` which are used to pass messages between chains.
 │  └──  ...
 ├── modules         // Substrate Runtime Modules (a.k.a Pallets)
 │  ├── ethereum     // Ethereum PoA Header Sync Module
-│  ├── substrate    // Substrate Based Chain Header Sync Module
+│  ├── grandpa      // On-Chain GRANDPA Light Client
 │  ├── messages     // Cross Chain Message Passing
```
tomusdrw, line 90:
```suggestion
│  ├── messages     // Cross Chain Message Passing
│  ├── dispatch        // Target-Chain Message Dispatching (Execution)
```

bin/rialto/runtime/Cargo.toml
```diff
@@ -93,9 +93,9 @@ std = [
 	"pallet-aura/std",
 	"pallet-balances/std",
 	"pallet-bridge-eth-poa/std",
-	"pallet-bridge-call-dispatch/std",
+	"pallet-bridge-dispatch/std",
```
tomusdrw, line 96: The rename affects sorting, I think it would be good to fix in all `Cargo.toml` files (unfortunately fmt is not able to handle that afaik).

tomusdrw (approved):
lgtm!

HCastano:
>     1. There's "Substrate Finality Verifier Pallet" line that may need to be fixed in the top-level comment of `modules/grandpa/src/lib.rs`

I could've sworn I fixed this one...

>     2. Dispatch pallet is still referenced as "call dispatch" or "call-dispatch" from within several places (documentation). But since it is actually dispatching calls, this is ok imo

Yeah, I saw some places in the docs where it looked better as "Call dispatch". We can always change this later if we decide that the docs are awkward

bin/rialto/runtime/Cargo.toml
HCastano: I fixed the ones touched in this PR. It was actually pretty easy with [Vim's `sort` utility](https://vim.fandom.com/wiki/Sort_lines).

You can select the range you want (either with numbers or in visual mode) and then call `:sort`.

PR #839, opened 2021-03-23
Fix RUST_LOG in Rialto deployment compose file

svyatonik:


svyatonik:
(will merge without approval - it's a lear issue + I'm going to restart deployments anyway now)

PR #840, opened 2021-03-23
Changed how relay loops are started

svyatonik:
This is required both for #833 (otherwise process with 4 relays will provide 4 same global metrics, spending 4x time on sysinfo calls) and for "standalone" metrics (see https://github.com/paritytech/parity-bridges-common/pull/822#discussion_r594091027).


relays/generic/exchange/src/exchange_loop.rs
```diff
@@ -85,50 +85,32 @@ pub async fn run<P: TransactionProofPipeline>(
...
-		source_client,
-		target_client,
-		|source_client, target_client| {
+
+	relay_utils::relay_loop(source_client, target_client)
+		.with_metrics(format!("{}_to_{}_Exchange", P::SOURCE_NAME, P::TARGET_NAME))
+		.loop_metric(ExchangeLoopMetrics::default())?
+		.standalone_metric(GlobalMetrics::default())?
+		.expose(metrics_params)
+		.await?
```
svyatonik, line 96: I'd prefer to have additional tabs here and in other loop files, but fmt prevents me from doing this :/

relays/generic/utils/src/metrics/global.rs
```diff
@@ -0,0 +1,109 @@
+// Copyright 2019-2021 Parity Technologies (UK) Ltd.
```
svyatonik, line 1: There's now single file in this folder, but I have another one to add in the next PRs

relays/utils/src/relay_loop.rs
```diff
@@ -32,60 +34,164 @@ pub trait Client: Clone + Send + Sync {
...
+			loop_metric: None,
+		}
+	}
+
+	/// Run relay loop.
+	///
+	/// This function represents an outer loop, which in turn calls provided `loop_run` function to do
+	/// actual job. When `loop_run` returns, this outer loop reconnects to failed client (source,
+	/// target or both) and calls `loop_run` again.
+	pub async fn run<R, F>(mut self, loop_run: R) -> Result<(), String>
```
HCastano, line 84:
```suggestion
	pub async fn run<R, F>(mut self, run_loop: R) -> Result<(), String>
```

HCastano: From what I understand this doesn't need to be any sort of loop though, it really just needs to be something which returns a `Future`, right? Maybe we shouldn't even have loop in the name here

relays/utils/src/relay_loop.rs
```diff
@@ -32,60 +34,164 @@ pub trait Client: Clone + Send + Sync {
...
+
+		Ok(())
+	}
+}
+
+impl<SC, TC, LM> LoopMetrics<SC, TC, LM> {
+	/// Add relay loop metrics.
+	///
+	/// Loop metrics will be passed to the loop callback.
+	pub fn loop_metric<NewLM: Metrics>(self, loop_metric: NewLM) -> Result<LoopMetrics<SC, TC, NewLM>, String> {
```
HCastano, line 148:
```suggestion
	pub fn loop_metric<M: Metrics>(self, loop_metric: M) -> Result<LoopMetrics<SC, TC, M>, String> {
```

relays/utils/src/metrics.rs
```diff
@@ -31,62 +32,32 @@ pub struct MetricsParams {
 }
 
 /// Metrics API.
-pub trait Metrics {
+pub trait Metrics: Clone + Send + Sync + 'static {
```
HCastano, line 35: Why does this need to be `'static` now?

HCastano (approved):
>  I'd prefer to have additional tabs here and in other loop files, but fmt prevents me from doing this :/

I would say that RustFmt's word is law, even if we don't agree with it.

Also, looks like there are some Clippy warnings which need to be fixed

relays/utils/src/metrics.rs
svyatonik: It is now used in [`async_std::task::spawn()`](https://docs.rs/async-std/1.9.0/async_std/task/fn.spawn.html), which required future to be `'static`

relays/utils/src/relay_loop.rs
svyatonik: Yes, you may use it to run arbitrary futures, I see no point in using abstract names here. It is only used (and most probably will only be used) to run relay loops. I bet you'll be the first one to object if I'll rename this whole functionality to something like `fn run_future_with_retry()` :)

PR #841, opened 2021-03-23
Rococo (mocky)Runtime and Client

tomusdrw:
This PR adds a Rococo-Client and a partially-mocked Rococo runtime to the repository.

Currently clients depend heavily on the runtime itself, but we can't import the Rococo runtime easily, since it lives in the Polkadot repo.
I really wanted to keep the relayer code in the bridges repo, so started thinking how difficult would be to only import these parts of Rococo runtime that we actually need to the repo. It turned out to be not as bad as I thought, we basically need a dummy version of `SignedExtensions` and the `Call` enum (or actually only the parts of the call enum that we are really interested about).

I will follow up on this with a similar implementation for `Westend` and a relayer between `Rococo` and `Westend` (the last PR will include required `Call` enum variants).
My intention with the relayer is to move the CLI code into a macro, so that we can easily implement `Westend` -> `Millau` subcommand or `Rococo <> Westend` relayer without duplicating the code (As it is currently).


HCastano (approved):
Looks fine, but it's unclear to me how we're doing to deal with changes in the Rococo runtime. For example, how will we know that our mock runtime doesn't match the Rococo runtime anymore? Do we just have to wait for relayers to fail on a live deployment?

primitives/chain-rococo/Cargo.toml
```diff
@@ -7,22 +7,31 @@ edition = "2018"
...
+[dev-dependencies]
+hex = "0.4"
 
 [features]
 default = ["std"]
 std = [
 	"bp-messages/std",
 	"bp-polkadot-core/std",
 	"bp-runtime/std",
+	"parity-scale-codec/std",
```
svyatonik, line 33: also `frame-support/std` and `sp-runtime/std`?

primitives/polkadot-core/Cargo.toml
```diff
@@ -7,6 +7,9 @@ edition = "2018"
 license = "GPL-3.0-or-later WITH Classpath-exception-2.0"
 
 [dependencies]
+parity-scale-codec = { version = "2.0.0", default-features = false, features = ["derive"] }
+parity-util-mem = { version = "0.9.0", default-features = false, features = ["primitive-types"] }
+hash256-std-hasher = { version = "0.15.2", default-features = false }
```
svyatonik, line 12: Are you sure these are used anywhere?

primitives/polkadot-core/src/lib.rs
```diff
@@ -160,14 +167,94 @@ pub type AccountId = <AccountPublic as IdentifyAccount>::AccountId;
...
+/// A type of the data encoded as part of the transaction.
+pub type SignedExtra = ((), (), (), sp_runtime::generic::Era, Nonce, (), Balance);
+
+/// Parameters which are part of the payload used to produce transaction signature,
+/// but don't end up in the transaction itself (i.e. inherent part of the runtime).
+pub type AdditionalSigned = (u32, u32, Hash, Hash, (), (), ());
+
+/// A simplified version of signed extensions meant for producing signed transactions
+/// and signed payload in the client code.
+#[derive(PartialEq, Eq, Clone, Debug)]
```
svyatonik, line 190: `Debug` -> `RuntimeDebug`?

svyatonik (approved):

primitives/chain-rococo/src/lib.rs
```diff
@@ -21,10 +21,60 @@
...
 use sp_std::prelude::*;
+use sp_version::RuntimeVersion;
 
 pub use bp_polkadot_core::*;
 
+pub type UncheckedExtrinsic = bp_polkadot_core::UncheckedExtrinsic<Call>;
+
+pub const VERSION: RuntimeVersion = RuntimeVersion {
+	spec_name: sp_version::create_runtime_str!("polkadot"),
+	impl_name: sp_version::create_runtime_str!("parity-polkadot"),
```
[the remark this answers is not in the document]
tomusdrw: I want the version to match exactly what's in polkadot.

primitives/polkadot-core/Cargo.toml
tomusdrw: Only `parity_scale_codec` indeed. Probably some leftovers from some other approaches I've tried.

primitives/polkadot-core/src/lib.rs
tomusdrw: It's not supposed to be used in `Runtime` anyway, but yeah, let me switch to `RuntimeDebug`.

tomusdrw (commented):
> Looks fine, but it's unclear to me how we're doing to deal with changes in the Rococo runtime. For example, how will we know that our mock runtime doesn't match the Rococo runtime anymore? Do we just have to wait for relayers to fail on a live deployment?

We already have a bunch of assumptions that we have to keep in sync within the runtime - i.e. the Bridge pallet is not really able to import the entire other network runtime so we have a bunch of "mocks". This extends that concept to relayers.

I'm imagining that we will add a bunch of tests in the `polkadot` repository, to make sure that things that are actually used are kept in sync. For instance we will make sure that `RuntimeVersion` is the same (failing test will be a clear indication that something has to be updated in the mock runtime as well) and that the `Call` enum encoding matches.

PR #842, opened 2021-03-23
Larger Test Keyring Support

HCastano:
Another thing I realized needed to be more flexible while working through #815.

In #835 I added the ability to create justifications with a tunable number of forks.
However, this only works for the case where `forks <= authorities.len()` since each fork
has a pre-commit which needs to be signed by an authority and the same authority can't
sign off on two different pre-commits (they would be equivocating).

Our existing test keyring with friendly names (Alice, Bob, etc.) is not great for one
particular benchmark where we want to scale up the number of authorities past the number
of enum variants.

This PR adds a small wrapper and a helper function to generate an "abitrary" (really just
`u8::MAX`) number of accounts which we can then use during benchmarking.

You might ask: is `u8` enough for benchmarking if Polkadot/Kusama are going to support
1000 validators? I think this is fine since we're looking for _trends_ in benchmarking
data (e.g how this scales for every _extra_ validator and pre-commit), and `u8::MAX`
data-points seems like a reasonable amount to me. If you guys disagree I can change this,
but it'll make generating `SecretKey`s for the test accounts a more complicated.

### DIFF
--- a/primitives/test-utils/src/keyring.rs
+++ b/primitives/test-utils/src/keyring.rs
@@ -46,16 +46,7 @@ pub trait Keyring {
 	}
 }
 
-#[derive(RuntimeDebug, Clone, Copy)]
-pub struct Account(pub u8);
-
-impl Keyring for Account {
-	fn secret(&self) -> SecretKey {
-		SecretKey::from_bytes(&[self.0; 32]).expect("A static array of the correct length is a known good.")
-	}
-}
-
-/// Set of test accounts.
+/// Set of test accounts with friendly names.
 #[derive(RuntimeDebug, Clone, Copy)]
 pub enum TestKeyring {
 	Alice,
@@ -69,6 +60,16 @@ pub enum TestKeyring {
 impl Keyring for TestKeyring {
 	fn secret(&self) -> SecretKey {
 		SecretKey::from_bytes(&[*self as u8; 32]).expect("A static array of the correct length is a known good.")
+	}
+}
+
+/// A test account which can be used to sign messages.
+#[derive(RuntimeDebug, Clone, Copy)]
+pub struct Account(pub u8);
+
+impl Keyring for Account {
+	fn secret(&self) -> SecretKey {
+		SecretKey::from_bytes(&[self.0; 32]).expect("A static array of the correct length is a known good.")
 	}
 }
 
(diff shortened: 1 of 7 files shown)


svyatonik (approved):

primitives/test-utils/src/keyring.rs
```diff
@@ -0,0 +1,120 @@
...
+/// Convenience function to get a list of Grandpa authorities.
+pub fn authority_list() -> AuthorityList {
+	test_keyring()
+		.iter()
+		.map(|(id, w)| (AuthorityId::from(*id), *w))
+		.collect()
+}
+
+/// Get the corresponding identities from the keyring for the "standard" authority set.
+pub fn test_keyring() -> Vec<(TestKeyring, u64)> {
```
tomusdrw, line 104: What's the `u64` here it's hard to guess without a context, maybe at least add something to the docs.

primitives/test-utils/src/keyring.rs
```diff
@@ -0,0 +1,120 @@
...
+
+/// Get a list of "unique" accounts.
+pub fn accounts(len: u8) -> Vec<Account> {
+	let mut v = vec![];
+	for i in 0..len {
+		v.push(Account(i));
+	}
+
+	v
+}
```
tomusdrw, line 120:
```suggestion
pub fn accounts(len: u8) -> Vec<Account> {
	0..len.map(Account).collect()
}
```

primitives/test-utils/src/keyring.rs
```diff
@@ -0,0 +1,120 @@
...
+//! Utilities for working with test accounts.
+
+use ed25519_dalek::{Keypair, PublicKey, SecretKey, Signature};
+use finality_grandpa::voter_set::VoterSet;
+use sp_application_crypto::Public;
+use sp_finality_grandpa::{AuthorityId, AuthorityList};
+use sp_runtime::RuntimeDebug;
+
+/// Used to indicate if a type is able to cryptographically sign messages.
+pub trait Signer {
```
tomusdrw, line 26: I don't like that we need to introduce the trait. Can we instead keep just `TestKeyring` and introduce `constants` that represent "named" accounts? Is it really that important to distinguish between `TestKeyring::Alice` and `Account(0)` and introduce additional complexity of handling generic types?

tomusdrw (commented):
> If you guys disagree I can change this,
but it'll make generating SecretKeys for the test accounts a more complicated.

IMHO it's fine for now, but we should run the tests with larger number of validators in the future. The _trends_ are simply easier to observe when differences are high - a linear and quadratic or quadratic and exponential might not be easy to observe if the sample is too small.

Not a fan of adding extra generic parameters to the `test-utils` though.

PR #845, opened 2021-03-26
Correctly format SS58-prefixed addresses in the CLI

tomusdrw:
CC @hbulgarini

### DIFF
--- a/relays/bin-substrate/src/cli.rs
+++ b/relays/bin-substrate/src/cli.rs
@@ -286,7 +286,7 @@ impl AccountId {
 		if original.ss58_format != self.ss58_format {
 			log::warn!(
 				target: "bridge",
-				"Address {} does not seem to match {}'s SS58 format (got: {}, want: {}).\nConverted to: {}",
+				"Address {} does not seem to match {}'s SS58 format (got: {}, expected: {}).\nConverted to: {}",
 				original,
 				T::NAME,
 				original.ss58_format,
--- a/relays/bin-substrate/src/cli.rs
+++ b/relays/bin-substrate/src/cli.rs
@@ -449,7 +449,7 @@ macro_rules! declare_chain_options {
 mod tests {
 	use std::str::FromStr;
 
-    use super::*;
+	use super::*;
 
 	#[test]
 	fn should_format_addresses_with_ss58_format() {
@@ -467,12 +467,8 @@ mod tests {
 			.map(|s| AccountId::from_str(s).unwrap())
 			.collect::<Vec<_>>();
 
-		let actual = parsed
-			.iter()
-			.map(|a| format!("{}", a))
-			.collect::<Vec<_>>();
+		let actual = parsed.iter().map(|a| format!("{}", a)).collect::<Vec<_>>();
 
 		assert_eq!(actual, expected)
 	}
 }
-
--- a/relays/bin-substrate/src/cli.rs
+++ b/relays/bin-substrate/src/cli.rs
@@ -16,6 +16,8 @@
 
 //! Deal with CLI args of substrate-to-substrate relay.
 
+use std::convert::TryInto;
+
 use bp_messages::LaneId;
 use codec::{Decode, Encode};
 use sp_runtime::app_crypto::Ss58Codec;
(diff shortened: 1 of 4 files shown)


relays/bin-substrate/src/cli.rs
```diff
@@ -414,3 +444,31 @@ macro_rules! declare_chain_options {
...
+
+	use super::*;
+
+	#[test]
+	fn should_format_addresses_with_ss58_format() {
+		// given
+		let rialto1 = "5sauUXUfPjmwxSgmb3tZ5d6yx24eZX4wWJ2JtVUBaQqFbvEU";
+		let rialto2 = "5rERgaT1Z8nM3et2epA5i1VtEBfp5wkhwHtVE8HK7BRbjAH2";
+		let millau1 = "752paRyW1EGfq9YLTSSqcSJ5hqnBDidBmaftGhBo8fy6ypW9";
+		let millau2 = "74GNQjmkcfstRftSQPJgMREchqHM56EvAUXRc266cZ1NYVW5";
```
HCastano, line 460: It might be useful to have a testcase where the formatting fails

relays/utils/src/initialize.rs
```diff
@@ -16,40 +16,81 @@
 
 //! Relayer initialization functions.
 
-use std::io::Write;
+use std::{fmt::Display, io::Write};
```
HCastano, line 19: Why where these logging changes necessary for this PR?

relays/bin-substrate/src/cli.rs
```diff
@@ -239,50 +241,78 @@ arg_enum! {
...
+		self.ss58_format = T::ss58_format().try_into().expect(SS58_FORMAT_PROOF);
+		log::debug!("{} SS58 format: {} (RAW: {})", self, self.ss58_format, self.account);
+		if original.ss58_format != self.ss58_format {
 			log::warn!(
 				target: "bridge",
-				"Following address: {} does not seem to match {}'s format, got: {}",
-				self.account,
-				net,
-				self.version,
+				"Address {} does not seem to match {}'s SS58 format (got: {}, want: {}).\nConverted to: {}",
```
HCastano, line 289:
```suggestion
				"Address {} does not seem to match {}'s SS58 format (got: {}, expected: {}).\nConverted to: {}",
```

nit

HCastano (approved):
Looks good.

I would've liked a description of the PR though. Would've been nice to have a bit of context behind this change and maybe why it's relevant to Hector's work.

relays/bin-substrate/src/cli.rs
tomusdrw: Not sure I get what you mean, do you want to trigger an error coming from `FromStr` implementation? It will only fail if the address is simply not a valid address, but that's something that is tested in the substrate repo already.

Currently the implementation is able to read address with any SS58 prefix and allows the caller to reformat (`enforce_chain`) it to desired format. I've added a test for this.

relays/utils/src/initialize.rs
tomusdrw: After #843 the warning is now being displayed and it looks gross with the timestamp.
I can extract these changes into a separate PR if you prefer.

tomusdrw (commented):
> I would've liked a description of the PR though. Would've been nice to have a bit of context behind this change and maybe why it's relevant to Hector's work.

Sure, sorry about that!

So currently we were relying on Substrate's `AccountId32` type which implements `Display` and handles `SS58Format`, but always converts the account to `DEFAULT_SS58_PREFIX` format (a global state variable).
So when the CLI displayed accounts they were not matching expected accounts in the UI. Hector reported that and I've decided to create a wrapper `AccountId` type, which displays accounts in the exact prefix that's specified, allowing the caller to convert between formats.
I think this is much cleaner than previous `into_rialto()` design and also in-line with another relayer work that I'm doing (making it easy to run for multiple chains without code duplication).

relays/utils/src/initialize.rs
HCastano: Yeah, I think this was best left for another PR. It's fine though, no need to split it out now.

Your loss though - one less green square for you 😄

PR #847, opened 2021-03-29
Use GrandpaJustification instead of Vec<u8> in Pallet API

HCastano:
Today Tomek asked why we use `Vec<u8>` instead of `GrandpaJustification<H>` in the
dispatchable API of `pallet-bridge-grandpa`. I couldn't remember why, so I went about
exploring if we could change the API use to `GrandpaJustification` instead. I've managed
to succesfully sync headers between Rialto and Millau with these changes, so it looks
like there's no problem.

The reason why this is useful it because it allows to do some inspection on the incoming
justification for weight purposes (fees are calculated using
`justification.commit.pre_commits` and `justification.vote_ancestries`).

Thanks for the suggestion Slava.

~~This PR is still a draft for a few reasons.~~

~~It's unclear to me if it's worth removing/replacing the relayers [justification wrapper](https://github.com/paritytech/parity-bridges-common/blob/master/relays/client-substrate/src/finality_source.rs#L34)
with `GrandpaJustification`.~~

~~This has a lot of cascading effects since implementations of `FinalityProofPipeline` use
this wrapper. For example, [`SubstrateFinalityToSubstrate`](https://github.com/paritytech/parity-bridges-common/blob/master/relays/bin-substrate/src/finality_pipeline.rs#L94),
which is used by Millau, Rialto, and Westend.~~

~~Another reason is error handling when decoding raw justifications. I have
`expect("TODO")`s in places right now because I don't know how to deal with the failure
case. Since those places are basically the last step before we submit a call to the node
I don't think we can fail to decode justifications (otherwise it means we've been sending
invalid data, which we haven't since our test network runs fine).~~

~~I think @svyatonik is the best person to answer these questions, but input from anyone is
cool too.~~


svyatonik:
This https://github.com/paritytech/parity-bridges-common/blob/master/relays/client-substrate/src/finality_source.rs#L38 needs to be replaced with decoded justification => you'll get decoded justification in the end && won't need these expects

relays/client-substrate/src/finality_source.rs
```diff
@@ -31,22 +32,27 @@ use std::{marker::PhantomData, pin::Pin};
...
+	/// Decoded GRANDPA justification.
+	justification: GrandpaJustification<H>,
 }
 
-impl<Number> Justification<Number> {
-	/// Extract raw justification.
-	pub fn into_inner(self) -> sp_runtime::Justification {
-		self.raw_justification
+impl<H: HeaderT> Justification<H> {
+	/// Extract raw encoded justification.
```
svyatonik, line 43: This method is not used anywhere && you can get the same result with `.justification().encode()`, so there's no reason to keep it imo.

svyatonik (approved):

relays/client-substrate/src/finality_source.rs
```diff
@@ -31,22 +32,27 @@ use std::{marker::PhantomData, pin::Pin};
 
 /// Wrapped raw Justification.
 #[derive(Debug, Clone)]
-pub struct Justification<Number> {
+pub struct Justification<H: HeaderT> {
```
tomusdrw, line 35: Is this struct still necessary? The doc comment says that `target_header_number` is decoded from the `raw_justification`, seems that you can get what you need from `GrandpaJustification.commit.target_number`. Please remove.

tomusdrw (commented):
LGTM apart from unnecessary wrapper struct. @andresilva could you take a look at this and let us know if there are any reasons to keep raw `Vec<u8>` instead apart from API-level compatibility (maybe some decode bomb?)

relays/client-substrate/src/finality_source.rs
svyatonik: It is required for `impl<H: HeaderT> FinalityProof<H::Number> for Justification<H>` (`FinalityProof` is from the crate that knows nothing about GRANDPA and `GrandpaJustification` is from the crate that shall not reference the `finality-relay` crate). Alternatively, we may extract `FinalityProof` trait to the `bp-header-chain` and impl it for the `GrandpaJustification` there. But eventually I hope that justification will be exposed from `sp-finality-grandpa` && we will be able to remove all our justification-related copypaste => so we'll need this wrapper again.

HCastano: I ended up removing the wrapper despite your last point Slava. I'm not sure when, or even if, we'll get the primitives exposed via `sp-finality-grandpa`. 

We can revisit this if that happens.

andresilva:
Passing `GrandpaJustification` instead of a `Vec<u8>` makes sense to me, off the top of my head I don't remember any reason for it being `Vec<u8>` in the first place. In the grandpa pallet we also pass along the concrete `EquivocationProof` type when reporting equivocations (which includes prevotes/precommits) rather than an encoded blob.

> But eventually I hope that justification will be exposed from `sp-finality-grandpa`

I think this is doable, and yeah avoiding duplication would be the best path. I added it to my TODOs.

tomusdrw (approved):

PR #848, opened 2021-03-30
Pre-create metrics registry before loop is started + administrative metrics

svyatonik:
### Part 1: binaries are now able to change set of metrics, exposed by the loop

Right now single relay process has at most one active relay loop. So all loops are preparing metrics and exposing metrics endpoint in their code (like [here](https://github.com/paritytech/parity-bridges-common/blob/f29d49b243491c04da487d4ec47db0247ed5b04d/relays/finality/src/finality_loop.rs#L103) and [here](https://github.com/paritytech/parity-bridges-common/blob/f29d49b243491c04da487d4ec47db0247ed5b04d/relays/messages/src/message_lane_loop.rs#L224)). With coming [multiple-loops-in-single-process](https://github.com/paritytech/parity-bridges-common/pull/833#issuecomment-803758458), we can't use same approach because we need to expose single endpoint (=> metrics prefix must also be shared). This PR expands approach introduced in #840 - loops may now receive pre-created-and-populated metrics registry and add their specific metrics there.

Related change in the code is expanded `MetricsParams` struct - now it has additional shared reference to the metrics `registry`. So we may add more metrics from the relay binary, before starting the loop. Later we'll be able to share the same registry by multiple loops.

### Part 2: administrative metrics

There are now three standalone metrics, which will be used by the 'administrative dashboard' on real-chains setup. They are:
1) Substrate-specific `FloatStorageValueMetric`, which reads some `sp_arithmetic::Fixed*` value from the runtime storage at best finalized block and exposes this value as float gauge. We'll be using it to watch [token conversion rates](https://github.com/paritytech/parity-bridges-common/blob/f29d49b243491c04da487d4ec47db0247ed5b04d/bin/rialto/runtime/src/millau_messages.rs#L40) used by messages mechanism;
2) Substrate-specific `StorageProofOverheadMetric`, which exposes (approximate) number of extra bytes in the single value storage proof of the given chain. Since Substrate has no RPC methods that are returning proofs, I'm using our messages RPC pallet. So this metric can only be used on chains that have messages pallet deployed + have at least single incoming message received at selected lane;
3) `FloatJsonValueMetric` metric, which makes HTTP GET request to given URL, parses response as json and then reads float value from given json path. We'll be using this metric to compute tokens conversion rate and then compare this rate to the value from (1).

We have discussed that we'll need such metrics before. But there are some things that may need to be discussed:
1) do we really need this 'administrative dashboard'? Imo yes - I'm not following conversion rates + I do not want to get e.g. proof size from some other tool - we already have dashboards with alerts, so new mechanism/tool isn't really required;
2) do we need a separate tool that will expose these metrics, or just reuse existing relayers? I'd vote for separate tool, but: (2a) we need to support "part1" anyway and (2b) we already have everything required (cli, clients, metrics) in relay. So now it makes sense to me to reuse relayers for now and think of a separate tool later;
3) do we really need `FloatJsonValueMetric`, given that there's e.g. https://github.com/prometheus-community/json_exporter? I'm not sure - I just don't want another process/container just for that. Metric code seems pretty dumb (request -> response -> parse) && enough for us.

### Part 3: administrative metrics in existing loops

To test both part1 and part2, I've added new metrics to existing loops:
1) Millau storage proof overhead and Rialto -> Millau conversion rate metrics are exposed by the Millau -> Rialto messages relay;
2) Rialto storage proof overhead and Millau -> Rialto conversion rate metrics are exposed by the Rialto -> Millau messages relay;
3) Polkadot/USD and Kusama/USD rates (which may be used to compute P -> K and K -> P rates) are exposed by the Westend -> Millau headers relay. This loop has nothing common with these tokens, but at least it mentions next-level Westend chain.

I haven't added 'administrative' dashboard, though - it may be added in a separate PR.

### Part 4: conclusion

Parts are ordered by their importance and also every part depends on all previous parts. Combined, they show my current vision of the future 'administrative' dashboard. But maybe there are better ways to build this dashboard? Or maybe we just need some separate tool and a cron job instead of dashboard? I'm unsure now - that's why all 3 parts are combined into single PR.

Of all parts only part1 is actually required for anything else than this dashboard. So I could split PR or just drop part2 and part3 if you suggest some better way.


relays/utils/Cargo.toml
```diff
@@ -10,10 +10,13 @@ ansi_term = "0.12"
 async-std = "1.6.5"
 async-trait = "0.1.40"
 backoff = "0.2"
+chttp = "0.5"
```
adoerr, line 13: Any specific reason for a new http client? Or in other words, any specfic reason not to use `hyper` which is a dependency of Substrate anyways?

relays/utils/src/metrics.rs
```diff
@@ -61,11 +75,64 @@ pub trait StandaloneMetrics: Metrics {
...
+		MetricsAddress {
 			host: "127.0.0.1".into(),
 			port: 9616,
 		}
 	}
 }
+
+impl MetricsParams {
+	/// Creates metrics params so that metrics are not exposed.
+	pub fn disabled() -> Self {
```
adoerr, line 89: `disabled()` should be used instead of `default()`, I guess.

relays/utils/src/relay_loop.rs
```diff
@@ -24,6 +24,9 @@ use substrate_prometheus_endpoint::{init_prometheus, Registry};
 /// Default pause between reconnect attempts.
 pub const RECONNECT_DELAY: Duration = Duration::from_secs(10);
 
+/// Generic relay loop metrics that may be reused by any loop.
+pub type GenericLoopMetrics = LoopMetrics<(), (), ()>;
```
adoerr, line 28: Looks like this alias isn't used anythere, so why have it at all. We can always add it, once it is acutally used.

adoerr (approved):
Some comments regarding making metric parameters mandatory. Overall, LGTM.

relays/bin-ethereum/src/ethereum_exchange.rs
```diff
@@ -67,7 +67,7 @@ pub struct EthereumExchangeParams {
 	/// Relay working mode.
 	pub mode: ExchangeRelayMode,
 	/// Metrics parameters.
-	pub metrics_params: Option<MetricsParams>,
+	pub metrics_params: MetricsParams,
```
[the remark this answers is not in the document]
svyatonik: `Option<S>` != `S { Option<>, Option<> }`. Fields may be `None` because of different reasons:
1) `address` is `None` when you don't need to expose metrics - either user has disabled them, or they are exposed somewhere else;
2) `registry` is `Some` when it's either shared by multiple loops, or some metrics were pre-registered. If it's `None` you're the only loop and shall use your own prefix when registry is created.

The structure has `disabled()` method, which may be used e.g. by tests. In binaries you first deal with `MetricsAddress`, which is both `Default` and is passed inside `Option<>` + there's `impl From<Option<MetricsAddress>> for MetricsParams`

relays/utils/Cargo.toml
svyatonik: Afair `hyper` requires tokio runtime, which we don't use + different `hyper` versions are requiring different tokio runtimes, making it harder to maintain. The `chttp` is runtime-agnostic => easier to use

svyatonik: > hyper which is a dependency of Substrate anyways?

We're not in the substrate binary ;)

relays/utils/src/metrics.rs
svyatonik: Yes, or `From<Option<MetricsAddress>>` - `MetricsAddress` is used in cli

relays/utils/Cargo.toml
adoerr: Good point, wasn't sure about `tokio`. Don't like it to much anyways :-)

relays/bin-ethereum/src/ethereum_exchange.rs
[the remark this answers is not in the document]
adoerr: Fair enough.

relays/client-substrate/src/metrics/float_storage_value.rs
```diff
@@ -0,0 +1,94 @@
...
+const UPDATE_INTERVAL_IN_BLOCKS: u32 = 5;
+
+/// Metric that represents fixed-point runtime storage value as float gauge.
+#[derive(Clone, Debug)]
+pub struct FloatStorageValueMetric<C: Chain, T: Clone> {
+	client: Client<C>,
+	storage_key: StorageKey,
+	maybe_default_value: Option<T>,
+	metric: Gauge<F64>,
+	_phantom: PhantomData<T>,
```
HCastano, line 38: Is this needed? `T` is already used by `maybe_default_value`

relays/client-substrate/src/metrics/float_storage_value.rs
```diff
@@ -0,0 +1,94 @@
...
+		C::AVERAGE_BLOCK_INTERVAL * UPDATE_INTERVAL_IN_BLOCKS
+	}
+
+	async fn update(&self) {
+		relay_utils::metrics::set_gauge_value(
+			&self.metric,
+			self.client
+				.storage_value::<T>(self.storage_key.clone())
+				.await
+				.map(|maybe_storage_value| {
```
HCastano, line 86: I think this can be turned into a `map_or().map()` chain

relays/utils/src/metrics/float_json_value.rs
```diff
@@ -0,0 +1,124 @@
+// Copyright 2019-2020 Parity Technologies (UK) Ltd.
```
HCastano, line 1:
```suggestion
// Copyright 2019-2021 Parity Technologies (UK) Ltd.
```

relays/bin-substrate/src/rialto_millau/westend_headers_to_millau.rs
```diff
@@ -61,13 +62,29 @@ pub async fn run(
...
 		westend_client,
-		millau_client,
-		metrics_params,
+		millau_client.clone(),
+		relay_utils::relay_metrics(
+			finality_relay::metrics_prefix::<WestendFinalityToMillau>(),
+			metrics_params.address,
+		)
+		.standalone_metric(FloatJsonValueMetric::new(
+			"https://api.coingecko.com/api/v3/simple/price?ids=Polkadot&vs_currencies=usd".into(),
```
HCastano, line 76: Is Millau not listed on CoinGecko yet? We should look into it 😏

On a more serious note, I would make a `NOTE` here indicating why we're using DOT/KSM price feeds here just like you noted in the PR description

HCastano (approved):
To answer some of your questions. I think having these extra admin metrics is totally fine. I also think it's fine to have them as part of the relayer instead of being done in a separate process.

I think it would make sense to split this into a different process if we notice that the overhead for collecting these metrics is impacting performance, usability, or maintainability in some way. Right now I don't think this is the case. Same goes for the `FloatJsonValueMetric` vs. Prometheus `json_exporter` question you asked.

I'm looking forward to seeing the Grafana dashboard with this info - that'll show us if this PR was actually worthwhile 😛

relays/utils/src/metrics.rs
```diff
@@ -61,11 +75,64 @@ pub trait StandaloneMetrics: Metrics {
...
+/// Set value of gauge metric.
+///
+/// If value is `Ok(None)` or `Err(_)`, metric would have default value.
+pub fn set_gauge_value<T: Default + Debug, V: Atomic<T = T>, E: Debug>(gauge: &Gauge<V>, value: Result<Option<T>, E>) {
+	gauge.set(match value {
+		Ok(Some(value)) => {
+			log::trace!(
+				target: "bridge-metrics",
+				"Updated value of metric '{:?}': {:?}",
+				gauge.desc().first().map(|d| &d.fq_name),
```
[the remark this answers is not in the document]
svyatonik: `desc` is a method of `Collector` trait, which may represent several metrics (e.g. gauge vec). In this case we deal with gauge, which only has single desc.

relays/utils/src/metrics.rs
```diff
@@ -14,23 +14,37 @@
...
 	pub host: String,
 	/// Serve HTTP requests at given port.
 	pub port: u16,
 }
 
+/// Prometheus endpoint MetricsParams.
+#[derive(Debug, Clone)]
+pub struct MetricsParams {
+	/// Interface and TCP port to be used when exposing Prometheus metrics.
+	pub address: Option<MetricsAddress>,
```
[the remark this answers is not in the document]
svyatonik: I've answered [here](https://github.com/paritytech/parity-bridges-common/pull/848#discussion_r605400767)

relays/client-substrate/src/metrics/mod.rs
```diff
@@ -0,0 +1,23 @@
+// Copyright 2019-2020 Parity Technologies (UK) Ltd.
```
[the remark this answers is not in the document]
svyatonik: Do you see any point to only update this only in modified files? If not, I suggest to do that in a separate PR, like it is done e.g. in Substrate/Polkadot.

relays/utils/src/metrics.rs
```diff
@@ -14,23 +14,37 @@
...
+use std::{fmt::Debug, time::Duration};
 
+mod float_json_value;
 mod global;
 
-/// Prometheus endpoint MetricsParams.
+/// Unparsed address that needs to be used to expose Prometheus metrics.
 #[derive(Debug, Clone)]
-pub struct MetricsParams {
+pub struct MetricsAddress {
```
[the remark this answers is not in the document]
svyatonik: Yes, probably. But it has default value, which is currently used by eth<->sub relay, so probably in a separate PR later.

svyatonik:
(this PR has a lot of conflicts with master + I believe a lot of conflicts with other opened relayer-related PRs, so I'm going to merge after these other PRs will be merged)

relays/client-substrate/src/metrics/mod.rs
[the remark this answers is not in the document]
HCastano: I've only been pointing this out for files which have been newly added in a PR.

I guess we could more closely follow the Substrate/Polkadot approach and `s/2020/2021/g` for all files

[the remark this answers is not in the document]
HCastano: I've opened #864 for this. No more bikeshedding from me here :)

PR #734, opened 2021-02-16
Rework relay CLI a bit

tomusdrw:
This PR extracts some common parts of the CLI commands to avoid duplication and also changes the way CLI is structured a bit.

Currently the CLI is flat:
```
$ ./target/release/substrate-relay --help
substrate-relay 0.1.0
Substrate-to-Substrate relay

USAGE:
    substrate-relay <SUBCOMMAND>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

SUBCOMMANDS:
    help                                          Prints this message or the help of the given subcommand(s)
    initialize-millau-headers-bridge-in-rialto    Initialize Millau headers bridge in Rialto
    initialize-rialto-headers-bridge-in-millau    Initialize Rialto headers bridge in Millau
    millau-headers-to-rialto                      Relay Millau headers to Rialto
    millau-messages-to-rialto                     Serve given lane of Millau -> Rialto messages
    rialto-headers-to-millau                      Relay Rialto headers to Millau
    rialto-messages-to-millau                     Serve given lane of Rialto -> Millau messages
    submit-millau-to-rialto-message               Submit message to given Millau -> Rialto lane
    submit-rialto-to-millau-message               Submit message to given Rialto -> Millau lane
```

IMHO it becomes already a bit difficult to parse long strings, and since I'm planning to add even more commands (see #607, #608, #609) I've changed it to have a bit more structure:

```
$ ./target/release/substrate-relay --help
substrate-relay 0.1.0
Substrate-to-Substrate relay

USAGE:
    substrate-relay <SUBCOMMAND>

FLAGS:
    -h, --help       
            Prints help information

    -V, --version    
            Prints version information


SUBCOMMANDS:
    help              Prints this message or the help of the given subcommand(s)
    init-bridge       Initialize on-chain bridge pallet with current header data
    relay-headers     Start headers relay between two chains
    relay-messages    Start messages relay between two chains
    send-message      Send custom message over the bridge
```

Initially I tried to split the CLI into network related commands (like `$ relay millau to-rialto headers`), but this quickly get's out of hand and is not user friendly.
Instead in the top level you will get different functions of the relayer binary first and then you see networks.

- [ ] I wanted to avoid doing too much changes, so the commands are still in a single module, but I think the next step might be extracting commands/networks/bridges into separate modules.

[...]

PR #849, opened 2021-03-30
Make relay CLI generic

tomusdrw:
The purpose of this PR is to make it easy to add new chains. It's a follow up on #734.

Initially I thought that we can orient the CLI around various bridges/chains and simply duplicate the code responsible for handling the CLI commands (see `src/rialto_millau` module). Later on I figure that it's quite a lot of code to duplicate, but was still hoping that we can get around this by using some macros.
However since the code is split into multiple files it's not that trivial to macro-ify the code + it heavily reduces readability and ability to debug errors.

This PR introduces a different approach, where we try to keep the code generic enough so that a lot of code can be shared without introducing a macro, but all the things that are difficult to make generic, or depend on particular sourceNetwork-targetNetwork bridge pair are resolved via macro-shared code.

The most important part of the PR is the `select_bridge!` macro, which allows us to write some `$generic` code, without actually naming all of the types (which would be hard).
I've refactored `RelayHeaders` and `InitBridge` for now, but since the PR is getting quite big, I've decided to make it a draft and let you decide if I should continue or if we rather review & merge like this and change the rest in a follow up PR.

There is a bunch of unaddressed and unassigned `TODOs` for now, and most likely some bash scripts require updating as well.

Follow up tasks:
- [ ] Remove target-specific stuff from CliChain.
- [ ] Support multiple bridges in BridgeSendMessage Call.
- [ ] Remove `source_*` `target_*` CLI methods (e.g. add `Into<ConnectionParams>`)
- [ ] Consider using structopt renames for Singing and Connection parameters.
- [ ] Create arg-enum-like macro with kebab-case support or use existing proc-macro crates.


relays/bin-substrate/src/cli/mod.rs
```diff
@@ -302,15 +290,48 @@ impl AccountId {
...
+	const RUNTIME_VERSION: sp_version::RuntimeVersion;
+
+	/// Crypto keypair type used to send messages.
+	///
+	/// In case of chains supporting multiple cryptos, pick one used by the CLI.
+	type KeyPair: sp_core::crypto::Pair;
+
+	/// Bridge Message Payload type.
+	///
+	/// TODO [ToDr] This should be removed in favour of target-specifc types.
```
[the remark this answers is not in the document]
tomusdrw: In case we have multiple bridges the `MessagePayload` type will be different.
For instance for `RialtoToWestend` and `RialtoToMillau` the `MessagePayload` and `encode_call` will look quite differently.
The TODO is there cause it's not the right way to abstract the code, however since right now we don't have a chain with two bridges that include MessageLanes it's good enough, but I already have an idea how to improve it, hence the TODO.

relays/bin-substrate/src/cli/mod.rs
```diff
@@ -302,15 +290,48 @@ impl AccountId {
...
+	type MessagePayload;
 
 	/// Numeric value of SS58 format.
 	fn ss58_format() -> u16;
+
+	/// Parse CLI call and encode it to be dispatched on this specific chain.
+	fn encode_call(call: crate::rialto_millau::cli::Call) -> Result<Self::Call, String>;
+
+	/// Construct message payload to be sent over the bridge.
+	fn encode_message(message: crate::rialto_millau::cli::MessagePayload) -> Result<Self::MessagePayload, String>;
```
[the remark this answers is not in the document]
tomusdrw: Same there.

relays/bin-substrate/src/rialto_millau/mod.rs
```diff
@@ -769,59 +682,80 @@ impl cli::ToRialtoMessage {
...
-		Ok(call)
+	fn target_signing_params(params: TargetSigningParams) -> Result<Self::KeyPair, String> {
+		Self::KeyPair::from_string(&params.target_signer, params.target_signer_password.as_deref())
+			.map_err(|e| format!("Failed to parse target-signer: {:?}", e))
 	}
 }
 
-impl AccountChain for Rialto {
-	const NAME: &'static str = "Rialto";
+impl CliChain for Westend {
```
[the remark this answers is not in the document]
tomusdrw: How so? We don't really support Westend, the fact that we can't implement all methods means the abstraction is not fine-grained enough (i.e. there are sub-commands that are not even available for westend, but the trait is shared between sub-commands), but I would rather go with coarse abstraction and later on split it if it turns out to be correct.

relays/bin-substrate/src/rialto_millau/cli.rs
```diff
@@ -382,47 +291,11 @@ pub enum ToRialtoMessage {
...
-		recipient: AccountId,
-		/// Amount of target tokens to send in target chain base currency units.
-		#[structopt(long)]
-		amount: bp_millau::Balance,
+		amount: Balance,
 	},
-	/// A call to the Rialto Bridge Messages pallet to send a message over the bridge.
-	RialtoSendMessage {
+	// TODO [#853] Support multiple bridges.
+	/// A call to the specific Message Lane pallet to send a message over the bridge.
```
svyatonik, line 297: Quote from #853:

```
Currently the Call CLI enum, supports BridgeSendMessage variant. However in case there are multiple bridges on one source chain, we need to specify what Target chain we want to send the message to.
```

I'm a bit confused now - I thought that this variant is to call over the bridge the other bridge (deployed on the target chain) to send other message. So it may be `Source --- (M1) ---> Target --- (M2) ---> Source` or `Source1 --- (M1) ---> Target --- (M2) ---> Source2` iiuc. So the quote should be the "multiple bridges on the target chain", no? 

I guess it was introduced in the PR about printing encoded messages, so it isn't actually used in messages sending. In practice it is only used by the `encode-call` subcommand. But since it is available to the `send-message` command as well, maybe it worth at least to expand docs a bit - e.g instead of "specific Message Lane pallet" say "Message Lane pallet on the target chain", for payload change it to the "encoded Call of the source chain", etc.

P.S.: 'Message Lane pallet' is also changed to 'Messages pallet' everywhere ;)

svyatonik (approved):

relays/bin-substrate/src/cli/init_bridge.rs
```diff
@@ -0,0 +1,137 @@
...
+	/// Bridge to initialize.
+	pub enum InitBridgeName {
+		MillauToRialto,
+		RialtoToMillau,
+		WestendToMillau,
+	}
+}
+
+macro_rules! select_bridge {
+	($bridge: expr, $generic: tt) => {
```
adoerr, line 51: `generic` is generic here in the sense that it can be an arbitrary closure?

relays/bin-substrate/src/cli/relay_headers.rs
```diff
@@ -0,0 +1,91 @@
...
+	/// Headers relay bridge.
+	pub enum RelayHeadersBridge {
+		MillauToRialto,
+		RialtoToMillau,
+		WestendToMillau,
+	}
+}
+
+macro_rules! select_bridge {
+	($bridge: expr, $generic: tt) => {
```
adoerr, line 48: Same here w.r.t. to `generic`.

relays/client-substrate/src/error.rs
```diff
@@ -42,6 +42,19 @@ pub enum Error {
 	Custom(String),
 }
 
+impl std::error::Error for Error {
```
adoerr, line 45: Perhaps we can agree on using the `thiserror` / `anyhow` combo for all our errors and start migrate towards this.

adoerr (commented):
For me, this goes into the right direction. If anything, I want to see more of this approach. My kind of ideal in the end would be having a `bridge` procedural macro, similar to the `pallet` macro. Basically generating a specific bridge implementation, batteries (CLI) included.  

So I 'approve' but keep this as comment only.

relays/bin-substrate/src/rialto_millau/cli.rs
tomusdrw: Yes, that's a good point. Indeed I was thinking about it from the `encode-call` perspective, I'll update the docs.

relays/bin-substrate/src/cli/relay_headers.rs
tomusdrw: It's actually not a closure, but a piece of arbitrary code that will use generic parameters `type Source =/ type Target = ` that is set in the particular branch of `match` statement.

relays/client-substrate/src/error.rs
tomusdrw: Yes, logged here https://github.com/paritytech/parity-bridges-common/issues/857

Issue #855, opened 2021-03-31
Move CLI commands to own modules and DRY them.


tomusdrw:
Follow up of #849 

Sub-commands to convert:
- [ ] `RelayMessages`
- [ ] `EncodeCall`
- [ ] `SendMessage`
- [ ] `EncodeMessagePayload`
- [ ] `EstimateFee`
- [ ] `DeriveAccount`

tomusdrw:
After either of #859 and #860 being merged I'll extract the common bridge enum and `select_bridge` macro to avoid duplication.

tomusdrw:
I believe this is now all solved?

Issue #854, opened 2021-03-31
Fix `CliChain` abstraction


tomusdrw:
Follow up of #849.

After the PR lands, `CliChain` trait includes things like `type MessagePayload` and `fn encode_call/encode_message`. The problem with this abstraction is that it's not generic enough (see #853), it only works cause currently we don't have a bridge which can send messages to two different target chains.

The goal is to fix this abstraction - most likely it's not `CliChain` responsibility at all I think this can be nicely generalised with the `select_bridge` macro and some structural code unification.

svyatonik:
Irrelevant

PR #859, opened 2021-04-02
CLI: Encode Call & Multiple Bridge Instances.

### LINKED ISSUE #853: Support sending messages to multiple bridges.
Follow up of #849

Currently the `Call` CLI enum, supports `BridgeSendMessage` variant. However in case there are multiple bridges on one source chain, we need to specify what `Target` chain we want to send the message to.
The code to generate such message will then need to be target-specific (see #854).

The problem is that target-chain possibilities are specific to the source chain (i.e. not all pairs are valid).
I see two possible solutions for now:
1. Add a generic `BridgeInstance` parameter, which can either be an instance index of stringified chain name. (both are a bit clunky usage-wise)
2. Change sub-commands that generate calls (send-message, encode-call) to select bridge instead of single network.

I like (2) the most, I think for `send-message` we already know the target chain, so perhaps just a matter of changing `encode-call` API.

tomusdrw:
Related #855
Related #854 
Closes #853  

This PR moves `encode-call` sub-command to it's own module and adds the `select_bridge` macro logic as well.
Additionally it removes `CliChain::encode_call` (which was a temporary measure anyway) and adds a separate trait for encoding calls. Instead of selecting a network, this command is now unified and requires a bridge as well to have full bridge context. (Temporarily) `preprocess_call` method is introduced to fill extra parameters that are related to specific instance of the bridge, currently we simply pass `bridge_pallet_index` from the context where we know which bridge we operate in.

Following: rewrite of `send-message` & `encod-message-payload` sub-commands and possibly fully solving #854 - I think after this `preprocess_call` will also not be required.


relays/bin-substrate/src/cli/encode_call.rs
```diff
@@ -0,0 +1,248 @@
...
+		recipient: AccountId,
+		/// Amount of target tokens to send in target chain base currency units.
+		#[structopt(long)]
+		amount: Balance,
+	},
+	/// A call to the specific Bridge Messages pallet to queue message to be sent over a bridge.
+	BridgeSendMessage {
+		/// An index of the pallet to encode.
+		#[structopt(skip = 255)]
+		pallet_index: u8,
```
tomusdrw, line 65: That's arguably a bit hacky, but the only alternative I see is to have two separate enums (one for CLI and another one for `encode_call`) potentially generated by a macro to avoid duplication.

relays/bin-substrate/src/cli/encode_call.rs
```diff
@@ -0,0 +1,248 @@
...
+	let service_bytes = 1 + 1 + 4;
+	maximal_call_size - service_bytes
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+
+	#[test]
+	fn should_encode_transfer_call() {
```
HCastano, line 227: Is it worth having tests for the other supported calls?

relays/bin-substrate/src/cli/encode_call.rs
```diff
@@ -0,0 +1,248 @@
...
+
+			Ok(encoded)
+		})
+	}
+
+	/// Run the command.
+	pub async fn run(mut self) -> anyhow::Result<()> {
+		println!("{:?}", self.encode()?);
+		Ok(())
+	}
```
HCastano, line 148: Does this really need to be `async`?

HCastano (commented):
I left a couple of questions

relays/bin-substrate/src/cli/encode_call.rs
```diff
@@ -0,0 +1,248 @@
...
+	fn pallet_index(&self) -> u8 {
+		match self {
+			Self::MillauToRialto => MILLAU_TO_RIALTO_INDEX,
+			Self::RialtoToMillau => RIALTO_TO_MILLAU_INDEX,
+		}
+	}
+}
+
+pub const RIALTO_TO_MILLAU_INDEX: u8 = 0;
+pub const MILLAU_TO_RIALTO_INDEX: u8 = 0;
```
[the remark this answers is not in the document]
tomusdrw: These depend on the runtime configuration of specfiic source chains. So if `Millau` is the first bridge pallet in `Rialto` thn it's `0` index, if `Rialto` was also bridged with, say, `Westend` then `Westend` would have index 1.

relays/bin-substrate/src/cli/encode_call.rs
tomusdrw: No, but it makes the code in `src/cli/mod.rs` nicer, cause all sub-commands are expected to be async. If I could I would make it a trait, but `async fn` traits are not supported yet.

relays/bin-substrate/src/cli/encode_call.rs
tomusdrw: Yes, although note that there were no tests earlier and it's not strictly the goal of this PR. Will log it down and someone can pick it up (https://github.com/paritytech/parity-bridges-common/issues/863)

relays/bin-substrate/src/cli/encode_call.rs
HCastano: There is the [`async_trait`](https://github.com/dtolnay/async-trait) crate.

But I guess this is fine as it is, we can always change it in a different PR

relays/bin-substrate/src/cli/encode_call.rs
HCastano: Thanks for the write up. The relationship between `bridge` and `bridge_instance_index` makes more sense now.  

Changing `pallet_index` to `bridge_instance_index` is definitely a good call. Some of the confusion was also the use of `0` for `bridge_instance_index`, but it's usage should become more clear as we add more bridges with different indices.

HCastano (approved):

PR #860, opened 2021-04-02
CLI: Derive Account

tomusdrw:
Related #855 

This PR moves `derive-account` sub-command to its own module.

### DIFF
--- a/relays/bin-substrate/src/cli/derive_account.rs
+++ b/relays/bin-substrate/src/cli/derive_account.rs
@@ -1,0 +1,129 @@
+// Copyright 2019-2021 Parity Technologies (UK) Ltd.
+// This file is part of Parity Bridges Common.
+
+// Parity Bridges Common is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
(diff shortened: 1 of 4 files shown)


relays/bin-substrate/src/cli/derive_account.rs
```diff
@@ -0,0 +1,129 @@
...
+	/// Parse CLI arguments and derive account.
+	///
+	/// Returns both the Source account in correct SS58 format and the derived account.
+	fn derive_account(&self) -> (AccountId, AccountId) {
+		select_bridge!(self.bridge, {
+			let mut account = self.account.clone();
+			account.enforce_chain::<Source>();
+			let acc = bp_runtime::SourceAccount::Account(account.raw_id());
+			let id = bp_millau::derive_account_from_rialto_id(acc);
+			let derived_account = AccountId::from_raw::<Target>(id);
```
HCastano, line 74: Should't this be part of the macro code? For example, in the case of Millau to Rialto we would want:

```rust
let id = bp_rialto::derive_account_from_millau_id(acc);
```

tomusdrw: Great catch, I trusted myself too much that after #849 the code sections are properly split. It didn't trigger any compilation errors, cause the account types are the same on both chains.

HCastano (approved):

Issue #857, opened 2021-04-01
Unify error enums & generation


tomusdrw:
Currently we mostly use manually implemented `Error` enums across the codebase. These enums do not necessarily implement `std::error::Error` trait, which is fine for runtime stuff, but would be useful for relayer code.
The relayer code is using `String` errors though, cause we mostly don't care about anything specific, but need to propagate it up in the stack.

Recently in #813 `anyhow` got introduced and now with #849 there is even more error-specific refactorings being done.
@adoerr [suggested](https://github.com/paritytech/parity-bridges-common/pull/849#discussion_r605403959) to use `anyhow / thiserror` combo to unify error handling which I think is the right idea.

Basically the rule is:
- `anyhow` for the last mile CLI stuff, where we really don't care about the error.
- `thiserror` for non-cli libraries that are used outside of the runtime to avoid manual `Display/std::io::Error` implementation.

tomusdrw:
Assuming closed with #1094

PR #861, opened 2021-04-02
CLI: De-duplicate signing/connection params.

### LINKED ISSUE #852: De-duplicate `Source/Target Signing/Connection Params`
Follow up on #849 

Currently `Source/Target Signing/Connection Params` are generated via macro and have different field names and they represent different types to have structopt distinguish CLI flags.

Perhaps it's possible to avoid having two different types and use `structopt` rename feature to make sure that both kinds of flag can be used in one sub-command..
Alternatively we can keep distinct Source/Traget types but unify field names and use rename to have them be different CLI flags.

The latter can help with removing `source|target_chain_client` and `source|target_signing_params` methods duplication. Perhaps a `Into<ConnectionParams>` and `TryInto<KeyPair>`  could be implemented?

Also open for any other solution that removes either macro or duplication or best both :)

tomusdrw:
Resolves #852 
Related #857

### DIFF
--- a/relays/bin-substrate/src/cli/init_bridge.rs
+++ b/relays/bin-substrate/src/cli/init_bridge.rs
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
 
-use crate::cli::{CliChain, SourceConnectionParams, TargetConnectionParams, TargetSigningParams};
+use crate::cli::{SourceConnectionParams, TargetConnectionParams, TargetSigningParams};
 use bp_runtime::Chain as ChainBase;
 use codec::Encode;
 use pallet_bridge_grandpa::InitializationData;
@@ -108,10 +108,9 @@ impl InitBridge {
 	/// Run the command.
 	pub async fn run(self) -> anyhow::Result<()> {
 		select_bridge!(self.bridge, {
-			let source_client = crate::rialto_millau::source_chain_client::<Source>(self.source).await?;
-			let target_client = crate::rialto_millau::target_chain_client::<Target>(self.target).await?;
-			let target_sign =
-				Target::target_signing_params(self.target_sign).map_err(|e| anyhow::format_err!("{}", e))?;
+			let source_client = self.source.into_client::<Source>().await?;
+			let target_client = self.target.into_client::<Target>().await?;
+			let target_sign = self.target_sign.into_keypair::<Target>()?;
 
 			crate::headers_initialize::initialize(
 				source_client,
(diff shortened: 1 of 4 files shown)


relays/bin-substrate/src/cli/init_bridge.rs
```diff
@@ -108,10 +108,9 @@ impl InitBridge {
 	/// Run the command.
 	pub async fn run(self) -> anyhow::Result<()> {
 		select_bridge!(self.bridge, {
-			let source_client = crate::rialto_millau::source_chain_client::<Source>(self.source).await?;
-			let target_client = crate::rialto_millau::target_chain_client::<Target>(self.target).await?;
-			let target_sign =
-				Target::target_signing_params(self.target_sign).map_err(|e| anyhow::format_err!("{}", e))?;
+			let source_client = self.source.into_client::<Source>().await?;
+			let target_client = self.target.into_client::<Target>().await?;
+			let target_sign = self.target_sign.into_keypair::<Target>()?;
```
HCastano, line 113: Slightly off-topic, but how does the Rust compiler know to resolve macros in the correct order?

For example, the `into_client()` and `into_keypair()` methods only exist after the `declare_chain_options!` macro has been evaluated. How does the compiler know to evaluate `declare_chain_options!` first and only after that code's been generated evaluate `select_bridge!`?

relays/bin-substrate/src/rialto_millau/mod.rs
```diff
@@ -57,10 +54,10 @@ async fn run_relay_messages(command: cli::RelayMessages) -> Result<(), String> {
 			type Source = Millau;
 			type Target = Rialto;
 
-			let source_client = source_chain_client::<Source>(source).await?;
-			let source_sign = Source::source_signing_params(source_sign)?;
-			let target_client = target_chain_client::<Target>(target).await?;
-			let target_sign = Target::target_signing_params(target_sign)?;
+			let source_client = source.into_client::<Source>().await.map_err(format_err)?;
```
HCastano, line 57: Could we not change `run_relay_messages` to return an `anyhow::Result` here instead?

HCastano (approved):

relays/bin-substrate/src/cli/init_bridge.rs
tomusdrw: Macros are expanded before the compilation, so the code is generated first and only then compiled, so it's not really an issue here.

The interesting part is in what order macros are declared and resolved, and interestingly it's simply top-down in the order they are listed. (i.e. `macro_rules` need to be before usage).

relays/bin-substrate/src/rialto_millau/mod.rs
tomusdrw: Yes, but `run_relay_messages` is removed in #858 anyway, so I didn't want to cause more conflicts.

relays/bin-substrate/src/cli/init_bridge.rs
HCastano: > Macros are expanded before the compilation, so the code is generated first and only then compiled, so it's not really an issue here.

Actually yeah, that makes a lot of sense

> The interesting part is in what order macros are declared and resolved, and interestingly it's simply top-down in the order they are listed. (i.e. macro_rules need to be before usage).

This is kinda dumb, but given the first thing you said also makes sense

PR #862, opened 2021-04-03
Use Vote Ancestries Directly in Weight Calculations

HCastano:
This is a follow up to #815, in which I introduced benchmarks for
`pallet-bridge-grandpa`. In #815 I was calculating weight based on: number of
pre-commits, and number of headers in a _single_ fork. This approach turned out to be a
bit unintuitive, and led to some bugs during weight refunds which were luckily spotted by
Tomek.

The main thing this PR does is change the weight calculation to use the _total_ number of
vote ancestries in a justification instead of the _average_ number for a single fork.
This, alongside the API changes in #847, allows us to more easily and accurately declare
weights for the `submit_finality_proof` call.

One thing to note is that we don't use the `forks` parameter as part of the weight
calculations anymore. Pre-commit signature verification is really a function of the
number of authorities, and `vote_ancestries` are a single `Vec<Header>` anyways, so I
don't think the effect of `forks` is significant. However, I do want to follow up on this
with some different benchmarks just to double check.


primitives/test-utils/src/lib.rs
```diff
@@ -87,23 +94,34 @@ pub fn make_justification_for_header<H: HeaderT>(params: JustificationGeneratorP
...
 
-	assert!(depth != 0, "Can't have a chain of zero length.");
+	assert!(votes >= forks, "Need at least one header per fork.");
 	assert!(
 		forks as usize <= authorities.len(),
 		"If we have more forks than authorities we can't create valid pre-commits for all the forks."
 	);
 
+	// Roughly, how many vote ancestries do we want per fork
+	let target_depth = (votes + forks - 1) / forks;
```
adoerr, line 112: Since the divisor (`forks`) is a parameter, this should perhaps be safeguarded against accidental division by '0'.

adoerr (approved):
One minor grumble, other than that, LGTM

modules/grandpa/src/lib.rs
```diff
@@ -141,8 +119,8 @@ pub mod pallet {
 		/// If successful in verification, it will write the target header to the underlying storage
 		/// pallet.
 		#[pallet::weight(T::WeightInfo::submit_finality_proof(
-			T::MaxBridgedSessionLength::get().as_() as u32,
-			T::MaxBridgedValidatorCount::get(),
+			justification.commit.precommits.len() as u32,
```
svyatonik, line 122: Is this correct order of arguments? From the trait it seems that the first argument is `v` (`let v in 1..MAX_VOTE_ANCESTRIES;`).

svyatonik (approved):

modules/grandpa/src/lib.rs
HCastano: Oh wow, you're right these are flipped. Good catch!

PR #864, opened 2021-04-05
Update Copyright Date to 2021

HCastano:
This bumps the date on our license to 2021 throughout the repo. With this change we
should also be mindful to update the date every year, which we weren't quite doing before
(date was only being bumped for new files).

### DIFF
--- a/bin/millau/node/build.rs
+++ b/bin/millau/node/build.rs
@@ -1,4 +1,4 @@
-// Copyright 2019-2020 Parity Technologies (UK) Ltd.
+// Copyright 2019-2021 Parity Technologies (UK) Ltd.
 // This file is part of Parity Bridges Common.
 
 // Parity Bridges Common is free software: you can redistribute it and/or modify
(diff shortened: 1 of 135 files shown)


tomusdrw:
AFAIR, it's totally fine to update copyright only if the content changes, it's even fine to not update the date it just means it will be copyrighted a bit shorter than it should have. TBH I'm quite sick and tired with the date stuff (note it really does not matter for copyright laws afaik, but ianal) and I'd prefer if it was just uniform across the whole codebase and updated with a script every year (or derived from git history).

HCastano:
> AFAIR, it's totally fine to update copyright only if the content changes, it's even fine to not update the date it just means it will be copyrighted a bit shorter than it should have. TBH I'm quite sick and tired with the date stuff (note it really does not matter for copyright laws afaik, but ianal) 

The main motivation here isn't how "correct" this is w.r.t copyright laws (but I don't think this is totally wrong, since it's what Substrate and Polkadot do), but more so consistency in process. Sometimes people copy a header and forget to update the date, or files are moved and in ambiguous as to whether or not the date should change. By bumping everything to `$CURRENT_YEAR` in one go its unambiguous what a developer should do.

> I'd prefer if it was just uniform across the whole codebase and updated with a script every year (or derived from git history).

It's pretty uniform right now, what inconsistencies are you referring to?

svyatonik (approved):

adoerr:
What I'm wondering is, whether the whole copyright notice actually has to be present in each and every source file. An alternative would be to just have the SPDX license identifier and a reference/pointer to the actual copyright notice. This way, only one file would have to be updated.

PR #865, opened 2021-04-05
Fix SS58 Warnings in Message Generator Scripts

HCastano:
~~The message generator scripts got missed with some recent CLI updates. This updates them
to make sure they indicate a `remark-payload` when neccessary.~~

Tomek sniped these changes (and the fixes suggested by Slava) with #868.

Gonna salvage my green square by morphing this into just fixing some warnings related to wrong SS58 account encodings.


deployments/bridges/rialto-millau/entrypoints/relay-messages-to-millau-generator-entrypoint.sh
```diff
@@ -75,6 +78,7 @@ do
 			--lane $MESSAGE_LANE \
 			--origin Target \
 			remark \
+			--remark-payload $PAYLOAD \
```
svyatonik, line 81: There are a few issues with this (which are mostly introduced in other PRs, but probably must be fixed here):
1) `remark-payload` is marked as [`conflicts_with("remark_size")`](https://github.com/paritytech/parity-bridges-common/blob/4d30e66cd7c322a81206a0e2eac21e9d295cca8d/relays/bin-substrate/src/cli/encode_call.rs#L46), so you can't (actually can, but see (2)) use both here - in this case the `--remark_size` [will be ignored](https://github.com/paritytech/parity-bridges-common/blob/4d30e66cd7c322a81206a0e2eac21e9d295cca8d/relays/bin-substrate/src/cli/encode_call.rs#L168);
2) [`conflicts_with("remark_size")`](https://github.com/paritytech/parity-bridges-common/blob/4d30e66cd7c322a81206a0e2eac21e9d295cca8d/relays/bin-substrate/src/cli/encode_call.rs#L46) and [`conflicts_with("remark_payload")`](https://github.com/paritytech/parity-bridges-common/blob/4d30e66cd7c322a81206a0e2eac21e9d295cca8d/relays/bin-substrate/src/cli/encode_call.rs#L168) should actually be `conflicts_with("remark-size")` and `conflicts_with("remark-payload")` (i.e. underscore -> hyphen). Right now this declaration doesn't work;
3) since `remark_payload` is now required, I don't see any point in keeping [this payload generation](https://github.com/paritytech/parity-bridges-common/blob/4d30e66cd7c322a81206a0e2eac21e9d295cca8d/relays/bin-substrate/src/cli/encode_call.rs#L191). But actually I'd prefer to keep this and mark `remark_payload` as `Optional<>`.

svyatonik (approved):

tomusdrw (changes requested):
That's incorrect. `remark-payload` should not be required, it should be possible to do either:
```bash
$0 remark # default remark payload generated
$0 remark --remark-size <SIZE> # remark payload of given size is generated
$0 remark --remark-payload <BYTES> # explicit remark payload is being sent
``` 

I'm pretty sure I tested that, but will take a look again and add some test cases to make sure it works correctly, but the change was meant to be 100% backwards compatible.

EDIT: Indeed it does not work correctly, I don't know what the hell I was testing :D #868 to the rescue.

tomusdrw (approved):

PR #866, opened 2021-04-05
Remove Substrate Pallet

HCastano:
With the addition of #628 and #783 we have no need for `pallet-substrate-bridge` anymore.
`pallet-bridge-grandpa` covers all our requirements while providing a safe interface. The
time has come to (again) remove `pallet-substrate-bridge` (I say again because I think
this is my second or third PR which `git rm`'s it, lol).

CC @MaciejBaj in case you're still using this.


relays/bin-substrate/src/cli/encode_call.rs
```diff
@@ -243,7 +243,7 @@ mod tests {
 		// then
 		assert_eq!(
 			format!("{:?}", hex),
-			"0x0d00d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27de5c0"
+			"0x0c00d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27de5c0"
```
HCastano, line 246: Tbh I don't understand why this changed, does anyone else know?

svyatonik: `Call` is an `enum`, where every variant is a single pallet - e.g. for some runtime it may be `enum Call { System, BridgeSubstrate BridgeGrandpa, BridgeMessages, .. }`. You've removed pallet `BridgeSubstrate` from the runtime - next pallets indices have changed (here: from `0x0d` to `0x0c`) => encoding has changed. That's the reason why we can't simply insert pallets in the middle of `Call`, why we can't add callable methods before callable methods and can't non-trivially change structs used in callable method arguments without bumping runtime `transaction_version`.

svyatonik (approved):

adoerr (approved):
LGTM

PR #867, opened 2021-04-06
Workaround buildah.

tomusdrw:
Based on a workaround proposed here:
https://github.com/redhat-actions/buildah-build/issues/45#issuecomment-812043531


svyatonik (approved):
Hope it'll work

Issue #863, opened 2021-04-03
Add CLI tests for other `Call` variants


tomusdrw:
Follow up on #859

Context: https://github.com/paritytech/parity-bridges-common/pull/859/files#r606399268

oxarbitrage:
Hey, i was wondering if this was resolved fully by https://github.com/paritytech/parity-bridges-common/pull/868 or if there are still tests to be done.

tomusdrw:
So AFAICT there are tests for `Remark` and `Transfer`, but we still have `Raw` and `BridgeSendMessage` in the enum, so might be worth adding tests for these.

svyatonik:
irrelevant after #1376

PR #868, opened 2021-04-06
Fix `encode-call` CLI options

tomusdrw:
Related #863 
Replaces #865

I've added a bunch of remark tests to make sure that the CLI parsing works correctly. The change was supposed to be backward compatible, but I've incorrectly set `conflicts_with` and forgot about `default_value`.


relays/bin-substrate/src/cli/encode_call.rs
```diff
@@ -246,4 +246,72 @@ mod tests {
...
+	#[test]
+	fn should_encode_remark_with_default_payload() {
+		// given
+		let mut encode_call = EncodeCall::from_iter(vec!["encode-call", "RialtoToMillau", "remark"]);
+
+		// when
+		let hex = encode_call.encode().unwrap();
+
+		// then
+		assert!(format!("{:?}", hex).starts_with("0x080154556e69782074696d653a"));
```
svyatonik, line 259: Default payload is using current time, right? So this test will always fail

svyatonik (approved):

relays/bin-substrate/src/cli/encode_call.rs
tomusdrw: Note that this one is not exact, but only checks the beginning (`.starts_with`). The timestamp comes after some prefix, so we only check the prefix.

HCastano (approved):

PR #870, opened 2021-04-06
Fix Undeterministic CLI Encoding Test

HCastano:
Looks like a test introduced in #868 is not deterministic. You can see our `master` build
failed [here](https://github.com/paritytech/parity-bridges-common/runs/2279623635) even though the tests in the PR passed.

This PR updates the test to check for both possible error variants, and as long as we
match on one the test will be considered successful.

### DIFF
--- a/relays/bin-substrate/src/cli/encode_call.rs
+++ b/relays/bin-substrate/src/cli/encode_call.rs
@@ -268,12 +268,8 @@ mod tests {
 
 		// then
 		assert_eq!(err.kind, structopt::clap::ErrorKind::ArgumentConflict);
-		assert_eq!(
-			err.info,
-			Some(vec![
-				"remark-size".to_string(),
-				"--remark-payload <remark-payload>".to_string()
-			])
-		);
+
+		let info = err.info.unwrap();
+		assert!(info.contains(&"remark-payload".to_string()) | info.contains(&"remark-size".to_string()))
 	}
 }


relays/bin-substrate/src/cli/encode_call.rs
```diff
@@ -268,12 +268,8 @@ mod tests {
...
-		assert_eq!(
-			err.info,
-			Some(vec![
-				"remark-size".to_string(),
-				"--remark-payload <remark-payload>".to_string()
-			])
-		);
+
+		let info = err.info.unwrap();
+		assert!(info.contains(&"remark-payload".to_string()) | info.contains(&"remark-size".to_string()))
```
adoerr, line 273: Should the `err.info` contain both, `remark-payload` and `remark-size`. So, it would be `&` instead of `|`.

adoerr (approved):
LGTM

relays/bin-substrate/src/cli/encode_call.rs
tomusdrw: Actually I think it would be best to simply sort both vectors before comparing.

tomusdrw (approved):

relays/bin-substrate/src/cli/encode_call.rs
HCastano: In our case `err.info` will be either `Some(["remark-payload", "--remark-size <remark-size>"])` or `Some(["remark-size", "--remark-payload <remark-payload>"])`.

This means we can't check for both variants (e.g the first vector doesn't `contain` "remark-size"), and because of this sorting the vector won't help.

PR #875, opened 2021-04-08
Westend<>Rococo Headers Relay

HCastano:
This PR lays the groundwork for the Rococo<>Westend relay.

~Still in progress since I need to figure out how to make a mock generic Call type for
`submit_finality_proof` (and probably fix some other issues which I haven't encountered
of yet).~


tomusdrw:
I think it's ready for review, let's get the bridge out :)

primitives/chain-rococo/src/lib.rs
```diff
@@ -42,8 +43,26 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 };
 
 #[derive(parity_scale_codec::Encode, parity_scale_codec::Decode, Debug, PartialEq, Eq, Clone)]
+#[allow(non_camel_case_types)]
 pub enum Call {
-	MockPallet,
+	/// Westend bridge pallet.
+	///
+	/// Note this index has to be kept in-sync with Rococo runtime configuration:
+	/// https://github.com/paritytech/polkadot/blob/master/runtime/rococo/src/lib.rs#L272
```
HCastano, line 51: I don't think this points to what you want it to point to. 

I imagine you want this to point to the Westend bridge pallet instance of Rococo's `construct_runtime!` macro, but it's pointing elsewhere.

We may need to be more explicit in the comment, or tag it to a commit.

primitives/chain-rococo/src/lib.rs
```diff
@@ -42,8 +43,26 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
...
+	#[codec(index = 40)]
+	BridgeGrandpaWestend(BridgeGrandpaWestendCall),
+}
+
+#[derive(parity_scale_codec::Encode, parity_scale_codec::Decode, Debug, PartialEq, Eq, Clone)]
+#[allow(non_camel_case_types)]
+pub enum BridgeGrandpaWestendCall {
+	#[codec(index = 0)]
```
HCastano, line 59: We should have a similar comment here. For example, what if somebody flips the extrinsic declarations in the pallet and doesn't update this?

primitives/chain-westend/src/lib.rs
```diff
@@ -43,8 +44,26 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 };
 
 #[derive(parity_scale_codec::Encode, parity_scale_codec::Decode, Debug, PartialEq, Eq, Clone)]
+#[allow(non_camel_case_types)]
 pub enum Call {
-	MockModule,
+	/// Rococo bridge pallet.
+	///
+	/// Note this index has to be kept in-sync with westend runtime configuration:
+	/// https://github.com/paritytech/polkadot/blob/master/runtime/westend/src/lib.rs#L809
```
HCastano, line 52: Same thing about the link location

primitives/chain-rococo/src/lib.rs
```diff
@@ -42,8 +43,26 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 };
 
 #[derive(parity_scale_codec::Encode, parity_scale_codec::Decode, Debug, PartialEq, Eq, Clone)]
+#[allow(non_camel_case_types)]
 pub enum Call {
```
HCastano, line 47: We should have a doc comment here with what our goals are for this enum for future readers.

E.g this should be SCALE compatible with what's in the runtime without necessarily using the runtime

HCastano (commented):
I can't approve my own PR, but looks good :)

relays/bin-substrate/src/rialto_millau/rococo_headers_to_westend.rs
```diff
@@ -0,0 +1,79 @@
...
+/// Rococo-to-Westend finality sync pipeline.
+pub(crate) type RococoFinalityToWestend = SubstrateFinalityToSubstrate<Rococo, Westend, WestendSigningParams>;
+
+impl SubstrateFinalitySyncPipeline for RococoFinalityToWestend {
+	const BEST_FINALIZED_SOURCE_HEADER_ID_AT_TARGET: &'static str = bp_rococo::BEST_FINALIZED_ROCOCO_HEADER_METHOD;
+
+	type TargetChain = Westend;
+
+	fn customize_metrics(params: MetricsParams) -> anyhow::Result<MetricsParams> {
+		Ok(
```
svyatonik, line 38: I'd either remove them from Westend -> Millau headers relay (would need a change in Grafana/Prometheus config), or from here. Why would we need two samples of "just-for-test" metrics?

relays/bin-substrate/src/rialto_millau/westend_headers_to_rococo.rs
```diff
@@ -0,0 +1,79 @@
...
+/// Westend-to-Rococo finality sync pipeline.
+pub(crate) type WestendFinalityToRococo = SubstrateFinalityToSubstrate<Westend, Rococo, RococoSigningParams>;
+
+impl SubstrateFinalitySyncPipeline for WestendFinalityToRococo {
+	const BEST_FINALIZED_SOURCE_HEADER_ID_AT_TARGET: &'static str = bp_westend::BEST_FINALIZED_WESTEND_HEADER_METHOD;
+
+	type TargetChain = Rococo;
+
+	fn customize_metrics(params: MetricsParams) -> anyhow::Result<MetricsParams> {
+		Ok(
```
svyatonik, line 38: Also please consider removing these metrics from here.

svyatonik (approved):

relays/bin-substrate/src/rialto_millau/rococo_headers_to_westend.rs
tomusdrw: I expect that it's going to be different people looking at these metrics (and they will be visibile in dfferent places).
1. We have bridge-team-internal Grafana and our team looking at Westend<>Millau stuff.
2. We have parity-internal Grafana and our devops team looking at Rococo<>Westend.

Hence I'd keep both. I extracted a function though, so the code is not triplicated.

primitives/chain-rococo/src/lib.rs
tomusdrw: Added comments and test.

primitives/chain-westend/src/lib.rs
tomusdrw: Fixed.

primitives/chain-rococo/src/lib.rs
tomusdrw: Added a test for this.

primitives/chain-rococo/src/lib.rs
tomusdrw: Comment updated, please take a look.

PR #878, opened 2021-04-08
Complex headers+messages Millau<->Rialto relay

svyatonik:
related #817

Brief overview of changes in this PR:
1) `relays/bin-substrate/src/on_demand_headers.rs` is a successor of the same file from https://github.com/paritytech/parity-bridges-common/pull/833 . However, instead of `fn activate(activate: bool)` the `OnDemandHeadersRelay` now has a `fn require_finalized_header(header_id)`. That's because there are now two loops using the same on-demand relay && activate/deactivate pattern doesn't work well with this. So there's now background task that is trying to sync headers up to the maximal passed `header_id`;
2) `relays/bin-substrate/src/cli/relay_headers_and_messages.rs` - this is the file that starts the new headers+messages relay. The cli subcommand is `./bin/substrate-relay relay-headers-and-messages millau-rialto`. Command line options are the same as for regular messages relay. The only exclusion that in complex relay there's no 'source' and 'target' chains. Internally I'm using `left` and `right` terminology, but for cli options it is counter-intuitive - the user must associate e.g. `Millau` with `left` (or `source`) chain. So I've reintroduced chain-specific connection and signing parameters, but they're now local to this file. The full cli looks like:
```
./bin/substrate-relay --millau-host=127.0.0.1 --millau-port=10944 --millau-signer=//Charlie --rialto-host=127.0.0.1 --rialto-port=9944 --rialto-singer=//Charlie --lane=00000000 --prometheus-port=9700
```
3) I've already explained why I think we still need all 3 kinds of relays, but will duplicate it here. This complex relay is perfect for our current situation - when there's no concurrency between relayers && there's exactly one bridge application running (i.e. single lane). If one of these conditions will change, a separate headers/messages relayers may be required. E.g. if there'll be other messages relayers, then they'll probably rely on our headers+messages relayer running, instead of manually relaying headers. If we'll open couple of other message lanes, then it'll make sense to run headers+messages relay that will handle multiple lanes. Finally, if there'll be some other bridge application running, then it may start its own headers relay.
4) I was opening separate PRs for changes that are related to metrics, but there's still one change that is left in this PR (could extract into a separate PR on demand, though). The idea is that we should have the single messages dashboard that works both for dedicated messages relay and for complex relay. So metrics should have the same `name` in both relays. Previously all metrics were `Default`, now they're created using `fn new()` that accepts optional metrics prefix. I've also removed `fn register()` method from `trait Metrics` - metrics are now registered when they're created;
5) only single instance of standalone metric is spawned - this works for `GlobalMetrics`, which is exposed by every relay.

Left for future PRs:
1) actually use this relay in deployments - I'm planning to leave Westend -> Millau headers relay running, but replace all Millau <-> Rialto relays with this one. Probably also worth running separate messages relay for lane 0x00000001 just to be sure that it works;
2) remove Millau <-> Rialto header relays dashboards - they will show no data and generate alerts with this relay running;
2) tests for on-demand-headers relay.

Some ideas and possible problems generated during this work:
1) we may change simple proof-of-storage used by messages pallet to header + justification + proof-of-storage. Then we may do not start the headers relay at all. This however introduces more problems than it solves - e.g. grandpa pallet must be able to import (or at least imitate import) of headers that are behind the best finalized. What with the authorities change - we can't trust old authorities set anymore? What if someone is still importing headers - then we'll just resubmit the same information to the chain? The weight of messages/confirmations delivery transaction will increase;
2) this complex relay may be extended to support multiple lanes. It's quite an easy change, but the PR is already too big, so I haven't introduced it here;
3) **NEEDS ATTENTION**: my understanding of the current "protection" against long-range attacks is that we're always running headers relay and it will see and submit header with authorities set change almost instantly. But what if there are no messages for, say, a month. Then we won't submit new headers at all and old authorities set may submit their own headers and do whatever they want with the bridge. You can say that they may do the same even with headers relay running, but: (1) authorities funds are still bonded e.g. for 28 days on polkadot => we may request slashing from council or whoever who could make this happen (otoh the profit from the attack may be larger than the loss from the slashing) and (2) we have an ability to halt the bridge + will probably have an alerts for cases like that, so it won't last forever. So if I'm correct and that's ^^^ the only guarantee that we have, I suggest also running headers relay for P<->K deployments, that will synchronize at least 1 header every `N` minutes.


relays/bin-substrate/src/on_demand_headers.rs
```diff
@@ -0,0 +1,255 @@
...
+						break
+					},
+				}
+			},
+			_ = relay_exited_rx => {
+				// there could be a situation when we're receiving exit signals after we
+				// have already stopped relay or when we have already started new relay.
+				// but it isn't critical, because even if we'll accidentally stop new relay
+				// we'll restart it almost immediately
+				stop_on_demand_headers_relay(active_headers_relay.take()).await;
```
adoerr, line 149: Shouldn't we keep track of header relay tasks properly? I feel a bit uneasy to cancel whatever `JoinHandle` happens to be assigned to `active_headers_relay`. Are we absolutely positive that there is never more than one active header relay task at a time (the one assigned to `active_headers_relay`). I guess I'm worried of a potential task resource leak here.

svyatonik: >  never more than one active header relay task at a time

Yes. Did you find a way to have two?

adoerr: No, but I'm sure Murphy will ;-)

relays/bin-substrate/src/cli/relay_headers_and_messages.rs
```diff
@@ -0,0 +1,183 @@
...
+
+				$generic
+			}
+		}
+	};
+}
+
+// All supported chains.
+declare_chain_options!(Millau, millau);
+declare_chain_options!(Rialto, rialto);
```
tomusdrw, line 110: I'd actually prefer the `rialto-` & `millau-` options for all sub-commands but I didn't have ideas how to do that in a generic way without duplicating code. I think this is a right call to introduce them here. If we ever add more bridges here and still don't figure out good solution for dynamic options, we could add ugly `first_chain`, `second_chain` or `left/right` as you suggested.

tomusdrw (approved):
I only gave it a quick look, but lgtm.

>  Probably also worth running separate messages relay for lane 0x00000001 just to be sure that it works;

Agree.

> we may change simple proof-of-storage used by messages pallet to header + justification + proof-of-storage. Then we may do not start the headers relay at all. 

Perhaps let's consider batching headers + messages instead if this causes issues?

> this complex relay may be extended to support multiple lanes. It's quite an easy change, but the PR is already too big, so I haven't introduced it here;

:+1: 

>  So if I'm correct and that's ^^^ the only guarantee that we have, I suggest also running headers relay for P<->K deployments, that will synchronize at least 1 header every N minutes.

Yes your understanding is 100% correct. I guess to support your last proposal we need to introduce some kind of "periodic/delayed" headers relayer mode?

svyatonik:
> Perhaps let's consider batching headers + messages instead if this causes issues?

It was the original idea - to use "message proof + header + justification" instead of simple "message proof". It actually introduces some problems I've written above. So I decided not to go this way

> Yes your understanding is 100% correct. I guess to support your last proposal we need to introduce some kind of "periodic/delayed" headers relayer mode?

Wwe have had something like that in poa relay (backup relay mode). I'm not yet sure - whether it is better to have separate headers relay process here (then we may have that nonces problem again), or just change this on-demand code a bit (so that it'll submit header periodically). But if my assumption is valid, then will definitely do that before P+K.

tomusdrw:
> It was the original idea - to use "message proof + header + justification" instead of simple "message proof". It actually introduces some problems I've written above. So I decided not to go this way

Actually I meant keeping the two things separate, but using `utility.batch` to put it into the same transaction, afaiu you were suggesting to add headers+justifications to messages pallet, no?

svyatonik:
> Actually I meant keeping the two things separate, but using utility.batch to put it into the same transaction, afaiu you were suggesting to add headers+justifications to messages pallet, no?

Ok, I misunderstood then. But it actually is the same as with changing the storage proof type. I.e. you can't be sure that your transaction will succeed - e.g. the other tx may have already imported header `N+1`, while you have built your proof on top of header `N`. It may be even worse if that `N+1` has set new authorities set - we have already dropped old set, so justification can't be imported.

svyatonik:
(there are some issues after merging with master => back to draft for now)

PR #881, opened 2021-04-09
Rialto+Millau maintenance dashboard

svyatonik:
This is a prototype for future P<>K maintenance dashboard. Right now it is a bit weird - even though it is related to Millau+Rialto, it shows Polkadot<>Kusama conversion rates + it also has no any alerts turned on. And alerts are main thing why it is required. But it at least illustrates that our "maintenance" metrics are working.


tomusdrw (approved):

HCastano (approved):
In the future, attaching an image of the dashboards would be nice (I know you already sent it to us on Element, but not everyone has access to this).

PR #887, opened 2021-04-12
Revert to older nightly.

tomusdrw:


adoerr (approved):
I assume, this solves `clippy` complaining about `clang-sys`?

LGTM

PR #893, opened 2021-04-13
Use Substrate state_getReadProof RPC method to get storage proofs

svyatonik:
This method is there for at least a year - I do not know how I missed it :/ Found it yesterday, when I was going to add method like that to Substrate.

So this PR removes our custom messages RPC crate. Instead we're now using `state_getReadProof` method from `messages_source.rs`/`messages_target.rs`. Also storage proof overhead metric now uses the proof of ":code:" storage value, so it should work for all chains now.


relays/bin-substrate/src/messages_target.rs
```diff
@@ -166,10 +179,13 @@ where
 		id: TargetHeaderIdOf<P>,
 	) -> Result<(TargetHeaderIdOf<P>, P::MessagesReceivingProof), SubstrateError> {
 		let (id, relayers_state) = self.unrewarded_relayers_state(id).await?;
+		let inbound_data_key = pallet_bridge_messages::storage_keys::inbound_lane_data_key::<R, I>(&self.lane_id);
```
tomusdrw, line 182: This will cause issues for Rococo & Westend. Can we avoid depending on the runtime directly and just add a function to `bp_rialto/bp_millau/bp_messages` crates which generates the storage key?

tomusdrw: I guess it's good for now, there is probably a bunch of other things to fix first anyway.

tomusdrw (approved):
lgtm! <3 less code :)

relays/bin-substrate/src/messages_target.rs
svyatonik: I really wanted to avoid manually crafting storage keys. And the main thing is that we don't need runtime here at all - it's here just to satisfy compiler - only instance is used to build key. I'll see what can be done here

svyatonik: So as [this comment](https://github.com/paritytech/parity-bridges-common/blob/7dace6246ec438ea025ad042270f049c1c2ccbe1/modules/messages/src/lib.rs#L656) explains, we may use any runtime that implements `Config<I>`. Presumably for all W/R/P/K chains we'll be using the same instance => we may use e.g. `millau_runtime::Runtime` everywhere. Alternatively I may just copypaste [this code](https://github.com/paritytech/substrate/blob/dcce8b4cfb461cbb2033021926e623b52c95fc74/frame/support/src/storage/generator/map.rs#L74) to the messages pallet, but as I said I wanted to avoid that.

relays/client-substrate/src/metrics/storage_proof_overhead.rs
```diff
@@ -19,60 +19,39 @@ use crate::client::Client;
...
-/// Regular Substrate node does not provide any RPC endpoints that return storage proofs.
-/// So here we're using our own `pallet-bridge-messages-rpc` RPC API, which returns proof
-/// of the inbound message lane state. Then we simply subtract size of this state from
-/// the size of storage proof to compute metric value.
-///
-/// There are two things to keep in mind when using this metric:
-///
-/// 1) it'll only work on inbound lanes that have already accepted at least one message;
-/// 2) the overhead may be slightly different for other values, but this metric gives a good estimation.
+/// Ther's one thing to keep in mind when using this metric: the overhead may be slightly
```
HCastano, line 33:
```suggestion
/// There's one thing to keep in mind when using this metric: the overhead may be slightly
```

HCastano (approved):
I also ❤️ less code

PR #896, opened 2021-04-13
Support more than `u8::max_value` GRANDPA validators

### LINKED ISSUE #846: Benchmark GRANDPA Pallet with more than `u8::MAX` Validators
Right now the benchmarks for `pallet-bridge-grandpa` are limited to `u8::MAX` validators.
It would be useful to be able to use a higher number of validators since it'll allow us
to better see how justification verification scales.

The reason why we're limited to `u8::MAX` is because when we generate secret keys for
each validator we use a `[u8; 32]` as the seed, with all the array elements being the
same `u8` value.

```rust
pub struct Account(pub u8);

impl Account {
    ...
    SecretKey::from_bytes(&[self.0; 32]).expect("...")
    ...
}
```

We'll need to change the scheme used to generate the seed so that more validators can be
supported.

tomusdrw:
Resolves #846

### DIFF
--- a/scripts/update-weights.sh
+++ b/scripts/update-weights.sh
@@ -1,8 +1,10 @@
 #!/bin/sh
 
+set -exu
+
 # Run this script from root of the repo
 
-cargo run --manifest-path=bin/rialto/node/Cargo.toml --release --features=runtime-benchmarks -- benchmark \
+time cargo run --manifest-path=bin/rialto/node/Cargo.toml --release --features=runtime-benchmarks -- benchmark \
 	--chain=local \
 	--steps=50 \
 	--repeat=20 \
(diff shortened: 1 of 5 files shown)


burdges:
Just fyi, we might do somewhat above 1024 eventually, just depends upon how our concentration bounds analysis for multiple relay chains turns out, so currently unclear.

scripts/update-weights.sh
```diff
@@ -1,9 +1,13 @@
...
+#
+# Run this script from root of the repo.
 
-# Run this script from root of the repo
+set -eux
 
-cargo run --manifest-path=bin/rialto/node/Cargo.toml --release --features=runtime-benchmarks -- benchmark \
-	--chain=local \
+time cargo run --release -p rialto-bridge-node --features=runtime-benchmarks -- benchmark \
+	--chain=dev \
```
HCastano, line 10: Changed it to the `dev` chain since that's what the Substrate benches bot uses

HCastano (approved):

PR #897, opened 2021-04-13
CLI: naming clean-up.

tomusdrw:
Change `rialto_millau` module into `chains` module, and split chain-specific code to separate `.rs` files.


HCastano (approved):
Looks good!

I'd be careful with the merge order here. I think it will be less painful to merge #875 first, and then merge this - but it's up to you.

Issue #587, opened 2020-12-18
bench-bot integration


svyatonik:
Bot: https://github.com/paritytech/bench-bot

That is not super-important for this repo, since we only have testnets here. And accurate weights are not super-important for testnets. But still it may be useful to see how pallet weights are changing just by adding PR comment. Also - if we'll deploy bot on reference machine, we'll be able to at least estimate cost of Polkadot/Kusama transactions before opening PR to these repos.

Right now I'm going to execute benchmarks locally just to generate that `WeightInfo` trait. Polkadot/Kusama have their own bots, so we'll get accurate weights there anyway.

tomusdrw:
Bumping the priority to nice to have, cause I think it shouldn't be that difficult to integrate, and I agree it's really useful even for testnet.

svyatonik:
https://github.com/paritytech/devops/issues/783

EmmanuellNorbertTulbure:
Not relevant for Rialto - Millau

PR #898, opened 2021-04-13
Update weights

tomusdrw:
It seems that my laptop is much worse than the previous machines used for generating these. We should probably consider doing #587 to use standardized hardware.


modules/grandpa/src/weights.rs
```diff
@@ -69,29 +60,29 @@ pub struct RialtoWeight<T>(PhantomData<T>);
 impl<T: frame_system::Config> WeightInfo for RialtoWeight<T> {
 	fn submit_finality_proof(v: u32, p: u32) -> Weight {
 		(0 as Weight)
-			.saturating_add((160_060_000 as Weight).saturating_mul(v as Weight))
-			.saturating_add((640_223_000 as Weight).saturating_mul(p as Weight))
-			.saturating_add(T::DbWeight::get().reads(5 as Weight))
-			.saturating_add(T::DbWeight::get().writes(3 as Weight))
```
svyatonik, line 75: @HCastano so what I meant [here](https://github.com/paritytech/parity-bridges-common/pull/890#issuecomment-818829027) is this - current benchmarks are not doing pruning. These two additional writes are writes to `ImportedHashes` and `ImportedHashesPointer`. But there's no additional write for header removal (`<ImportedHeaders<T, I>>::remove(hash);`). I suggest to change benchmarks to include that.

svyatonik (approved):
@todr To avoid fixing these test failures (this may escalate after this test is fixed), I could re-run benchmarks on my laptop, if you wish. Or you may just revert messages weight updates. #587 is definitely a solution

tomusdrw:
I've re-run the benchmarks on more standardized hardware (details shared internally) and now the tests seem to pass. I'll look into pruning next.

HCastano (approved):

PR #905, opened 2021-04-14
Use complex headers+messages relay in test deployments

### LINKED ISSUE #817: Headers + messages relay

### LINKED ISSUE #904: Add another message lane and serve it using standalone messages relay
When we'll have #878 used in our test deployments, the lane `00000000` will be served by complex headers relay. So we won't have standalone messages relay running => it may break as it has happened multiple times already. So the suggestion is to add another lane (along with the dashboard + alerts) `00000001` and serve it using standalone messages relay.

svyatonik:
closes #817
closes #904 

This PR replaces 4 relays (Rialto -> Millau headers, Rialto -> Millau messages, Millau -> Rialto headers, Millau -> Rialto messages) with single complex headers+messages Rialto-Millau relay. So now we have:
1) Rialto-Millau headers and message lane 00000000 served by complex relay;
2) Westend->Millau headers served by standalone headers relay;
3) Rialto<>Millau message lane 00000001 served by standalone messages relay.

For lane 00000001 there are two additional widgets in messages dashboard:
UpdatedDashboard

I haven't added all 5 widgets for this lane - it isn't super overloaded. There are two alerts in every direction - when messages are not delivered/generated and when messages are not confirmed. There's no Rialto<>Millau headers dashboards for now - they're already built into messages dashboards.


HCastano (approved):
The dashboard is looking pretty crowded. If possible, I'd move the metrics related to lane 1 into their own dashboard.

Also, while running them locally I had no data for lane 1 show up in the dashboards. Could somebody else double check to see if it's an issue with my setup?

svyatonik:
> The dashboard is looking pretty crowded. If possible, I'd move the metrics related to lane 1 into their own dashboard.

That's possible, but in this set of widgets some are depend on each other. So nothing can't be relayed at lane 00000001 if headers are not relayed. And headers are not relayed if messages at lane 00000000 are not generated. You may actually see these pauses by looking at those widgets.

> Also, while running them locally I had no data for lane 1 show up in the dashboards. Could somebody else double check to see if it's an issue with my setup?

That's most probably because you'r using paritytech/* images which reject all messages to 00000001

tomusdrw (approved):

HCastano:
> That's most probably because you'r using paritytech/* images which reject all messages to 00000001

Ah you're right - I only built a local `substrate-relay` image

PR #906, opened 2021-04-15
Speed up Grandpa Benchmark Setup

HCastano:
At the moment, benchmarking `pallet-bridge-grandpa` takes quite a long time, even on
beefy hardware. Through some digging I was able to figure out that most of the time was
actually spent in the setup portion of the benchmark, rather than running the actual
benchmarks.

I found that the culprit of the slowdown was `bp_test_utils::generate_chain()`, which had
a quadratic complexity due an unecessary use of recursion within
`bp_test_utils::test_header()`.

This PR removes the recursion from `test_header()` and significantly improves benchmark
times for `pallet-bridge-grandpa`.

Life Pro Tip Sidenote: The runtime benchmark framework supports arbitrary code. I was able
check performance and produce graphs of the setup code by adding a new bench and copying
the setup into the body of the bench. No "advanced" profiling tools required.

### Before
old

### After
new


adoerr (approved):
LGTM

tomusdrw (approved):
Nice!

PR #911, opened 2021-04-19
pay dispatch fee at target chain

### LINKED ISSUE #377: Refund relayer with 'weight' that remains after calls are dispatched
origin:
https://github.com/paritytech/parity-bridges-common/pull/339#discussion_r494872788
https://github.com/paritytech/parity-bridges-common/pull/339#discussion_r494879969

related:
https://github.com/paritytech/parity-bridges-common/pull/383#discussion_r495938789

### LINKED ISSUE #561: Pay dispatch fee on the target chain.

svyatonik:
closes #377 
closes #561 

Things to do in this PR (quite a lot - opening just to show some progress):
- [x] actually return unspent weight from delivery call;
- [x] test that transfer actually happens when using our `messages` adapter;
- [x] update benchmarks - the worst delivery case is now when fee is paid at the target chain
- [x] (most probably) weight of pay-dispatch-fee-at-target must not be paid by submitter when message is prepaid at source + we should refund relayer with this fee if message is already prepaid in delivery tx. This would require additional constant in chain primitives, which is a bit unfortunate.

Things to do in follow-up PRs:
- [ ] support pay-at-target-chain option in `encode-call`, `send-message` and other subcommands;
- [ ] actually generate pay-at-target-chain messages in test deployments.


modules/dispatch/src/lib.rs
```diff
@@ -102,6 +102,13 @@ pub struct MessagePayload<SourceChainAccountId, TargetChainAccountPublic, Target
 	pub weight: Weight,
 	/// Call origin to be used during dispatch.
 	pub origin: CallOrigin<SourceChainAccountId, TargetChainAccountPublic, TargetChainSignature>,
+	/// If true, then the sender has decided to pay dispatch fee at the target chain.
+	///
+	/// The fee is paid right before the message is dispatched. So in case of any other
+	/// issues (like invalid call encoding, invalid signature, ...) we won't do any direct
+	/// transfers. Instead, we're returning fee related to this message dispatch to the
+	/// relayer.
+	pub pay_dispatch_fee_at_target_chain: bool,
```
tomusdrw, line 111: To make it more future proof I'd opt for an enum:

```suggestion
	pub dispatch_fee:  DispatchFee::{SourceChain, TargetChain},
```

this will allow us to add other options in the future (like paying with non-primary token) without breaking backward compatibility of payload format.

modules/dispatch/src/lib.rs
```diff
@@ -191,22 +201,27 @@ decl_module! {
...
+impl<T: Config<I>, I: Instance> MessageDispatch<T::AccountId, T::MessageId> for Pallet<T, I> {
 	type Message =
 		MessagePayload<T::SourceChainAccountId, T::TargetChainAccountPublic, T::TargetChainSignature, T::EncodedCall>;
 
 	fn dispatch_weight(message: &Self::Message) -> Weight {
 		message.weight
 	}
 
-	fn dispatch(bridge: InstanceId, id: T::MessageId, message: Result<Self::Message, ()>) {
+	fn dispatch<P: FnOnce(&T::AccountId, Weight) -> Result<(), ()>>(
```
tomusdrw, line 212: Might be good to add a comment what kind of `Weight` is being returned.

bin/runtime-common/src/messages.rs
```diff
@@ -326,8 +330,17 @@ pub mod source {
...
 		// the fee (in Bridged tokens) of all transactions that are made on the Bridged chain
-		let delivery_transaction =
-			BridgedChain::<B>::estimate_delivery_transaction(&payload.call, payload.weight.into());
+		//
+		// if we're going to pay dispatch fee at the target chain, then we don't include weight
+		// of the message dispatch in the delivery transaction cost
+		let delivery_transaction = BridgedChain::<B>::estimate_delivery_transaction(
+			&payload.call,
+			if payload.pay_dispatch_fee_at_target_chain {
+				0.into()
```
tomusdrw, line 339: I wonder if we should include the (potentially failed) transfer cost or emitting even from dispatch pallet here.

modules/messages/src/inbound_lane.rs
```diff
@@ -90,9 +90,10 @@ impl<S: InboundLaneStorage> InboundLane<S> {
 	}
 
 	/// Receive new message.
-	pub fn receive_message<P: MessageDispatch<S::MessageFee>>(
+	pub fn receive_message<P: MessageDispatch<AccountId, S::MessageFee>, AccountId>(
 		&mut self,
-		relayer: S::Relayer,
+		relayer_at_brdiged_chain: &S::Relayer,
```
tomusdrw, line 95:
```suggestion
		relayer_at_bridged_chain: &S::Relayer,
```

modules/messages/src/inbound_lane.rs
```diff
@@ -114,25 +115,29 @@ impl<S: InboundLaneStorage> InboundLane<S> {
...
 		}
 
 		self.storage.set_data(data);
 
-		P::dispatch(DispatchMessage {
-			key: MessageKey {
-				lane_id: self.storage.id(),
-				nonce,
+		P::dispatch(
+			relayer_at_this_chain,
```
tomusdrw, line 132: We should probably add a note somewhere, that relayers should make sure that their derived accounts are above `ED` cause otherwise they might not be able to collect any fees.

primitives/messages/src/target_chain.rs
```diff
@@ -100,7 +100,17 @@ pub trait MessageDispatch<Fee> {
...
+	///
+	/// If your configuration allows paying dispatch fee at the target chain, then
+	/// it must be paid inside this method to the `relayer_account`.
+	///
+	/// Returns 'unused' dispatch weight that will be deducted from total delivery transaction
+	/// weight, thus reducing the transaction cost. This shall not be zero in two cases:
+	///
+	/// 1) if message has been dispatched successfully, but post-dispatch weight is less than
+	///    the weight, declared by the message sender;
+	/// 2) if message has not been dispatched at all.
```
tomusdrw, line 112: I was thinking a bit about burning part of that weight, or returning part of the fee to the message sender, but I concluded that it doesn't make much sense for now. Leaving the comment here for posterity though.

1. In case of target-chain fee payment, messages senders are heavily incentivised to specify the lowest possible weight, cause they pay up-front. Any unused weight is benefiting the relayers now by increasing their reward for accepting target-chain fee payment, which inherently increases the risk for them. So that seems to be quite aligned.
2. In case of source-chain fee payment, the situation is quite similar - in such case relayers risk is increased by price fluctuation though.
3. Hence message senders does not really benefit directly from having the dispatch consume less weight than declared, cause they pay for dispatch upfront. Burning part of the unused weight would only reduce reward for the relayers (which are exposed to a bunch of risks already) and would unnecessarily inflate block consumption.
4. Returning part of unused fee to senders reduces their incentive to calculate the fee correctly and increases risk for the relayers.

Overall we might consider doing a more detailed analysis, perhaps there is some fair split between returning/burning/leaving to the relayer, but taking into account real-world data will be a great help for this, so I'd rather collect it first :)
My gut feeling is that relayers are more on the upside here, but I think it's acceptable for the first deployment.

tomusdrw (commented):
Nice!

modules/dispatch/src/lib.rs
svyatonik: It's trait implementationt. The documentation is in the trait declaration

bin/runtime-common/src/messages.rs
svyatonik: This is only about message dispatch weight. All extra costs related to dispatch are already included in the delivery tx weight.

modules/messages/src/inbound_lane.rs
svyatonik: Yes, but here (at target chain) this problem is not super-important - the relayer account is the account that submits delivery tx. So it is guaranteed to have some funds (at least to pay for the tx). Te edge case is when relayer account balance drops under ED as a result of the delivery tx and fee is less than ED (this unlikely to happen with adequate relayer, because he'll be unable to submit any more transactions).

tomusdrw: Ah good point, I forgot about this.

svyatonik:
Some notes about weight of the delivery transaction - it has been increased from `1_000_000_000` to `1_500_000_000` (`DEFAULT_MESSAGE_DELIVERY_TX_WEIGHT`) in this PR. That's because of additional `Currency::transfer` call. This is the cost of worst-case single message delivery tx, where dispatch fee is paid at the target chain. However, the actual post-dispatch fee that the delivery call now returns is decreased by this additional cost (`PAY_INBOUND_DISPATCH_FEE_WEIGHT`) if fee has already been paid at the source chain.

As a result of ^^^, we now able to deliver only `704` instead of `1020` messages in single delivery tx. This may be 'fixed' by adding another parameter (something like `number_of_pay_fee_calls`) to the delivery tx, but it'll complicate a relayer code a bit (which is probably ok) + will add another pre-dispatch check in the call (which is also probably ok). The milestone of this fix actually depends on the expected use-case at real chains - if we'll have most/all messages that are paid at the target chain, then it's ok to leave as is for now + fix later. If we expect mixed/pay-at-the-source-chain usage, then it is better to fix before deployments.

modules/messages/src/benchmarking.rs
```diff
@@ -242,7 +247,8 @@ benchmarks_instance! {
 	// * proof does not include outbound lane state proof;
 	// * inbound lane already has state, so it needs to be read and decoded;
 	// * message is successfully dispatched;
-	// * message requires all heavy checks done by dispatcher.
+	// * message requires all heavy checks done by dispatcher;
+	// * message dispatch fee is pait at target (this) chain.
```
tomusdrw, line 251:
```suggestion
	// * message dispatch fee is paid at target (this) chain.
```

modules/messages/src/benchmarking.rs
```diff
@@ -257,20 +263,23 @@ benchmarks_instance! {
...
+		assert!(T::is_message_dispatched(21));
 	}
 
 	// Benchmark `receive_messages_proof` extrinsic with two minimal-weight messages and following conditions:
 	// * proof does not include outbound lane state proof;
 	// * inbound lane already has state, so it needs to be read and decoded;
 	// * message is successfully dispatched;
-	// * message requires all heavy checks done by dispatcher.
+	// * message requires all heavy checks done by dispatcher;
+	// * message dispatch fee is pait at target (this) chain.
```
tomusdrw, line 282:
```suggestion
	// * message dispatch fee is paid at target (this) chain.
```

modules/messages/src/benchmarking.rs
```diff
@@ -288,20 +297,23 @@ benchmarks_instance! {
...
+		assert!(T::is_message_dispatched(22));
 	}
 
 	// Benchmark `receive_messages_proof` extrinsic with single minimal-weight message and following conditions:
 	// * proof includes outbound lane state proof;
 	// * inbound lane already has state, so it needs to be read and decoded;
 	// * message is successfully dispatched;
-	// * message requires all heavy checks done by dispatcher.
+	// * message requires all heavy checks done by dispatcher;
+	// * message dispatch fee is pait at target (this) chain.
```
tomusdrw, line 316:
```suggestion
	// * message dispatch fee is paid at target (this) chain.
```

modules/messages/src/benchmarking.rs
```diff
@@ -389,13 +405,48 @@ benchmarks_instance! {
...
+		);
+		assert!(T::is_message_dispatched(21));
+	}
+
+	// Benchmark `receive_messages_proof` extrinsic with single minimal-weight message and following conditions:
+	// * proof does not include outbound lane state proof;
+	// * inbound lane already has state, so it needs to be read and decoded;
+	// * message is successfully dispatched;
+	// * message requires all heavy checks done by dispatcher;
+	// * message dispatch fee is pait at source (bridged) chain.
```
tomusdrw, line 424:
```suggestion
	// * message dispatch fee is paid at source (bridged) chain.
```

modules/messages/src/lib.rs
```diff
@@ -507,20 +530,57 @@ decl_module! {
...
+					let dispatch_weight = T::MessageDispatch::dispatch_weight(&message);
+					let receival_result = lane.receive_message::<T::MessageDispatch, T::AccountId>(
+						&relayer_id_at_bridged_chain,
+						&relayer_id_at_this_chain,
+						message.key.nonce,
+						message.data,
+					);
+
+					// note that we're returning unspent weight to relayer even if message has been
+					// rejected by the lane. This allows relayers to submit spam transactions with
```
tomusdrw, line 542: Why do we do that? I imagine this may cause issues in the future if we for instance want to prioritise bridge transactions or give them a separate (fair) amount of block space (I imagine bridge hub with 3 bridges, were the block weight is split equally between them so they don't have to compete between each other; more sophisticated scheme would also split the unused weight of other bridges).

AFAIU this would lower the cost of a potential griefing attack here. I guess the reasoning is to alleviate excessive costs for multiple concurrent relayers?

modules/messages/src/lib.rs
```diff
@@ -507,20 +530,57 @@ decl_module! {
...
+					// delivery transaction cost anyway. And base cost covers everything except
+					// dispatch, so we have a balance here.
+					let (unspent_weight, refund_pay_dispatch_fee) = match receival_result {
+						ReceivalResult::Dispatched(dispatch_result) => {
+							valid_messages += 1;
+							(dispatch_result.unspent_weight, !dispatch_result.dispatch_fee_paid_during_dispatch)
+						},
+						ReceivalResult::InvalidNonce
+							| ReceivalResult::TooManyUnrewardedRelayers
+							| ReceivalResult::TooManyUnconfirmedMessages => (dispatch_weight, true),
```
tomusdrw, line 554: Why do we assume the message is paid at the source chain here? Wouldn't it be safer to refund less?

tomusdrw (approved):
looks good to me! Couple of small grumbles and questions. I'm super sorry it took my so long to take a look at this.

modules/messages/src/lib.rs
svyatonik: Yes, this is to lower costs for concurrent relayers. And yes - this may cause a problem if something like that ^^^ would be implemented. But right now it imo solves more problems than it creates - relayer pays the cost of proof parsing, which in case of duplicate messages should be near to 100% of the whole tx cost (duplicated messages are never passed to outside of this module). So the goal is to make post-dispatch weight equal to weight of the same-proof-transaction that declares to bring zero messages.

modules/messages/src/lib.rs
svyatonik: Weight formula (=> pre-dispatch weight) includes the cost of `Currency::transfer` call (or whatever happens on the integration level when dispatch fee is paid on target chain) for **every message**. Since we are not dispatching the message , this call is not happened => therefore we're deducting it from the final cost

modules/messages/src/lib.rs
tomusdrw: Makes sense, thanks for the explanations!

PR #913, opened 2021-04-20
(temporarily) Switch to jsonrpsee master

### LINKED ISSUE #909: Subscription leak in complex relay
Our ws client uses default configuration where it supports at most 256 parallel requests (https://github.com/paritytech/jsonrpsee/blob/75260c8435fa1cb9d143b63f2ea3df2dda4217eb/ws-client/src/client.rs#L139). On-demand headers relay makes new subscription every time it starts. After 255 restarts, the client starts returning error `MaxSlotsExceeded`, which means that there are 256 parallel requests. So it seems that subscription isn't gracefully dropped when on-demand headers relay is stopped.

FWIW: the quick workaround will be to increase `max_concurrent_requests` [here](https://github.com/paritytech/parity-bridges-common/blob/master/relays/client-substrate/src/client.rs#L109).

svyatonik:
closes #909 

I've been able to `cargo build -p polkadot` using this patch, so it should not break polkadot codebase, unless I'm missing something.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -182,16 +182,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-attributes"
-version = "1.1.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a3203e79f4dd9bdda415ed03cf14dae5a2bf775c683a00f94e9cd1faf0f596e5"
-dependencies = [
- "quote",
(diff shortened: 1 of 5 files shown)


tomusdrw (approved):

svyatonik:
(sry - forgot to remove change of param used to test the fix)

PR #914, opened 2021-04-20
update jsonrpsee to 0.2.0-alpha.5

niklasad1:
//cc @svyatonik 

You should probably test this before merging, we did some major refactoring of the `jsonrpc related types` and require significant changes in the `ws-client`.

Technically, we have released `jsonrpsee root` crate but as you only use the `ws-client` it would be bring in some additional deps.
As we have feature-gated it by client, server, macros but if you have better ideas please comment in https://github.com/paritytech/jsonrpsee/issues/271 :)


HCastano (approved):
> You should probably test this before merging, we did some major refactoring of the jsonrpc related types and require significant changes in the ws-client.

Our test network will pull the changes in tonight - if it breaks it breaks ¯\\_(ツ)_/¯

> Technically, we have released jsonrpsee root crate but as you only use the ws-client it would be bring in some additional deps.
> As we have feature-gated it by client, server, macros but if you have better ideas please comment in paritytech/jsonrpsee#271 :)

I personally liked the granularity of being able to pull in just the `ws-client`, but having `client`, `server`, and `macro` split seems reasonable to me as well.

PR #917, opened 2021-04-21
Add properties to Chain Spec

tomusdrw:
This fixes token display in the UI and when using `polkadot/api`.


bin/millau/node/src/chain_spec.rs
```diff
@@ -67,6 +67,15 @@ pub fn get_authority_keys_from_seed(s: &str) -> (AccountId, AuraId, GrandpaId) {
 impl Alternative {
 	/// Get an actual chain config from one of the alternatives.
 	pub(crate) fn load(self) -> ChainSpec {
+		let properties = Some(
+			serde_json::json!({
+				"tokenDecimals": 9,
```
HCastano, line 72: I don't think this is a problem since `tokenDecimals` is mostly a UI thing (how many base units make up one token), but I want to double check something. Since Millau and Rialto have different balance types (`u64` vs `u128`), should we have different `tokenDecimals` for each to account for that?

HCastano (approved):

bin/millau/node/src/chain_spec.rs
tomusdrw: Since the token is worthless it really doesn't matter :) AFAIR despite different balance type capacity the initial issuance is the same on both chains, so at least we will now have the same values on both chains. I'm up to set it to a arbitrary value for tests though, any suggestions? :)

HCastano: I'm okay leaving it for now, we can always experiment with the values later

PR #918, opened 2021-04-21
Add pruning to bechmarks & update weights.

tomusdrw:
Follow up on #898


svyatonik (approved):

bin/rialto/runtime/src/lib.rs
```diff
@@ -414,7 +414,20 @@ parameter_types! {
...
 
+#[cfg(feature = "runtime-benchmarks")]
+parameter_types! {
+	// Number of headers to keep in benchmarks.
+	//
+	// In benchmarks we always populate with full number of `HeadersToKeep` to make sure that
+	// pruning is taken into account.
+	// This is lower than regular value, to speed up benchmarking setup.
+	pub const HeadersToKeep: u32 = 1024;
+}
```
tomusdrw, line 427: This code is optional, but keeping it reduces the time to run benchmarks ~30% (from 70minutes to 54minutes).

modules/grandpa/src/lib.rs
```diff
@@ -427,6 +415,29 @@ pub mod pallet {
 		)
 	}
 
+	/// Import a previously verified header to the storage.
+	///
+	/// Note this function solely takes care of updating the storage and does not verify
+	/// the validaty of such import.
```
HCastano, line 421: I would add a note that this function also performs pruning if necessary.

modules/grandpa/src/lib.rs
```diff
@@ -427,6 +415,29 @@ pub mod pallet {
...
+	pub(crate) fn insert_header<T: Config<I>, I: 'static>(
+		header: BridgedHeader<T, I>,
+		hash: Option<BridgedBlockHash<T, I>>,
```
HCastano, line 424: I don't think it makes sense for this to be an `Option` - it'd just force the caller to provide a `hash`

modules/grandpa/src/lib.rs
```diff
@@ -441,15 +452,38 @@ pub mod pallet {
...
+		while number < end_number {
+			number = number + sp_runtime::traits::One::one();
+			let header = <BridgedHeader<T, I>>::new(
+				number,
+				Default::default(),
+				Default::default(),
+				Default::default(),
+				Default::default(),
+			);
+			insert_header::<T, I>(header, None);
```
HCastano, line 483:
```suggestion
			insert_header::<T, I>(header, header.hash());
```

modules/grandpa/src/lib.rs
```diff
@@ -441,15 +452,38 @@ pub mod pallet {
...
 		<IsHalted<T, I>>::put(is_halted);
 	}
 
+	#[cfg(feature = "runtime-benchmarks")]
+	pub(crate) fn bootstrap_bridge<T: Config<I>, I: 'static>(
+		init_params: super::InitializationData<BridgedHeader<T, I>>,
+	) {
+		let inject_headers = T::HeadersToKeep::get();
+		let start_number = *init_params.header.number();
+		let end_number = start_number + inject_headers.into();
```
HCastano, line 470:
```suggestion
		let end_number = start_number + T::HeadersToKeep::get().into();
```

I think this is more readable than `inject_headers`

modules/grandpa/src/benchmarking.rs
```diff
@@ -66,6 +67,10 @@ const MAX_VOTE_ANCESTRIES: u32 = 1000;
 // number of validators.
 const MAX_VALIDATOR_SET_SIZE: u32 = 1024;
 
+fn header_number<T: Config<I>, I: 'static, R: From<u32>>() -> R {
```
HCastano, line 70: Does `R` stand for something clever here? I would've expected `N` for number, lol

HCastano, line 70:
```suggestion
// In order to make sure pruning capabilities are being triggered we...
fn header_number<T: Config<I>, I: 'static, R: From<u32>>() -> R {
```

bin/rialto/runtime/src/lib.rs
```diff
@@ -414,7 +414,20 @@ parameter_types! {
...
 
+#[cfg(feature = "runtime-benchmarks")]
+parameter_types! {
+	// Number of headers to keep in benchmarks.
+	//
+	// In benchmarks we always populate with full number of `HeadersToKeep` to make sure that
+	// pruning is taken into account.
+	// This is lower than regular value, to speed up benchmarking setup.
+	pub const HeadersToKeep: u32 = 1024;
```
HCastano, line 426:
```suggestion
	// pruning is taken into account.
	//
	// Note: This is lower than regular value, to speed up benchmarking setup.
	pub const HeadersToKeep: u32 = 1024;
```

bin/rialto/runtime/src/lib.rs
HCastano: Is the 70 mins for the full (messages + header) bench suite?

HCastano (approved):
Some small things but looks good!

modules/grandpa/src/benchmarking.rs
tomusdrw: it's for "return value", fine to change it to `N` :)

tomusdrw:
> Is the 70 mins for the full (messages + header) bench suite?

Yup, on the standardized hardware :)

HCastano (approved):

HCastano:
> > Is the 70 mins for the full (messages + header) bench suite?
> 
> Yup, on the standardized hardware :)

Ouf that's rough. Looks like there' still some work to be done on improving the bench times

PR #921, opened 2021-04-22
Fix issue with on-demand headers relay not starting

svyatonik:
The reason of failures [like that](https://github.com/paritytech/parity-bridges-common/pull/920/checks?check_run_id=2403423846) is that when there are more messages to deliver than unrewarded-relayers vec capacity AND all source headers that have 'produced' messages are known to the target chain THEN the delivery strategy was not asking for new source headers that had reward confirmation => it was unable to deliver all messages.

### DIFF
--- a/relays/messages/src/message_race_delivery.rs
+++ b/relays/messages/src/message_race_delivery.rs
@@ -292,7 +292,24 @@ impl<P: MessageLane> RaceStrategy<SourceHeaderIdOf<P>, TargetHeaderIdOf<P>, P::M
 	}
 
 	fn required_source_header_at_target(&self, current_best: &SourceHeaderIdOf<P>) -> Option<SourceHeaderIdOf<P>> {
-		self.strategy.required_source_header_at_target(current_best)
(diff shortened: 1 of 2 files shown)


relays/messages/src/message_race_delivery.rs
```diff
@@ -292,7 +292,24 @@ impl<P: MessageLane> RaceStrategy<SourceHeaderIdOf<P>, TargetHeaderIdOf<P>, P::M
...
+			(Some(id1), Some(id2)) => {
+				if id1.0 > id2.0 {
+					Some(id1)
+				} else {
+					Some(id2)
+				}
+			}
+			(Some(id1), None) => Some(id1),
+			(None, Some(id2)) => Some(id2),
+			(None, None) => None,
```
tomusdrw, line 311:
```suggestion
			(Some(id1), Some(id2)) => Some(if id1.0 > id2.0 { id1 } else { id2 }),
			(a, b) => a.or(b),
```

tomusdrw (approved):
I'm not going to pretend I fully understand what's going on here 😅, but it seems reasonable and I trust you on the solution.

PR #923, opened 2021-04-22
Relay at least one header for every source chain session

### LINKED ISSUE #903: Submit at least 1 header every N minutes (when complex relay is used)
See (3) from section ["Some ideas and possible problems"](https://github.com/paritytech/parity-bridges-common/pull/878#issue-611454693) for details.

TLDR: since we have now complex relay that relays both messages and headers (when there are new messages) AND we rely on headers relay always running (for bridge security) THEN we need to sometimes submit new headers even if there's no new messages.

svyatonik:
closes #903 

The `N` from the issue is the average session length. The unbonding period is much larger than session length (if I'm not missing something), so it should be fine. On Rialto session length is 4 blocks, at Milau ~5 minutes.

### DIFF
--- a/relays/bin-substrate/src/on_demand_headers.rs
+++ b/relays/bin-substrate/src/on_demand_headers.rs
@@ -288,7 +288,7 @@ fn select_on_demand_relay_action<C: Chain>(
 	let best_finalized_source_header_at_source =
 		best_finalized_source_header_at_source.unwrap_or(best_finalized_source_header_at_target);
 
-	// if there are too many source headers missing from the target node, requie some
+	// if there are too many source headers missing from the target node, require some
 	// new headers at target
 	//
 	// why do we need that? When complex headers+messages relay is used, it'll normally only relay
--- a/relays/bin-substrate/src/on_demand_headers.rs
+++ b/relays/bin-substrate/src/on_demand_headers.rs
@@ -20,13 +20,18 @@ use crate::finality_pipeline::{SubstrateFinalitySyncPipeline, SubstrateFinalityT
 use crate::finality_target::SubstrateFinalityTarget;
 
 use bp_header_chain::justification::GrandpaJustification;
-use finality_relay::TargetClient as FinalityTargetClient;
+use finality_relay::{
+	FinalitySyncPipeline, SourceClient as FinalitySourceClient, TargetClient as FinalityTargetClient,
+};
 use futures::{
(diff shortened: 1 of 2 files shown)


relays/bin-substrate/src/on_demand_headers.rs
```diff
@@ -186,14 +199,127 @@ async fn background_task<SourceChain, TargetChain, TargetSign>(
...
+				error,
+			);
+
+			None
+		})
+}
+
+/// What to do with the on-demand relay task?
+#[derive(Debug, PartialEq)]
+enum OnDemandRelayAction {
```
HCastano, line 266: I like this better than the `(bool, bool)` stuff we had before 👍

relays/bin-substrate/src/on_demand_headers.rs
```diff
@@ -186,14 +199,127 @@ async fn background_task<SourceChain, TargetChain, TargetSign>(
...
+	// unneeded transactions
+	let best_finalized_source_header_at_target =
+		best_finalized_source_header_at_target.unwrap_or(required_source_header_at_target);
+
+	// if we have been unable to read header number from the source, then let's assume
+	// that it is the same as at the target
+	let best_finalized_source_header_at_source =
+		best_finalized_source_header_at_source.unwrap_or(best_finalized_source_header_at_target);
+
+	// if there are too many source headers missing from the target node, requie some
```
HCastano, line 291:
```suggestion
	// if there are too many source headers missing from the target node, require some
```

relays/bin-substrate/src/on_demand_headers.rs
```diff
@@ -186,14 +199,127 @@ async fn background_task<SourceChain, TargetChain, TargetSign>(
...
+) -> Option<SourceChain::BlockNumber>
+where
+	SubstrateFinalityTarget<TargetChain, P>: FinalityTargetClient<P>,
+	P: FinalitySyncPipeline<Number = SourceChain::BlockNumber>,
+{
+	finality_target
+		.best_finalized_source_block_number()
+		.await
+		.map(Some)
+		.unwrap_or_else(|error| {
```
HCastano, line 252: This is pretty strange, but I can't find a good combinator to replace this. Maybe something with `map_or_else`?

HCastano (approved):
Haven't tested it locally, but it looks reasonable 👍

PR #832, opened 2021-03-17
Configuration of cargo-spellcheck

TomaszWaszczyk:
My second try to configure the tool ;-)

PR #924, opened 2021-04-22
Add backbone configuration of cargo-spellcheck

TomaszWaszczyk:


.github/workflows/lint.yml
```diff
@@ -41,3 +41,28 @@ jobs:
...
+          submodules:   recursive
+      
+      - name:           Add cargo-spellcheck
+        run:            cargo install cargo-spellcheck
+      
+      - name:           Run spellcheck
+        uses:           actions-rs/cargo@master
+        with:
+          command:      spellcheck
+          args:         check  -m 1 -vv
```
tomusdrw, line 68:
```suggestion
          args:         check  -vv
```

Since the job is going to fail every time before we fix all of the errors I'd rather not mark every PR green, cause it's not clear how fast we can deal with it.

Ideally we would make a snapshot of all errors and just fail on newly introduced issues, but that might be too involved. I know you are planning to fix all errors anyway, hence I'm fine with having this job be kind of useless now, but at the very end let's switch to `-m 1`.

tomusdrw (approved):

.github/workflows/lint.yml
TomaszWaszczyk: Yup, but I need to narrow source to analysis, try to do it like (in order to keep green checks):

`
search_dirs = ["./modules/currency-exchange"]
`

And to start correction from module `currency-exchange` but cargo-spellcheck does not take into consideration my way of thinking how to configure it ;/

tomusdrw: I think `search_dirs` is more for extra dictionaries, not really for narrowing down the content. However looking at this example:
```
exec cargo spellcheck -m 99 $(git diff-index --cached --name-only --diff-filter=AM HEAD)
```
We can try to add modules here one by one, perhaps we can start with:
```
cargo spellcheck -m 1 -vv $(find modules/currency-exchange -name *.rs)
```
?

HCastano:
@TomaszWaszczyk Is the core of this different from #832? Or is the only difference that this doesn't actually fix any of the spelling/grammar warnings from `cargo-spellcheck` (since that can be done in the future)?

FYI, both PRs use the same `TomaszWaszczyk:master` branch which makes it a bit hard for collaboration. I can't easily check out this PR since Git already thinks I already have the branch locally. In the future I'd recommend using different branches for different PRs - even if they're closely related.

TomaszWaszczyk:
The idea is to make more smaller changes instead of one big/huge PR.

> FYI, both PRs use the same TomaszWaszczyk:master branch which makes it a bit hard for collaboration. I can't easily check out this PR since Git already thinks I already have the branch locally. In the future I'd recommend using different branches for different PRs - even if they're closely related.

Roger that, next PR will be on different branches. Thanks for the hint! ;-)

.github/workflows/lint.yml
```diff
@@ -41,3 +41,28 @@ jobs:
...
+          submodules:   recursive
+      
+      - name:           Add cargo-spellcheck
+        run:            cargo install cargo-spellcheck
+      
+      - name:           Run spellcheck
+        uses:           actions-rs/cargo@master
+        with:
+          command:      spellcheck
+          args:         check  -m 1 -vv $(find modules/currency-exchange -name *.rs)
```
HCastano, line 68:
```suggestion
          args:         check -m 1 -vv $(find modules/currency-exchange -name *.rs)
```

One nitpick (and I think it was even introduced by me in the other PR 😅 )

HCastano (approved):
As long as the CI passes, this should be good 👍

TomaszWaszczyk:
I am really much more stressed than usually to merge the PR. But well... it is time to go out of my comfort zone (PR without merging ;D..)

.github/workflows/lint.yml
```diff
@@ -41,3 +41,28 @@ jobs:
...
+          submodules:   recursive
+      
+      - name:           Add cargo-spellcheck
+        run:            cargo install cargo-spellcheck
+      
+      - name:           Run spellcheck
+        uses:           actions-rs/cargo@master
+        with:
+          command:      spellcheck
+          args:         check -m 1 -vv $(find modules/currency-exchange -name *.rs)
```
HCastano, line 68: This doesn't find anything when I run this locally. I think what we want here is:

```suggestion
          args:         check -m 1 -vv $(find modules/currency-exchange/src -name "*.rs")
```

HCastano:
I can't reproduce the CI failure locally, @tomusdrw any ideas?

tomusdrw (approved):

TomaszWaszczyk:
I am really curious how have You concluded the commit `Use bash.` was thinking about it around ~two days and the output from all that thinking was exactly nothing. Maybe not nothing, my best guess was that my shell and Github's shell are not the same but the commit break my mind. No idea.

Cool that `Auto-merge is enabled` will be easier and more time efficient.

tomusdrw:
@TomaszWaszczyk the difference is that `run` is executed within a shell (bash) so we can use hacks like `$(find ...)`. just passing `args` to the `cargo` action was not sufficient, cause it wasn't going through any shell and was literally being passed as arguments to the `cargo spellcheck` command.

PR #925, opened 2021-04-22
Add Wococo<>Rococo Header Relayer

HCastano:
I took the fastest approach I could for this, which was to simply add new modules for
Wococo related code. However, given that Wococo and Rococo use the same runtime there is
definitely room to remove code duplication. With that said, I don't think it's worthwhile
to improve things just yet - I'd rather focus on making sure this works first.

~~Currently a draft since I haven't tested it with the Polkadot side.~~

I have a Polkadot branch working with this 🎉


tomusdrw (approved):

HCastano:
Gonna merge this since the changes introduced between Tomek's review and now were trivial (e.g `spec_version` bump)

PR #927, opened 2021-04-23
Make sure that relayers have dates in logs.

tomusdrw:
In #845 I accidentally disabled dates from relayers as well, I somehow thought that they use `initialize_relay` function and forgot that we share the same CLI. This PR tries to bring the old behavior back.


adoerr (approved):
LGTM

HCastano (approved):

PR #935, opened 2021-04-27
Relay basic single-bit message dispatch results back to the source chain

### LINKED ISSUE #547: Delivery confirmation should contain dispatch result as well.
Currently on the `source` chain we only know if the message was delivered or not, we can't tell if it was dispatched successfuly.

I think the reason for that is that initially we were considering splitting delivery & dispatch or having a more complicated scheduling system, so these two things would be completely disjoint.

I'm more towards coupling the two (as it is implemented now) and allowing more complicated mechanisms to be built on top if needed. For instance a scheduler could be a "high-level app", where the delivery&dispatch would report it was always successful, cause it only adds the item to the dispatch queue. Anyone using this kind of high-level app would be required to build their own dispatch result notification if they need it.

Having dispatch results on the source chain would allow schemes like the one described in #543 (second variant).

svyatonik:
closes #547
on top of #911 => draft

This PR introduces basic mechanism to solve #547 that Tomek has recently suggested. So for every delivered message confirmation would contain single-bit which will mean (in our configuration) whether message (in ur config: call) has been successfully dispatched or not. At the source chain the `MessagesDelivered` event would contain these dispatch results. So the message sender may check events (also see TODOs below) to check result of their messages.

More sophisticated applications may introduce their own relays to relay extended dispatch results, but for simplest bridge apps single bit should be enough && I agree that it isn't super-heavy to maintain.

TODOs left for doing in this PR:
- [x] weights must be updated after #911;

TODOs left for future PRs:
- [x] ~instead of examining events (=> additional overload in `on_block_finalize()`) we need to introduce `trait OnMessagesDelivered { fn on_messages_delivered(messages: &DeliveredMessages) {} }`. The method will be called only on blocks with confirmation transactions, so it must be easier to use~ done; 
- [ ] example pallet (built into Rialto or Millau) that will send its messages (from time to time on or demand) and will do perform some actions if dispatch result has been successful.


primitives/messages/src/source_chain.rs
```diff
@@ -135,6 +135,15 @@ pub trait MessageDeliveryAndDispatchPayment<AccountId, Balance> {
 	}
 }
 
+/// Handler for delivered messages.
+#[impl_trait_for_tuples::impl_for_tuples(30)]
+pub trait OnMessagesDelivered {
+	/// Called when we receive confirmation that our messages have been delivered to the
+	/// target chain. The confirmation aso has single bit dispatch result for every
+	/// confirmed message (see `DeliveredMessages` for details).
+	fn on_messages_delivered(_lane: &LaneId, _messages: &DeliveredMessages) {}
```
svyatonik, line 144: This'll need to do something with weights - I'm not yet sure how to deal with this. Change benchmarks to worst-case result and then refund? But what is worst case when there are e.g. multiple apps using different lanes for their purposes. And because of refund policy we'll need to decrease max number of messages in the confirmation tx. Probably just reserve single read and single write for the every registered callback? I'm not yet sure - will leave it for future PRs as well

primitives/messages/src/source_chain.rs
```diff
@@ -135,6 +135,15 @@ pub trait MessageDeliveryAndDispatchPayment<AccountId, Balance> {
 	}
 }
 
+/// Handler for delivered messages.
+#[impl_trait_for_tuples::impl_for_tuples(30)]
+pub trait OnMessagesDelivered {
+	/// Called when we receive confirmation that our messages have been delivered to the
+	/// target chain. The confirmation aso has single bit dispatch result for every
```
tomusdrw, line 142:
```suggestion
	/// target chain. The confirmation also has single bit dispatch result for every
```

primitives/messages/src/source_chain.rs
tomusdrw: Yeah, I think the general sentiment about these kind of handlers is that they should be as simple as possible. Having one read + one write seems reasonable. In future we can prepare some tooling to kind of "subscribe" to interesting messages and for instance have them processed in `on_initialize` of the pallet or some manual trigger (extrinsic).

Might be good to add this assumption to the docs though, so that implementers are aware of that.

primitives/runtime/src/messages.rs
```diff
@@ -36,6 +36,12 @@ pub enum DispatchFeePayment {
 /// Message dispatch result.
 #[derive(Encode, Decode, RuntimeDebug, Clone, PartialEq, Eq)]
 pub struct MessageDispatchResult {
+	/// Dispatch result flag. This flag is relayed back to the source chain and, generally
+	/// speaking, may bring any (that fits in single bit) information from the dispatcher at
+	/// the target chain to the message submitter at the source chain. If you're using immediate
+	/// call dispatcher, then it'll be result of the dispatch - `true` if dispatch has succeeded
+	/// and `false` otherwise.
+	pub dispatch_result: bool,
```
tomusdrw, line 44:
```suggestion
	pub dispatched: bool,
```
I find it a bit confusing to have `dispatch_result` and `bool` - it kind of signifies that there is some extra result embedded in case that's true.

I'd call this simply `dispatched: true/false` cause it's part of `MessageDispatchResult` struct already. It's just a personal preference, there is nothing wrong really with `dispatch_result`, so if you don't find my arguments convincing feel free to leave as-is.

primitives/messages/src/source_chain.rs
```diff
@@ -135,6 +135,15 @@ pub trait MessageDeliveryAndDispatchPayment<AccountId, Balance> {
 	}
 }
 
+/// Handler for delivered messages.
+#[impl_trait_for_tuples::impl_for_tuples(30)]
+pub trait OnMessagesDelivered {
```
tomusdrw, line 140: Since the pallet is used on both source and target chains it's sometimes hard to identify which associated types are for which side.

A generic name like `OnMessagesDelivered` imho is equally valid as a callback for messages being delivered to the target chain and confirmations being delivered back to the source chain.

I'd vote for calling the handler (and the associated type) `OnDeliveryConfirmed`, cause I think it doesn't have this kind of ambiguity.
```suggestion
pub trait OnDeliveryConfirmed {
```

modules/messages/src/weights.rs
```diff
@@ -188,110 +188,110 @@ impl<T: frame_system::Config> WeightInfo for RialtoWeight<T> {
...
-			.saturating_add((9_000 as Weight).saturating_mul(i as Weight))
+		(486_301_000 as Weight)
+			.saturating_add((10_000 as Weight).saturating_mul(i as Weight))
 			.saturating_add(RocksDbWeight::get().reads(5 as Weight))
 			.saturating_add(RocksDbWeight::get().writes(3 as Weight))
 	}
 	fn receive_message_proofs_with_large_leaf(i: u32) -> Weight {
-		(175_300_000 as Weight)
-			.saturating_add((6_000 as Weight).saturating_mul(i as Weight))
+		(178_139_000 as Weight)
```
tomusdrw, line 274: Which machine did you use to generate the weights? They are surprisingly similar to the previous ones :)

tomusdrw (approved):
lgtm!

modules/messages/src/weights.rs
svyatonik: My laptop. Previous weights were also generated here, so it's expected

primitives/messages/src/source_chain.rs
svyatonik: #1012

PR #936, opened 2021-04-27
Reconnect to failed client in on-demand relay background task

svyatonik:
Normally connection errors are processed by relay loop & whenever connection is lost, we're reconnecting to the failed client. But on-demand background task is technically not a relay loop, so connection errors weren't processed there. So there may be a case (which has actually happened last 2 days) when: (1) on-demand relay has been inactive (2) we were failing to read best block number from source/target nodes => on-demand client was never started. What I currently don't understand is why the same client was working fine with other requests - probably that's because of that not-yet-patched-in-our-codebase "Invalid request ID" error in jsonrpsee? But this patch should solve the stall issue


tomusdrw (approved):
looks reasonable to me!

PR #937, opened 2021-04-27
Do not panic if pallet is not yet initialized

svyatonik:
Right now there's a panic in `fn submit_finality_proof()` call if pallet has not been initialized yet


HCastano (commented):
Do you have a "real world" workflow where this happens? Because from what I remember there was no way to make this happen with the normal bootstrapping mechanisms (e.g `GenesisConfig` or `initialize_bridge()`)

svyatonik:
Just start chains and run headers relay without initializing the bridge - there are no any guarantees that prevent this expect from failing.

tomusdrw (approved):
With runtime I'd rather be on the safe side always and avoid panicking at all cost.

PR #938, opened 2021-04-27
Update to jsonrpsee alpha6

svyatonik:
this most probably fixes the "Invalid request id"
please do not manually deploy it - I want to see whether relayer with #936 is able to work with unpatched jsonrpsee


adoerr (approved):
LGTM

PR #940, opened 2021-04-27
Remove Westend<>Rococo header sync

### LINKED ISSUE #929: Remove Westend<>Rococo bridge traces
It's not feasible currently, so we should remove it. Note that Westend -> Millau headers sync can stay.

HCastano:
Turns out that bridging to Westend<>Rococo isn't the right testing approach for us at the
moment, so this removes the code related to that bridge.

Closes #929


primitives/chain-rococo/src/lib.rs
```diff
@@ -55,14 +55,14 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 /// See: https://github.com/paritytech/polkadot/blob/master/runtime/rococo/src/lib.rs
 #[derive(parity_scale_codec::Encode, parity_scale_codec::Decode, Debug, PartialEq, Eq, Clone)]
 pub enum Call {
-	/// Westend bridge pallet.
+	/// Generic bridge pallet.
```
HCastano, line 58: I left this enum and inner type in tact since I'm using it as part of the Wococo bridge

adoerr (approved):
LGTM

PR #944, opened 2021-04-28
Token-swap-over-bridge pallet

### LINKED ISSUE #543: Token swap pallet
It would be cool to have a very simple mechanism of swapping the tokens between bridged chains (think KSM<>DOT).

We could use `atomic-swap` pallet from Substrate repo, but with bridges we can use simpler mechanisms to achieve the same.

Order matching would happen off-chain, the pallet should only be responsible for actually performing an atomic exchange of the tokens.

There should be two workflows possible, depending on whether the pallet is included on both chains or not.

## If both chains include the `Swap` pallet.
1. `source::Alice` and `target::Bob` agree to exchange `source(X) tokens <-> target(Y) tokens` and they pick a `random_nonce`.
2. `let transaction_id = hash(random_nonce ++ source::Alice ++ target::Bob ++ source(X) ++ target(Y))`
3. Both send a similar transaction on their respective chains. Let's analyze Alice's case.
4. `source::Alice` sends a transaction to the pallet:
```rust
source::Swap::exchange(random_nonce, target::Bob, source(X), target(Y));
```
5. The pallet then:
5.1. Lock's Alice's `source(X)` tokens (or transfers them to `source::derive_account_id(transaction_id)`) and stores that fact.
5.2. Send a transaction over the bridge on Alice's behalf (`CallOrigin::SourceAccount(source::Alice)`) with these details:
```rust
target::Swap::claim(random_nonce, target::Bob, source(X), target(Y));
```
6. Depending on the target pallet state (Bob's sent his `exchange` or not) there are two things happening:
6.1. The `target::Swap` pallet records the `claim` under `transaction_id` if Bob has not sent the exchange transaction. If Bob sends the `target::Swap::exchange(random_nonce, source::Alice, source(X), target(Y))` transaction, the tokens are immediately transferred to Alice (like in 6.2) and a `source::Swap::claim` call is sent over the bridge.
6.2. The `target::Swap` performs a transfer for `target(Y)` tokens from `target::derive_account_id(transaction_id` to `target::derive_account_id(source::Alice)` or unlocks Bob's tokens and sends them to `target::derive_account_id(source::Alice)`.
7. Data stored in the pallet in both `target::Swap::claim` and `source::Swap::exchange` expire after a period of `1 day` and `2 days` respectively and the funds locked in `exchange` are redeemable by the original owner.

There is a risk for Alice to have her funds locked for 2 days if Bob decides not to publish his transaction. This could be solved bif Bob shared his signature over a "cancel order" with Alice, which she can use to cancel the transaction. This signature would have to be wrapped in a `target `chain transaction, and would cancel the claim on the target chain & send a message over the bridge to cancel the `exchange` on the source chain.

## If only `source` chain includes the `Swap` pallet.
The workflow:
1. `source::Alice` and `target::Bob` agree to exchange `source(X) tokens <-> target(Y) tokens` and choose unique `random_nonce`.
2. `source::Alice` requests from `target::Bob` a signature for `signed_message` defined as below:
```rust
let transaction_id = hash(random_nonce ++ source::Alice ++ target::Bob ++ source(X) ++ target(Y))
let bob_account_id_on_source = source::derive_account_id(target::Bob);
let alice_account_id_on_target = target::derive_account_id(source::Alice);
// the account id we derive from the transaction id
let lock_account_id = target::derive_account_id(transaction_id)
// a signature which can be used with `CallOrigin::TargetAccount(lock_account_id, target::Bob, bob_signature)`
// this will allow the bridge to perform a transfer on Bob's behalf.
let bob_transfer = target::Transfer(target::Bob, alice_account_id_on_target, Y);
let bob_signed_message = ( bob_transfer ++ lock_account_id );
```
4. `source::Alice` calls into the pallet with:
```rust
source::Swap::start_swap(random_nonce, target::Bob, source(X), source(Y), bob_transfer, bob_signature);
```
5. The pallet locks `source::Alice`'s funds (or transfers them to `lock_account_id`) and sends the `bob_transfer` message over the bridge using `CallOrigin::TargetAccount(lock_account_id, target::Bob, bob_signature)`.

6. Currently neither the delivery mechanism nor dispatch doesn't relay the result of dispatch back to the source chain (i.e. on the source chain we only know the message was delivered, but we don't know if it dispatched successfully). I think it's worth changing (shouldn't be that complex) - but there is an alternative solution, which requires Bob sending locking his funds in a temporary account first (with some option to transfer back by sending Alice's signed-call on the source chain).

7. Assuming after some time we know the result of the dispatch on the source chain, we can now proceed to either:
7.1. Bob claiming Alice's locked funds. Since Bob interacts with the target chain, he would send a message over the bridge (using `CallOrigin::SourceAccount(target::Bob)`). This call should only succeed if the transfer dispatch was succesful.
```rust
source::Swap::claim_swap(random_nonce, source::Alice, source(X), target(Y))
```
7.2. Alice unlocking her funds by sending `source::Swap::cancel_swap(random_nonce, target::Bob, source(X), source(Y));` which should only succeed if the transfer dispatch failed. 

In this scheme Alice pays all the fees and is more in risk in case Bob doesn't have enough funds on the target chain (or front-runs the bridge message), but there is no time-locking involved. In a happy case the "cost" of the swap is very similar - i.e. each party sends exactly one transaction on their respective chains.

svyatonik:
closes #543 (second option is implemented)
on top of #935 => draft

It won't probably be reviewed (or I'll be unable to respond to review) until at least mid May - I'm opening it early, because I have some questions + some changes for messages pallet that may be illustrated with this PR.

So the actual set of questions:
1) ~I'm not sure why we need this unique `random_nonce` mentioned in #543 @tomusdrw . I guess it is somewhat linked to the replay protection, but I don't see how it can be exploited. Could you elaborate? Without reviewing the code - just the explanation on why we need it~;
2) ~I'm not yet sure how to deal with weights of `OnMessagesDelivered` callbacks (see https://github.com/paritytech/parity-bridges-common/pull/935#discussion_r622178967). Most probably it'll be pre-hardcoded to `1 db read + 1 write per delivered message`. Then all callbacks must be implemented as something like: `for message in &messages { if check_if_we_are_waiting_this_message(message) { remember_to_do_something_in_next_on_initialize(); } }`~ #1012.

TODOs for this PR:
1) ~deal with [this comment](https://github.com/paritytech/parity-bridges-common/pull/944#issuecomment-829088415) - so iiuc `transaction_id` must be used to protect against Alice starting the claim again. I've missed that part when implementing~ https://github.com/paritytech/parity-bridges-common/pull/944/commits/cda7e8f7889e2007994f78939a711ff9121389b1

TODOs for parallel PRs:
1) add runtime-internal function to the messages pallet;
2) ~add `enum MessageOrigin { Runtime, Root, User }` and pass it to the `LaneMessageVerifier::verify_message()` instead of the current `sender`. This would allow to implement lanes that are dedicated to some applications or general usage lanes~ This probably can be solved by using custom
```rust
	#[pallet::origin]
	#[derive(RuntimeDebug, Clone, Eq, PartialEq, Encode, Decode)]
	pub struct Origin<T: Config<I>, I: 'static = ()> {
		/// Sender account.
		sending_on_behalf_of: T::AccountId,
		/// Phantom marker.
		_phantom: sp_std::marker::PhantomData<I>,
	}
```
and then passing this origin to the `send_message` function. So this is most probably for the runtime-integration PR;

TODOs for future PRs:
1) integrate into runtime;
2) generate token swap calls.


svyatonik:
Just my current thoughts (no need to comment now).

So from what I understand nonce should prevent Alice from resubmitting the same swap again after it has been claimed/cancelled. With current pallet it is possible. So if Bob is using some permanent account, Alice may swap until Bob's account depletes. I don't yet see how nonce may solve this problem (given current pallet implementation) - Alice may reuse the same signature in all `create_swap` calls. So probably pallet needs to remember the `nonce`. OTOH the pallet isn't verifying `bob_signature` now, so we can't check if `nonce` matches the nonce from signature. Then the pallet needs to remember used signatures. but it'll require a storage entry for every completed swap => may be used as a way of bloating runtime storage. Will need to think how to do that properly.

modules/token-swap/src/lib.rs
```diff
@@ -0,0 +1,958 @@
...
+		///
+		/// **WARNING**: the submitter of this transaction is responsible for verifying:
+		///
+		/// 1) that the `bridged_currency_transfer` represents a valid token transfer call that transfers
+		///    `swap.target_balance_at_bridged_chain` to his `source_account_at_bridged_chain` account;
+		/// 2) that either the `source_account_at_bridged_chain` already exists, or the
+		///    `swap.target_balance_at_bridged_chain` is above existential deposit of the Bridged chain;
+		/// 3) the `target_public_at_bridged_chain` matches the `swap.target_account_at_bridged_chain`;
+		/// 4) the `bridged_currency_transfer_signature` is valid and generated by the owner of the
+		///    `target_public_at_bridged_chain` account.
```
tomusdrw, line 189:
```suggestion
		///    `target_public_at_bridged_chain` account (read more about [`CallOrigin::TargetAccount`]).
```
Maybe we can refer to `CallOrigin::TargetAccount` docs for more details about the signature?

modules/token-swap/src/lib.rs
```diff
@@ -0,0 +1,958 @@
...
+		///
+		/// Violating rule#1 will lead to losing your `source_balance_at_this_chain` tokens. Violating other
```
tomusdrw, line 191: Hmm, AFAICT if the swap is in `Failed` state you can still cancel it, no? So no funds should be lost.

primitives/token-swap/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+	/// The `target_account_at_bridged_chain` is temporary and only have funds for single swap.
+	///
+	/// ***WARNING**: if `target_account_at_bridged_chain` still exists after the swap has been
+	/// completed (either by claiming or cancelling), the `source_account_at_this_chain` will be able
+	/// to restart the swap again and repeat the swap until `target_account_at_bridged_chain` depletes.
+	TemporaryTargetAccountAtBridgedChain,
+	/// This swap type prevents `source_account_at_this_chain` from restarting the swap after it has
+	/// been completed. There are two consequences:
+	///
+	/// 1) the `source_account_at_this_chain` won't be able to call `start_swap` after given <ThisBlockNumber>;
```
tomusdrw, line 36: I might be missing something, but I feel this crucial condition is not enforced in `create_swap`.

tomusdrw (commented):
Looks good! One critical thing (`crate_swap` enforcing `LockClaim` block number) and some smaller suggestions.

modules/token-swap/src/lib.rs
svyatonik: If you've signed any call that has succeeded at the target chain (e.g. `system::remark`), you can't cancel it on this chain.

primitives/token-swap/src/lib.rs
svyatonik: No, you are not missing something - this check must be added. Thanks for spotting this!

modules/token-swap/src/lib.rs
```diff
@@ -0,0 +1,958 @@
...
+		/// Tokens balance type at the Bridged chain.
+		type BridgedBalance: Parameter;
+		/// Account identifier type at the Bridged chain.
+		type BridgedAccountId: Parameter;
+		/// Account public key type at the Bridged chain.
+		type BridgedAccountPublic: Parameter;
+		/// Account signature type at the Bridged chain.
+		type BridgedAccountSignature: Parameter;
+		/// Converter from raw hash (derived from Bridged chain account) to This chain account.
+		type FromBridgedToThisAccountIdConverter: Convert<H256, Self::AccountId>;
```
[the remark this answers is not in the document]
svyatonik: I can move `Balance`, `AccountId`, `Public` and `Signature` fields from `relay_substrate_client::Chain` to `bp_runtime::Chain`. But that would also mean that even if you only deploying GRANPA pallet you'll need to specify those types as well. Other fields are too specific for messaging or this pallet needs.

Let's do that in a separate PR, though - it'll touch (I assume) too many files. So list of PRs to get this pallet ready:
1) this ^^ refactoring;
2) integrating into Millau or Rialto runtimes;
3) #962 
4) weights

modules/token-swap/src/lib.rs
tomusdrw: Oh, yeah, you're right. Actually given we allow arbitrary calls here, I'd suggest having the user provide `weight` as well. This will enable calling into proxy accounts, mutlisigs, etc. Also one less thing the pallet needs to know about the remote chain.

modules/token-swap/src/lib.rs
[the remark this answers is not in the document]
tomusdrw: Makes sense!

modules/token-swap/src/lib.rs
svyatonik: Right. I've also moved `bridged_chain_spec_version` to arguments

tomusdrw (approved):

PR #947, opened 2021-04-29
Bump Substrate WASM builder

HCastano:
Looks like we're using an outdated version. I'm also have some suspicions that this is
causing some issues with our Subtree builds in Polkadot since they're on version 2.0.

### DIFF
--- a/bin/millau/node/src/chain_spec.rs
+++ b/bin/millau/node/src/chain_spec.rs
@@ -164,7 +164,7 @@ fn testnet_genesis(
 ) -> GenesisConfig {
 	GenesisConfig {
 		frame_system: SystemConfig {
-			code: WASM_BINARY.to_vec(),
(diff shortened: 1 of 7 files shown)


bin/millau/node/src/chain_spec.rs
```diff
@@ -164,7 +164,7 @@ fn testnet_genesis(
 ) -> GenesisConfig {
 	GenesisConfig {
 		frame_system: SystemConfig {
-			code: WASM_BINARY.to_vec(),
+			code: WASM_BINARY.expect("Millau development WASM not available").to_vec(),
```
HCastano, line 167: Tbh I'm not sure how this can happen. I tried `SKIP_WASM_BUILD=1 cargo run -p rialto-bridge-node -- --dev` but nothing terrible happened 🤷‍♂️

tomusdrw (approved):

PR #951, opened 2021-05-02
Endow accounts and add `bridgeIds` to chainspec.

### LINKED ISSUE #943: Add balances to derived DEV accounts
Currently the derived accounts of Alice, Bob, Charlie, etc do not have any balance on the target chain, it would cool to preload them with some Target tokens (but not all of them).

tomusdrw:
Closes #943 

Endows a bit more accounts for testing (Bob, Dave, Ferdie and Alice/Charile/Eve on other chain) and also adds chain spec property which defines configured `bridgeIds`, which will be useful for https://github.com/paritytech/parity-bridges-ui/issues/86

### DIFF
--- a/bin/millau/node/src/chain_spec.rs
+++ b/bin/millau/node/src/chain_spec.rs
@@ -71,6 +71,9 @@ impl Alternative {
 			serde_json::json!({
 				"tokenDecimals": 9,
 				"tokenSymbol": "MLAU",
+				"bridgeIds": {
+					"Rialto": bp_runtime::RIALTO_BRIDGE_INSTANCE,
+				}
 			})
 			.as_object()
 			.expect("Map given; qed")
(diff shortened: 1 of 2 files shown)


svyatonik (approved):
Probably worth endowing all `[Alice; Eve]`-derived accounts on both chains just for simplicity?

tomusdrw:
> Probably worth endowing all [Alice; Eve]-derived accounts on both chains just for simplicity?

I was thinking about this initially, but this originally comes from the purpose of demoing UI and it actually looks better if we have distinct accounts having balances.

Primarly I want it to be clear that `RialtoAlice` and `derived(RialtoAlice)` has nothing to do with `MillauAlice` and `derived(MillauAlice)`. By having different balances on these accounts it's a bit easier to observe.

HCastano (approved):

PR #958, opened 2021-05-04
Bump Substrate to `b094edaf`

HCastano:


bin/millau/node/src/cli.rs
```diff
@@ -63,7 +63,7 @@ pub enum Subcommand {
 	Revert(sc_cli::RevertCmd),
 
 	/// Inspect blocks or extrinsics.
-	Inspect(node_inspect::cli::InspectCmd),
+	Inspect(node_inspect::cli::InspectKeyCmd),
```
tomusdrw, line 66: :thinking:

HCastano: @tomusdrw What's with the 🤔?

tomusdrw: I guess this search & replace didn't go that well. 
https://github.com/paritytech/substrate/pull/8678/commits/3700a10625767ee9a76731fcb57b5638a5c33081

`node_inspect` has nothing to do with keys and should not have been touched in th original PR. CC @NukeManDan

NukeManDan: My mistake. I should have caught this, likely why the struct naming was called out in that PR. All good now?

NukeManDan: Ah or is a new PR to fix this issue needed in both places?

HCastano: @NukeManDan So we'll need a Substrate PR to revert the changes, and then a PR which bumps Substrate in this repo (and in that PR changes the name back)

NukeManDan: https://github.com/paritytech/substrate/pull/8787
Once this lands, I will open a PR to bump substrate and the name change.

tomusdrw: Awesome, thanks! We will fix that during our regular substrate-bump cycle, so you don't have to do it unless you really want to :) If you run into any other issues (that might be related to substrate bump - let us know).

PR #959, opened 2021-05-04
Fix hunspell typos - Dispatch crate

TomaszWaszczyk:


.github/workflows/lint.yml
```diff
@@ -62,4 +62,4 @@ jobs:
         run:            cargo install cargo-spellcheck
 
       - name:           Run spellcheck
-        run:            cargo spellcheck check -m 1 -vv $(find modules/currency-exchange/src -name "*.rs")
+        run:            cargo spellcheck check -m 1 -vv $(find modules/dispatch/src -name "*.rs")
```
tomusdrw, line 65:
```suggestion
        run:            cargo spellcheck check -m 1 -vv $(find modules/dispatch/src -name "*.rs") $(find modules/currency-exchange/src -name "*.rs")
```

We should probably keep the old checks as well to make sure we don't introduce new typos.

tomusdrw (approved):
lgtm!

.github/workflows/lint.yml
TomaszWaszczyk: In the future just add new `find` command? Thanks for the hint!

.github/workflows/lint.yml
```diff
@@ -62,4 +62,4 @@ jobs:
         run:            cargo install cargo-spellcheck
 
       - name:           Run spellcheck
-        run:            cargo spellcheck check -m 1 -vv $(find modules/currency-exchange/src -name "*.rs")
+        run:            cargo spellcheck check -m 1 -vv $(find modules/dispatch/src -name "*.rs") $(find modules/currency-exchange/src -name "*.rs")
```
HCastano, line 65: I think we should be fixing typos for `modules/*` instead of doing each one by one. We can then add `bin/*`, then `relays/*`, etc. in order to keep the typos PRs relatively manageable in size

HCastano (commented):
Looks fine, but @TomaszWaszczyk can you please squash the commits up to and including 79ee76bffd2d5f8cbd42652911bc587872407daa? Those commits are from #924 and shouldn't be part of this PR

.github/workflows/lint.yml
TomaszWaszczyk: Roger, will do that in the way.

HCastano:
Any updates here @TomaszWaszczyk?

TomaszWaszczyk:
In the second half of the week will deliver new commits, too much work.

tomusdrw:
@TomaszWaszczyk please reopen when you have time to take a look at this. I've made a PR to your fork with gitlab fixes.

PR #963, opened 2021-05-12
Pass source and target chain ids to account_ownership_digest

### LINKED ISSUE #950: Account ownership digest should include target chain instance.
Currently, when we construct `account_ownership_digest` it consists of:
```
target_call,
source_account_id,
target_spec_version,
source_instance_id,
````

However this means that in case there are two networks that are bridged with `source` and both happen to have the same `spec_version` and some subset of `target_call` the signature might be re-used on both chains.

To prevent that we might want to add `target_instance_id` to the digest as well, to make sure it is valid only for this one pair.

svyatonik:
closes #950 

The only change that affects logic is that additional argument to `fn account_ownership_digest()`. I've also renamed `InstanceId` to `ChainId` (this is what all other changed lines are about), because that's how it is used now. It was introduced initially for message lane RPCs (that are gone now) to identify **instance of the messages module**. But we have started to use it everywhere to identify chains, so this name is more appropriate now. If this rename causes problems with UI or some other PRs - could just rollback the rename for now.

### DIFF
--- a/bin/millau/node/src/chain_spec.rs
+++ b/bin/millau/node/src/chain_spec.rs
@@ -72,7 +72,7 @@ impl Alternative {
 				"tokenDecimals": 9,
 				"tokenSymbol": "MLAU",
 				"bridgeIds": {
-					"Rialto": bp_runtime::RIALTO_BRIDGE_INSTANCE,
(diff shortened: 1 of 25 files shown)


HCastano (approved):
cc @hbulgarini

tomusdrw (approved):

PR #964, opened 2021-05-12
Serve multiple lanes with single complex relay instance

svyatonik:
follow-up for #878 

Complex relay is now able to serve multiple message lanes. Following command starts relay that serves lanes 00000000 and 00000001 in both directions:
```bash
./bin/substrate-relay relay-headers-and-messages millau-rialto\
	--millau-host=$MILLAU_HOST\
	--millau-port=$MILLAU_PORT\
	--millau-signer=//Charlie\
	--rialto-host=$RIALTO_HOST\
	--rialto-port=$RIALTO_PORT\
	--rialto-signer=//Charlie\
	--lane=00000000\
	--lane=00000001\
	--prometheus-port=9700
```

We may add another lane in test deployments just to prove that it works, but we already had problems with account depletion. So probably not the best idea to do that now

### DIFF
--- a/relays/bin-substrate/src/cli/relay_headers_and_messages.rs
+++ b/relays/bin-substrate/src/cli/relay_headers_and_messages.rs
@@ -41,7 +41,7 @@ pub enum RelayHeadersAndMessages {
 #[derive(StructOpt)]
 pub struct HeadersAndMessagesSharedParams {
 	/// Hex-encoded lane identifiers that should be served by the complex relay.
-	#[structopt(long)]
+	#[structopt(long, default_value = "00000000")]
 	lane: Vec<HexLaneId>,
 	#[structopt(flatten)]
 	prometheus_params: PrometheusParams,
--- a/relays/bin-substrate/src/cli/relay_headers_and_messages.rs
+++ b/relays/bin-substrate/src/cli/relay_headers_and_messages.rs
@@ -40,9 +40,9 @@ pub enum RelayHeadersAndMessages {
 /// Parameters that have the same names across all bridges.
 #[derive(StructOpt)]
 pub struct HeadersAndMessagesSharedParams {
-	/// Hex-encoded lane id that should be served by the relay. Defaults to `00000000`.
-	#[structopt(long, default_value = "00000000")]
-	lane: HexLaneId,
+	/// Hex-encoded lane identifiers that should be served by the complex relay.
+	#[structopt(long)]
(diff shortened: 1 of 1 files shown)


relays/bin-substrate/src/cli/relay_headers_and_messages.rs
```diff
@@ -143,46 +143,48 @@ impl RelayHeadersAndMessages {
...
-				target_to_source_headers_relay: Some(left_to_right_on_demand_headers),
-				lane_id: lane,
-				metrics_params: metrics_params
-					.clone()
-					.disable()
-					.metrics_prefix(messages_relay::message_lane_loop::metrics_prefix::<RightToLeftMessages>(&lane)),
-			})
-			.map_err(|e| anyhow::format_err!("{}", e))
-			.boxed();
+			let mut message_relays = Vec::with_capacity(lanes.len() * 2);
```
HCastano, line 146:
```suggestion
			// Need 2x capacity since we consider both directions for each lane
			let mut message_relays = Vec::with_capacity(lanes.len() * 2);
```

relays/bin-substrate/src/cli/relay_headers_and_messages.rs
```diff
@@ -40,9 +40,9 @@ pub enum RelayHeadersAndMessages {
 /// Parameters that have the same names across all bridges.
 #[derive(StructOpt)]
 pub struct HeadersAndMessagesSharedParams {
-	/// Hex-encoded lane id that should be served by the relay. Defaults to `00000000`.
-	#[structopt(long, default_value = "00000000")]
```
HCastano, line 44: Could we not continue to have a default that's just `vec!["00000000"]`?

HCastano (approved):

relays/bin-substrate/src/cli/relay_headers_and_messages.rs
svyatonik: Sorry - I was unable to interpret "not continue to have" part :) I guess you wanted to return default value for this param, right? I was going to remove it, because there's actually nothing special in lane 00000000 && imo it makes no sense to have default for this. But there are other commands where this default value is still used, so I've reverted it for now. If you mean something else, please respond here

HCastano: Haha, yeah giving it a second read it does sound confusing, sorry! But you got it, I wanted to keep the default value there.

PR #965, opened 2021-05-12
Start finality relay spec-version guards for Rococo <> Wococo finality relays

svyatonik:
I'll add another guard (`abort_when_account_balance_decreased`) in follow-up PR

### DIFF
--- a/relays/bin-substrate/src/finality_pipeline.rs
+++ b/relays/bin-substrate/src/finality_pipeline.rs
@@ -46,6 +46,9 @@ pub trait SubstrateFinalitySyncPipeline: FinalitySyncPipeline {
 		Ok(params)
 	}
 
+	/// Start finality relay guards.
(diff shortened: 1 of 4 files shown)


relays/bin-substrate/src/finality_pipeline.rs
```diff
@@ -46,6 +46,9 @@ pub trait SubstrateFinalitySyncPipeline: FinalitySyncPipeline {
 		Ok(params)
 	}
 
+	/// Start finality relay guards.
+	fn start_relay_guards(_target_client: &Client<Self::TargetChain>) {}
```
HCastano, line 50: Is the empty default impl intentional? If so, can you add a comment explaining why?

HCastano (approved):
I totally forgot this was a thing (since we don't have it on our testnets), good addition 👍

tomusdrw (approved):

PR #973, opened 2021-05-19
Avoid hashing headers twice in verify_justification

svyatonik:


tomusdrw (approved):

HCastano (approved):
I know it's a small change, but maybe worth upstreaming?

PR #974, opened 2021-05-21
Increase pause before starting relays

svyatonik:
This morning containers `deployments_relay-millau-rialto_1`, `deployments_relay-messages-millau-to-rialto-lane-00000001_1`, `deployments_relay-messages-rialto-to-millau-lane-00000001_1` and `deployments_relay-headers-westend-to-millau_1` have failed to start because `curl http://*-node-*:9933` to Millau/Rialto nodes have exited with non-zero code (connection refused). I was thought initially that something has changed with the latest Substrate update - e.g. RPCs are not exposed over HTTP anymore. But it looks like after some time `curl` succeeds. So probably RPCs are now started with delay? Haven't found anything in Substrate commits + can't reproduce this locally. So increasing delay seems the only possible solutions to avoid this.

### DIFF
--- a/deployments/bridges/poa-rialto/entrypoints/relay-headers-poa-to-rialto-entrypoint.sh
+++ b/deployments/bridges/poa-rialto/entrypoints/relay-headers-poa-to-rialto-entrypoint.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -xeu
 
-sleep 3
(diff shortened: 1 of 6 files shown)


tomusdrw:
Might be just that our test vm is being busy with other things.

tomusdrw (approved):

PR #977, opened 2021-05-21
Add JSON type definitions for Rococo<>Wococo bridge

HCastano:


adoerr (approved):

deployments/types/rococo-wococo.json
```diff
@@ -0,0 +1,52 @@
...
+			"AuthoritiesChange": "Vec<AuthorityId>",
+			"ChangesTrieRoot": "RococoBlockHash",
+			"SealV0": "SealV0",
+			"Consensus": "Consensus",
+			"Seal": "Seal",
+			"PreRuntime": "PreRuntime"
+		}
+	},
+	"--2": "Wococo Types",
+	"WococoBalance": "u128",
```
tomusdrw, line 28: Isn't Rococo and Wococo always going to be the same? I feel this is a bit duplicated.

deployments/types/rococo-wococo.json
```diff
@@ -0,0 +1,52 @@
...
+		"_enum": {
+			"Other": "Vec<u8>",
+			"AuthoritiesChange": "Vec<AuthorityId>",
+			"ChangesTrieRoot": "WococoBlockHash",
+			"SealV0": "SealV0",
+			"Consensus": "Consensus",
+			"Seal": "Seal",
+			"PreRuntime": "PreRuntime"
+		}
+	}
```
tomusdrw, line 51: This is most likely not necessary - it was only really needed for Millau, because the regular `Header` type has smaller hash size hardcoded.

While this is not wrong obviously, I think it might make sense to keep the types minimal, so that it's easier to integrate them in the future with polkadot.js/api core types.

deployments/types/rococo.json
```diff
@@ -0,0 +1,14 @@
+{
+	"Fee": "RococoBalance",
+	"Balance": "RococoBalance",
+	"BlockHash": "RococoBlockHash",
+	"BlockNumber": "RococoBlockNumber",
+	"BridgedBlockHash": "WococoBlockHash",
```
tomusdrw, line 6: I've started wondering how are we going to handle multiple bridges with **different** types here, but it's more of a Metadata/UI problem really.

PR #979, opened 2021-05-25
Custom justification verification

svyatonik:
This PR introduces custom GRANDPA justification verification function. I suggest to use this function instead of original `fn finality_grandpa::validate_commit()`. Original implementation uses same structs for both voting and validation, therefore there are many internal buffers and data structures that are only used during voting.

I've also added a set of tests to demonstrate that the current version of code works exactly as the original one. Technically justification is an offchain struct, so having small differences in the implementation should not be a consensus issue (ofc it'll be bad if there's an actual bug in the code, which will reject valid or accept invalid justification). So we may add some more optimizations later. Like e.g. original implementation will process all precommits even if threshold is reached - we may just stop after this. But for the beginning I'd like both to be the same.

I've ran [customized benchmarks](https://gist.githubusercontent.com/svyatonik/91d0b226ec35cdb5fd1d1aaeaa5c222d/) and got [this result](https://docs.google.com/spreadsheets/d/156UV-cSyp7NP-iDSysIRFfk0poxMYlNEVOgw98aBJqE/edit?usp=sharing). You can see that new version is: (1) ~10-30% faster than previous when `v` (number of headers in ancestry) grows (2) ~5-10% when `p` (number of precommits) grows and (3) ~99% faster for large values of `p` and `v`.


primitives/header-chain/tests/implementation_match.rs
```diff
@@ -0,0 +1,317 @@
...
+/// Implementation of `finality_grandpa::Chain` that is used in tests.
+struct AncestryChain(bp_header_chain::justification::AncestryChain<TestHeader>);
+
+impl AncestryChain {
+	fn new(ancestry: &[TestHeader]) -> Self {
+		Self(bp_header_chain::justification::AncestryChain::new(ancestry))
+	}
+}
+
+impl finality_grandpa::Chain<TestHash, TestNumber> for AncestryChain {
```
HCastano, line 46: If this implementation isn't up to date with what's in `finality-grandpa` our checks could fall out of line with what's considered "valid", is that right?

HCastano (commented):
I've got one major concern with the validity of the `AncestryChain` after use, but other than that I think it looks good.

It would definitely be good to get another set of eyes on it given how crucial this is. Maybe @andresilva if you have time, but @adoerr will do too 😉

primitives/header-chain/src/justification.rs
```diff
@@ -63,119 +87,136 @@ pub fn verify_justification<Header: HeaderT>(
...
+		//
+		// there's a lot of code in the `VoteGraph::insert` method inside `finality-grandpa` crate,
+		// but in the end it is only used to find GHOST, which we don't care about. The only thing
+		// that we care about is that the justification target has enough weight
+		cumulative_weight = cumulative_weight.checked_add(authority_info.weight().0.into()).expect(
+			"sum of weights of ALL authorities is expected not to overflow - this is guaranteed by\
+				existence of VoterSet;\
+				the order of loop conditions guarantees that we can account vote from same authority\
+				multiple times;\
+				thus we'll never overflow the u64::MAX;\
```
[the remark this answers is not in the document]
svyatonik: (1) total weight of all authorities in the `VoterSet` can't overflow and (2) we only accept vote from single authority (it is ignored above) => it can't overflow.

primitives/header-chain/tests/implementation_match.rs
svyatonik: There's no implementation of `Chain` in `finality-grandpa`. Is that what you mean?

primitives/header-chain/src/justification.rs
```diff
@@ -20,28 +20,52 @@
...
+	/// The authority has provided an invalid signature.
 	InvalidAuthoritySignature,
-	/// The justification has precommit for the header that has no route from the target header.
-	InvalidPrecommitAncestryProof,
-	/// The justification has 'unused' headers in its precommit ancestries.
-	InvalidPrecommitAncestries,
+	/// The justification contains precommit for header that has lower number than the commit header.
+	PrecommitTargetHasLowerNumberThanCommitTarget,
+	/// The justification contains precommit for header that is not a descendant of the commit header.
+	PrecommitTargetIsNotCommitDescendant,
```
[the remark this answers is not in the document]
svyatonik: I've replaced both with `PrecommitIsNotCommitDescendant`

primitives/header-chain/tests/implementation_match.rs
HCastano: Sorry, I meant in [Substrate](https://github.com/paritytech/substrate/blob/master/client/finality-grandpa/src/justification.rs#L224), not `finality-grandpa`

HCastano (approved):

primitives/header-chain/tests/implementation_match.rs
svyatonik: In theory - yes. But in practice I don't think that the meaning of `fn ancestry()` would change :) There's also a lot of other places where we may diverge. But again - I'm not sure whether it is critical issue, or not.

For example - AFAIU the valid justification is justification that presents 2/3+1 signatures for descendants of finality target. But what if e.g. someone would add **additional** invalid signature to the justification. So if validators set is 10 validators, we need 7 correct signatures to prove finality of header. What if there's eighth signature in justification and this signature is invalid? Right now both original (Substrate + finality-grandpa) and our implementations would reject such justification. But even though this justification contains invalid data, it still proves finality (it has 7 valid signatures out of 10). If this will ever change in original implementation (it'll ignore invalid vote and accept such justification), the node would accept such justification and the relayer that is connected to this node will try to submit such justification to the runtime. This transaction will fail (becaue) and it'll keep retrying forever.

This ^^^ is bad, so I've decided to go with same implementations for now.

TLDR: having different implementations of `finality_grandpa::Chain` is not the only thing that may lead to the failure + it is unlikely to diverge imo.

adoerr (approved):
Code as such, LGTM.

AFAIU, the GRANDPA implementation is Substrate has to be considered being the baseline. So making sure, observable behavior matches the Substrate implementation seems reasonable.  It is not the ideal solution, but the best we can do, right now.

PR #982, opened 2021-05-28
Flag for rejecting all outbound messages

### LINKED ISSUE #946: Add a way to prevent new messages from being queued to message lane
The idea for this might be soft-disabling the bridge. I.e. we allow existing messages to be processed, but we disallow adding new messages. This can be useful in case we want to temporarily disable or migrate the bridge to new pallets.

I guess it can be implemented in many different ways (flag, mode-enum, limit, per-lane limits, etc) so I'm not enforcing any specific one - it's up to the implementer to choose what's most reasonable.

svyatonik:
closes #946 (not actually, because flag is runtime-specific and we'll need it for P<>K bridge; but we'll be mostly copying code from our testnets => closing now)

I could also extract it to the `bp-runtime-common` && introduce a set of 'common' flags. This would require either a way to merge two enums (something like `decl_outer_parameter!`) or nested enums - `enum CommonParameters { RejectOutboundMessages(bool), RuntimeSpecificParameters { enum RialtoSpecificParameter { MillauToRialtoConversionRate(FixedU128) } } }`.


HCastano (approved):
> I could also extract it to the bp-runtime-common && introduce a set of 'common' flags. 

This could work, but I would wait until we implement this into some Polkadot based runtimes before splitting it out

bin/millau/runtime/src/rialto_messages.rs
```diff
@@ -40,6 +40,8 @@ pub const INITIAL_RIALTO_TO_MILLAU_CONVERSION_RATE: FixedU128 = FixedU128::from_
 parameter_types! {
 	/// Rialto to Millau conversion rate. Initially we treat both tokens as equal.
 	pub storage RialtoToMillauConversionRate: FixedU128 = INITIAL_RIALTO_TO_MILLAU_CONVERSION_RATE;
+	/// If `true`, all Rialto -> Millau messages are rejected.
+	pub storage RejectOutboundMessages: bool = false;
```
tomusdrw, line 44: Why not add this parameter directly to the pallet? This is not a bridge-specific parameter and as you mentioned it will need to be copy-pasted for every bridge instance.

We already have `IsHalted` in the pallet now. I think the two could either be merged in one `Config` object with two bool values or we could introduce `enum OperatingMode { Full, Halted, Inbound }`

The reason is that while the current solution reduces pallet-internal complexity it adds extra integration-level complexity and I feel that's already quite high.

svyatonik: I'm considering this as 'our way' to ease upgrade process. Other may choose their own paths to solve that. Or, if they have non-upgradeable chains (like Rialto and Millau), they may just not use parameters like that. I would avoid adding any functionality that may need to be customized to the pallet itself - the pallet itself is not trivial && having potentially unused stuff there is undesirable.

svyatonik: Actually - what's more important is that additional storage read. So since I don't want to make messages more expensive, I'll implement your option then

modules/messages/src/lib.rs
```diff
@@ -196,8 +196,8 @@ decl_storage! {
 		/// runtime methods may still be used to do that (i.e. democracy::referendum to update halt
 		/// flag directly or call the `halt_operations`).
 		pub PalletOwner get(fn module_owner): Option<T::AccountId>;
-		/// If true, all pallet transactions are failed immediately.
-		pub IsHalted get(fn is_halted) config(): bool;
+		/// Pallet operating mode.
```
HCastano, line 199:
```suggestion
		/// The current operating mode of the pallet.
		/// 
		/// Depending on the mode either all, some, or no transactions will be allowed.
```

modules/messages/src/lib.rs
```diff
@@ -695,9 +695,18 @@ fn ensure_owner_or_root<T: Config<I>, I: Instance>(origin: T::Origin) -> Result<
...
-fn ensure_operational<T: Config<I>, I: Instance>() -> Result<(), Error<T, I>> {
-	if IsHalted::<I>::get() {
+/// Ensure that the pallet is in normal operational mode.
+fn ensure_normal_operating_mode<T: Config<I>, I: Instance>() -> Result<(), Error<T, I>> {
+	if PalletOperatingMode::<I>::get() != OperatingMode::Normal {
+		Err(Error::<T, I>::Halted)
+	} else {
+		Ok(())
+	}
+}
```
HCastano, line 705:
```suggestion
	if PalletOperatingMode::<I>::get() == OperatingMode::Normal {
		Ok(())
	} else {
		Err(Error::<T, I>::Halted)
	}
}
```

nit: But it matches the check that the function says it does

HCastano (approved):

tomusdrw (approved):

PR #984, opened 2021-06-03
Add script to setup bench machine.

tomusdrw:
There is a bit of repeated work after I order a fresh bechmarking instance, made a simple script to automate that.

### DIFF
--- a/scripts/update-weights-setup.sh
+++ b/scripts/update-weights-setup.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+set -exu
+
+# Set up the standardized machine and run `update-weights.sh` script.
+# The system is assumed to be pristine Ubuntu 20.04 and we install
+# all required dependencies.
+
+# To avoid interruptions you might want to run this script in `screen` cause it will take a while
+# to finish.
+
+# We start off with upgrading the system
+apt update && apt dist-upgrade
+
+# and installing `git` and other required deps.
+apt install -y git clang curl libssl-dev llvm libudev-dev screen
+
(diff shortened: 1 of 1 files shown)


scripts/update-weights-setup.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+set -exu
+
+# Set up the standardized machine and run `update-weights.sh` script.
+# The system is assumed to be pristine Ubuntu 20.04 and we install
+# all required dependencies.
```
HCastano, line 7: We should probably note down the expected machine specs somewhere, but not here since this script doesn't care as long as it's Ubuntu

scripts/update-weights-setup.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# To avoid interruptions you might want to run this script in `screen` cause it will take a while
```
HCastano, line 9:
```suggestion
# To avoid interruptions you might want to run this script in `tmux` cause it will take a while
```

FTFY 😉

HCastano (approved):

scripts/update-weights-setup.sh
tomusdrw: We have it on our internal notion page (Deployments). I wasn't sure if we should advertise a specific provider :)

HCastano: We can list specs without listing a provider (not that it matters imo). It would be useful for people who need a reference for their own Substrate based chains.

Just a thought though, this can be merged without that.

PR #987, opened 2021-06-07
Make sure GRANDPA shares state with RPC.

ruseinov:
The current setup does not work, because the state is always empty when queried through RPC.

### DIFF
--- a/bin/millau/node/src/service.rs
+++ b/bin/millau/node/src/service.rs
@@ -349,7 +349,7 @@ pub fn new_full(mut config: Configuration) -> Result<TaskManager, ServiceError> 
 			network,
 			voting_rule: sc_finality_grandpa::VotingRulesBuilder::default().build(),
 			prometheus_registry,
-			shared_voter_state: shared_voter_state.clone(),
+			shared_voter_state,
 			telemetry: telemetry.as_ref().map(|x| x.handle()),
 		};
 
--- a/bin/millau/node/src/service.rs
+++ b/bin/millau/node/src/service.rs
@@ -33,7 +33,7 @@ use sc_client_api::{ExecutorProvider, RemoteBackend};
 use sc_consensus_aura::{ImportQueueParams, SlotProportion, StartAuraParams};
 use sc_executor::native_executor_instance;
 pub use sc_executor::NativeExecutor;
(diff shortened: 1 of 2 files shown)


bin/millau/node/src/service.rs
```diff
@@ -230,7 +231,7 @@ pub fn new_full(mut config: Configuration) -> Result<TaskManager, ServiceError>
 
 		let justification_stream = grandpa_link.justification_stream();
 		let shared_authority_set = grandpa_link.shared_authority_set().clone();
-		let shared_voter_state = sc_finality_grandpa::SharedVoterState::empty();
+		let shared_voter_state = shared_voter_state.clone();
```
tomusdrw, line 234: I guess we need the same fix for `rialto` node?

tomusdrw (commented):
Thanks!

bin/millau/node/src/service.rs
ruseinov: Yeah, will do

ruseinov: done

tomusdrw:
@joao-paulo-parity seems that the CLA bot is missing, could you take a look?

HCastano (commented):
~~The "Update Guide" section would also need to be updated to reflect this change, since it deviates from what the `node-template` provides.~~

Nevermind, the changes are within the `rpc_extensions_builder`.

I also haven't found any indication in the relayer or node logs that anything was off w.r.t the finality RPCs, can I get a second look from you @svyatonik?

Even though nothing terrible happened by not having this, we should probably have the relayer emit a `WARN` log if the Grandpa state is empty

ruseinov:
Done!

ruseinov:
> The "Update Guide" section would also need to be updated to reflect this change, since it deviates from what the `node-template` provides.

node-template does not provide the RPC for GRANDPA at all AFAIK, or it didn't a couple of months ago.

tomusdrw:
> I also haven't found any indication in the relayer or node logs that anything was off w.r.t the finality RPCs,

@HCastano it's `shared_voter_state`, not finality notifications, so afaiu it was only affecting vote-related RPCs, not the ones our relayers are using.

svyatonik (approved):

PR #989, opened 2021-06-08
Increase pause before starting relays (again)

svyatonik:
See #974 for details - the same has happened today + it isn't reproduced when started locally + it has happened exactly after Substrate reference has been updated. On test deployments it seems that the 'correct' sleep value is somewhere between 12-14s now.


adoerr (approved):

HCastano (approved):
If it was the bump in Substrate, it's hard to tell what's caused this sluggishness.

It would be nice to find out though, I don't think we can just keep bumping the delay indefinitely (assuming we don't improve the deployment hardware)

PR #990, opened 2021-06-08
Move ChainWithBalances::NativeBalance -> Chain::Balance

svyatonik:
Required for greedy relayer.
This means that the relayer now assumes that all chains have some base token, which is used to: (1) pay transactions fee (2) pay message delivery and dispatch fee and (3) reward relayers. If we'll ever met case where this isn't true (e.g. when relayer is rewarded with weight for its future transactions), then this would need to be changed. But now it is easier to continue with this assumption.

### DIFF
--- a/relays/client-millau/src/lib.rs
+++ b/relays/client-millau/src/lib.rs
@@ -44,11 +44,10 @@ impl Chain for Millau {
 	type Index = millau_runtime::Index;
 	type SignedBlock = millau_runtime::SignedBlock;
 	type Call = millau_runtime::Call;
+	type Balance = millau_runtime::Balance;
 }
 
 impl ChainWithBalances for Millau {
-	type NativeBalance = millau_runtime::Balance;
-
 	fn account_info_storage_key(account_id: &Self::AccountId) -> StorageKey {
 		use frame_support::storage::generator::StorageMap;
 		StorageKey(frame_system::Account::<millau_runtime::Runtime>::storage_map_final_key(
--- a/relays/client-substrate/src/chain.rs
+++ b/relays/client-substrate/src/chain.rs
@@ -54,14 +54,16 @@ pub trait Chain: ChainBase + Clone {
 	type SignedBlock: Member + Serialize + DeserializeOwned + BlockWithJustification<Self::Header>;
 	/// The aggregated `Call` type.
 	type Call: Dispatchable + Debug;
+	/// Balance of an account in native tokens.
+	///
+	/// The chain may suport multiple tokens, but this particular type is for token that is used
+	/// to pay for transaction dispatch, to reward different relayers (headers, messages), etc.
+	type Balance: Parameter + Member + DeserializeOwned + Clone + Copy + CheckedSub + PartialOrd + Zero;
 }
 
 /// Substrate-based chain with `frame_system::Config::AccountData` set to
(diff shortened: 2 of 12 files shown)


relays/client-millau/src/lib.rs
```diff
@@ -44,11 +44,10 @@ impl Chain for Millau {
 	type Index = millau_runtime::Index;
 	type SignedBlock = millau_runtime::SignedBlock;
 	type Call = millau_runtime::Call;
+	type Balance = millau_runtime::Balance;
 }
 
 impl ChainWithBalances for Millau {
```
tomusdrw, line 50: Can we eradicate this trait completely now and move the `account_info_storage_key` function to `Chain` as well?

relays/client-substrate/src/chain.rs
```diff
@@ -54,14 +54,16 @@ pub trait Chain: ChainBase + Clone {
 	type SignedBlock: Member + Serialize + DeserializeOwned + BlockWithJustification<Self::Header>;
 	/// The aggregated `Call` type.
 	type Call: Dispatchable + Debug;
+	/// Balance of an account in native tokens.
+	///
+	/// The chain may suport multiple tokens, but this particular type is for token that is used
+	/// to pay for transaction dispatch, to reward different relayers (headers, messages), etc.
+	type Balance: Parameter + Member + DeserializeOwned + Clone + Copy + CheckedSub + PartialOrd + Zero;
```
tomusdrw, line 61: Does it have to be the native token? Would it work if there are multiple instances of `balances` pallet?

relays/client-millau/src/lib.rs
svyatonik: This trait 'protects' us from chains that aren't storing `pallet_balances::AccountData` directly in the storage. I could imagine that there may be additional info stored for every account - account storage root, ref count, ... And `AccountData` is either stored as a part of this data, or is ephemeral. By 'protect' here I mean that we're able to work with such chains, but some functionality (right now it is only relay guards) is unavailable. Moving `account_info_storage_key` would mean that we are not supporting such chains anymore. At least we can't 'detect' supported chains at compilation time.

That said, if you feel that this trait may be removed - please do that in a separate PR. I feel that the current PR itself introduces a quite important change, which deserves its own discussion.

relays/client-substrate/src/chain.rs
svyatonik: Honestly - I'm not sure what's the difference between `TOKEN1` and `TOKEN2` (different instances of the balance/assets pallet). Which token is native then? :) The type I'm talking here is the type that is used to pay transaction fees, relayer rewards. Depending on the runtime implementation, this may be any instance of the balance pallet. But what this PR does, is that it brings knowledge that all payments that he may see during its work will be conducted in the token that is defined here.

svyatonik: So why it was not required before - because relayer was just ignoring fees and rewards. Greedy relayer would need to know that => that's why I've moved that in a base trait.

HCastano: So maybe we shouldn't be calling them "native tokens", and instead call them "payment tokens".

It's totally reasonable for a runtime to have many instances for the balance pallet, and perhaps accept any of those tokens as a way to pay fees.

HCastano (approved):

PR #995, opened 2021-06-09
Messages relayer operating mode

svyatonik:
This is the core PR of 'greedy relayer' stuff. It introduces new `enum RelayerMode {}`, which has three options now:
1) `Altruistic` - the current behavior where relayer don't care about delivery/dispatch costs. It is the only available mode now. All others will be supported through cli option in follow-up PRs
2) `NoLosses` - the relayer will keep delivering messages as long as the reward is larger than delivery+confirmation transactions cost. This will be default mode in cli;
3) `MaximalReward` - basically `NoLosses`, with single exception. The exception is if there are `M1(reward=10, cost=5), M2(reward=3, cost=7), M3(reward=2, cost=1)`, then `NoLosses` would select to deliver all messages (total reward=10+3+2=15 total cost=5+7+1=13), but `MaximalReward` would only deliver `M1`, because total reward for delivering only `M1` (it is 10-5=5) is larger than for delivering either `M1, M2` (10+3-5-7=1), or `M1, M2, M3` (15-13=2). OTOH once `M1` is delivered, it will immediately submit `M2, M3`, because their reward is larger than their cost. So it may be useful if relayer wants to win the delivery race, but unuseful long term (and if there's no any race). So I'm leaving it in initial version just to share my thoughts - will remove if you agree it is unuseful.

There are also some changes related to making `select_nonces_to_deliver` + extracting some code from `select_nonces_to_deliver` to separate `select_nonces_for_delivery_transaction`. Follow up PRs would: (1) change existing RPCs to support reading message fees and (2) add cli support for selecting relayer mode


relays/bin-substrate/src/messages_source.rs
```diff
@@ -242,6 +243,10 @@ where
 			target_to_source_headers_relay.require_finalized_header(id).await;
 		}
 	}
+
+	async fn estimate_confirmation_transaction(&self) -> P::SourceChainBalance {
+		num_traits::Zero::zero() // TODO: will be introduced in follow-up PR(s)
```
HCastano, line 248: Can you make an issue for this?

relays/bin-substrate/src/messages_source.rs
```diff
@@ -341,9 +346,10 @@ fn make_message_weights_map<C: Chain>(
 
 		weights_map.insert(
 			nonce,
-			MessageWeights {
-				weight,
+			MessageDetails {
+				dispatch_weight: weight,
 				size: size as _,
+				reward: num_traits::Zero::zero(), // TODO: will be introduced in follow-up PR(s)
```
HCastano, line 352: Same here

relays/client-substrate/src/chain.rs
```diff
@@ -58,7 +58,7 @@ pub trait Chain: ChainBase + Clone {
 	///
 	/// The chain may suport multiple tokens, but this particular type is for token that is used
 	/// to pay for transaction dispatch, to reward different relayers (headers, messages), etc.
-	type Balance: Parameter + Member + DeserializeOwned + Clone + Copy + CheckedSub + PartialOrd + Zero;
+	type Balance: Parameter + Member + DeserializeOwned + Clone + Copy + CheckedSub + PartialOrd + Saturating + Zero;
```
HCastano, line 61: According to [the docs](https://docs.rs/num-traits/0.2.14/num_traits/ops/saturating/trait.Saturating.html) the `Saturating` trait is deprecated. We should use the others (`SaturatingAdd`, `SaturatingSub`, etc.) as suggested

relays/messages/src/message_lane_loop.rs
```diff
@@ -74,20 +87,24 @@ pub struct MessageDeliveryParams {
...
+/// Message details.
 #[derive(Debug, Clone, Copy, PartialEq)]
-pub struct MessageWeights {
+pub struct MessageDetails<SourceChainBalance> {
 	/// Message dispatch weight.
-	pub weight: Weight,
+	pub dispatch_weight: Weight,
 	/// Message size (number of bytes in encoded payload).
-	pub size: usize,
+	pub size: u32,
```
HCastano, line 101: Why was size changed from a `usize`?

relays/messages/src/message_race_delivery.rs
```diff
@@ -222,7 +222,11 @@ struct DeliveryRaceTargetNoncesData {
 }
 
 /// Messages delivery strategy.
-struct MessageDeliveryStrategy<P: MessageLane> {
+struct MessageDeliveryStrategy<P: MessageLane, SC, TC> {
+	/// TODO
```
HCastano, line 226: Leftover TODOs

relays/messages/src/message_race_delivery.rs
```diff
@@ -473,87 +495,215 @@ impl<P: MessageLane> RaceStrategy<SourceHeaderIdOf<P>, TargetHeaderIdOf<P>, P::M
...
+			new_selected_weight if selected_count == 0 => {
+				log::warn!(
+					target: "bridge",
+					"Going to submit message delivery transaction with declared dispatch \
+					weight {:?} that overflows maximal configured weight {}",
+					new_selected_weight,
+					max_messages_weight_in_single_batch,
+				);
+				new_selected_weight.unwrap_or(Weight::MAX)
+			}
```
HCastano, line 587: I'm confused, why would you want to send a message in this case? Wouldn't it be better to print a warning and not send it?

relays/messages/src/message_race_delivery.rs
```diff
@@ -473,87 +495,215 @@ impl<P: MessageLane> RaceStrategy<SourceHeaderIdOf<P>, TargetHeaderIdOf<P>, P::M
...
+			_ => break,
```
HCastano, line 588: Should we print a warning here about the weight overflow?

relays/messages/src/message_race_delivery.rs
```diff
@@ -473,87 +495,215 @@ impl<P: MessageLane> RaceStrategy<SourceHeaderIdOf<P>, TargetHeaderIdOf<P>, P::M
...
+
+		// limit number of messages in the batch
+		let new_selected_count = selected_count + 1;
+		if new_selected_count > max_messages_in_this_batch {
+			break;
+		}
+
+		// now the message has passed all 'strong' checks, and we CAN deliver it. But do we WANT
+		// to deliver it? It depends on the relayer strategy.
+		match relayer_mode {
```
HCastano, line 615: Can you extract all this code into a separate function?

HCastano (commented):
I need to take another look at this PR tomorrow.

As far as the `MaximalReward` option goes, I'm not convinced it that useful. It doesn't seem to me that it offers better rewards than`NoLosses`. However, I think it would be interesting to write a little simulation which shows the profit/loss for sending messages with the different strategies, maybe it would be better in the long run.

Also

> OTOH once M1 is delivered, it will immediately submit M2, M3, because their reward is larger than their cost. 

Here you mean it would only submit `M3`, right? Submitting `M2` would result in a loss

svyatonik:
> Here you mean it would only submit M3, right? Submitting M2 would result in a loss

Relayer can only submit message if all previous messages are also delivered. So relaying only `M3`, without `M2` is impossible. But you're right - actually relayer won't relay `M2+M3`, because their total cost is larger than total reward.

relays/messages/src/message_race_delivery.rs
svyatonik: No, it's a regular situation - there are simply too many undelivered messages

relays/messages/src/message_race_delivery.rs
svyatonik: This a moved (not new) code. There's a comment about that above:
```rust
		// Since we (hopefully) have some reserves in `max_messages_weight_in_single_batch`
		// and `max_messages_size_in_single_batch`, we may still try to submit transaction
		// with single message if message overflows these limits. The worst case would be if
		// transaction will be rejected by the target runtime, but at least we have tried.
```

relays/messages/src/message_lane_loop.rs
svyatonik: It's returned as `u32` from RPC and then is used as `u32` in a weight formula.

relays/messages/src/message_race_delivery.rs
```diff
@@ -881,21 +1049,21 @@ mod tests {
...
-			strategy.select_nonces_to_deliver(&state),
+			strategy.select_nonces_to_deliver(state).await,
 			Some(((20..=23), proof_parameters(true, 4)))
 		);
 	}
 
-	#[test]
-	fn source_header_is_requied_when_confirmations_are_required() {
+	#[async_std::test]
+	async fn source_header_is_requied_when_confirmations_are_required() {
```
HCastano, line 1058:
```suggestion
	async fn source_header_is_required_when_confirmations_are_required() {
```

relays/messages/src/message_race_delivery.rs
```diff
@@ -927,4 +1095,36 @@ mod tests {
 			Some(header_id(2))
 		);
 	}
+
+	#[async_std::test]
+	async fn no_losses_relayer_is_delivering_messages_if_cost_is_equal_to_reward() {
```
HCastano, line 1100: We could probably use a few more tests for the `NoLosses` strategy imo

HCastano (approved):

PR #998, opened 2021-06-10
Rococo <> Wococo relayer balance guard

svyatonik:
Apart from introducing balance guards, this PR also removes some duplicate stuff from wococo primitives, instead reexporting it from the rococo primitives. Could revert it, but since they're literally equivalent chains, imo it is better than duplicating `WeightToFee`, `VERSION`, ...


relays/bin-substrate/src/chains/wococo_headers_to_rococo.rs
```diff
@@ -65,3 +76,41 @@ impl SubstrateFinalitySyncPipeline for WococoFinalityToRococo {
...
+		// Worst case: we're submitting proof for every source header. Since we submit every header, the number of
+		// headers in ancestry proof is near to 0 (let's round up to 2). And the number of authorities is 1024,
+		// which is (now) larger than on any existing chain => normally there'll be ~1024*2/3+1 commits.
+		const AVG_VOTES_ANCESTRIES_LEN: u32 = 2;
+		const AVG_PRECOMMITS_LEN: u32 = 1024 * 2 / 3 + 1;
+		let number_of_source_headers_per_day: bp_wococo::Balance = bp_wococo::DAYS as _;
+		let single_source_header_submit_call_weight =
+			RococoGrandpaPalletWeights::submit_finality_proof(AVG_VOTES_ANCESTRIES_LEN, AVG_PRECOMMITS_LEN);
+		// for simplicity - add extra weight for base tx fee + fee that is paid for the tx size + adjusted fee
+		let single_source_header_submit_tx_weight = single_source_header_submit_call_weight * 3 / 2;
```
svyatonik, line 106: Without this `*1.5`, the cost is ~740 KSM per day. With it it is > 1000 KSM. So the final const value is 1500 KSM / day (left some reserve there).

primitives/chain-rococo/src/lib.rs
```diff
@@ -43,6 +44,23 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	transaction_version: 0,
 };
 
+// NOTE: This needs to be kept up to date with the Rococo runtime found in the Polkadot repo.
```
HCastano, line 47: Another thing that has to be kept up to date with the Polkadot repo 🙃

relays/bin-substrate/src/chains/wococo_headers_to_rococo.rs
HCastano: 750 tokens/day is unreasonably high. We can't deploy the bridge with these kinds of costs imo.

I know this is a worst case calculation, but it's still incredibly expensive

relays/bin-substrate/src/chains/wococo_headers_to_rococo.rs
```diff
@@ -65,3 +76,41 @@ impl SubstrateFinalitySyncPipeline for WococoFinalityToRococo {
...
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+	use frame_support::weights::WeightToFeePolynomial;
+	use pallet_bridge_grandpa::weights::WeightInfo;
+
+	#[test]
+	fn maximal_balance_decrease_per_day_is_sane() {
+		// Rococo/Wooco GRANDPA pallet weights. They're now using Rialto weights => using `RialtoWeight` is justified.
```
HCastano, line 88:
```suggestion
		// Rococo/Wococo GRANDPA pallet weights. They're now using Rialto weights => using `RialtoWeight` is justified.
```

relays/bin-substrate/src/chains/wococo_headers_to_rococo.rs
svyatonik: Isn't that what you're currently deploying on W<>R? At least it was my impression :) And this PR is about W<>R. On Kusama we'll be hopefully deploying complex relay.

HCastano: Yeah, you're right it's not that important on W<>R, but the goal would be to eventually have it on P<>K where it would be unacceptable at this cost.

> On Kusama we'll be hopefully deploying complex relay.

I think the high costs still stand. If anything, depending on how much messages cost to send it could be even worse (assuming that every finalized header contains a new message to send).

We'll really need to watch the costs with our R<>W deployment and optimize costs accordingly

relays/bin-substrate/src/chains/wococo_headers_to_rococo.rs
```diff
@@ -21,11 +21,17 @@ use crate::finality_pipeline::{SubstrateFinalitySyncPipeline, SubstrateFinalityT
...
+use relay_substrate_client::{Chain, TransactionSignScheme};
 use relay_utils::metrics::MetricsParams;
 use relay_wococo_client::{SyncHeader as WococoSyncHeader, Wococo};
 use sp_core::{Bytes, Pair};
 
+/// Maximal saturating difference between `balance(now)` and `balance(now-24h)` to treat
+/// relay as gone wild.
+///
+/// See `maximal_balance_decrease_per_day_is_sane` test for details.
+pub(crate) const MAXIMAL_BALANCE_DECREASE_PER_DAY: bp_rococo::Balance = 1_500_000_000_000_000;
```
HCastano, line 33:
```suggestion
/// See `maximal_balance_decrease_per_day_is_sane` test for details.
/// Note that this is in base units, so this corresponds to `1500 UNITS`
pub(crate) const MAXIMAL_BALANCE_DECREASE_PER_DAY: bp_rococo::Balance = 1_500_000_000_000_000;
```

HCastano (approved):

tomusdrw (approved):

PR #999, opened 2021-06-10
Bump Rococo and Wococo spec versions

HCastano:
Right now the spec version our relayers are using it behind what's deployed on the live
networks.


svyatonik (approved):

tomusdrw (approved):

tomusdrw:
IS it used on-chain as well? Could it be controlled by the pallet owner to make it easier to upgrade without upgrading the entire runtime?

HCastano:
> IS it used on-chain as well? Could it be controlled by the pallet owner to make it easier to upgrade without upgrading the entire runtime?

Can you elaborate on what you mean here?

tomusdrw:
> Can you elaborate on what you mean here?

What I mean is that it's pain in the ass to update just one numeric value and trigger the entire build time either for upgrading the runtime (if it's used on-chain) or just to build a new relayer binary.

I think we should have a way to override this setting:
- in on-chain code (if there is any) via owner-originated transaction
- in relayer code via CLI flag

PR #1010, opened 2021-06-18
Use plain source_queue view when selecting nonces for delivery

### LINKED ISSUE #1008: Messages relayer creates too large delivery transactions sometimes
Bug was introduced by #995 && happened today:
1) at source block `100` relayer sees message with nonce `500`. Message has large weight or size;
2) at source block `101` relayer sees message with nonce `501`. Message has large weight or size;
3) relayer sees that source block `101` is finalized at the target chain;
4) relayer selects messages `100` and `101` for its delivery transaction;
5) the delivery transaction is rejected by the node (exhausts block limits) and relayer keep trying it over and over (with timeouts) until it consider itself "stalled" and restarts;
6) after restarts relayer sees both messages `500` and `501` at source block `200` and only selects `500` for delivery => we are not getting alerts.

There are a few ways how to fix it => I'm just logging the issue now instead of opening PR.

svyatonik:
closes #1008

### DIFF
--- a/relays/messages/src/message_race_strategy.rs
+++ b/relays/messages/src/message_race_strategy.rs
@@ -473,10 +473,7 @@ mod tests {
 		fn source_queue_nonces(
 			source_queue: &SourceRangesQueue<TestSourceHeaderHash, TestSourceHeaderNumber, SourceNoncesRange>,
 		) -> Vec<MessageNonce> {
-			source_queue
-				.iter()
-				.flat_map(|(_, range)| range.clone())
-				.collect()
+			source_queue.iter().flat_map(|(_, range)| range.clone()).collect()
 		}
 
 		strategy.remove_le_nonces_from_source_queue(1);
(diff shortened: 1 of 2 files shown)


relays/messages/src/message_race_strategy.rs
```diff
@@ -107,47 +107,26 @@ where
...
-						nonces_end = Some(range_to_requeue.begin() - 1);
-					}
-					self.source_queue.push_front((queued_at, range_to_requeue));
-					break;
-				}
-				None => {
-					nonces_end = Some(queued_range_end);
-				}
+			if let Some(range_to_requeue) = queued_range.greater_than(nonce) {
+				self.source_queue.push_front((queued_at, range_to_requeue));
```
tomusdrw, line 126: Just to confirm - `greater_than` does not return empty ranges, right?

tomusdrw (approved):
lgtm!

relays/messages/src/message_race_strategy.rs
svyatonik: Yes, it returns `None` if there are nonces > given nonce (i.e. if range is empty).

Issue #620, opened 2021-01-05
Allow Root to use CallOrigin::SourceAccount


tomusdrw:
Currently `SourceAccount` call origin can only be used by regular accounts, but `Root`/`Runtime Dev` can use it anyway by crafitng a special origin when calling that function. This is cumbersome to use and we should consider allowing Root to simply use any `SourceOrigin` it wants.

kaichaosun:
I'd like to work on this issue, 
when say `SourceOrigin`, do you mean any account, pallet origin from the source chain?
If yes, I'll probably add a `SourceChainAccountId` to `SourceRoot`, hope I'm in the right direction.

tomusdrw:
@kaichaosun thanks for looking into this. Adding account id to `SourceRoot` is one of the options indeed, I was thinking more about changing this verification [here](https://github.com/paritytech/parity-bridges-common/blob/fc5cf7dd9e8c0ecf8c11956b8851e44a0c1af639/modules/dispatch/src/lib.rs#L355) to allow `Root` as well.

@svyatonik which variant do you prefer? I'd leave `SourceRoot` intact, since even if we add account id there the root account can do whatever it wants via `sudo_as`, hence I'm not sure if it gives us much on the target chain side to distinguish between `SourceAccount(account)` and `SourceRoot(account)` and it does complicate the handling a bit.

kaichaosun:
Thanks for pointing out both options, @tomusdrw .
It's much clean by changing the verification logic at the source chain side. 
But at target chain side, it loses information whether `SourceAccount(account)` is from a real account or assigned by the root.
Are there scenarios that need to distinguish them? Another option could cover this requirement is adding an extra field `forced_account` to `SourceAccount(account)`, this may be overkill at current stage.

PR #1011, opened 2021-06-18
Allow Root to assume SourceAccount.

kaichaosun:
Fixed https://github.com/paritytech/parity-bridges-common/issues/620

Another thing I notice is, [SourceRoot](https://github.com/paritytech/parity-bridges-common/blob/bf52fff13fbb1d3fc6a868d209cddddf2b8f79ef/relays/bin-substrate/src/cli/send_message.rs#L99) is not available in send_message CLI, is there an issue tracking this feature?

### DIFF
--- a/modules/dispatch/src/lib.rs
+++ b/modules/dispatch/src/lib.rs
@@ -308,7 +308,8 @@ where
 		}
 		CallOrigin::SourceAccount(ref source_account_id) => {
 			ensure!(
-				sender_origin == &RawOrigin::Signed(source_account_id.clone()),
+				sender_origin == &RawOrigin::Signed(source_account_id.clone()) ||
+					sender_origin == &RawOrigin::Root,
 				BadOrigin
 			);
 			Ok(Some(source_account_id.clone()))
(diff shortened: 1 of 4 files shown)


svyatonik:
If you'll pass `--assume-account=<some-account>` using CLI, it won't pass the check you've altered in this PR, right? Because you may only pass this check (`verify_message_origin`) if the transaction signer is the same `source_account_id` from the `CallOrigin::SourceAccount()` OR if you're the root (that's what you've changed in this PR). But our CLI is not sending messages using root origin (as you've pointe), so this option seems unneeded now. Let's probably only leave your first commit (and `cargo fmt`)?

kaichaosun:
I see, need to enable sudo call in CLI to support root origin call.

After some testing from polkadotjs/apps, the sudo call failed to pass the `pay_delivery_and_dispatch_fee` since [Root origin is not allowed to pay fee](https://github.com/paritytech/parity-bridges-common/blob/604eb1cd0941ba1e17d083fd198df1f19955abf8/modules/messages/src/instant_payments.rs#L77).
Does it make sense if I add an extra param `payer` for `send_message` dispatable, as Root origin from democracy/council is not supposed to hold assets from what I understand?

svyatonik:
> Does it make sense if I add an extra param payer for send_message dispatable, as Root origin from democracy/council is not supposed to hold assets from what I understand?

Let's leave it for other PRs - there's a way now to configure that now (by changing that `RootAccount`) + there's #962 which will separate call origin from fee payer. Hopefully we'll be able to solve it in some general way

svyatonik (approved):

tomusdrw (approved):
Nice, thanks! I was suspecting this is going to be a bigger change, but the smaller the better 😅

Issue #997, opened 2021-06-10
Support 'greedy' mode in relayer


svyatonik:
Since there are multiple related PRs, I need this issue to use in TODOs. There are/will be following PRs:
- [x] #990
- [x] #995
- [x] #996
- [ ] Estimate transaction costs
- [ ] CLI options

PR #1015, opened 2021-06-22
Estimate transaction fee RPC call + estimate message delivery and confirmation transactions

svyatonik:
related #997


relays/client-substrate/src/chain.rs
```diff
@@ -58,7 +63,20 @@ pub trait Chain: ChainBase + Clone {
...
+		+ Member
+		+ DeserializeOwned
+		+ Clone
+		+ Copy
+		+ Bounded
+		+ CheckedSub
+		+ PartialOrd
+		+ SaturatingAdd
+		+ Zero
+		+ std::convert::TryFrom<sp_core::U256>;
```
svyatonik, line 79: This `TryFrom` is required, because `payment_queryFeeDetails` returns `InclusionFee<NumberOrHex>` and `NumberOrHex` holds `u64`/`U256`. Not sure what happens on chain with e.g. `U512`

relays/bin-substrate/src/messages_source.rs
```diff
@@ -245,10 +252,40 @@ where
 	}
 
 	async fn estimate_confirmation_transaction(&self) -> P::SourceChainBalance {
-		num_traits::Zero::zero() // TODO: https://github.com/paritytech/parity-bridges-common/issues/997
+		self.client
```
svyatonik, line 255: Actually, this may be called from `fn new()` and then just returned here, but it'll add additional assumptions - like that call weights may not be updated while relayer is running.

relays/bin-substrate/src/chains/rialto_messages_to_millau.rs
```diff
@@ -128,13 +128,15 @@ impl SubstrateMessageLane for RialtoMessagesToMillau {
 /// Rialto node as messages source.
 type RialtoSourceClient = SubstrateMessagesSource<
 	Rialto,
+	Millau,
 	RialtoMessagesToMillau,
```
tomusdrw, line 132: Couldn't the lane contain all other necessary generic types? It seems it's bound with them anyway?

relays/client-substrate/src/chain.rs
tomusdrw: I think in other parts of substrate we already require that the balance is at most `128bits`.

svyatonik: I haven't seen that, but it is good to know

relays/bin-substrate/src/chains/rialto_messages_to_millau.rs
svyatonik: This requires some refactoring - simply using associated types from lane causes `cycle detected when using associated type bounds`. But ok - I'll try to solve it here.

tomusdrw: No worries, let's log this and leave for a potential follow up PR.

tomusdrw (approved):

relays/bin-substrate/src/chains/rialto_messages_to_millau.rs
svyatonik: I've tried some attempts, but all are adding a lot of changes to this PR, so right - let's keep it for other PRs: #1033

PR #1016, opened 2021-06-22
Greedy relayer don't need message dispatch to be prepaid if it is paid at the target chain

svyatonik:
related #997


tomusdrw (approved):
looks sensible

Issue #885, opened 2021-04-12
CLI: Move `--target-sign` to Origins.


tomusdrw:
Currently for some CLI variants (`send-message` and `encode-message`), the `--target-sign` flag is required despite it being useful only in case `--origin Target` is selected as well.
We should move the flag to be part `origin` enum instead and make it explicit or at least mark `target-sign` as not required in case `Source` origin is used.

PR #1018, opened 2021-06-22
Make target signer optional when sending message.

kaichaosun:
Fix https://github.com/paritytech/parity-bridges-common/issues/885

I choose this option after finding that,

* `arg_enum` does not support struct variant,
* `structopt` field flagged with `flatten` is not able to use `require_if`
* `TargetSigningParams` is reused by a few other places, extract it from `declare_chain_options` macro seems too much change.

Also `encode-message` list in issue is not using `target_signer`, please correct me if I'm wrong.

### DIFF
--- a/relays/bin-substrate/src/cli/send_message.rs
+++ b/relays/bin-substrate/src/cli/send_message.rs
@@ -264,8 +264,6 @@ mod tests {
 			"1234",
 			"--source-signer",
 			"//Alice",
-			"--target-signer",
-			"//Bob",
 			"remark",
 			"--remark-payload",
 			"1234",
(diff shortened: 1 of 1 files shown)


svyatonik (approved):
If I'm not missing something, you may just use something like that to avoid using `required_if`:
```rust
...
struct SendMessage {
...
	#[structopt(flatten)]
	origin: Origins,
...
}

#[derive(StructOpt, Debug, PartialEq, Eq)]
pub enum Origins {
  Source,
  Target {
	#[structopt(flatten)]
	target_sign: TargetSigningParams,
  }
}
```

This way structopt would reject cli  where origin is not `Target`, but target account/password is specified. Right now you're requiring it when necessary, but not rejecting it when it is not used. That said, it isn't super-important imo

kaichaosun:
I tried the new option in [this branch](https://github.com/paritytech/parity-bridges-common/compare/optional-target-sign-2?expand=1), but it's not work as expected.

If use `--origin Source` with `send-message`, 
get error: `Found argument '--origin' which wasn't expected, or isn't valid in this context`.
if use `source`, 
then get error: `Found argument 'remark' which wasn't expected, or isn't valid in this context`, which probably means only 1 subcommand is allowed.

svyatonik:
Ok, thanks! Let's merge as is then :)

Issue #1009, opened 2021-06-17
Deal with InsufficientBalance messages in Rialto<>Millau test deployments


svyatonik:
Errors like that
```
2021-06-17 05:02:00.003  TRACE tokio-runtime-worker runtime::bridge-dispatch: Message [114, 108, 116, 111]/([0, 0, 0, 0], 4) has been dispatched. Weight: 206909000 of 206909000. Result: Err(DispatchErrorWithPostInfo { post_info: PostDispatchInfo { actual_weight: None, pays_fee: Pays::Yes }, error: Module { index: 9, error: 2, message: Some("InsufficientBalance") } })    
```

appear after 2 minutes bridge start. So either accounts that are used for sending are incorrect (i.e. not endowed at all), or the transfer amount is too large.

PR #1020, opened 2021-06-23
Fix insufficient balance when send message.

kaichaosun:
Fix https://github.com/paritytech/parity-bridges-common/issues/1009

* `Dave` and its derived target account is used in deployment scripts [1](https://github.com/paritytech/parity-bridges-common/blob/ad5876f05a80c95f9fcefb62bbf2c9b891a788f8/deployments/bridges/rialto-millau/entrypoints/relay-messages-to-millau-generator-entrypoint.sh#L19) and [2](https://github.com/paritytech/parity-bridges-common/blob/ad5876f05a80c95f9fcefb62bbf2c9b891a788f8/deployments/bridges/rialto-millau/entrypoints/relay-messages-to-rialto-generator-entrypoint.sh#L19). The specific `InsufficientBalance` issue can be resolved by adding `derive_account_from_rialto_id(...dave)` in Millau chainspec. I think fund more derived accounts could do no harm to the testnet, so add a few other derived accounts on both sides.
* ~~change the dev script with lower amount, in case it used in test env~~
* ~~bump up the genesis free balance so that it can run more than a few days. (2^50/1000_000_000/3600/24 = 13)~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ the `relays` which are used to pass messages between chains.
 │  └──  ...
 ├── relays          // Application for sending headers and messages between chains
 │  └──  ...
-└── scripts         // Useful development and maintenence scripts
+└── scripts         // Useful development and maintenance scripts
 ```
 
 ## Running the Bridge
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ the `relays` which are used to pass messages between chains.
 │  └──  ...
 ├── relays          // Application for sending headers and messages between chains
 │  └──  ...
-└── scripts         // Useful development and maintenance scripts
+└── scripts         // Useful development and maintenence scripts
 ```
 
 ## Running the Bridge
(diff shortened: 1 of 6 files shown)


svyatonik:
Thanks. The scripts you've changed are using for manual testing iirc. The scripts that are used in our test-deployments are in *generator* files in [this folder](https://github.com/paritytech/parity-bridges-common/tree/master/deployments/bridges/rialto-millau/entrypoints). There's [document](https://github.com/paritytech/parity-bridges-common/tree/master/deployments) on how to start this locally - you'll be able to get logs locally. I've seen these errors after ~2 minutes running, so you should be able to check almost immediately.

kaichaosun:
I notice the original [discussion](https://github.com/paritytech/parity-bridges-common/pull/951#issuecomment-830854533), is it better if we distinguish the accounts on both sides with the actual balance, like balance of endowed account in Rialto is 2^50, and Millau is 2^51?

svyatonik (approved):

svyatonik:
> I notice the previous [discussion](https://github.com/paritytech/parity-bridges-common/pull/951#issuecomment-830854533) on not all derived accounts are endowed, is it better if we distinguish the accounts on both sides with the actual balance, like balance of endowed account in Rialto is 2^50, and Millau is 2^51?

Let's ask @tomusdrw - I haven't seen how it looks in the UI, so can't judge here.

PR #1027, opened 2021-06-23
Update `InboundLaneData` JS types definition

tomusdrw:
We've added the dispatch result in #935, so changing custom type definitions to match the Rust code now.


hbulgarini (approved):
I have verified in the UI that after this update, the type issues have gone in the console. Thanks!

PR #1029, opened 2021-06-23
Update the wasm builder to substrate master

chevdor:
This PR updates the dependencies on substrate-wasm-builder from 3.0.0 to master. This allows benefiting from automatic generation of compressed wasm in adddition to the existing one.

After doing that change, I noticed that there was a single dependency left on 3.0.0, so I updated it as well have have everything in sync.


chevdor:
I am leaving this PR "as is" for now since the issues reported by the CI don't seem to be related to the changes.

HCastano:
> I am leaving this PR "as is" for now since the issues reported by the CI don't seem to be related to the changes.

I think they are related. Can you bump Substrate (`cargo update -p sp-io`) and try again?

PR #1030, opened 2021-06-24
Rococo <> Wococo messages relay

svyatonik:
This PR has two parts && I can extract second part into a separate PR if it'll be easier to review like that.

Part 1: first 5 commits are here mostly to solve that non-standard situation with R<>W bridge when we have single runtime, but two chains. 

https://github.com/paritytech/parity-bridges-common/commit/c6ef9805ceaa6b872f4b4024e622b29f8b40af13 - actually we only need to know instance of messages module at the bridged chain. So it makes sense to move it from `ChainWithMessages` to `MessagesBridge`;

https://github.com/paritytech/parity-bridges-common/commit/2714bafefdf26a4f724f1803b1a06567373b5198  - I'm going to introduce `struct RococoLikeChain` that is generic over instances in the Rococo runtime, but having this constant in `ChainWithMessages` makes it impossible (and you can't use `[u8; 4]` as generic parameters);

https://github.com/paritytech/parity-bridges-common/commit/80411bad44a1e8799aeecf2f3c45d02895a6b091 - changes scale of generic parameter (Bridge -> Call);

https://github.com/paritytech/parity-bridges-common/commit/52fdf6f5ce415d5371161558e7f623d9d6b8926a - to ease integrations for runtimes that are not using messages parameters at all. Not sure if we'll need any parameters in R<>W bridge;

https://github.com/paritytech/parity-bridges-common/commit/9a424724685491c7e09cacc9e966dbc7909de673 - wanted to avoid that, but seems there's no other way. So now we're constructing runtime storage map keys in the messages module. Since we can't reference Rococo runtime, we can't use `fn storage_map_final_key` directly. And since we'll be using at least two instances in the Rococo runtime, we can't use Rialto/Millau runtime;

Part 2: here the main change (apart from adding required files to the `substrate-relay` binary) is the move of `Call` enum from `bp-rococo`/`bp-wococo` to `relay-rococo-client`/`relay-wococo-client`, because it need to reference `bridges-runtime-common`. I thought that it is better not to reference this crate from primitives.

I've added support for sending messages to be able to test relay without ui. Have no plans to support other subcommands, though - like estimate fee, encode call.


relays/client-rococo/src/runtime.rs
```diff
@@ -0,0 +1,134 @@
...
+/// The enum represents a subset of possible `Call`s we can send to Rococo chain.
+/// Ideally this code would be auto-generated from Metadata, because we want to
+/// avoid depending directly on the ENTIRE runtime just to get the encoding of `Dispatchable`s.
+///
+/// All entries here (like pretty much in the entire file) must be kept in sync with Rococo
+/// `construct_runtime`, so that we maintain SCALE-compatibility.
+///
+/// See: https://github.com/paritytech/polkadot/blob/master/runtime/rococo/src/lib.rs
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
+pub enum Call {
```
tomusdrw, line 66: The reason why I initially envisioned that in `primitives` is to make it easy for the other runtime to actually import `primitives` and create this call from a pallet in a (near) future.
That said, getting the testnet out is much higher priority than imaginary future use cases :)

tomusdrw (approved):

PR #1035, opened 2021-06-29
Use real conversion rate in greedy relayer strategy

svyatonik:
related #997 

The only thing that is missing in greedy-relayer after this PR is the cli option. However, it'll be better to extract some parts of #1005 to associate our test chains with value of some other chains before doing that. Otherwise greedy relayer will never deliver any messages.

### DIFF
--- a/relays/bin-substrate/src/messages_lane.rs
+++ b/relays/bin-substrate/src/messages_lane.rs
@@ -201,6 +201,15 @@ pub struct StandaloneMessagesMetrics {
 	pub source_to_base_conversion_rate: Option<F64SharedRef>,
 }
 
+impl StandaloneMessagesMetrics {
+	/// Return conversion rate from target to source tokens.
+	pub async fn target_to_source_conversion_rate(&self) -> Option<f64> {
+		let target_to_base_conversion_rate = (*self.target_to_base_conversion_rate.as_ref()?.read().await)?;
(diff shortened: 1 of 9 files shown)


relays/bin-substrate/src/messages_lane.rs
```diff
@@ -201,6 +201,15 @@ pub struct StandaloneMessagesMetrics {
 	pub source_to_base_conversion_rate: Option<F64SharedRef>,
 }
 
+impl StandaloneMessagesMetrics {
+	/// Return conversion rate from target to source tokens.
+	pub async fn target_to_source_conversion_rate(&self) -> Option<f64> {
+		let target_to_base_conversion_rate = (*self.target_to_base_conversion_rate.as_ref()?.read().await)?;
+		let source_to_base_conversion_rate = (*self.source_to_base_conversion_rate.as_ref()?.read().await)?;
+		Some(target_to_base_conversion_rate / source_to_base_conversion_rate)
```
tomusdrw, line 209: Might panic in case `source_to_base_conversion_rate` is `0`. Should we return `None` instead?

tomusdrw (approved):

relays/bin-substrate/src/messages_lane.rs
svyatonik: Yes, thanks. It shouldn't panic, but any negative, zero or non-normal float values may cause some issues elsewhere. So I've added check to the metric itself, which reads these values.

PR #1036, opened 2021-06-29
Use kebab-case for bridge arguments

### LINKED ISSUE #851: Use kebab-case for bridge arguments
Follow up on #849.

Currently the `structop::clap::arg_enum!` macro generates `FromStr` and `Display` + `fn variants` for an enum that represents a chain pair (a bridge). However the enum simply uses variant name for it's stringified version, so:
```rust
enum Bridge {
  RialtoToWestend,
  RialtoToMillau,
}
```
Leads to CLI arguments that look like that: `init-bridge RialtoToWestend` instead of `init-bridge rialto-to-westend`.

We rather prefer the latter, but without a need to manually implement anything. We should consider either:
1. Using some proc-macro libraries to get what we want (`derive_more`?, `serde`?)
2. Implement simple macro that does that for us.

halfpint104:
This PR closed #851. Changing CLI arguments for message sending, bridge initialisation and header relaying to be kebab case.

### DIFF
--- a/relays/bin-substrate/Cargo.toml
+++ b/relays/bin-substrate/Cargo.toml
@@ -17,6 +17,7 @@ num-format = "0.4"
 num-traits = "0.2"
 paste = "1.0"
 structopt = "0.3"
+strum = { version = "0.21.0", features = ["derive"] }
 
 # Bridge dependencies
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8069,7 +8069,7 @@ dependencies = [
  "lazy_static",
  "sp-core",
  "sp-runtime",
- "strum",
+ "strum 0.20.0",
 ]
 
 [[package]]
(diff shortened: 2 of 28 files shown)


tomusdrw (approved):

tomusdrw:
Awesome! Thanks a lot!

relays/bin-substrate/Cargo.toml
```diff
@@ -17,6 +17,7 @@ num-format = "0.4"
 num-traits = "0.2"
 paste = "1.0"
 structopt = "0.3"
+strum = { version = "0.21.0", features = ["derive"] }
```
HCastano, line 20: Why did you choose `strum` over `serde` or `derive_more`?

HCastano (approved):
Thanks!

PR #1038, opened 2021-07-01
Give real value to Rialto and Millau tokens

svyatonik:
This maybe seems a controversial PR and actually I'm personally not sure if that's the right way. That's why a separate PR for this kinda "small" change. Here's my plan:
1) we assign Rialto and Millau tokens values that are always almost 1:1;
2) we switch our test-deployments relays to greedy mode. Everything should work without any issues (of course if the 'greedy' code is correct), because rate stored in the runtime is also 1:1;
3) (optionally) once in a day we are changing rate to something that differs from 1:1, meaning that greedy relay would stop relaying messages & we'll notice that from logs (or even with alerts);
4) when/if we'll accept #1005, we'll be able to follow some tokens that are not 1:1 (polkadot+kusama) && see how relayer behaves in such environment. This will probably require additional relayer (test-only) functionality to increase stuck messages fees.

### DIFF
--- a/relays/bin-substrate/src/chains/millau_messages_to_rialto.rs
+++ b/relays/bin-substrate/src/chains/millau_messages_to_rialto.rs
@@ -225,8 +225,8 @@ pub(crate) fn add_standalone_metrics(
 		metrics_prefix,
 		metrics_params,
 		source_client,
-		None,
-		None,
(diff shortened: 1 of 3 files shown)


tomusdrw (approved):
Looks reasonable to me. Thanks for adding the extensive comment.

PR #1039, opened 2021-07-01
Confirm delivery detects when more than expected messages are confirmed

svyatonik:
This may only happen when storage of one of chains is corrupted - e.g. when target chain has 'received' `100` messages and contains no entries in unrewarded relayers vector. And the source chain has sent `100` messages, but the last confirmed message is `99`. Then the relayer may craft valid transaction, declaring that it is going to confirm `0` messages and it'll pass [this check](https://github.com/paritytech/parity-bridges-common/blob/035bee87153aef51fcf3d2605f28b3bf0cb063a2/modules/messages/src/lib.rs#L616)

### DIFF
--- a/modules/messages/src/benchmarking.rs
+++ b/modules/messages/src/benchmarking.rs
@@ -904,7 +904,7 @@ fn confirm_message_delivery<T: Config<I>, I: Instance>(nonce: MessageNonce) {
 		});
 	}
 	assert!(matches!(
-		outbound_lane.confirm_delivery(nonce, &relayers),
+		outbound_lane.confirm_delivery(nonce - latest_received_nonce, nonce, &relayers),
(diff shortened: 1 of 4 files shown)


tomusdrw (approved):
seems okay, but I don't get the full picture here really. Why is protecting from this particular corrupted state scenario necessary? I'm imagining that if we allow state corruption to be part of the model, there is probably many additional other attack vectors we should be protecting against, no? Why choose this specific one? Is it more likely to happen somehow?

svyatonik:
This particular 'corruption' may lead to situation when extrinsic has pre-dispatch weight `N*T`, where `N` is number of messages in the 'unrewarded relayers' vector. And its actual post-dispatch weight will be `(N+X)*T`, so it'll be larger than pre-dispatch weight => relayer will pay less-than-required fee + block production will slow down + ...

PR #1040, opened 2021-07-01
pre and post dispatch weights of OnDeliveryConfirmed callback

### LINKED ISSUE #1012: Update delivery confirmation transaction weights to account for delivery callbacks
#935 has introduced `OnDeliveryConfirmed` callback, which is called as a part of delivery confirmation transaction. The idea (https://github.com/paritytech/parity-bridges-common/pull/935#discussion_r622178967) is that this callback can only perform single db read + single db write for every confirmed message. However, some messages may be simply ignored by this callback, or we may have runtime without callbacks. So this callback must return spent/unspent weight.

This will decrease (I guess significantly) maximal number of messages that may be confirmed in a single delivery transaction. If it'll be true, then it may worth to call this callback from `on_initialize` instead.

svyatonik:
closes #1012 

All weight-related tests have passed, meaning that the confirmation transaction weight have enough reserve for additional db read+write.

### DIFF
--- a/modules/messages/README.md
+++ b/modules/messages/README.md
@@ -354,7 +354,7 @@ Both conditions are verified by `pallet_bridge_messages::ensure_weights_are_corr
 `pallet_bridge_messages::ensure_able_to_receive_messages` functions, which must be called from every
 runtime's tests.
 
-### Post-dispatch weight refunds of the `receive_messages_proof` call
+#### Post-dispatch weight refunds of the `receive_messages_proof` call
 
 Weight formula of the `receive_messages_proof` call assumes that the dispatch fee of every message is
 paid at the target chain (where call is executed), that every message will be dispatched and that
@@ -388,6 +388,7 @@ The weight formula is:
 Weight = BaseWeight + MessagesCount * MessageConfirmationWeight
        + RelayersCount * RelayerRewardWeight
        + Max(0, ActualProofSize - ExpectedProofSize) * ProofByteDeliveryWeight
+       + MessagesCount * (DbReadWeight + DbWriteWeight)
 ```
 
 Where:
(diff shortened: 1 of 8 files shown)


modules/messages/src/lib.rs
```diff
@@ -641,7 +662,37 @@ decl_module! {
...
+							"Messages delivery callback has returned unspent weight to refund the submitter: \
+							{} - {} = {}",
+							preliminary_callback_overhead,
+							actual_callback_weight,
+							difference,
+						);
+						actual_weight = actual_weight - difference;
+					},
+					None => {
+						debug_assert!(false, "The delivery confirmation callback is wrong");
```
svyatonik, line 686: This is something I'm not comfortable with, but Substrate pallets are full of debug_asserts, so it should be fine

modules/messages/README.md
```diff
@@ -403,6 +404,15 @@ Where:
 | `ExpectedProofSize`       | `EXTRA_STORAGE_PROOF_SIZE`                                                                                            | Size of proof that we are expecting                                                                                                                                                                     |
 | `ProofByteDeliveryWeight` | `(receive_single_message_proof_16_kb - receive_single_message_proof_1_kb) / (15 * 1024)`                              | Weight of processing every additional proof byte over `ExpectedProofSize` limit. We're using the same formula, as for message delivery, because proof mechanism is assumed to be the same in both cases |
 
+#### Post-dispatch weight refunds of the `receive_messages_delivery_proof` call
+
+Weight formula of the `receive_messages_delivery_proof` call assumes that all messages in the proof
+are actually delivered (so there are no already confirmed messages) and every messages is processed
+by the `OnDeliveryConfirmed` callback. This means that for every message, we're adding single db read
+weight and single db write weight. If, by some reason, messaes are not processed by the
```
tomusdrw, line 412:
```suggestion
weight and single db write weight. If, by some reason, messages are not processed by the
```

modules/messages/src/lib.rs
tomusdrw: Wow, interesting. If that's the convention, let's keep it - I guess it's there just to know that it is never triggered in tests (Assuming they are run in debug mode not release)?

tomusdrw (approved):

PR #1042, opened 2021-07-02
Cli option for greedy relayer + run no-losses relayer by default

### LINKED ISSUE #997: Support 'greedy' mode in relayer

svyatonik:
closes #997

This PR adds cli option for selecting relayer mode (altruistic || no-losses). no-losses relayer is the default option (this will break wococo <> rococo, because they have no value => conversion rate is unknown => no messages will be delivered => R<>W relays need to specify `--relayer-mode=altruistic`).

What may happen after this PR is merged:
1) messages delivery will break if there's an issue in greedy relayer code (locally it is running fine, though);
2) messages delivery will start breaking from time to time if price of tokens, associated with chains will start to diverge from ideal 1:1 ratio.

I've also added some traces to be able to diagnose possible issues from logs.

### DIFF
--- a/relays/bin-substrate/src/cli/relay_messages.rs
+++ b/relays/bin-substrate/src/cli/relay_messages.rs
@@ -22,7 +22,26 @@ use crate::cli::{
 use crate::messages_lane::MessagesRelayParams;
 use crate::select_full_bridge;
 use structopt::StructOpt;
-use strum::VariantNames;
+use strum::{EnumString, EnumVariantNames, VariantNames};
+
+/// Relayer operating mode.
+#[derive(Debug, EnumString, EnumVariantNames, Clone, Copy, PartialEq)]
+#[strum(serialize_all = "kebab_case")]
+pub enum RelayerMode {
+	/// The relayer doesn't care about rewards.
+	Altruistic,
+	/// The relayer will deliver all messages and confirmations as long as he's not losing any funds.
+	NoLosses,
+}
+
+impl From<RelayerMode> for messages_relay::message_lane_loop::RelayerMode {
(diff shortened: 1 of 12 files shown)


relays/bin-substrate/src/cli/relay_messages.rs
```diff
@@ -22,7 +22,26 @@ use crate::cli::{
...
+use strum::{EnumString, EnumVariantNames, VariantNames};
+
+/// Relayer operating mode.
+#[derive(Debug, EnumString, EnumVariantNames, Clone, Copy, PartialEq)]
+#[strum(serialize_all = "kebab_case")]
+pub enum RelayerMode {
+	/// The relayer doesn't care about rewards.
+	Altruistic,
+	/// The relayer will deliver all messages and confirmations as long as he's not losing any funds.
+	NoLosses,
```
tomusdrw, line 34:
```suggestion
	Rational,
```
I'd vote for `Rational` instead of `NoLosses`. Couple of reasons for that:
1. I find the plural and the "no losses" thing a bit weird. AFAIR we still can't guarantee that in case multiple relayers are running it will indeed never lose any funds.
2. `Rational` as "Rational Economic Actor" in game theory dissertations
3. I find it easier to mentally compare `Altruistic` and `Rational` than `Altruistic` and `NoLosses`. Both are adjectives that describe a behavior.

tomusdrw (approved):
lgtm!

HCastano (approved):

PR #1043, opened 2021-07-05
Pay dispatch fee at target chain cli option

svyatonik:
1) I haven't added this option to `estimate-fee` and `encode-message` - I'm not sure if we still need these commands + they already have some default values (e.g. for target dispatch origin);
2) this actually has no effect on tx fee in our test deployments, because chains are configured to have zero fee multiplier (the weight-dependent portion of tx fee is computed as 'multiplier * weight'). It'll have effect on R/W/P/K, though

### DIFF
--- a/deployments/bridges/rialto-millau/entrypoints/relay-messages-to-millau-generator-entrypoint.sh
+++ b/deployments/bridges/rialto-millau/entrypoints/relay-messages-to-millau-generator-entrypoint.sh
@@ -46,6 +46,7 @@ do
 		$SEND_MESSAGE \
 			--lane $SECONDARY_MESSAGE_LANE \
 			--origin Target \
+			--dispatch-fee-payment at-target-chain \
 			remark
 	fi
 
--- a/deployments/bridges/rialto-millau/entrypoints/relay-messages-to-rialto-generator-entrypoint.sh
+++ b/deployments/bridges/rialto-millau/entrypoints/relay-messages-to-rialto-generator-entrypoint.sh
@@ -46,6 +46,7 @@ do
 		$SEND_MESSAGE \
 			--lane $SECONDARY_MESSAGE_LANE \
 			--origin Target \
+			--dispatch-fee-payment at-target-chain \
 			remark
 	fi
 
--- a/relays/bin-substrate/src/cli/send_message.rs
+++ b/relays/bin-substrate/src/cli/send_message.rs
@@ -73,7 +73,7 @@ pub struct SendMessage {
 	#[structopt(
 		long,
 		possible_values = DispatchFeePayment::VARIANTS,
-		case_insensitive=true,
+		case_insensitive = true,
 		default_value = "at-source-chain",
 	)]
 	dispatch_fee_payment: DispatchFeePayment,
--- a/relays/bin-substrate/src/cli/send_message.rs
+++ b/relays/bin-substrate/src/cli/send_message.rs
@@ -22,7 +22,6 @@ use crate::cli::{
 	TargetSigningParams,
 };
 use bp_message_dispatch::{CallOrigin, MessagePayload};
-use bp_runtime::messages::DispatchFeePayment;
 use codec::Encode;
 use frame_support::weights::Weight;
 use relay_substrate_client::{Chain, TransactionSignScheme};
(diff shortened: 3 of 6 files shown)


relays/bin-substrate/src/cli/send_message.rs
```diff
@@ -22,15 +22,33 @@ use crate::cli::{
...
 use structopt::StructOpt;
-use strum::VariantNames;
+use strum::{EnumString, EnumVariantNames, VariantNames};
+
+/// Relayer operating mode.
+#[derive(Debug, EnumString, EnumVariantNames, Clone, Copy, PartialEq, Eq)]
+#[strum(serialize_all = "kebab_case")]
+pub enum DispatchFeePayment {
+	/// The dispacth fee is paid at the source chain.
+	AtSourceChain,
```
tomusdrw, line 39: Can we rename (in `strum`?) so that the CLI options are just `source` and `target`?

deployments/bridges/rialto-millau/entrypoints/relay-messages-to-millau-generator-entrypoint.sh
```diff
@@ -46,6 +46,7 @@ do
 		$SEND_MESSAGE \
 			--lane $SECONDARY_MESSAGE_LANE \
 			--origin Target \
+			--dispatch-fee-payment at-target-chain \
```
tomusdrw, line 49: Would that always work? Do we have enough funds in the companion account?

deployments/bridges/rialto-millau/entrypoints/relay-messages-to-rialto-generator-entrypoint.sh
```diff
@@ -46,6 +46,7 @@ do
 		$SEND_MESSAGE \
 			--lane $SECONDARY_MESSAGE_LANE \
 			--origin Target \
+			--dispatch-fee-payment at-target-chain \
```
tomusdrw, line 49: same question here.

relays/bin-substrate/src/cli/send_message.rs
```diff
@@ -51,6 +69,14 @@ pub struct SendMessage {
 	/// Hex-encoded lane id. Defaults to `00000000`.
 	#[structopt(long, default_value = "00000000")]
 	lane: HexLaneId,
+	/// Where dispatch fee is paid?
+	#[structopt(
+		long,
+		possible_values = DispatchFeePayment::VARIANTS,
+		case_insensitive=true,
```
tomusdrw, line 76:
```suggestion
		case_insensitive = true,
```

tomusdrw (approved):

tomusdrw:
No need to add the option to other commands for now IMHO. They were there mostly to help the UI team figure out the encodings.

relays/bin-substrate/src/cli/send_message.rs
svyatonik: Yes, but what's the reasoning? IMO it is better to have same id everywhere than trying to remember how selected variant is called in CLI

deployments/bridges/rialto-millau/entrypoints/relay-messages-to-millau-generator-entrypoint.sh
svyatonik: The dispatch origin on target chain is //Dave, so it should have enough funds. I can't guarantee if it'll be enough for all daily messages, but probably it will.

Actually, I just had a thought - this PR should have no effect on our chains, because `multiplier` [here](https://github.com/paritytech/substrate/blob/19c6dd2f2ab8dea8d5734f2a0f246c935b76e75e/frame/transaction-payment/src/lib.rs#L473) is always zero. But actually it will have an effect, because multiplier is not used [here](https://github.com/paritytech/parity-bridges-common/blob/20d08204a2a93fef5c93c70c488f9049086fdb97/bin/runtime-common/src/messages.rs#L531). So the answer to your question would be: (1) yes we have enough funds because amount required for this PR is `0` (2) current implementation has an issue, though, so it'll be non-zero - I'll fix it in a parallel PR.

svyatonik:
> No need to add the option to other commands for now IMHO. They were there mostly to help the UI team figure out the encodings.

WDYT about removing these options then - there's a lot of code dedicated to these subcommands.

deployments/bridges/rialto-millau/entrypoints/relay-messages-to-millau-generator-entrypoint.sh
svyatonik: The alternative would be to change multiplier in Millau/Rialto to non-zero. E.g. to `1`, or even use `1` iniitally and `SlowAdjustingFeeUpdate` from polkadot runtime. I've been using hardcoded version for some experiments some time ago, but probably changing to somewhat similar to production code will be better. Anyway - I'm going do that in a separate PR - it's not a trivial change.

tomusdrw:
> WDYT about removing these options then - there's a lot of code dedicated to these subcommands?

Frankly, I'd prefer to keep them and make it on-par with the UI (the keyword in my previous comment is "for now" :)). The UI will always be lagging behind the core stuff and for instance it would be nice to enable token swap, by using "Custom Call" UI feature and creating the encoding via CLI.
Also it will help the UI team make sure the encoding is exactly as expected in the future, when they start implementing additional features like that.

relays/bin-substrate/src/cli/send_message.rs
tomusdrw: It's about CLI user experience. There is already a lot of options to pass and shorter terms are easier to remember than multi-word sentences. I'm okay renaming the enums too, but being more explicit in the code do add readability value IMHO.

svyatonik: There are two type of users imo:
- if you're familiar with relay code, then you should have clear association between this cli option and enum and it'll be easier to you to just remember **single** variant;
- if you're not familiar with relay code, then you'll learn cli options either from examples or from clip help, where you'll find everything you need. And since you have no any initial knowledge, there'll be no shorter vs longer confusion. And I really doubt that remembering `at-source-chain` vs `source` is a problem. I'd rather argue that understanding what `--dispatch-fee-payment` means is a problem. And then seeing `--dispatch-fee-payment=at-source-chain` makes it easier to understand than seeing '--dispatch-fee-payment'=source`.

PR #1044, opened 2021-07-05
Enable weight fee adjustment in Rialto/Millau chains

svyatonik:
Consequences:
1) basically all transactions costs are doubled, so we may expect insufficient balance errors after this PR will be merged;
2) I've added another messages parameter that won't be updated in testnets, but we'll need it in P<>K.

Just an interesting fact that I've noticed (though it is obvious if you think about it) - the delivery fee that the sender pays for the pay-dispatch-fee-at-target-chain message may be larger than the cost of the same message with pay-dispatch-fee-at-source-chain, though it is counter-intuitive. That's because delivery transaction cost is increased by the weight of the currency transfer call (that is the transfer from target dispatch origin account to the relayer account). So not even the pay-at-target chain is always more expensive (here I mean total cost on both chains) than the pay-at-source chain, but for sometimes it isn't even cheaper at the source chain.

### DIFF
--- a/bin/runtime-common/src/messages.rs
+++ b/bin/runtime-common/src/messages.rs
@@ -38,7 +38,7 @@ use frame_support::{
 };
 use hash_db::Hasher;
 use sp_runtime::{
-	traits::{AtLeast32BitUnsigned, CheckedAdd, CheckedDiv, CheckedMul},
+	traits::{AtLeast32BitUnsigned, CheckedAdd, CheckedDiv, CheckedMul, Saturating, Zero},
 	FixedPointNumber, FixedPointOperand, FixedU128,
 };
 use sp_std::{cmp::PartialOrd, convert::TryFrom, fmt::Debug, marker::PhantomData, ops::RangeInclusive, vec::Vec};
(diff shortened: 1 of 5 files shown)


bin/runtime-common/src/messages.rs
```diff
@@ -525,13 +526,20 @@ pub mod target {
...
 				|dispatch_origin, dispatch_weight| {
-					ThisCurrency::transfer(
-						dispatch_origin,
-						relayer_account,
-						ThisRuntime::WeightToFee::calc(&dispatch_weight),
-						ExistenceRequirement::AllowDeath,
-					)
-					.map_err(drop)
+					let unadjusted_weight_fee = ThisRuntime::WeightToFee::calc(&dispatch_weight);
+					let fee_multiplier = pallet_transaction_payment::Pallet::<ThisRuntime>::next_fee_multiplier();
```
tomusdrw, line 530: Hmm, while this seems innocent and is definitely what we want to do, I started thinking about potential attack vector towards bridge senders.
The relayers can attempt to increase the fee multiplier (by sending remarks for instance) and only then dispatch the message. Do you think this will be an issue in practice? Does the cost of such attack make it feasbile?

tomusdrw (approved):
lgtm!

bin/runtime-common/src/messages.rs
svyatonik: If I get everything right, then there's absolutely no reasons for relayers to do that. Basically - this code is executed as a part of delivery transaction. The fee of delivery transaction is computed as `BaseTxFee + PerByteFee * TransactionLength + Multiplier * WeightToFee(BaseDeliveryTransactionWeight + MessageDispatchWeight)`. This transfer compensates `Multiplier * WeightToFee(MessageDispatchWeight)` amount that the relayer has just spent (inside the same transaction). So even if relayer will be increasing multiplier, he won't get any profit, because he has just spent the same amount (computed using increased multiplier) on transaction fees.

tomusdrw: Ugh, yeah, you're totally right. I was somehow thinking that the payment here might be significantly more than what the relayer does actually pay.

Issue #201, opened 2020-07-17
Add Kovan Bridge to Rococo Parachain Runtime


tomusdrw:
Since we'd like to deploy Kovan bridge on Rococo Parachain, we must add the bridge pallet to Rococo runtime.

Related: https://github.com/paritytech/parity-bridges-common/issues/202

tomusdrw:
Closing, not really planned in a foreseeable future.

PR #1047, opened 2021-07-06
Add UI to the deployment

tomusdrw:
This PR adds an additional docker service to the compose file which contains the UI static files hosted by a nginx server.

The image is based on Nikos' work https://github.com/paritytech/parity-bridges-ui/pull/197
Since the image is not being published regularly yet (see https://github.com/paritytech/parity-bridges-ui/issues/201) I'm using the test version.

There is one additional thing that should be addressed (but can be done independently from this PR imho):
- [x] Currently the image has nodes addresses "hardcoded" we need to be able to configure that with `environment` variables so that it can be pointing to local instances instead of the hosted ones. I've added a note about that in [#201](https://github.com/paritytech/parity-bridges-ui/issues/201).


svyatonik (approved):
> Currently the image has nodes addresses "hardcoded"

This may be very confusing for those who are running deployments locally :) Hope there aren't many

tomusdrw:
> This may be very confusing for those who are running deployments locally :) Hope there aren't many

Yes. I'm considering changing this PR to a draft until https://github.com/paritytech/parity-bridges-ui/issues/201 is resolved.

deployments/ui/docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '3.5'
+services:
+  bridges-ui:
+    # TODO [ToDr] switch to paritytech after https://github.com/paritytech/parity-bridges-ui/issues/201
+    image: wirednkod13/parity-bridges-ui
```
hbulgarini, line 5:
```suggestion
    image: paritytech/parity-bridges-ui
```

hbulgarini (commented):
You can go ahead and replace the UI docker image with the one pushed on `paritytech` hub

hbulgarini:
>  Currently the image has nodes addresses "hardcoded" we need to be able to configure that with environment variables so that it can be pointing to local instances instead of the hosted ones. I've added a note about that in #201.

If you agree with proposed approach on PR https://github.com/paritytech/parity-bridges-ui/pull/273 this can be solved already.

tomusdrw:
> If you agree with proposed approach on PR paritytech/parity-bridges-ui#273 this can be solved already.

This doesn't work unfortunately. The `.env` file affects the image that is being built, but docker compose references an already-built image. We need something that will modify the paths in the runtime (i.e. before starting the container). I have an even more hacky solution somewhere on a branch already, will try to finish this up soonish.

hbulgarini (approved):
The PR looks good to me even if it still is under draft status.

PR #1054, opened 2021-07-15
Ignore nonstandard_macro_braces clippy warning

svyatonik:


svyatonik:
Related issue: https://github.com/rust-lang/rust-clippy/issues/7434
And re that ICE: https://github.com/rust-lang/rust-clippy/issues/7423 - should hopefully be fixed in tomorrow nightly

svyatonik:
(this PR is a bit too late - everything is fixed in nightly as of today)

PR #1059, opened 2021-07-19
Allow reading suri && password override from file

### LINKED ISSUE #1000: Pass signer using keyfile or env variable.
To avoid leaking the secret in CLI logs it's expected to pass secrets (signer SURI, etc) either as env variables or as a filename.

svyatonik:
closes #1000

The easiest way would be to use [Environment variable fallback](https://docs.rs/structopt/0.3.22/structopt/#environment-variable-fallback). But we're now using options like`--source-signer`, and it's not super intuitive (imho easy to make an error) to have a variable named `LEFT_SIGNER` and know that it means signer at Millau. This is especially important if you're starting multiple relay processes from single script. So imo seeing `--source-signer-file=millau-signer-suri` is better than falling back to `LEFT_SIGNER`.

Still, everything that is read from file may be overrided by the explicit cli-options. E.g. passing `--source-signer=//Alice --source-signer-file=bob-suri` will mean that `//Alice` will be used. May make it conflicting options, if it looks better to someone.

### DIFF
--- a/relays/bin-substrate/src/cli/mod.rs
+++ b/relays/bin-substrate/src/cli/mod.rs
@@ -374,24 +374,61 @@ macro_rules! declare_chain_options {
 			}
 
 			#[doc = $chain " signing params."]
-			#[derive(StructOpt, Debug, PartialEq, Eq)]
+			#[derive(StructOpt, Debug, PartialEq, Eq, Clone)]
 			pub struct [<$chain SigningParams>] {
 				#[doc = "The SURI of secret key to use when transactions are submitted to the " $chain " node."]
 				#[structopt(long)]
-				pub [<$chain_prefix _signer>]: String,
+				pub [<$chain_prefix _signer>]: Option<String>,
 				#[doc = "The password for the SURI of secret key to use when transactions are submitted to the " $chain " node."]
 				#[structopt(long)]
 				pub [<$chain_prefix _signer_password>]: Option<String>,
+
+				#[doc = "Path to the file, that contains SURI of secret key to use when transactions are submitted to the " $chain " node. Can be overrided with " $chain_prefix " option."]
+				#[structopt(long)]
+				pub [<$chain_prefix _signer_file>]: Option<std::path::PathBuf>,
+				#[doc = "Path to the file, that password for the SURI of secret key to use when transactions are submitted to the " $chain " node. Can be overrided with " $chain_prefix " option."]
+				#[structopt(long)]
+				pub [<$chain_prefix _signer_password_file>]: Option<std::path::PathBuf>,
 			}
 
 			impl [<$chain SigningParams>] {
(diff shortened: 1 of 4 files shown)


relays/bin-substrate/src/cli/send_message.rs
```diff
@@ -60,12 +60,8 @@ pub struct SendMessage {
...
 	#[structopt(flatten)]
 	source_sign: SourceSigningParams,
-	/// The SURI of secret key to use when transactions are submitted to the Target node.
-	#[structopt(long, required_if("origin", "Target"))]
-	target_signer: Option<String>,
-	/// The password for the SURI of secret key to use when transactions are submitted to the Target node.
-	#[structopt(long)]
-	target_signer_password: Option<String>,
+	#[structopt(flatten)]
+	target_sign: TargetSigningParams,
```
svyatonik, line 64: This change means that now this particular cli option is checked during runtime. It isn't super-intuitive, but I haven't found a way to reuse existing struct && I do not want to create another one

relays/bin-substrate/src/cli/mod.rs
```diff
@@ -374,24 +374,61 @@ macro_rules! declare_chain_options {
...
 			pub struct [<$chain SigningParams>] {
 				#[doc = "The SURI of secret key to use when transactions are submitted to the " $chain " node."]
 				#[structopt(long)]
-				pub [<$chain_prefix _signer>]: String,
+				pub [<$chain_prefix _signer>]: Option<String>,
 				#[doc = "The password for the SURI of secret key to use when transactions are submitted to the " $chain " node."]
 				#[structopt(long)]
 				pub [<$chain_prefix _signer_password>]: Option<String>,
+
+				#[doc = "Path to the file, that contains SURI of secret key to use when transactions are submitted to the " $chain " node. Can be overrided with " $chain_prefix " option."]
```
tomusdrw, line 386:
```suggestion
				#[doc = "Path to the file, that contains SURI of secret key to use when transactions are submitted to the " $chain " node. Can be overridden with " $chain_prefix "_signer option."]
```
?

relays/bin-substrate/src/cli/mod.rs
```diff
@@ -374,24 +374,61 @@ macro_rules! declare_chain_options {
...
+				#[structopt(long)]
+				pub [<$chain_prefix _signer_file>]: Option<std::path::PathBuf>,
+				#[doc = "Path to the file, that password for the SURI of secret key to use when transactions are submitted to the " $chain " node. Can be overrided with " $chain_prefix " option."]
```
tomusdrw, line 389:
```suggestion
				#[doc = "Path to the file, that password for the SURI of secret key to use when transactions are submitted to the " $chain " node. Can be overridden with " $chain_prefix "_signer_password option."]
```

tomusdrw (approved):

PR #1061, opened 2021-07-20
Use tokio runtime to execute jsonrpsee futures

svyatonik:


svyatonik:
(I'm merging this, as it breaks our deployments && I want to see them green tomorrow)

PR #1064, opened 2021-07-21
Fix Westend -> Millau sync

svyatonik:
We read justifications are read from the stream using `stream.next().now_or_never()`. In current code, every time `next` is called on the stream, additional tokio task is spawned. And because of that, only one justification is read from the stream, every time we awake. Even though the stream has more unread justifications. This PR changes the way how `stream::next()` is implemented - instead of spawning tokio task every time we call `next`, it is only called once (when stream is created) and keep running, filling justifications channel. This channel is polled from `stream::next()`, meaning that we only deal with single executor there.

I've also added couple of traces to help debugging similar issues in the future

### DIFF
--- a/relays/client-substrate/src/client.rs
+++ b/relays/client-substrate/src/client.rs
@@ -23,8 +23,9 @@ use crate::{ConnectionParams, Error, Result};
 use async_std::sync::{Arc, Mutex};
 use codec::Decode;
 use frame_system::AccountInfo;
+use futures::{SinkExt, StreamExt};
 use jsonrpsee_ws_client::{traits::SubscriptionClient, v2::params::JsonRpcParams, DeserializeOwned};
-use jsonrpsee_ws_client::{Subscription, WsClient as RpcClient, WsClientBuilder as RpcClientBuilder};
+use jsonrpsee_ws_client::{WsClient as RpcClient, WsClientBuilder as RpcClientBuilder};
 use num_traits::{Bounded, Zero};
 use pallet_balances::AccountData;
 use pallet_transaction_payment::InclusionFee;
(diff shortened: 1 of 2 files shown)


relays/client-substrate/src/client.rs
```diff
@@ -397,10 +426,8 @@ impl<C: Chain> Client<C> {
 impl JustificationsSubscription {
 	/// Return next justification from the subscription.
 	pub async fn next(&self) -> Result<Option<Bytes>> {
-		let subscription = self.1.clone();
-		self.0
-			.spawn(async move { subscription.lock().await.next().await })
-			.await?
-			.map_err(Error::RpcError)
+		let mut receiver = self.0.lock().await;
+		let justification = receiver.next().await;
```
tomusdrw, line 430: Does it mean we keep the `lock` active when we wait for the next justification? Why do we have multiple tasks for that if only one of them is really making progress at the same time anyway? (i.e. why is it `Arc<Mutex` and we have multiple `JustificationSubscription` instances).

svyatonik: We don't have multiple `JustificationSubscription` instances. `JustificationSubscription` needs to be `Send + Sync` => `Arc<Mutex<>>`. It is polled using `now_or_never()`, so if either `lock()` or `next()` are not `Poll::Ready(_)`, it'll immediately return

tomusdrw: I see. Although `Mutex` seems to be `Send+Sync` already, I guess we can now drop the `Arc` since we never move the mutex to a different future (as in previous code).

svyatonik:
(merging to see working test deployments tomorrow)

Issue #1002, opened 2021-06-11
Token mint pallet based on specific event


tomusdrw:
The idea is to create a pallet that could be used in a one-way bridge. The pallet will verify a specific event on the bridged chain (for instance "burn" or "bond") and mint tokens on the home chain with the same amount.

The use cases in mind:

1. Give testnet (Rococo / Westend) tokens to people who bonded KSM on Kusama.
2. Move testnet tokens between networks (Westend -> Rococo) transition.

Issueing tokens based on tokens `bond`ed is obviously gameable, but it's rate-limitted and costly. One needs to wait for unbounding period, send some additional transactions and have the lost opportunity cost (staking rewards), so I think it's still interesting.

If the pallet was made generic enough it could be an arbitrary "if event -> then action" kind of pallet. So it should look up a particular event (or events) and then trigger some action on the home chain. There should be no ordering required or guaranteed, calls should be expected to be made by regular users that are interested.
To prevent multi-claims we can temporarily store already-claimed events and prune them at the same time the finality pallet is being pruned.

This pallet should not be based on messaging protocol, but rather just on the finality pallet.

tomusdrw:
Another use case is deploying two one way bridges with potentially different security:
1. Fully trustless one way bridge based on this pallet (transfer to multisig -> mint)
2. Multisig based bridge to transfer tokens back on the main chain.

The advantage is that the network we bridge to only needs to support the multisig pallet and no extra bridge-specific pallets need to be added - In fact the chain is not even aware it's bridged with.
If the bridge chain is using some form of PoS and the multisig is controlled by the same set of validators (would require multisig ownership changes during authority set hand offs though) the security is essentially as high as it can be - by having tokens on the bridged chain you already trust the authorities there not to collude, same as with the multisig. Colluding on multisig transfers could also be made a slashable offence on the original chain to prevent validators from colluding only on the bridge side not the PoS chain itself.

EmmanuellNorbertTulbure:
Obsolete

PR #1068, opened 2021-07-22
Parachains finality pallet

svyatonik:
I've been considering generic pallet that'll e able to solve both parachains finality + https://github.com/paritytech/parity-bridges-common/issues/1002#issue-918820254 . That's because both pallets are simple receiving storage proof => parsing proof => optionally check that the proof is duplicate => doing some actions, based on this proof. But the generic palle will be **too generic** - imo it is better to have parachains finality logic separated from some generic code.

This is very WIP, although current code already gives general idea of what it'll look like. Main questions that still needs to be solved is keeping old parachain headers (aka heads) as we do with relay chain headers + add similar pruning.


primitives/runtime/src/storage_keys.rs
```diff
@@ -0,0 +1,49 @@
+// Copyright 2019-2021 Parity Technologies (UK) Ltd.
```
svyatonik, line 1: This whole file is: (1) wrong, as all our pallets (and `paras` pallet) are based on frame v2, which means that the storage keys are computed slightly differently (2) obsolete, because in master there are **correct** versions of these functions declared at the crate level.

I could fix it in this PR, or in follow-up PRs.

modules/parachains/src/lib.rs
```diff
@@ -0,0 +1,594 @@
...
+		/// The setting is there to prevent growing the on-chain state indefinitely. Note
+		/// the setting does not relate to parachain block numbers - we will simply keep as much
+		/// items in the storage, so it doesn't guarantee any fixed timeframe for heads.
+		#[pallet::constant]
+		type HeadsToKeep: Get<u32>;
+	}
+
+	/// Best parachain heads.
+	#[pallet::storage]
+	pub type BestParaHeads<T: Config<I>, I: 'static = ()> =
```
svyatonik, line 95: Identity hasher for `ParaId` should probably be replaced with blake2 here and below

modules/parachains/src/lib.rs
```diff
@@ -0,0 +1,594 @@
...
+
+			// now parse storage proof and read parachain heads
+			pallet_bridge_grandpa::Pallet::<T, T::BridgesGrandpaPalletInstance>::parse_finalized_storage_proof(
+				relay_block_hash,
+				sp_trie::StorageProof::new(parachain_heads_proof),
+				move |storage| {
+					for parachain in parachains {
+						let parachain_head = match Pallet::<T, I>::read_parachain_head(&storage, parachain) {
+							Some(parachain_head) => parachain_head,
+							None => {
```
svyatonik, line 150: Actually it should be supported - parachain may be offboarded and its head may be removed. So there should be 3 possible outcomes from `read_parachain_head`: `Err(_)`, `Ok(Some(_))` and `Ok(None)`. Right now the first (which is relayer error) is replaced with `Ok(None)`, which is handled incorrectly.

primitives/runtime/src/storage_keys.rs
svyatonik: fixed

modules/parachains/src/lib.rs
svyatonik: I've changed all to be the strongest `Blake2_128Concat` for `ParaId` and `ParaHash`, as they can be (in theory) selected by the submitter. As for the insertion index, I've changed to `Twox64Concat`

modules/parachains/src/lib.rs
svyatonik: I'd prefer to deal with it in follow-up PRs, as it requires some more thinking + there are already two PRs on top of that one.

primitives/polkadot-core/src/parachains.rs
```diff
@@ -0,0 +1,43 @@
...
+//! reference polkadot crates from pallets. That's because bridges repository is
+//! included in the polkadot repository and included pallets are used by polkadot
+//! chains. Having pallets that are referencing polkadot, would mean that there may
+//! be two versions of polkadot crates included in the runtime. Which is bad.
+
+use sp_core::Hasher;
+use sp_std::vec::Vec;
+
+/// Parachain id.
+pub type ParaId = u32;
```
svyatonik, line 29: All parachain pallets from Polkadot are using `polkadot_parachain::primitives::Id` to store parachain id. The difference is that it is compact-encoded `u32` => it should be fixed

primitives/polkadot-core/src/parachains.rs
```diff
@@ -0,0 +1,43 @@
...
+
+/// Parachain head, which is (at least in Cumulus) a SCALE-encoded parachain header.
+pub type ParaHead = Vec<u8>;
```
svyatonik, line 32: Parachains pallet from Polkadot is using `polkadot_parachain::primitives::HeadData` to store parachain heads. The difference is that they're encoding parachain header twice, i.e. it is prefixed with encoded bytevec length => it should be fixed.

svyatonik:
@AurevoirXavier @fewensa If you're really keen on this parachain pallets+relay+bridge, then it'll be really helpful if you could add your review to these three dependent PRs - this one, #1199 and #1218 . Please ask any questions directly in PRs (better option) or DM me on riot (you've been already chatting with me there).

The most important design question: there are two options to use parachains finality - one is a separate pallet+relay and another one is simply adding parachain head proof to existing proofs. The latter looks simpler - i.e. right now proof of messages looks like `(bridged_header_hash, message_storage_proof, lane, range_of_nonces)`. We could then simply add a `(parachain_id, parachain_head_storage_proof)` there and everything should work. But in this case we won't be able to use results of previous `parachain_head_storage_proof` verification when/if another messages proof is submitted. We may add some cache, but it'll result exactly in a separate pallet. Another thing is that this separate pallet is just another layer on top of our `bridge-grandpa` (and following `bridge-beefy`) pallet. Both are sharing (almost) the same API and we're not mixing all possible cases in a single pallet - e.g. we may combine relay<>parachain, relay<>relay and parachain<>parachain bridge by using different set of isolated pallets, while only changing the "glue" code.

HackFisher:
> But in this case we won't be able to use results of previous parachain_head_storage_proof verification when/if another messages proof is submitted.

I don't get this part, do you mean a new parachain_head_storage_proof could be submitted before some message proof get included using relative older parachain header?

If that's the case, bridge grandpa pallet may also have similar issue, and onchain cache looks simple and straight forward to me.

svyatonik:
> > But in this case we won't be able to use results of previous parachain_head_storage_proof verification when/if another messages proof is submitted.
> 
> I don't get this part, do you mean a new parachain_head_storage_proof could be submitted before some message proof get included using relative older parachain header?
> 
> If that's the case, bridge grandpa pallet may also have similar issue, and onchain cache looks simple and straight forward to me.

GRANDPA pallet works exactly like this cache :) Let me explain in details. So message proof = `(bridged_header_hash, message_storage_proof, lane, range_of_nonces)`. To verify `message_storage_proof`, you'll need to know a storage trie root, which is a part of `bridged_header_hash` header. And you need to be sure that this header is finalized.

So imagine you don't have a bridge GRANDPA pallet in the runtime (let's omit validator set changes for simplicity). Then the messages pallet may still be used, but the message proof changes to `(bridged_header, bridged_header_grandpa_justification, message_storage_proof, lane, range_of_nonces)`. So it includes finality proof now. And if you need to prove several same-header messages using several transactions, it would mean that you'll need to submit and verify justification multiple times.

That's where GRANDPA pallet helps and where it works like cache - once finality proof is verified, you may just provide header hash in your message proof. and when this proof is verified, you may just read the header from the cache (runtime storage), without reverifying finality proof again.

Now imagine you're bridging with parachain and you don't have both GRANDPA pallet and parachains finality pallet. Then the message proof would be `(bridged_relay_header, bridged_relay_header_grandpa_justification, bridged_parachain_header, bridged_parachain_header_storage_proof, message_storage_proof, lane, range_of_nonces)`. And you would need to verify 3 proofs every time you're reusing the same parachain header in the message proof. So what this separate parachains finality pallet does, is removing the `bridged_parachain_header, bridged_parachain_header_storage_proof` from this tuple, instead providing a trusted cached map of `bridged_parachain_header_hash => bridged_parachain_header`. So it also works like cache, but with a parallel API, like bridge GRANDPA pallet does.

Let me know if it makes sense to you.

eskimor (approved):
Looks pretty good as far as I can tell.

modules/parachains/src/lib.rs
```diff
@@ -0,0 +1,594 @@
...
+					}
+				},
+			)
+			.map_err(|_| Error::<T, I>::InvalidStorageProof)?;
+
+			// TODO: there may be parachains we are not interested in - so we only need to accept
+			// intersection of `parachains-interesting-to-us` and `parachains`
+
+			// TODO: if some parachain is no more interesting to us, we should start pruning its
+			// heads
```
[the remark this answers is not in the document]
svyatonik: Yep, thanks - I'll open issues for that

acatangiu (commented):
Looks good, just nits and typos :+1:

serban300 (commented):
I tried to ramp-up as a bit on how parachains work and I did multiple iterations on this code. However I still have limited experience with this. But for me, the design looks good. I can't find any issue with it. I just left 2 nits.

modules/parachains/src/lib.rs
```diff
@@ -0,0 +1,601 @@
...
+/// Best known parachain head as it is stored in the runtime storage.
+#[derive(Decode, Encode, PartialEq, RuntimeDebug, TypeInfo)]
+pub struct BestParaHead {
+	/// Number of relay block where this head has been updated.
+	pub at_relay_block_number: RelayBlockNumber,
+	/// Hash of parachain head.
+	pub head_hash: ParaHash,
+	/// Current ring buffer position for this parachain.
+	pub next_imported_hash_position: u32,
+}
```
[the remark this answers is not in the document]
svyatonik: Yeah, that's a good idea imo

PR #1071, opened 2021-07-27
Merge two weight-related loops in messages pallet

### LINKED ISSUE #919: Merge two loops in `receive_messages_proof`
There are now two loops ([loop1](https://github.com/paritytech/parity-bridges-common/pull/911/files#diff-66ac6a25a52411800da13cc98b1aeb99cfb3e16966920a12b38f5d649c2b635cR491) and [loop2](https://github.com/paritytech/parity-bridges-common/pull/911/files#diff-66ac6a25a52411800da13cc98b1aeb99cfb3e16966920a12b38f5d649c2b635cR514)) in the `receive_messages_proof` call. The first just checks whether `dispatch_weight` argument covers dispatch of all passed messages. And the second one is an actual dispatch. Given that now (see #911) we're able to return actual post-dispatch weight from this call, it makes sense to merge two loops in one and keep dispatching messages until `dispatch_weight` covers their dispatch.

svyatonik:
closes #919 

This change breaks existing pallet logic - previously we were rejecting transactions which had declared `dispatch_weight` less than total pre-dispatch weight of all messages. Now we'll be trying to dispatch messages until `dispatch_weight` depletes.

### DIFF
--- a/modules/messages/src/lib.rs
+++ b/modules/messages/src/lib.rs
@@ -186,8 +186,6 @@ decl_error! {
 		TooManyMessagesInTheProof,
 		/// Invalid messages has been submitted.
 		InvalidMessagesProof,
-		/// Invalid messages dispatch weight has been declared by the relayer.
-		InvalidMessagesDispatchWeight,
 		/// Invalid messages delivery proof has been submitted.
 		InvalidMessagesDeliveryProof,
 		/// The bridged chain has invalid `UnrewardedRelayers` in its storage (fatal for the lane).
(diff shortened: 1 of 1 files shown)


tomusdrw (approved):
Why not fail the dispatch like previously? Is it because we would need to revert all transactions, right?

I feel this is reasonable, wondering if there are no consequences of relayers producing a larger-than-required proof, but I guess they need to pay for that anyway.

I wonder if similarly we could skip already existing messages if they were already dispatched. That way the relayer (or 3 different relayers) could produce 3 transactions containing the same set of messages (say `N..N + 30`), but declare `dispatch_weight` for ~ 10 of them. These transactions could then all be included, ideally covering all 30 transactions in the set.

Basically the relayers could opt-in to pay a little bit higher fee (for potentially redundant data), but at least get some rewards in case there is a lot of competitors. wdyt, @svyatonik?

svyatonik:
> Why not fail the dispatch like previously? Is it because we would need to revert all transactions, right?

Yes - we have already dispatched some calls and we can't simply returning `Err(_)` is agains Substrate pallet patterns. We could open nested transaction, but I haven't seen anything wrong in switching this behavior.

> I feel this is reasonable, wondering if there are no consequences of relayers producing a larger-than-required proof, but I guess they need to pay for that anyway.

Yes, they're paying for every proof byte anyway + we deal only with message dispatch weight here.

> I wonder if similarly we could skip already existing messages if they were already dispatched. That way the relayer (or 3 different relayers) could produce 3 transactions containing the same set of messages (say N..N + 30), but declare dispatch_weight for ~ 10 of them. These transactions could then all be included, ideally covering all 30 transactions in the set.

Already existing messages are skipped - `receive_message(duplicate_message)` returns `InvalidNonce` in this case and then `unspent_weight` is equal to `dispatch_weight` (https://github.com/paritytech/parity-bridges-common/blob/530f37a23be8744c96d2080240b51bf5763c45d3/modules/messages/src/lib.rs#L581). 

> Basically the relayers could opt-in to pay a little bit higher fee (for potentially redundant data), but at least get some rewards in case there is a lot of competitors. wdyt, @svyatonik?

This is possible, but I don't think that we should treat this as a solution for concurrent relayers problem.

PR #1073, opened 2021-07-27
Emit mortal transactions from relay

### LINKED ISSUE #444: Use Mortal transactions for the relayer
Currently the relayer uses immortal transactions, while there is nothing super bad in this, it's would generally be safer if we used mortal transactions. This prevents replace attack in case the relayer call goes below Existential Deposit and is wiped out, but then re-used for some other purposes.

svyatonik:
closes #444

WIP. Our relay loops now have 'stall timeout', which means that if we have submitted tx, but see no state update for `stall_timeout`, then we consider this transaction lost and submit next (probably the same) transaction. The idea of this PR is to introduce transactions mortality, which would also mean that even if our previous transaction is not actually lost, it'll became spoiled => relayer won't lose his funds.

This PR will only introduce mortality for finality relay transactions. If it'll be accepted, other loops will be updated in follow-up PRs.


tomusdrw (approved):
LGTM!

> This PR will only introduce mortality for finality relay transactions and specifically for Westend -> Millau relay (mortality = 4 blocks). If it'll be accepted, other loops will be updated in follow-up PRs.

Let's not forget about this, perhaps an issue?

PR #1079, opened 2021-07-30
Change ChainSpec::from_genesis for Rialto and Millau chains to reflect the chain names.

### LINKED ISSUE #1078: Change ChainSpec::from_genesis for Rialto and Millau chains to reflect the chain names.
At the moment Rialto and Millau are being created with "Local Testnet" name and this should be changed to retrieve the corresponding names as Rococo or Polkadot.

This change is needed so the bridge UI can get the name and get the bridgesId dynamically.
https://github.com/paritytech/parity-bridges-ui/blob/master/src/util/getConfigs.ts#L32

hbulgarini:
Closes #1078


bin/millau/node/src/chain_spec.rs
```diff
@@ -81,8 +81,8 @@ impl Alternative {
 		);
 		match self {
 			Alternative::Development => ChainSpec::from_genesis(
-				"Development",
-				"dev",
+				"Millau",
+				"millau",
```
hbulgarini, line 85: I figured out that i need to change the names as well for Development otherwise UI would not work with development nodes. In case this is wrong i would just revert this change.

hbulgarini:
cc @tomusdrw

svyatonik (approved):
Probably "Millaur Deverlopment"/"Millau_dev" for development chainspec?

bin/millau/node/src/chain_spec.rs
adoerr: I agree with @svyatonik, there should be some way to distinguish the development chain spec from the local test net one. Something like;
- `Millau Development` / `millau_dev`

and

- `Millau Local` / `millau_local`

adoerr (approved):

hbulgarini:
@svyatonik @adoerr , names changed as suggested. Thanks!

adoerr (approved):

Issue #1055, opened 2021-07-15
Migrate the common files in the substrate-relay to the new crate


fewensa:
Hi:  I think we can migrate the common files to a new crate

includes:

- [finality_pipeline.rs](https://github.com/paritytech/parity-bridges-common/blob/1734d005175843dd3860e491b7bf63c5d5bb9221/relays/bin-substrate/src/finality_pipeline.rs)
- [finality_target.rs](https://github.com/paritytech/parity-bridges-common/blob/1734d005175843dd3860e491b7bf63c5d5bb9221/relays/bin-substrate/src/finality_target.rs)
- [headers_initialize.rs](https://github.com/paritytech/parity-bridges-common/blob/1734d005175843dd3860e491b7bf63c5d5bb9221/relays/bin-substrate/src/headers_initialize.rs)
- [messages_lane.rs](https://github.com/paritytech/parity-bridges-common/blob/1734d005175843dd3860e491b7bf63c5d5bb9221/relays/bin-substrate/src/messages_lane.rs)
- [messages_source.rs](https://github.com/paritytech/parity-bridges-common/blob/1734d005175843dd3860e491b7bf63c5d5bb9221/relays/bin-substrate/src/messages_source.rs)
- [messages_target.rs](https://github.com/paritytech/parity-bridges-common/blob/1734d005175843dd3860e491b7bf63c5d5bb9221/relays/bin-substrate/src/messages_target.rs)
- [on_demand_headers.rs](https://github.com/paritytech/parity-bridges-common/blob/1734d005175843dd3860e491b7bf63c5d5bb9221/relays/bin-substrate/src/on_demand_headers.rs)

Because this will make it easier for bridge developers to reference these files instead of copying these files to other projects. I can modify it if this is feasible

tomusdrw:
Makes sense to me. I can also see `bin-substrate` becoming more generic to cater for use cases you might have. Would that be possible?

fewensa:
I will do this and create a pr

PR #1082, opened 2021-08-03
Common crate for substrate-relay

fewensa:
issue: https://github.com/paritytech/parity-bridges-common/issues/1055


relays/lib-substrate-relay/src/help.rs
```diff
@@ -0,0 +1,23 @@
+use relay_utils::metrics::{FloatJsonValueMetric, PrometheusError, Registry};
```
svyatonik, line 1: This file is missing license prelude.

svyatonik: Also - when I see a file named `help.rs`, I'm thinking that it is related to help subsystem. WDYT about renaming it to `metrics.rs` or `helpers.rs`?

svyatonik (commented):
Thank you! Have you considered moving it to `relay_substrate_client` instead of creating a new crate, though? Do you see any benefits of this new crate?

(I'll try to review this PR in details on thursday)

relays/lib-substrate-relay/src/help.rs
fewensa: Yes

fewensa:
I think the `relay_substrate_client ` is a tool for substrate client, irrelevant to relay. since they were all placed in bin-substrate before, I did not find a suitable library to place these files during the split, so I created a new one.
The biggest benefit should be more intuitive.

svyatonik (approved):

PR #1085, opened 2021-08-05
Bridge/This Chain Ids should be exposed as constants on pallet level.

### LINKED ISSUE #1077: Bridge/This Chain Ids should be exposed as constants on pallet level.
Currently the UI has no way to figure out what `CHAIN_ID` particular chain uses ([this stuff](https://github.com/TomaszWaszczyk/parity-bridges-common/blob/fae68c22646d9438aa5f2e680830715f78b45e40/primitives/runtime/src/lib.rs#L41)).

Chain Id is needed to construct the right payload and/or derive the companion account.

If the value was exposed as pallet constant ([like this](https://github.com/TomaszWaszczyk/parity-bridges-common/blob/fae68c22646d9438aa5f2e680830715f78b45e40/modules/grandpa/src/lib.rs#L97-L98)) it could be easily read by the UI.

We should consider annotating other constants in the messaging pallet.

hbulgarini:
Closes #1077

As the GH issue described this PR is moving the `CHAIN_ID` for the bridged chain from the message pallet and not from the `chain_spec`.


bin/millau/node/src/chain_spec.rs
```diff
@@ -70,10 +70,7 @@ impl Alternative {
 		let properties = Some(
 			serde_json::json!({
 				"tokenDecimals": 9,
-				"tokenSymbol": "MLAU",
-				"bridgeIds": {
-					"Rialto": bp_runtime::RIALTO_CHAIN_ID,
-				}
```
hbulgarini, line 76: I think this is not needed anymore since my understanding is that it was only used for the UI.

hbulgarini:
CC: @tomusdrw  @apopiak  @riusricardo

modules/messages/src/lib.rs
```diff
@@ -255,6 +258,9 @@ decl_module! {
 		/// Deposit one of this module's events by using the default implementation.
 		fn deposit_event() = default;
 
+		/// Gets the chain id value from the instance.
+		const bridged_id: ChainId = T::BridgedChainId::get();
```
svyatonik, line 262: I could see this name (`bridged_id`) in the metadata. Is that a common practice to use snake_case for identifying things in metadata? IIUC it is better to change it to `BridgedChainId` (perfectly fine to use `BridgedChainId` both for `Get`ter and this const).

svyatonik (approved):
LGTM, but you may consider renaming `bridged_id` to `BridgedChainId`. Also - please fix tests - you can use any id for that (in `mock.rs`). Thanks!

riusricardo (approved):
LGTM :+1:

modules/messages/src/lib.rs
hbulgarini: as the polkadotjs api transforms automatically from snake_case to camel_case i thought using snake_case was correct for this. I'm afraid that i don't have enough expertise to determine if this is a common practice or not. I prefer following your recommendation and switch to `BridgedChainId `

modules/messages/src/lib.rs
```diff
@@ -255,6 +258,9 @@ decl_module! {
 		/// Deposit one of this module's events by using the default implementation.
 		fn deposit_event() = default;
 
+		/// Gets the chain id value from the instance.
+		const BridgedChainId: ChainId = T::BridgedChainId::get();
```
hbulgarini, line 262: cc @svyatonik

modules/messages/src/mock.rs
```diff
@@ -142,6 +142,7 @@ parameter_types! {
 	pub const MaxUnrewardedRelayerEntriesAtInboundLane: u64 = 16;
 	pub const MaxUnconfirmedMessagesAtInboundLane: u64 = 32;
 	pub storage TokenConversionRate: FixedU128 = 1.into();
+  pub const TestBridgedChainId: bp_runtime::ChainId = *b"test";
```
hbulgarini, line 145: cc @svyatonik

Issue #1081, opened 2021-08-02
Upgrade Pallets to Frame V2


ferrell-code:
Extension of [polkadot#2882](https://github.com/paritytech/polkadot/issues/2882)

There are a number of custom Bridges pallets in this repo used in the polkadot repo and elsewhere. It is ideal to eventually have all active pallets migrated to the newest macro.

pallets using frame v1 in modules:
-  [x] dispatch
-  [x] messages
-  [x] shift-session-manager
-  [x] ethereum
-  [x] currency-exchange

svyatonik:
all pallets are using v2 now

PR #1088, opened 2021-08-10
[BREAKING] Migrate messages pallet to frame v2

svyatonik:
related #1081 

This migration (if it'll happen after Rococo/Wococo deployment) would require storage migrations in next runtime upgrade in these chains. Also - when this PR will be merged, the relay will became incompatible with Rococo/Wococo.

**TO AVOID STORAGE MIGRATIONS**

Make sure that your pallet is named `BridgeMessages` in `construct_runtime!()` as in:
```rust
construct_runtime!(
	pub enum Runtime where {
		...
		BridgeMessages: pallet_bridge_messages::{Pallet, Call, Storage, Event<T>},
		...
	}
);
```

If your pallet is named in other way, you can either: (1) rename it with upgrade (2) leave your name as-is, but then you'll need to migrate all storage entries of the pallet.

Please also note that pallet v1 can't "talk" to pallet v2. So if you'll have v1 on chain1 and v2 on chain2, then you need to coordinate both runtime upgrades.


svyatonik:
Also - there are some differences in metadata - I think they are not breaking, but;
1) in `ParameterUpdated` event arguments have changed from `arguments: ["Parameter"]` to `arguments: ["<T as Config<I>>::Parameter"]` (cosmetic?);
2) in `Dummy` event arguments have changed from `arguments: ["PhantomData<(AccountId, I)>"]` to `arguments: ["PhantomData<(<T as frame_system::Config>::AccountId` (cosmetic?);
3) not sure why, but old Rialto runtime has been returning empty errors metadata for messages pallet and updated returns all errors. Don't know why, but it shouldn't be a breaking change iiuc.

modules/messages/src/lib.rs
```diff
@@ -82,241 +75,177 @@ pub mod benchmarking;
...
-);
+	#[pallet::config]
+	pub trait Config<I: 'static = ()>: frame_system::Config {
+		// General types
 
-decl_module! {
-	pub struct Module<T: Config<I>, I: Instance = DefaultInstance> for enum Call where origin: T::Origin {
-		/// Deposit one of this module's events by using the default implementation.
-		fn deposit_event() = default;
+		/// They are overarching event type.
```
tomusdrw, line 90:
```suggestion
		/// The are overarching event type.
```

relays/bin-substrate/src/chains/wococo_messages_to_rococo.rs
```diff
@@ -226,13 +216,15 @@ pub async fn run(
...
 			ROCOCO_CHAIN_ID,
+			bp_wococo::WITH_ROCOCO_MESSAGES_PALLET_NAME,
 			params.target_to_source_headers_relay,
 		),
 		RococoTargetClient::new(
 			params.target_client,
 			lane,
 			lane_id,
 			WOCOCO_CHAIN_ID,
+			bp_rococo::WITH_WOCOCO_MESSAGES_PALLET_NAME,
```
tomusdrw, line 227: Couldn't all the constant parameters be part of a single trait, similarly to how this works in the runtime?

relays/lib-substrate-relay/src/messages_source.rs
```diff
@@ -42,63 +42,63 @@ use relay_substrate_client::{Chain, Client, Error as SubstrateError, HashOf, Hea
...
 pub type SubstrateMessagesProof<C> = (Weight, FromBridgedChainMessagesProof<HashOf<C>>);
 
 /// Substrate client as Substrate messages source.
-pub struct SubstrateMessagesSource<SC: Chain, TC: Chain, P: SubstrateMessageLane, I> {
+pub struct SubstrateMessagesSource<SC: Chain, TC: Chain, P: SubstrateMessageLane> {
 	client: Client<SC>,
 	lane: P,
 	lane_id: LaneId,
 	instance: ChainId,
+	messages_pallet_name: &'static str,
```
tomusdrw, line 58: Any reason to restrict this to a `'static str`? Maybe let's store `String` here?

tomusdrw (approved):

PR #1102, opened 2021-09-01
Remove unused deps

gilescope:
Some deps that did not appear to be used any longer.

Please shout out if you want to keep them!


svyatonik:
I don't think we need unused deps! :) Can you, please, merge master, to avoid CI errors? Thanks!

fuzz/storage-proof/README.md
```diff
@@ -6,7 +6,10 @@ Install dependencies:
 ```
 $ sudo apt install build-essential binutils-dev libunwind-dev
 ```
-
+or on nix:
```
tomusdrw, line 9: I see what you did here :)

tomusdrw (approved):
lgtm if the CI is happy :)

svyatonik:
Could you, please, just merge master instead of last two commits? :) Missing docs are already there plus it'll solve conflicts :) Thanks!

fuzz/storage-proof/README.md
gilescope: (I lost some time trying to install binutils, libunwind-dev etc. on nix not realising that I could just install honggfuzz.)

gilescope:
sorry - will do.

gilescope:
(I wouldn't normally force push but in this case it seemed the right call).

gilescope:
btw clippy ICEs on my machine when I run it against this repo. Have raise it to them.

svyatonik:
> btw clippy ICEs on my machine when I run it against this repo. Have raise it to them.

Could be https://github.com/rust-lang/rust-clippy/issues/7423 - I've had this a while ago. Now (did a `rustup update` yesterday) everything is fine.

Thanks!

Issue #1033, opened 2021-06-28
Refactor `SubstrateMessagesSource` and `SubstrateMessagesTarget` to avoid duplicate generic params


svyatonik:
https://github.com/paritytech/parity-bridges-common/pull/1015/files#r656095306

After #1015 `SubstrateMessagesSource` and `SubstrateMessagesTarget` has (apart from others) `SourceChain`, `TargetChain` and `MessageLane` parameters. `MessageLane` already has `SourceChain` and `TargetChain` associated types, so it makes sense to use these types instead of additional generic params. Directly using associated types is impossible, because of circular bounds requirements, so some other solution is required.

svyatonik:
closed by #1105

PR #1105, opened 2021-09-01
Refactor substrate messages source and substrate messages target

antonio-dropulic:
related to #1033

### DIFF
--- a/primitives/runtime/src/chain.rs
+++ b/primitives/runtime/src/chain.rs
@@ -91,6 +91,3 @@ pub type HeaderOf<C> = <C as Chain>::Header;
 
 /// Transaction era used by the chain.
 pub type TransactionEraOf<C> = crate::TransactionEra<BlockNumberOf<C>, HashOf<C>>;
-
-/// Number type used in the Header type used by the chain
-pub type HeaderNumberOf<C> = <<C as Chain>::Header as HeaderT>::Number;
--- a/primitives/runtime/src/chain.rs
+++ b/primitives/runtime/src/chain.rs
@@ -91,3 +91,6 @@ pub type HeaderOf<C> = <C as Chain>::Header;
 
 /// Transaction era used by the chain.
 pub type TransactionEraOf<C> = crate::TransactionEra<BlockNumberOf<C>, HashOf<C>>;
+
+/// Number type used in the Header type used by the chain
+pub type HeaderNumberOf<C> = <<C as Chain>::Header as HeaderT>::Number;
--- a/relays/lib-substrate-relay/src/messages_source.rs
+++ b/relays/lib-substrate-relay/src/messages_source.rs
@@ -25,7 +25,7 @@ use crate::on_demand_headers::OnDemandHeadersRelay;
 use async_trait::async_trait;
 use bp_messages::{LaneId, MessageNonce, UnrewardedRelayersState};
 use bp_runtime::messages::DispatchFeePayment;
-use bp_runtime::Chain as BaseChain;
+use bp_runtime::{BlockNumberOf as ChainBaseBlockNumberOf, HashOf as ChainBaseHashOf, HeaderOf as ChainBaseHeaderOf};
 use bridge_runtime_common::messages::{
 	source::FromBridgedChainMessagesDeliveryProof, target::FromBridgedChainMessagesProof,
 };
@@ -39,7 +39,9 @@ use messages_relay::{
 	},
 };
 use num_traits::{Bounded, Zero};
-use relay_substrate_client::{Chain, Client, Error as SubstrateError, HashOf, HeaderIdOf};
+use relay_substrate_client::{
+	Chain, ChainBalanceOf, ChainIndexOf, Client, Error as SubstrateError, HashOf, HeaderIdOf,
+};
 use relay_utils::{relay_loop::Client as RelayClient, BlockNumberBase, HeaderId};
 use sp_core::Bytes;
 use sp_runtime::{traits::Header as HeaderT, DeserializeOwned};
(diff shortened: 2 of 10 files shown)


relays/lib-substrate-relay/src/messages_target.rs
```diff
@@ -102,27 +98,28 @@ where
...
-	TC::BlockNumber: Copy,
-	TC::Header: DeserializeOwned,
-	TC::Index: DeserializeOwned,
-	<TC::Header as HeaderT>::Number: BlockNumberBase,
-	P: SubstrateMessageLane<SourceChain = SC, TargetChain = TC>,
+	<P::TargetChain as Chain>::Index: DeserializeOwned,
+	<P::TargetChain as BaseChain>::Hash: Copy,
+	<P::TargetChain as BaseChain>::BlockNumber: Copy,
+	<P::TargetChain as BaseChain>::Header: DeserializeOwned,
+	<<P::TargetChain as BaseChain>::Header as HeaderT>::Number: BlockNumberBase,
```
tomusdrw, line 118: To avoid repetition, maybe we could have type aliases that extract the right associated types? I.e.
```suggestion
	ChainIndexOf<P::TargetChain>: DeserializeOwned,
	BaseChainHashOf<P::TargetChain>: Copy,
	BaseChainBlockNumberOf<P::TargetChain>: Copy,
	BaseChainHeaderOf<P::TargetChain>: DeserializeOwned,
	BaseChainHeaderNumberOf<P::TargetChain>: BlockNumberBase,
```

tomusdrw (commented):
Nice, looks good to me!

tomusdrw:
@svyatonik could you also take a look?

relays/lib-substrate-relay/src/messages_target.rs
antonio-dropulic: Not sure if we can do this in stable rust. Here is a minimal example https://play.rust-lang.org/ to illustrate the problem

tomusdrw: The ```
type ChainIndexOf<T> = <T as Chain>::Index;
```
should work fine, we don't need the extra bounds.

Comment (antonio-dropulic) [reply relays/lib-substrate-relay/src/messages_target.rs:118]: It does work just fine. Still, I'm not a fan of replacing fully specified type names with type aliases. As someone just seeing the code for the first time it's just another layer of indirection. I have to learn to parse the type alias as rust. And I have to learn type names of the types in question since the naming can be ambiguous when one or more of the types have a multi word name. Also if the code base is to be uniform this adds a convention, for the sake of cca. 4 characters per replacement.

File: primitives/runtime/src/chain.rs [hunk 2]
```diff
@@ -85,3 +85,6 @@ pub type HasherOf<C> = <C as Chain>::Hasher;
 
 /// Header type used by the chain.
 pub type HeaderOf<C> = <C as Chain>::Header;
+
+/// Number type used in the Header type used by the chain
+pub type HeaderNumberOf<C> = <<C as Chain>::Header as HeaderT>::Number;
```
svyatonik, line 90: You could actually use `BlockNumerOf<>` instead of this, because `Chain::Header` is defined as `HeaderT<Number = Self::BlockNumber>` :)

relays/lib-substrate-relay/src/messages_source.rs
```diff
@@ -25,6 +25,7 @@ use crate::on_demand_headers::OnDemandHeadersRelay;
 use async_trait::async_trait;
 use bp_messages::{LaneId, MessageNonce, UnrewardedRelayersState};
 use bp_runtime::messages::DispatchFeePayment;
+use bp_runtime::{BlockNumberOf as ChainBaseBlockNumberOf, HashOf as ChainBaseHashOf, HeaderOf as ChainBaseHeaderOf};
```
svyatonik, line 28: IMO no need for `*Base*` here - `BlockNumberOf`, `HashOf` and `HeaderOf` should be enough.

relays/lib-substrate-relay/src/messages_source.rs
```diff
@@ -264,10 +262,10 @@ where
 			.estimate_extrinsic_fee(self.lane.make_messages_receiving_proof_transaction(
 				Zero::zero(),
 				HeaderId(Default::default(), Default::default()),
-				prepare_dummy_messages_delivery_proof::<SC, TC>(),
+				prepare_dummy_messages_delivery_proof::<P::SourceChain, P::TargetChain>(),
 			))
 			.await
-			.unwrap_or_else(|_| SC::Balance::max_value())
+			.unwrap_or_else(|_| <P::SourceChain as Chain>::Balance::max_value())
```
svyatonik, line 268:
```suggestion
			.unwrap_or_else(|_| BalanceOf::<P::SourceChain>::max_value())
```

relays/lib-substrate-relay/src/messages_target.rs
```diff
@@ -102,27 +101,28 @@ where
...
-	SC: Chain<
+	P: SubstrateMessageLane,
+	P::SourceChain: Chain<
 		Hash = <P::MessageLane as MessageLane>::SourceHeaderHash,
 		BlockNumber = <P::MessageLane as MessageLane>::SourceHeaderNumber,
 		Balance = <P::MessageLane as MessageLane>::SourceChainBalance,
 	>,
-	SC::Balance: TryFrom<TC::Balance> + Bounded,
-	TC: Chain<
+	<P::SourceChain as Chain>::Balance: TryFrom<<P::TargetChain as Chain>::Balance> + Bounded,
```
svyatonik, line 112:
```suggestion
	BalanceOf<P::SourceChain>: TryFrom<BalanceOf<P::TargetChain>> + Bounded,
```

svyatonik (approved):
Looks much better than before, thanks! :)

PR #1108, opened 2021-09-02
Rename param and update comment

AsceticBear:
In my understanding, `latest_received_nonce` is maintained in the `OutBoundLaneData` and `latest_delivered_nonce` is maintained in the `InBoundLaneData`.

https://github.com/paritytech/parity-bridges-common/blob/master/modules/messages/src/lib.rs#L608-L610L, Maybe use `latest_delivered_nonce` would be more accurate.


modules/messages/src/inbound_lane.rs
```diff
@@ -136,9 +136,7 @@ impl<S: InboundLaneStorage> InboundLane<S> {
 			return ReceivalResult::TooManyUnconfirmedMessages;
 		}
 
-		// dispatch message before updating anything in the storage. If dispatch would panic,
-		// (which should not happen in the runtime) then we simply won't consider message as
-		// delivered (no changes to the inbound lane storage have been made).
+		// then, dispatch message using pallet-bridge-dispatch
```
svyatonik, line 139: Messages pallet may work with any kind of dispatch mechanisms. E.g. in future we may switch to XCM (#214). So I'd prefer to keep comment more generic :)

svyatonik (approved):
> In my understanding, latest_received_nonce is maintained in the OutBoundLaneData and latest_delivered_nonce is maintained in the InBoundLaneData.

But in the end `latest_delivered_nonce` turns into `latest_received_nonce`. So probably worth renaming `OutboundLaneData::latest_received_nonce` to `OutboundLaneData::latest_delivered_nonce` as well, if it makes code more clear? That said, this PR LGTM, so I won't block it from merging.

Thanks for helping with that! :)

modules/messages/src/inbound_lane.rs
AsceticBear: That makes sense. Updated

PR #1109, opened 2021-09-03
Fix delivery transaction estimation used by rational relayer

svyatonik:
There are two fixes in this PR:
1) easy https://github.com/paritytech/parity-bridges-common/commit/960a918f1a010c74fd97f364a3259134ef543922 ;
2) complex https://github.com/paritytech/parity-bridges-common/commit/bca861cd86a9f9a74d330e4f675aef8f2cd1930e .

Re 2nd fix - after `PayDispatchFeeAtTargetChain` delivery tx includes additional  `Currency::transfer` cost, because message dispatch **may be** paid at the target chain. But when message is parsed and we see that dispatch has been paid at the source chain, we refund relayer with this cost. Last PR from greedy-relayer PRs set (#1005) has been created before this functionality has been added, so this wasn't well-supported by rational relayers.

So the fix seems simple - we know pre-dispatch cost of delivery tx (`payment_queryFeeDetails` RPC is called for that). And then we just need to subtract the cost that is refunded by messages module from this pre-dispatch cost. But the problem is that we only know weight of this payment. And the only way to convert weight -> fee is manual computation (unless I'm again missing something obvious).

So to convert weight to fee, we need to know: (a) `WeightToFee` of the target chain and (b) fee multiplier to compute adjusted weight fee. We already have data for (a) - so no problem here. There are two ways to get multiplier - one is read directly from the storage. We need to know storage key for that => another hardcoded constant, which I want to avoid. The second option only works with linear `WeightToFee` - we may compute `adjusted_weight_fee` for tx1 with dispatch weight `A` and then for tx2 with dispatch weight `A+100`. Then we may compute multiplier as `(tx2.adjusted_weight_fee - tx1.adjusted_weight_fee) / (WeightToFee(A+100) - WeightToFee(A))`. I've implemented this option, but actually it seems too hacky, so I could get back to hardcoded storage access if someone will have the same concern.


primitives/chain-rococo/src/lib.rs
```diff
@@ -97,6 +97,13 @@ pub const FROM_ROCOCO_LATEST_CONFIRMED_NONCE_METHOD: &str = "FromRococoInboundLa
 /// Name of the `FromRococoInboundLaneApi::unrewarded_relayers_state` runtime method.
 pub const FROM_ROCOCO_UNREWARDED_RELAYERS_STATE: &str = "FromRococoInboundLaneApi_unrewarded_relayers_state";
 
+/// Weight of pay-dispatch-fee operation for inbound messages at Rococo chain.
+///
+/// This value corresponds to the result of `pallet_bridge_messages::WeightInfoExt::pay_inbound_dispatch_fee_overhead()`
+/// call for your chain. Don't put too much reserve there, because it is used to **decrease**
+/// `DEFAULT_MESSAGE_DELIVERY_TX_WEIGHT` cost. So putting large reserve would make delivery transactions cheaper.
+pub const PAY_INBOUND_DISPATCH_FEE_WEIGHT: Weight = 600_000_000; // TODO: needs to be updated with real value
```
tomusdrw, line 105: What does the "real value" mean?

relays/lib-substrate-relay/src/messages_target.rs
```diff
@@ -258,27 +258,88 @@ where
...
+		let inclusion_fee_in_target_tokens = delivery_tx_fee.inclusion_fee();
+
+		// The pre-dispatch cost of delivery transaction includes additional fee to cover dispatch fee payment
+		// (Currency::transfer in regular deployment). But if message dispatch has already been paid
+		// at the Source chain, the delivery transaction will refund relayer with this additional cost.
+		// But `estimate_extrinsic_fee` obviously just returns pre-dispatch cost of the transaction. So
+		// if transaction delivers prepaid message, then it may happen that pre-dispatch cost is larger
+		// than reward and `Rational` relayer will refuse to deliver this message.
+		//
+		// The most ovious solution would be to deduct total weight of dispatch fee payments from the
```
tomusdrw, line 279:
```suggestion
		// The most obvious solution would be to deduct total weight of dispatch fee payments from the
```

relays/lib-substrate-relay/src/messages_target.rs
```diff
@@ -316,17 +377,121 @@ where
...
+fn compute_fee_multiplier<C: Chain>(
+	smaller_adjusted_weight_fee: BalanceOf<C>,
+	smaller_tx_weight: Weight,
+	larger_adjusted_weight_fee: BalanceOf<C>,
+	larger_tx_weight: Weight,
+) -> FixedU128 {
+	let adjusted_weight_fee_difference = larger_adjusted_weight_fee.saturating_sub(smaller_adjusted_weight_fee);
+	let smaller_tx_unadjusted_weight_fee = WeightToFeeOf::<C>::calc(&smaller_tx_weight);
+	let larger_tx_unadjusted_weight_fee = WeightToFeeOf::<C>::calc(&larger_tx_weight);
+	println!(
```
tomusdrw, line 398: `println!`?

relays/lib-substrate-relay/src/messages_target.rs
```diff
@@ -258,27 +258,88 @@ where
...
+				.estimate_extrinsic_fee(self.lane.make_messages_delivery_transaction(
+					Zero::zero(),
+					HeaderId(Default::default(), Default::default()),
+					nonces.clone(),
+					prepare_dummy_messages_proof::<P::SourceChain>(nonces.clone(), larger_dispatch_weight, total_size),
+				))
+				.await?;
+
+			compute_prepaid_messages_refund::<P>(
+				total_prepaid_nonces,
```
tomusdrw, line 302: Do I get this right, that we take into account cumulative reward from all pre-paid transactions here? Can you elaborate a little bit more on this? For instance a more concrete example would help.

primitives/chain-rococo/src/lib.rs
svyatonik: I mean value that we'll get on benchmarking machine for Rococo. But actually we're not interested in weights on Rococo, iiuc. So Rialto/Millau value should be fine her

relays/lib-substrate-relay/src/messages_target.rs
svyatonik: Delivery transaction contains some messages (a number of elements in the `nonces` range). The cost of dispatch of some of these messages is already paid `AtSourceChain` (number of such messages is what `total_prepaid_nonces` means). Dispatch of other messages is paid `AtTargetChain`. For the former group, delivery tx will deduct a weight of `Currency::transfer` call (payment for dispatch) from its pre-dispatch weight. This sum (we know the weight, but need to compute cost) is computed here (`expected_refund_in_target_tokens`).

Then (a bit above) there's also a regular call to the `estimate_extrinsic_fee`, which just estimates a cost of delivery transaction without taking into account any refunds (result is `inclusion_fee_in_target_tokens`). So final cost will be `inclusion_fee_in_target_tokens - expected_refund_in_target_tokens`.

tomusdrw (approved):

PR #1111, opened 2021-09-04
Remove fund account checks from upgrade.

tomusdrw:
Reverts some stuff from: https://github.com/paritytech/parity-bridges-common/pull/624

To prevent bricking the chain after upgrade, the check for relayer account is removed from `RuntimeUpgrade` hook and instead it's now checked on every payment.

While this might seem like unnecessary cost, it should be cheap (the account will be in cache) and instead of stalling the chain completely, it's possible to simply fix the incorrect state by topping up the relayer account.

### DIFF
--- a/modules/messages/src/instant_payments.rs
+++ b/modules/messages/src/instant_payments.rs
@@ -57,20 +57,15 @@ where
 {
 	type Error = &'static str;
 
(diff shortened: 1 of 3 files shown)


svyatonik (approved):

AurevoirXavier (approved):
Our testnet down due to this assert.

PR #1129, opened 2021-09-09
Add --mandatory-headers-only cli option to complex relay

### LINKED ISSUE #1104: Select GRANDPA relayer strategy for Kusama <> Polkadot bridge
We know that the cost of importing Polkadot/Kusama header will be high. And we don't want to run relay in its default mode where all source headers are synced to the target chain. On the other hand, mandatory-headers strategy is also not perfect - then message sender will need to wait for the whole session (4 hours for Polkadot -> Kusama messages and 1 hour for Kusama -> Polkadot messages + another interval for confirmations). So probably there's another solution?

Solutions I see now:
1) use default relayer mode (too expensive);
2) use mandatory-only relayer mode (too slow);
3) add cost of single header sync to the message fee - too expensive for message sender. Probably refund if there are several message senders (we need to define an interval between delivery tx is submitted);
4) split cost of headers sync across all message submitters (then we need some minimal number of messages that are required for delivery tx to be submitted?);
5) ...

svyatonik:
closes #1104 

I don't know of any way to test this at our deployments, because this option affects both directions of the bridge (and I think it is better to keep it this way for simplicity). So enabling it on Rialto<>Millau bridge, will lead to 10-minutes delays of Millau -> Rialto messages delivery => alerts. So - only manual testing here (did it locally). If we'll ever have mock clients, we could have better tests for that.

### DIFF
--- a/relays/bin-substrate/src/cli/relay_headers_and_messages.rs
+++ b/relays/bin-substrate/src/cli/relay_headers_and_messages.rs
@@ -57,6 +57,9 @@ pub struct HeadersAndMessagesSharedParams {
 	lane: Vec<HexLaneId>,
 	#[structopt(long, possible_values = RelayerMode::VARIANTS, case_insensitive = true, default_value = "rational")]
 	relayer_mode: RelayerMode,
+	/// If passed, only mandatory headers (headers that are changing the GRANDPA authorities set) are relayed.
+	#[structopt(long)]
+	only_mandatory_headers: bool,
 	#[structopt(flatten)]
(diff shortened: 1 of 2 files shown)


tomusdrw (approved):
Can we enable this for the one-way Westend bridge maybe?

svyatonik:
> Can we enable this for the one-way Westend bridge maybe?

Westend -> Millau bridge isn't using complex relay, because it has no messaging support. So we can't test it there :/

PR #1134, opened 2021-09-13
Add `OnMessageAccepted` callback

### LINKED ISSUE #1133: Add `OnMessageAccepted` callback function when message accepted
We can add a `OnMessageAccepted` call back function in `send_message` dispatch call,  just like `OnDeliveryConfirmed` in `receive_messages_delivery_proof`. Sometimes, other pallets want to trace the message that has been accepted and do something else.

AsceticBear:
Fix #1133  

~~There is an issue about how to cal pre-dispatch send_message weight after the call_back function is included. Now, I did not find a suitable way to add call_back weight to the pre-dispatch weight calculation, since we do not know how many call-back functions are configured in runtime and how much weight the call-back function consumes.  I have tried to strict the max consume weight of each call back function up to `single_message_callback_overhead`, then the call back function can not perform some more complex operations, but this seems a little trick.~~

~~How about set the max pre-dispatch weight of the call_back function to `n * single_message_callback_overhead`, then return some funds according to the actual weight return by the call back function.~~


svyatonik:
By `n` you mean number of callbacks (int the tuple), right? But then you need to define `n` somewhere. Can we say that the message may only be processed by a single callback => so  assume`n = 1`? This, however, has limited usage because of restrictions.

Alternatively we may add `fn maximal_accepted_callback_weight() -> Weight` method to the `WeightInfoExt` trait. Then refund caller if actual callback has returned lesser weight. This option is harder to use for runtime developer - e.g. you may need to do implement more benchmarks somewhere else to compute maximal callback weight. But it does not restrict maximal callback weight, which is good.

I'm fine with either way.

AsceticBear:
@svyatonik @tomusdrw This pr is ready for reivew

primitives/messages/src/source_chain.rs
```diff
@@ -173,6 +173,18 @@ impl OnDeliveryConfirmed for Tuple {
 	}
 }
 
+/// Handler for messages have been accepted
+pub trait OnMessageAccepted {
+	/// Called when a message has been accepted by message pallet.
+	fn on_messages_accepted(message: &MessageNonce) -> Weight;
```
svyatonik, line 179: Please add `lane: &LaneId` argument here - we support multiple message lanes, so callee should know this information to be able to identify message.

svyatonik (approved):
Looks good, thanks!

modules/messages/src/lib.rs
```diff
@@ -305,6 +308,36 @@ pub mod pallet {
 				payload: encoded_payload,
 				fee: delivery_and_dispatch_fee,
 			});
+			// Guaranteed to be called outside only when the message is accepted.
+			// We assume that the maximum weight call back used is `single_message_callback_overhead`, so do not perform
```
tomusdrw, line 312: This is a bit suboptimal for cases where `OnMessageAccepted` callback is not set - the user needs to pay for more weight than necessary initially. Ideally we would have a separate `WeightInfo` method for that.

modules/messages/src/lib.rs
```diff
@@ -305,6 +308,36 @@ pub mod pallet {
...
+				Some(difference) => {
+					log::trace!(
+						target: "runtime::bridge-messages",
+						"Messages accepted callback has returned unspent weight to refund the submitter: \
+						{} - {} = {}",
+						single_message_callback_overhead,
+						actual_callback_weight,
+						difference,
+					);
+					actual_weight -= difference;
```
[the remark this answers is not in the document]
AsceticBear: > Would be good to add a note why difference < actual_weight?

The log ahead this `actual_weight -= difference;` helps explain the logic here, so maybe no need to add more notes?

I change the old `-=` to a more safer way`saturating_sub`

modules/messages/src/lib.rs
```diff
@@ -305,6 +308,36 @@ pub mod pallet {
...
+				}
+				None => {
+					debug_assert!(false, "The accepted callback is wrong");
+					log::error!(
+						target: "runtime::bridge-messages",
+						"Messages accepted callback has returned more weight that it may spent: \
+						{} vs {}",
+						single_message_callback_overhead,
+						actual_callback_weight,
+					);
```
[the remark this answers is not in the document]
AsceticBear: I just keep this case the same as what `OnDeliveryConfirmed` does. But, I think you are right, it would be better to return with an error here. @svyatonik Is there any side effect here if we exit with an error directly?

modules/messages/src/lib.rs
AsceticBear: Even if we have a separate `weightInfo` method, the user also needs to pre-paid callback weight, right?

modules/messages/src/lib.rs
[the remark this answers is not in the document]
svyatonik: 1) you can't simply fail here, because storage has been modified already. You may wrap the whole code in transaction, though;
2) the only way to reach this line is to execute it in invalid runtime. What's better way to handle invalid runtime - to fail transaction (means that the submitter loses his fee he has paid for the tx + the lane is blocked) or to simply log an error? The similar case is when extrinsic weight formula returns something that literally says `I'll be executing for 1 second`, but actually it is executing for 2 seconds. We don't fail the tx in this case && the only difference is that here we log the error. So imo let's leave it as is

modules/messages/src/lib.rs
svyatonik: If it returns zero, then no :)

svyatonik: (second option here https://github.com/paritytech/parity-bridges-common/pull/1134#issuecomment-918877463 )

modules/messages/src/lib.rs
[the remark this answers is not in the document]
AsceticBear: I recheck the logic in the send_message call, maybe we can not directly return an error in this case, since the new message nonce has been generated and OutboundLaneData has been changed when the callback is invoked. If we directly return an error in this case, then we change lane important storage and return error finally, that broken substrate dispatch call rule。

modules/messages/src/lib.rs
AsceticBear: The second option is better indeed. But I have no idea about how to calculate the maximal accepted call back weight via benchmark in this case?  According to the block total weight?

svyatonik: Yes, this option is a bit harder for runtime developer - you'll need to add benchmarks + you'll need tests to ensure that this `maximal_accepted_callback_weight` always returns weight that is >= than the weight from benchmark. But for the simplest case (where you only allow single db write + single db read per message) you may return `DbWeight::get().reads_writes(1, 1)` from the `maximal_accepted_callback_weight` without any benchmarks. Default implementation (`impl WeightInfoExt for () { ... }` should return zero.

That said, I'm ok with current implementation - both options have their pros and cons. But @tomusdrw has a valid concern here.

AsceticBear: @tomusdrw Please take some time to review again.

modules/messages/src/lib.rs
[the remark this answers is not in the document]
tomusdrw: Well, it's always greater because `send_message_weight` always includes `single_message_callback_overhead`. But yeah, `saturating_sub` is good too.

modules/messages/src/weights_ext.rs
```diff
@@ -341,7 +344,7 @@ pub trait WeightInfoExt: WeightInfo {
 		Self::receive_single_message_proof().saturating_sub(Self::receive_single_prepaid_message_proof())
 	}
 
-	/// Returns pre-dispatch weight of single message delivery callback call.
+	/// Returns pre-dispatch weight of single callback call.
```
tomusdrw, line 347:
```suggestion
	/// Returns pre-dispatch weight of single callback call.
	///
	/// When benchmarking the weight please take into consideration both the `OnMessageAccepted` and `OnDeliveryConfirmed` callbacks. The method should return the greater of the two, because it's used to estimate the weight in both contexts.
```

tomusdrw (approved):
Please just update the docs to make sure we remember that the weight method is used in both contexts, and it looks good to me.

tomusdrw (approved):

PR #1139, opened 2021-09-14
Adding MultiAddress type and ValidationCodeHash

hbulgarini:
This PR is meant to add the corresponding Address type mainly for Rialto and setting `"Address": "AccountId"` for the rest.
As i mentioned before, based on this PR i was only able to see this change on Rialto and not on Millau, Rococo or Wococo.
https://github.com/paritytech/parity-bridges-common/pull/1053#discussion_r672265519

Then the UI started complaining about the missing type `ValidationCodeHash` which again it seems only applicable to Rialto runtime.


deployments/types-rococo.json
```diff
@@ -1,16 +1,17 @@
 {
   "--1": "Rococo Types",
+  "RococoAddress": "AccountId",
```
hbulgarini, line 3: As i removed the `Address` type from `common.js`, my understanding is that Rococo and Wococo contain the same `Address` type as Millau . In case this is wrong please let me know and i will revert the change or adjust it accordingly.

deployments/types/rialto.json
```diff
@@ -10,5 +11,8 @@
 		"_enum": {
 			"RialtoToMillauConversionRate": "u128"
 		}
+	},
+	"ValidationCodeHash": {
+		"Hash": "Hash"
 	}
```
tomusdrw, line 17:
```suggestion
	"ValidationCodeHash": "H256",
```

tomusdrw (approved):

deployments/types/rialto.json
hbulgarini: Addressed on here and generated new `types-rialto.json` with this change.

tomusdrw: This should be exactly as I've suggested. The type does not have "Hash" field, but rather is just a wrapper type around `H256`.

hbulgarini: sorry! i see the difference now. Changing it.

hbulgarini: solved.

hbulgarini:
@tomusdrw : could you please help on this CI job failure: https://gitlab.parity.io/parity/parity-bridges-common/-/jobs/1117845

I think is not related to this change.

PR #1152, opened 2021-09-22
Rename MessageId -> BridgeMessageId

### LINKED ISSUE #1151: Rename `MessageId`?
```
type MessageId = (LaneId, MessageNonce);
```

I found that Cumulus also defines a type named `MessageId`.

https://sourcegraph.com/github.com/polkadot-js/api/-/blob/packages/types/src/interfaces/cumulus/definitions.ts?L13

I have to overwrite it. Otherwise, I got:

```
Unable to retrieve the specified block details. Unable to decode storage system.events:: 
createType(Vec<EventRecord>):: decodeU8a: failed at 0x000000500a000001… 
on : {"phase":"Phase","event":"Event","topics":"Vec<Hash>"}:: decodeU8a: failed at 0x2c01706167727274… 
on topics: Vec<Hash>:: decodeU8a: failed at 0x0000000000000000… 
on : H256:: Expected at least 32 bytes (256 bits), found 28 bytes
```

tomusdrw:
Closes #1151

### DIFF
--- a/bin/millau/runtime/src/lib.rs
+++ b/bin/millau/runtime/src/lib.rs
@@ -211,7 +211,7 @@ impl pallet_aura::Config for Runtime {
 }
 impl pallet_bridge_dispatch::Config for Runtime {
 	type Event = Event;
-	type MessageId = (bp_messages::LaneId, bp_messages::MessageNonce);
+	type BridgeMessageId = (bp_messages::LaneId, bp_messages::MessageNonce);
 	type Call = Call;
 	type CallFilter = ();
 	type EncodedCall = crate::rialto_messages::FromRialtoEncodedCall;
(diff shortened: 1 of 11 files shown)


svyatonik (approved):
I believe that it solves the problem, but is there some general solution for that? Like if there'll be other bridge (Substrate <> BTC or Substrate <> Eth) in the runtime that will be using `BridgeMessageId`, then what? Or if there's second instance of the dispatch pallet that is using some triplet for identifying messages?

tomusdrw:
I believe this is all going to be sorted out with metadata v15 and `scale-info`. CC @ascjones can you confirm that it is possible to express different associated types for multiple pallet instances?

ascjones:
> I believe this is all going to be sorted out with metadata v15 and `scale-info`. CC @ascjones can you confirm that it is possible to express different associated types for multiple pallet instances?

Yes `scale-info` operates on concrete types rather than only type names.

PR #1153, opened 2021-09-22
Move calculation relayer reward into `MessageDeliveryAndDispatchPayment`

AsceticBear:
It's would be much better to take take the calculation reward code block out to the `MessageDeliveryAndDispatchPayment` implementation, make the reward strategy more generic. There is no logical change in reward implementation.


modules/messages/src/instant_payments.rs
```diff
@@ -85,16 +88,38 @@ where
...
+		received_range: &RangeInclusive<u64>,
 		relayer_fund_account: &T::AccountId,
-	) {
-		pay_relayers_rewards::<Currency, _>(
-			confirmation_relayer,
-			relayers_rewards,
-			relayer_fund_account,
-			GetConfirmationFee::get(),
-		);
+	) -> Result<(), Self::Error> {
```
svyatonik, line 96: Is there any specific reason to return `Result` now? I mean - the outside function would just ignore that error, because storage has already been altered && we don't rollback these changes.

modules/messages/src/mock.rs
```diff
@@ -346,14 +349,30 @@ impl MessageDeliveryAndDispatchPayment<AccountId, TestMessageFee> for TestMessag
...
 		_confirmation_relayer: &AccountId,
-		relayers_rewards: RelayersRewards<AccountId, TestMessageFee>,
+		received_range: &RangeInclusive<u64>,
 		_relayer_fund_account: &AccountId,
-	) {
-		for (relayer, reward) in relayers_rewards {
-			let key = (b":relayer-reward:", relayer, reward.reward).encode();
-			frame_support::storage::unhashed::put(&key, &true);
+	) -> Result<(), Self::Error> {
+		let mut relayers_rewards: RelayersRewards<_, TestMessageFee> = RelayersRewards::new();
```
svyatonik, line 358: It seems like almost exact copy of `InstantCurrencyPayments::pay_relayer_rewards`, right? Could you, please, make some shared function that prepares the `relayers_rewards` struct && deduplicate the code?

svyatonik (commented):
Seems good, thanks! Some feedback in comments

modules/messages/src/instant_payments.rs
AsceticBear: No extra specific reason now. I wonder if we can add the `[transactional]` tag to the message pallet dispatch call. So this return error would be useful.

modules/messages/src/lib.rs
```diff
@@ -620,40 +624,14 @@ pub mod pallet {
...
-			}
-
-			// if some new messages have been confirmed, reward relayers
-			if !relayers_rewards.is_empty() {
+				// if some new messages have been confirmed, reward relayers
 				let relayer_fund_account =
 					relayer_fund_account_id::<T::AccountId, T::AccountIdConverter>();
 				<T as Config<I>>::MessageDeliveryAndDispatchPayment::pay_relayers_rewards(
+					lane_id,
+					lane_data.relayers.clone(),
```
svyatonik, line 632: Is this clone required? Seems that `lane_data` isn't used anywhere below

svyatonik (approved):
Looks great, thank you! :) Could you, please, merge master - looks like we've solved that compilation issue there.

modules/messages/src/lib.rs
AsceticBear: Fixed. Please review again.

PR #1154, opened 2021-09-22
Box large arguments of GRANDPA pallet

svyatonik:
I've just found that our pallets are failing [this test](https://github.com/paritytech/polkadot/blob/master/runtime/polkadot/src/lib.rs#L1872). So I've boxed header-related arguments of GRANDPA pallet + added tests for pallets that are (to be) used by Polkadot/Kusama. In future we'll need to add blanket tests for Rialto/Millau calls (like Polkadot does), but now it fails because there are other pallets (poa?) that are failing this test.

We'll also need to think about boxing arguments of messages pallet - it is around 200 bytes now. But now it is not a super-high priority.

I assume that the call encoding stays the same (I'm not sure though) && it doesn't break existing relays. But I've left label because call signature has been changed.

It doesn't break runtimes (because encoding is the same as I assume), but it changes metadata. So I'm not sure - if breaksruntime fits here or not? What are we going to do with pallet deployed at Rococo/Wococo then? Would it lead to tx version bump or jsut spec version bump - I'm too far from this right now, but we need to take care of this when syncing code to polkadot repo

### DIFF
--- a/modules/grandpa/src/lib.rs
+++ b/modules/grandpa/src/lib.rs
@@ -598,7 +598,7 @@ pub(crate) fn find_forced_change<H: HeaderT>(
 #[cfg(feature = "runtime-benchmarks")]
 pub fn initialize_for_benchmarks<T: Config<I>, I: 'static>(header: BridgedHeader<T, I>) {
 	initialize_bridge::<T, I>(InitializationData {
-		header,
+		Box::new(header),
 		authority_list: sp_std::vec::Vec::new(), // we don't verify any proofs in external benchmarks
 		set_id: 0,
 		is_halted: false,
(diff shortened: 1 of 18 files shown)


modules/grandpa/src/lib.rs
```diff
@@ -130,7 +130,7 @@ pub mod pallet {
 		))]
 		pub fn submit_finality_proof(
 			origin: OriginFor<T>,
-			finality_target: BridgedHeader<T, I>,
+			finality_target: Box<BridgedHeader<T, I>>,
 			justification: GrandpaJustification<BridgedHeader<T, I>>,
```
tomusdrw, line 134: Isn't `Justification` bigger than the header?

tomusdrw (approved):

modules/grandpa/src/lib.rs
svyatonik: No, it keeps most of stuff on heap, so only `sizeof(Vec<>)` is accounted ~64bits.

PR #1158, opened 2021-09-23
Fix CI problems (caused by building multiple runtimes in parallel)

svyatonik:
I've ended up with:
1) skipping WASM build when clippy || check is running - we don't need wasm there;
2) adding explicit `cargo build --release` for all runtimes before `cargo build` && `cargo test` => runtimes are built sequentially, not in parallel;
3) running `cargo build` && `cargo test` with `SKIP_WASM_BUILD=1`, so artifacts from (2) are reused;
3) restoring sccache.

Build time has changed from ~21m to ~26m, but I have no other idea how to fix it now. `cargo metadata` is not helping - it has decreased failures rate, but they are still there.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -93,10 +93,8 @@ clippy-nightly:
   stage:                           lint
   <<:                              *docker-env
   <<:                              *test-refs
-  variables:
-    RUSTC_WRAPPER:                 ""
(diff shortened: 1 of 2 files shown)


svyatonik:
also updated polkadot+cumulus to latest commits before new metadata

Issue #1140, opened 2021-09-14
Clean up custom types that are not needed anymore.


hbulgarini:
Since polkadotjs api contains some of the custom types needed for setting up the bridge, some custom types are not needed anymore. For example, the header sync and messaging stuff.

tomusdrw:
@hbulgarini can you tak a look at this? Would be good to only have the essential types.

tomusdrw:
CC @hbulgarini

acatangiu:
This is obsolete AFAICT

PR #1159, opened 2021-09-24
Fix OutboundLaneData types

HackFisher:
Related https://github.com/darwinia-network/apps/issues/277#issuecomment-926319423

Reference: https://github.com/darwinia-network/parity-bridges-common/blob/7be2b478b49bc8e87ddfe034076e9e80e2f126f6/primitives/messages/src/lib.rs#L278-L287

### DIFF
--- a/deployments/types-millau.json
+++ b/deployments/types-millau.json
@@ -91,9 +91,9 @@
     "dispatch_results": "BitVec"
   },
   "OutboundLaneData": {
-    "latest_generated_nonce": "MessageNonce",
+    "oldest_unpruned_nonce": "MessageNonce",
     "latest_received_nonce": "MessageNonce",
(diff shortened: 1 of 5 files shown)


tomusdrw (approved):
Thanks! These types should not be necessary (see #1140) and the main fix should be in `polkadot.js/api`. Did you make a PR there already?

Let's merge this PR as well though, since it can serve as an override.

HackFisher:
Also made a PR to `polkadot.js/api` just now: https://github.com/polkadot-js/api/pull/3992

tomusdrw:
Awesome, thanks a lot!!!

tomusdrw (changes requested):
Actually, please update `common.json` file and run `build.sh` to regenerate the types properly, I just noticed it's missing from there.

tomusdrw (approved):

PR #1170, opened 2021-10-05
Register-parachain subcommand of substrate-relay

svyatonik:
I was initially thinking that it would be easier to add required storage keys to Rialto genesis. But it requires `rialto-parachain-collator` to be built and started (for genesis header export) first. So instead I'm adding a subcommand to relay.


primitives/chain-rialto-parachain/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+	RuntimeDebug,
+};
+use frame_system::limits;
+use sp_core::Hasher as HasherT;
+use sp_runtime::{
+	traits::{BlakeTwo256, IdentifyAccount, Verify},
+	MultiSignature, MultiSigner, Perbill,
+};
+
+/// Maximal weight of single Rialto parachain block.
```
svyatonik, line 35: Since relaychain && parachains must use polkadot primitives to work with Cumulus, we may consider reexporting some primitives from rialto primitives here.

relays/bin-substrate/src/cli/register_parachain.rs
```diff
@@ -0,0 +1,343 @@
...
+								.encode(),
+							)
+						},
+					)
+					.await?,
+			)
+			.await?;
+			log::info!(target: "bridge", "Reserved parachain id: {:?}", para_id);
+
+			// step 2: register parathread
```
tomusdrw, line 142: Couldn't these two steps be batched?

tomusdrw (approved):

relays/bin-substrate/src/cli/register_parachain.rs
svyatonik: We don't have utility pallet in our runtimes

PR #1178, opened 2021-10-11
Rialto parachain in test deployments

svyatonik:
Main idea behind this PR is to test if our test machine is able to run two relaychain, parachain and regular chain, or we need a separate instance to run parachain || relaychain. We don't have any means to bridge with parachains now => parachain is always started with main relay chain now.


deployments/run.sh
```diff
@@ -42,7 +42,7 @@ function show_help () {
   exit 1
 }
 
-RIALTO=' -f ./networks/rialto.yml'
+RIALTO=' -f ./networks/rialto.yml -f ./networks/rialto-parachain.yml'
```
svyatonik, line 45: Since we don't have any parachain<>parachain || parachain<>relaychain bridges now, parachain now will be started after main Rialto relaychain. In future we may start this only if P2P || P2R bridge is started.

PR #1180, opened 2021-10-13
Bump dependencies

svyatonik:
substrate: 90dd23e78851605180f3dc3bfcf43e2800354f79
polkadot: dd4b2e6a34a08a01b876d14641e99e7011be3463
cumulus: 9379cd6c1863ea846ad6c6a8cbbc99848dd5d693

~I can only guarantee that it compiles + messages between Rialto and Millau are delivered. I'll check if rialto-parachain works later (and fix if it doesn't work in other PR), because it takes a lot of time to verify it locally.~ All is good

### DIFF
--- a/relays/bin-substrate/src/chains/mod.rs
+++ b/relays/bin-substrate/src/chains/mod.rs
@@ -295,10 +295,11 @@ mod rococo_tests {
 			Box::new(header.clone()),
 			justification.clone(),
 		);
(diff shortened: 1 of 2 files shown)


svyatonik:
This will also break our UI, I believe. One reason is that I had to remove our custom `#[pallet::metadata]` attrs + this PR upgrades to new metadata version. There's no breaksui label, so I'm pinging @hbulgarini directly :)

tomusdrw:
Please merge whenever convenient. @hbulgarini is probably blocked on having polkadot.js/api support v14?

Issue #1138, opened 2021-09-14
Conversion rate between RLT <> MLAU must be updated before starting messages generation


svyatonik:
Right now it is 'solved'with [this](https://github.com/paritytech/parity-bridges-common/blob/6d802416e23d9aae6c1dce8504ffd9069cbb405d/deployments/bridges/rialto-millau/entrypoints/relay-messages-to-rialto-generator-entrypoint.sh#L39) pause. But it is too fragile - e.g.  if relayer has [failed to start](https://github.com/paritytech/parity-bridges-common/pull/1137), then you can't just simply restart it, because some messages were already generated using invalid conversion rate && rational relayer refuses to deliver it.

The proper fix would be e.g. to add subcommand that will run single iteration of [this loop](https://github.com/paritytech/parity-bridges-common/blob/6d802416e23d9aae6c1dce8504ffd9069cbb405d/relays/lib-substrate-relay/src/conversion_rate_update.rs) and waits until the rate is updated. It should be started before messages generation.

Alternative solution would be to replace [this RPC call](https://github.com/paritytech/parity-bridges-common/blob/6d802416e23d9aae6c1dce8504ffd9069cbb405d/relays/bin-substrate/src/cli/send_message.rs#L163) with manual computation. Or to add RPC method that accepts external conversion rate.

svyatonik:
Now estimate fee RPC method has additional `conversion_rate_override` argument and this argument is used by message generators in our test deployments

PR #1189, opened 2021-10-20
Override conversion rate in estimate-message-fee RPC

svyatonik:
related #1138 

This PR, if accepted would require changes in P<>K integration branch + our UI (cc @hbulgarini). So, probably, should be reviewed later. The general idea is that message submitter may know that actual conversion rate has been updated in real world && he'll use this updated rate when computing message fee. There are couple of caveats, though:
1) now all our lanes are paid and are assumed to be served by rational relayers, so it mean that even if actual conversion rate is now lesser than stored, the submitter still has to pay fee, computed with larger rate. Otherwise, send-message tx will fail. So fee should be computed as `max(estimate_without_fee_overrride(), estimate_with_fee_override())`;
2) given (1), in regular configurations, this is only useful if conversion-rate-update mechanism (either automatic or manual) is lagging (see #1138 for example). That's because relayer also looks at actual conversion rate. So in theory in scenarios where we assume all actors are honest (both senders and relayers), we may ignore on-chain conversion rate at all. But that's not the best idea.

This PR only adds an additional argument to RPC methods. If it'll be accepted, I'll change all our relayer subcommands to use real conversion rate when messages are submitted.


bin/runtime-common/src/messages.rs
```diff
@@ -787,8 +795,12 @@ mod tests {
 
 		fn bridged_balance_to_this_balance(
 			bridged_balance: BridgedChainBalance,
+			bridged_to_this_conversion_rate_override: Option<FixedU128>,
 		) -> ThisChainBalance {
-			ThisChainBalance(bridged_balance.0 * BRIDGED_CHAIN_TO_THIS_CHAIN_BALANCE_RATE as u32)
+			let conversion_rate = bridged_to_this_conversion_rate_override
+				.map(|r| r.to_float() as u32)
+				.unwrap_or(BRIDGED_CHAIN_TO_THIS_CHAIN_BALANCE_RATE);
+			ThisChainBalance(bridged_balance.0 * conversion_rate)
```
tomusdrw, line 803:
```suggestion
			ThisChainBalance(bridged_balance.0.saturating_mul(conversion_rate))
```

tomusdrw (approved):

bin/runtime-common/src/messages.rs
svyatonik: It's just a test, so it should be fine :)

tomusdrw: Oh, sorry! didn't notice it's a test code.

PR #1198, opened 2021-11-03
 Custom relay strategy

fewensa:
Help all bridgers implement their own relay strategy


AurevoirXavier:
`RelayStrategy` sounds better

relays/messages/src/message_race_delivery.rs
```diff
@@ -652,73 +641,27 @@ async fn select_nonces_for_delivery_transaction<P: MessageLane>(
...
-						);
-					})
-					.ok()?;
-
-				// if it is the first message that makes reward less than cost, let's log it
-				// if this message makes batch profitable again, let's log it
-				let is_total_reward_less_than_cost = total_reward < total_cost;
-				let prev_total_cost = total_cost;
-				let prev_total_reward = total_reward;
-				total_confirmations_cost =
```
svyatonik, line 685: Previous code:
```
let total_confirmations_cost = 0;
for _ in all_ready_nonces {
...
  total_confirmations_cost = total_confirmations_cost.saturating_add(&confirmation_transaction_cost);
...
}
```

New code:
```
for _ in all_ready_nonces {
...
  let total_confirmations_cost = 0;
  total_confirmations_cost = total_confirmations_cost.saturating_add(&confirmation_transaction_cost);
...
}
```

That's wrong. The same is for other nearby variables

relays/messages/src/relayer_strategy.rs
```diff
@@ -0,0 +1,171 @@
...
+	SourceClient: MessageLaneSourceClient<P>,
+	TargetClient: MessageLaneTargetClient<P>,
+> {
+	/// Relayer operating mode.
+	pub relayer_mode: RelayerMode,
+	/// source chain client
+	pub lane_source_client: SourceClient,
+	/// target chain client
+	pub lane_target_client: TargetClient,
+	pub hard_selected_begin_nonce: MessageNonce,
```
svyatonik, line 57: Please add docs for all public items

relays/messages/src/relayer_strategy.rs
```diff
@@ -0,0 +1,171 @@
...
+pub struct RelayerDecide<P: MessageLane> {
+	/// Whether to participate
+	pub participate: bool,
+	pub total_reward: Option<P::SourceChainBalance>,
+	pub total_cost: Option<P::SourceChainBalance>,
+}
+
+/// The default strategy
+#[derive(Clone)]
+pub struct DefaultRelayerStrategy {}
```
svyatonik, line 76:
```suggestion
pub struct DefaultRelayerStrategy;
```

relays/messages/src/relayer_strategy.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+/// Relayer reference data
+pub struct RelayerReference<
+	P: MessageLane,
+	SourceClient: MessageLaneSourceClient<P>,
+	TargetClient: MessageLaneTargetClient<P>,
+> {
+	/// Relayer operating mode.
+	pub relayer_mode: RelayerMode,
```
svyatonik, line 52: My guess is that you're going to add 3rd option for this enum, right? I mean - there's some third strategy that is currently missing from this enum && by replacing enum with trait, you make it easier to add it. But then we should get rid of enum completely - i.e. remove `enum RelayerMode` and split `DefaultRelayerStrategy` into `AltruisticRelayerStrategy` and `RationalRelayerStrategy`. Otherwise it seems strange :)

relays/messages/src/relayer_strategy.rs
```diff
@@ -0,0 +1,171 @@
...
+	pub ready_nonces_index: usize,
+	pub ready_nonce: MessageNonce,
+	pub ready_details: MessageDetails<P::SourceChainBalance>,
+}
+
+/// Relayer's decision
+pub struct RelayerDecide<P: MessageLane> {
+	/// Whether to participate
+	pub participate: bool,
+	pub total_reward: Option<P::SourceChainBalance>,
```
svyatonik, line 70: If you'll split `DefaultRelayerStrategy` into altruistic+rational strategies, then it'll also make sense to move these two `Options` to some strategy-context, or the strategy itself - i.e. your third strategy may be selecting nonces based on some other data, which may need accumulators.

svyatonik (commented):
The reason of this change is to ease process of adding new strategy, right? So now we have altruistic and rational, you need to add 3rd, but it requires forking `messages-relay`, right? If not - please explain :) If so, then idea seems good to me.

I've left some comments - one issue with accumulators && couple of design suggestions.

fewensa:
Yes, it is, we want to add another strategy to our bridger.

relays/messages/src/relayer_strategy.rs
fewensa: Good idea.

fewensa:
@svyatonik changed to follow your suggestions

relays/bin-substrate/src/cli/relay_headers_and_messages.rs
```diff
@@ -64,8 +63,6 @@ pub struct HeadersAndMessagesSharedParams {
 	/// Hex-encoded lane identifiers that should be served by the complex relay.
 	#[structopt(long, default_value = "00000000")]
 	lane: Vec<HexLaneId>,
-	#[structopt(long, possible_values = RelayerMode::VARIANTS, case_insensitive = true, default_value = "rational")]
```
svyatonik, line 67: Please keep it - we have plans to use both modes. You may need to use runtime relayer configuration (vs compile-time) because of that - i.e.  instead of
```rust
pub type MessageLaneWococoMessagesToRococo = SubstrateMessageLaneToSubstrate<
	Wococo,
	WococoSigningParams,
	Rococo,
	RococoSigningParams,
	AltruisticStrategy,
>;
```
you'll need something like that:
```rust
MessageRelayer::new(AltruisticStrategy)
```

relays/messages/src/relay_strategy/altruistic_strategy.rs
```diff
@@ -0,0 +1,163 @@
...
+			.flat_map(|(_, ready_nonces)| ready_nonces.iter())
+			.enumerate();
+		for (index, (_nonce, details)) in all_ready_nonces {
+			// Since we (hopefully) have some reserves in `max_messages_weight_in_single_batch`
+			// and `max_messages_size_in_single_batch`, we may still try to submit transaction
+			// with single message if message overflows these limits. The worst case would be if
+			// transaction will be rejected by the target runtime, but at least we have tried.
+
+			// limit messages in the batch by weight
+			let new_selected_weight = match selected_weight.checked_add(details.dispatch_weight) {
```
svyatonik, line 70: There are two types of checks in this loop - `hard` and `soft`. You can't violate hard checks in any strategy - i.e. if target chain weight has max transaction size set to 1MB, it means that you can't deliver 2MB of messages in a single tx. That's the 'hard' limit. The 'soft' limit is defined by the strategy. So I'd expect `*_strategy.rs` files to only contain soft checks. Hard checks must be done at the upper level and stay the same for all strategies.

That said, I expect `altruistic_strategy.rs` to only have something like `return true` or `return <something-that-says-Im-ready-to-deliver-all-nonces>` in its only method. And the `rational_strategy.rs` should only have checks related to rewards/losses. Is it possible to fix that, please?

relays/bin-substrate/src/cli/relay_headers_and_messages.rs
svyatonik: I'll expand a bit on why we need altruistic strategy - it is for two cases: (1) serving special lanes, where paid fee is zero and (2) backup plan for fixing cases when rational relayers refuse to deliver messages and regular lanes are stuck (e.g. because of sudden conversion rate change).

relays/messages/src/relay_strategy/altruistic_strategy.rs
fewensa: I changed it. can be reviewed.

relays/messages/src/relay_strategy/rational_strategy.rs
```diff
@@ -205,47 +179,20 @@ impl RelayStrategy for RationalStrategy {
...
-
-			hard_selected_count = soft_selected_count;
-		}
-
-		if hard_selected_count != 0 {
+		if soft_selected_count != 0 {
 			log::trace!(
 				target: "bridge",
-				"Expected reward from delivering nonces [{:?}; {:?}] is: {:?} - {:?} = {:?}",
+				"Expected reward from delivering nonces [{:?}; ?] is: {:?} - {:?} = {:?}",
```
fewensa, line 189: About there, because can not get `hard_selected_count`, it's in `EnforcementStrategy`, maybe we can add a new parameter to the decide function?

relays/messages/src/relay_strategy/altruistic_strategy.rs
fewensa: Oh, sorry, This is still wrong, please wait. let me fix it

fewensa: Done

fewensa: I don't know if there is a better solution, there is currently a big struct 
https://github.com/darwinia-network/parity-bridges-common/blob/845ac793b346aa594e0afcdff1d8ca62ca46d90c/relays/messages/src/relay_strategy/enforcement_strategy.rs?#L64-L89

relays/messages/src/relay_strategy/enforcement_strategy.rs
```diff
@@ -0,0 +1,221 @@
...
+		let mut soft_selected_count = 0;
+
+		let mut selected_weight: Weight = 0;
+		let mut selected_count: MessageNonce = 0;
+
+		let hard_selected_begin_nonce =
+			reference.nonces_queue[reference.nonces_queue_range.start].1.begin();
+
+		// relay reference
+		let mut relay_reference = RelayReference {
```
fewensa, line 65: I don't know if there is a better solution, there is currently a big struct

svyatonik (commented):
Thanks! I'll take a look in next few days - the change is non-trivial && I'll need to review it in details.

relays/messages/src/relay_strategy/enforcement_strategy.rs
```diff
@@ -0,0 +1,242 @@
...
+
+		let all_ready_nonces = reference
+			.nonces_queue
+			.range(reference.nonces_queue_range.clone())
+			.map(|(header_id, ready_nonces)| (header_id, ready_nonces.iter()))
+			.enumerate();
+
+		for (index, (header_id, nonces)) in all_ready_nonces {
+			relay_reference.index = index;
+			relay_reference.header_id = header_id.clone();
```
fewensa, line 100: We need to use the block number to determine whether the relay has timed out, so give up using flat_map and put the block number in the reference.

svyatonik (commented):
The code seems to be correct - I'll do a brief run before approving + merging, though :)

Couple of comments:
1) there are multiple warnings in the clippy + fmt jobs for this PR. They're introduced here - please fix it. I can't hit merge until everything is green :)
2) the `MixStrategy` is very similar to what've been `DefaultStrategy` in the very beginning :) It looks like a code that links two implementations - enum-based old and trait-based new. I'd prefer to have `AltruisticStrategy` + `RationalStrategy` only;
3) big structs that are passed between strategies. I've suggested strategies to have some context for their own fields. Like `total_cost` and `total_reward` fields of the `RelayerReference` struct are only used by the `RationalStrategy`. They're never touched by the `AltruisticStrategy` and they're never read or updated by that `EnforcementStrategy`. So it isn't clear when I look at all these strategies code - who's responsible for updating these fields and how they're related to selecting nonces.

Re (2) and (3) - we may add a separate issue for that and fix it in later PRs if you're blocked on this one.

fewensa:
About 2
I think `MixStrategy` is a simple solution. If we use `AltruisticStrate` or `RationalStrategy` directly, the code will become more complicated. We need to change 
[pub relay_strategy: Strategy,](https://github.com/darwinia-network/parity-bridges-common/blob/eaa4a1f4b9112be052ef501d456d3cf980e43929/relays/messages/src/message_lane_loop.rs?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L96) to pub relay_strategy: Box<Strategy>.
and there
https://github.com/darwinia-network/parity-bridges-common/blob/eaa4a1f4b9112be052ef501d456d3cf980e43929/relays/messages/src/message_race_delivery.rs?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L250
there
https://github.com/darwinia-network/parity-bridges-common/blob/eaa4a1f4b9112be052ef501d456d3cf980e43929/relays/bin-substrate/src/chains/kusama_messages_to_polkadot.rs?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L185
etc.

Maybe we can remove `RelauerMode` in messages-relay module. just in `substrate-relay` module. and move `MixStrategy` to `substrate-relay`. Or be sure to change to use `Box`?

About 3.
I'm sorry, I don't have a good idea to perfect this question, There is only one thing I think can be optimized, 

merge selected_reward and total_reward
merge selected_cost and total_cost

fewensa:
All tests and code format fixed.

svyatonik (approved):

svyatonik:
#1202 and #1203

PR #1220, opened 2021-11-23
Refactor finality relay helpers

svyatonik:
The end-goals of this (and follow-up messages related) PR is to ease adding and maintaining existing bridge relayer declarations. Right now every (even minor) change to generic finality/message relayers may lead to 17 files updates, which is a real PITA. This and follow-up message-related PRs move most of glue code from chain-specific `relays/bin-substrate/src/chains` files to generic implementations in the `relays/lib-substrate-relay/src`. The code left in `relays/bin-substrate/src/chains` will be mostly declarative, meaning (almost) no real code, just configuration.

Some notable changes:
1) I've removed price metrics from finality relays, because they're now exposed by message relays. We now have tokens associated with Millau and Rialto, so we may see it in maintain dashboard;
2) `SubmitFinalityProofCallBuilder` trait has two implementations - `DirectSubmitFinalityProofCallBuilder` for cases when we have direct access to the runtime (e.g. for Millau and Rialto). The other implementation(s) are for the cases when we don't have an access to the runtime and build calls using mock enums (like for all Polkadot-like chains). The latter is built using generate_mocked_submit_finality_proof_call_builder` macro - out of ideas how to make it declarative + without using macro;
3) some abstractions from `finality_pipeline.rs` are dropped.


tomusdrw (approved):
image
do like! :)

I think there is no way to avoid macros at this point, up till we get the tooling to support dynamic metadata.

PR #1227, opened 2021-11-26
Integrate BEEFY with Rialto & Millau runtimes

acatangiu:
Fixes https://github.com/paritytech/grandpa-bridge-gadget/issues/342

- Added Beefy pallet to Rialto runtime
- Added Beefy gadget to Rialto node
- Added MMR pallet to Rialto runtime
- Added Beefy pallet to Millau runtime
- Added Beefy gadget to Millau node
- Added MMR pallet to Millau runtime
- Added pallet_beefy_mmr to Millau runtime
- Added pallet_beefy_mmr to Rialto runtime
- Implemented MMR and BEEFY APIs in Rialto


bin/millau/node/src/service.rs
```diff
@@ -207,6 +207,7 @@ pub fn new_full(mut config: Configuration) -> Result<TaskManager, ServiceError>
 	}
```
svyatonik, line 207: There are short instructions on top of `service.rs` of both Millau and Rialto nodes. Could you, please, update these to (at least) mention that beefy integration should be re-added/preserved on update. These file are a major pain when we update Substrate/Polkadot refs => it is often easier to replace the whole file with upstream version and then fix issues/re-add something that we have customized.

bin/rialto/runtime/src/lib.rs
```diff
@@ -1143,7 +1228,7 @@ mod tests {
 
 	#[test]
 	fn call_size() {
-		const MAX_CALL_SIZE: usize = 230; // value from polkadot-runtime tests
+		const MAX_CALL_SIZE: usize = 232; // slightly bigger than polkadot-runtime tests
```
svyatonik, line 1231: I know that BEEFY has not yet reached the Polkadot or Kusama runtimes => it is safe to ignore this limit here. Bt I suggest to add an issue to BEEFY tracker to box large call arguments and revert this change when it's done. Otherwise we'll need to have some justification of why 232 is better than 230 when BEEFY reach Polkadot/Kusama :)

Meanwhile could you, please, return `230` here and replace blanket check with checks specific to pallets that are/will be included there soon(ish):
```rust
		assert!(core::mem::size_of::<pallet_bridge_grandpa::Call<Runtime>>() <= MAX_CALL_SIZE);
		assert!(core::mem::size_of::<pallet_bridge_messages::Call<Runtime>>() <= MAX_CALL_SIZE);
```

svyatonik (approved):
Looks good to me (given couple of optional grumbles). But it'd be better if @adoerr could also check this :)

In addition (probably in follow-up PR) I also suggest to add tracing for BEEFY pallets and gadget to our test deployment nodes ([here](https://github.com/paritytech/parity-bridges-common/blob/ffef6f89f97936c186e6bdf54c9a76a3d37d69c5/deployments/networks/millau.yml#L23) and [here](https://github.com/paritytech/parity-bridges-common/blob/ffef6f89f97936c186e6bdf54c9a76a3d37d69c5/deployments/networks/rialto.yml#L23)). Right now we don't even see if BEEFY is broken on our nodes + this may help us find some interesting long-running issues (given that we run errors-grep for nodes logs).

adoerr (approved):

bin/millau/node/Cargo.toml
```diff
@@ -23,6 +23,9 @@ pallet-bridge-messages = { path = "../../../modules/messages" }
 
 # Substrate Dependencies
 
+beefy-primitives = { git = "https://github.com/paritytech/substrate", branch = "master" }
+beefy-gadget = { git = "https://github.com/paritytech/substrate", branch = "master" }
+beefy-gadget-rpc = { git = "https://github.com/paritytech/substrate", branch = "master" }
```
tomusdrw, line 28: OCD: Please keep these sorted :)

bin/millau/node/src/chain_spec.rs
```diff
@@ -57,10 +58,11 @@ where
 }
 
 /// Helper function to generate an authority key for Aura
-pub fn get_authority_keys_from_seed(s: &str) -> (AccountId, AuraId, GrandpaId) {
+pub fn get_authority_keys_from_seed(s: &str) -> (AccountId, AuraId, BeefyId, GrandpaId) {
```
tomusdrw, line 61: nit: Any reason to add `BeefyId` before `GrandpaId`? Somehow feels more natural to have it at the end, but I guess you preferred to have things sorted?

bin/millau/runtime/src/lib.rs
```diff
@@ -462,6 +501,11 @@ construct_runtime!(
 		ShiftSessionManager: pallet_shift_session_manager::{Pallet},
 		RandomnessCollectiveFlip: pallet_randomness_collective_flip::{Pallet, Storage},
 
+		// Bridges support.
```
tomusdrw, line 504:
```suggestion
		// BEEFY Bridges support.
```

bin/rialto/node/Cargo.toml
```diff
@@ -28,10 +28,14 @@ rialto-runtime = { path = "../runtime" }
 
 # Substrate Dependencies
 
+beefy-primitives = { git = "https://github.com/paritytech/substrate", branch = "master" }
+beefy-gadget = { git = "https://github.com/paritytech/substrate", branch = "master" }
+beefy-gadget-rpc = { git = "https://github.com/paritytech/substrate", branch = "master" }
```
tomusdrw, line 33: sort please

bin/rialto/node/src/service.rs
```diff
@@ -117,8 +117,7 @@ type FullGrandpaLink = sc_finality_grandpa::LinkHalf<Block, FullClient, FullSele
 
 /// A set of APIs that polkadot-like runtimes must implement.
 ///
-/// This is the copy of `polkadot_service::RuntimeApiCollection` with some APIs removed
-/// (right now - MMR and BEEFY).
+/// This is a copy of `polkadot_service::RuntimeApiCollection`.
```
tomusdrw, line 120: Does it still need to be a copy now? I.e. could we re-use that module directly from polkadot?

bin/rialto/node/src/service.rs
```diff
@@ -713,6 +732,22 @@ where
 	let keystore_opt =
 		if role.is_authority() { Some(keystore_container.sync_keystore()) } else { None };
 
+	let beefy_params = beefy_gadget::BeefyParams {
+		client: client.clone(),
+		backend: backend.clone(),
+		key_store: keystore_opt.clone(),
+		network: network.clone(),
+		signed_commitment_sender,
+		min_block_delta: 4,
```
tomusdrw, line 741:
```suggestion
		min_block_delta: 2,
```

Let's make it faster.

bin/rialto/runtime/src/lib.rs
```diff
@@ -466,6 +507,11 @@ construct_runtime!(
 		Grandpa: pallet_grandpa::{Pallet, Call, Storage, Config, Event},
 		ShiftSessionManager: pallet_shift_session_manager::{Pallet},
 
+		// Bridges support.
```
tomusdrw, line 510:
```suggestion
		// BEEFY Bridges support.
```

bin/millau/runtime/src/lib.rs
```diff
@@ -606,6 +650,12 @@ impl_runtime_apis! {
 		}
 	}
 
+	impl beefy_primitives::BeefyApi<Block> for Runtime {
+		fn validator_set() -> ValidatorSet<BeefyId> {
+			Beefy::validator_set()
+		}
+	}
```
tomusdrw, line 657: `Mmr` runtime API missing?

bin/millau/node/src/service.rs
```diff
@@ -277,9 +280,15 @@ pub fn new_full(mut config: Configuration) -> Result<TaskManager, ServiceError>
 				shared_authority_set.clone(),
 				shared_voter_state.clone(),
 				justification_stream.clone(),
-				subscription_executor,
+				subscription_executor.clone(),
 				finality_proof_provider.clone(),
 			)));
+			io.extend_with(beefy_gadget_rpc::BeefyApi::to_delegate(
```
tomusdrw, line 286: Can we also add MMR custom RPC here?

bin/rialto/node/src/service.rs
```diff
@@ -321,9 +332,15 @@ where
 				shared_authority_set.clone(),
 				shared_voter_state,
 				justification_stream.clone(),
-				subscription_executor,
+				subscription_executor.clone(),
 				finality_proof_provider,
 			)));
+			io.extend_with(beefy_gadget_rpc::BeefyApi::to_delegate(
```
tomusdrw, line 338: Can we also add MMR custom RPC here?

tomusdrw (commented):
Looks good! Left couple of things that would be good to address.

bin/rialto/node/src/service.rs
acatangiu: Used the polkadot upstream definition :+1:

bin/millau/node/src/chain_spec.rs
tomusdrw: Makes sense!

tomusdrw: Makes sense, although we have to be careful with such changes for running networks - it's easier to migrate if we add something at the end than to change the ordering.

tomusdrw (approved):

tomusdrw:
@acatangiu seems that #1229 introduced some conflicts, would you mind updating to latest master?

PR #1229, opened 2021-11-29
update dependencies

antonio-dropulic:
<del>Had a lot of rustfmt changes, not sure if it's miss-configured somehow.</del>

### DIFF
--- a/bin/millau/node/src/command.rs
+++ b/bin/millau/node/src/command.rs
@@ -72,12 +72,12 @@ impl SubstrateCli for Cli {
 pub fn run() -> sc_cli::Result<()> {
 	let cli = Cli::from_args();
 	// make sure to set correct crypto version.
-	sp_core::crypto::set_default_ss58_version(sp_core::crypto::Ss58AddressFormat::Custom(
+	sp_core::crypto::set_default_ss58_version(sp_core::crypto::Ss58AddressFormat::custom(
 		millau_runtime::SS58Prefix::get() as u16,
 	));
 
 	match &cli.subcommand {
-		Some(Subcommand::Benchmark(cmd)) =>
+		Some(Subcommand::Benchmark(cmd)) => {
 			if cfg!(feature = "runtime-benchmarks") {
 				let runner = cli.create_runner(cmd)?;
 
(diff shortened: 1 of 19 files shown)


bin/millau/node/src/service.rs
```diff
@@ -107,15 +109,15 @@ pub fn new_partial(
...
+		sc_service::new_full_parts::<Block, RuntimeApi, _>(
+			&config,
 			telemetry.as_ref().map(|(_, telemetry)| telemetry.handle()),
 			executor,
 		)?;
 	let client = Arc::new(client);
 
 	let telemetry = telemetry.map(|(worker, telemetry)| {
-		task_manager.spawn_handle().spawn("telemetry", worker.run());
+		task_manager.spawn_handle().spawn("telemetry", None, worker.run());
```
antonio-dropulic, line 120: `SpawnTaskHandle` introduced groups. I've set all to default(None).

bin/millau/node/src/command.rs
```diff
@@ -88,70 +88,74 @@ pub fn run() -> sc_cli::Result<()> {
...
 			runner
 				.sync_run(|config| cmd.run::<Block, RuntimeApi, service::ExecutorDispatch>(config))
-		},
+		}
 		None => {
 			let runner = cli.create_runner(&cli.run)?;
 			runner.run_node_until_exit(|config| async move {
 				match config.role {
-					Role::Light => service::new_light(config),
+					Role::Light => {
```
antonio-dropulic, line 151: Should this be handled differently?

bin/millau/node/src/service.rs
svyatonik: Millau node is (mostly) a copy of https://github.com/paritytech/substrate/tree/master/bin/node-template . So if it is`None` there, then it should be fine

bin/millau/node/src/command.rs
```diff
@@ -147,7 +147,10 @@ pub fn run() -> sc_cli::Result<()> {
...
```
svyatonik, line 150: Let's just remove it as it is done in substrate node template: https://github.com/paritytech/substrate/blob/master/bin/node-template/node/src/command.rs#L113

tomusdrw (approved):

PR #1235, opened 2021-12-01
Add get origin client function

fewensa:
In this way, a healthy original client can be obtained, Only use `Any` to return, because there is no `relay-client-substrate` package dependent on this library.
This is not necessarily a good method, but it is indeed a solution.

After do that, If we create a customized relay strategy. we can get an original client from [`relay_reference`](https://github.com/paritytech/parity-bridges-common/blob/816ddd2dd2239012c6fd08b984bf4d97d5399131/relays/messages/src/relay_strategy/mod.rs#L66-L69)

```
let client: &relay_client_substrate::Client<RialtoChain> = relay_reference.lane_source_client.origin_client().downcast_ref().unwrap();
```

In doing so, the biggest advantage is that third-party applications can get a client that will not lose the connection, because this has been processed in the message-lane-loop .

### DIFF
--- a/relays/finality/src/finality_loop.rs
+++ b/relays/finality/src/finality_loop.rs
@@ -32,7 +32,6 @@ use relay_utils::{
 	MaybeConnectionError,
 };
 use std::{
-	any::Any,
 	pin::Pin,
 	time::{Duration, Instant},
 };
--- a/relays/finality/src/finality_loop.rs
+++ b/relays/finality/src/finality_loop.rs
@@ -87,9 +87,6 @@ pub trait SourceClient<P: FinalitySyncPipeline>: RelayClient {
 /// Target client used in finality synchronization loop.
 #[async_trait]
 pub trait TargetClient<P: FinalitySyncPipeline>: RelayClient {
-	/// Simplify to get origin client, But you need use downcast_ref to convert to you want types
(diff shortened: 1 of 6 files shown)


svyatonik:
> In doing so, the biggest advantage is that third-party applications can get a client that will not lose the connection, because this has been processed in the message-lane-loop .

Is it the only advantage? I assume that you're creating strategy using initial clients and then if disconnect occurs, loop is able to recover and your strategy isn't, right? Then imo the simplest way is to impl `TargetClient`/`SourceClient` for some struct like that (could probably be rewritten without Mutexes, but the idea is in common shared data of the cloneable client):

```rust
struct SharedSourceClient {
  shared_data: Arc<Mutex<SharedSourceClientData>>,
}

struct SharedSourceClientData {
  substrate_client: relay_substrate_client::Client<YourChain>,
}

impl RelayClient for SharedSourceClient {
  fn reconnect() {
    self.shared_data.lock().await.client.reconnect();
  }
}
```

Then you may simply return errors from your strategy and loop will do a reconnect.

If you have other usages for this `origin_client`, please comment. Thanks!

fewensa:
Yes, The purpose of adding this method is to get the client. I will try the method your suggestion.

PR #1245, opened 2021-12-07
runtimes: fix call_size() tests

acatangiu:
`call_size()` test now correctly validates that **bridges** pallets `Call` sizes stay within 200 bytes.

Rialto runtime `Call` total size is increased by the extra Beefy keys required in `pallet_session::Call`. Document that for clarity.

Fixes https://github.com/paritytech/grandpa-bridge-gadget/issues/359

### DIFF
--- a/bin/millau/runtime/src/lib.rs
+++ b/bin/millau/runtime/src/lib.rs
@@ -922,6 +922,15 @@ mod tests {
 
 	#[test]
 	fn call_size() {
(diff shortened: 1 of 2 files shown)


svyatonik (approved):
Cool, thanks! I was assuming it is related to some BEEFY pallets. Then it'll probably needs to be fixed in Substrate, when BEEFY will be close to deployment :) Alternatively - we have an excuse for this increase (since it isn't our pallets).

tomusdrw (approved):

PR #1248, opened 2021-12-10
Change submit transaction spec_version and transaction_version query from chain

fewensa:
After doing this, we don’t need to upgrade bridger after a chain runtime is upgraded.


svyatonik:
But it also means that if something changes in the bridge pallets (or in some other involved pallets) the relayer will keep creating old transactions, which may have different meaning in new `spec_version` or even different SCALE format in new `transaction_version`. In worst case relayer will keep submitting invalid transactions. which may even lead to funds loss if they are accepted to the pool. So in my opinion the this step requires manual approval => we even have `spec_version` guards (which abort relay process) enabled for some bridges. WDYT about it?

If you still need this patch, then I suggest to make it optional - e.g. probably change `TransactionSignScheme` trait to accept `Client<Self::Chain>` and read all required fields in this implementation for your chain. Alternatively - you may consider adding some CLI options for that.

fewensa:
I think `sign_transaction` accept `Client<Self::Chain>` isn't a good way.

for example.
https://github.com/paritytech/parity-bridges-common/blob/80c94292841ae3115e1ae8d9994601f82a087829/relays/bin-substrate/src/chains/kusama_messages_to_polkadot.rs#L64-L96


the `client` can not move to `prepare_extrinsic` function. because it used by `client.submit_signed_extrinsic` we need clone this.

I think keeping the current code is better. but your suggestions are good. So I want add 

```
#[derive(Clone, Debug)]
pub enum ChainRuntimeVersion {
	Auto,
	Custom(u32, u32),
	Bundle,
}
```
into relay_substrate_client::Client

But i won't let `sign_transaction` accept `Client<Self::Chain>`. many things will be changed. like sign_transaction mu be change to async function and  `submit_signed_extrinsic` must accept async `prepare_extrinsic` function.
What do you think?

svyatonik:
Client is lightweight + cloneable, so it shouldn't be a big problem. But the second option is ok to me too if it seems easier to you :)

fewensa:
Yes. I think the second option is easy to use. if not we need change 
https://github.com/paritytech/parity-bridges-common/blob/80c94292841ae3115e1ae8d9994601f82a087829/relays/client-substrate/src/client.rs#L347

to

```
prepare_extrinsic: impl FnOnce(HeaderIdOf<C>, C::Index) -> BoxFuture<relay_substrate_client::Result<Bytes>> + Send + 'static,
```
This is difficult to modify, and there are many changes.

relays/bin-substrate/src/cli/register_parachain.rs
```diff
@@ -83,6 +83,11 @@ macro_rules! select_bridge {
 
 				use bp_rialto::{PARAS_PALLET_NAME, PARAS_REGISTRAR_PALLET_NAME};
 
+				const RELAY_CHAIN_SPEC_VERSION: u32 = rialto_runtime::VERSION.spec_version;
+				const RELAY_CHAIN_TRANSACTION_VERSION: u32 = rialto_runtime::VERSION.spec_version;
+				const PARA_CHAIN_SPEC_VERSION: u32 = rialto_runtime::VERSION.spec_version;
+				const PARA_CHAIN_TRANSACTION_VERSION: u32 = rialto_runtime::VERSION.spec_version;
```
fewensa, line 89: I'm not sure it is correct.

fewensa:
I'm not found what reason about ci error for `continuous-integration/gitlab-spellcheck`

svyatonik:
> I'm not found what reason about ci error for `continuous-integration/gitlab-spellcheck`

Here's my local output:
<details>
<summary>Output</summary>

```
error: spellcheck(Hunspell)
    --> /home/svyatonik/dev/parity-bridges-common/relays/bin-substrate/src/cli/mod.rs:375
     |
 375 |  Custom spec_version and transaction_version
     |         ^^^^^^^^^^^^
     | - perversion or interspersion
     |
     |   Possible spelling mistake found.

error: spellcheck(Hunspell)
    --> /home/svyatonik/dev/parity-bridges-common/relays/bin-substrate/src/cli/mod.rs:375
     |
 375 |  Custom spec_version and transaction_version
     |                          ^^^^^^^^^^^^^^^^^^^
     | - transubstantiation
     |
     |   Possible spelling mistake found.

error: spellcheck(Hunspell)
    --> /home/svyatonik/dev/parity-bridges-common/relays/bin-substrate/src/cli/mod.rs:410
     |
 410 | The custom sepc_version for chain 
     |            ^^^^^^^^^^^^
     | - perversion
     |
     |   Possible spelling mistake found.

error: spellcheck(Hunspell)
    --> /home/svyatonik/dev/parity-bridges-common/relays/bin-substrate/src/cli/mod.rs:413
     |
 413 | The custom transaction_version for chain 
     |            ^^^^^^^^^^^^^^^^^^^
     | - transubstantiation
     |
     |   Possible spelling mistake found.

[2021-12-15T07:20:45Z INFO  cargo_spellcheck::action] ❌ /home/svyatonik/dev/parity-bridges-common/relays/bin-substrate/src/cli/mod.rs : 4
error: spellcheck(Hunspell)
   --> /home/svyatonik/dev/parity-bridges-common/relays/client-substrate/src/client.rs:69
    |
 69 |  the first is spec_version
    |               ^^^^^^^^^^^^
    | - perversion or interspersion
    |
    |   Possible spelling mistake found.

error: spellcheck(Hunspell)
   --> /home/svyatonik/dev/parity-bridges-common/relays/client-substrate/src/client.rs:70
    |
 70 |  the second is transaction_version
    |                ^^^^^^^^^^^^^^^^^^^
    | - transubstantiation
    |
    |   Possible spelling mistake found.

```
</details>


The idea is that (apart of `sepc_verstion` typo) all variable/function names should be enclosed in backquoutes.

We'll definitely need to do something with spellcheck logs at CI - I'll create an issue for that.

relays/bin-substrate/src/cli/register_parachain.rs
svyatonik: Could you, please, use `rialto_parachain_runtime::VERSION.spec_version` here?

relays/client-substrate/src/client.rs
```diff
@@ -178,6 +194,24 @@ impl<C: Chain> Client<C> {
 }
 
 impl<C: Chain> Client<C> {
+	/// Return spec version
```
svyatonik, line 197: I'd suggest to merge two methods - otherwise there's a risk of race when you read twice from non-upgraded and then upgraded chain.

svyatonik: + fields are always used together, if I get it correctly

relays/bin-substrate/src/cli/mod.rs
```diff
@@ -500,11 +531,32 @@ macro_rules! declare_chain_options {
 				/// Convert connection params into Substrate client.
 				pub async fn to_client<Chain: CliChain>(
 					&self,
+					bundle_spec_version: u32,
```
svyatonik, line 534: Could we replace these two parameters with `Option<RuntimeVersion>` here? `Option` because for some chains (where we won't submit any transaction) the version may be unknown (so it shall be treated as `ChainRuntimeVersion::Auto`). And using `RuntimeVersion` would allow us replace four consts with two, where `to_client()` is called. We'll probably move `VERSION` constants to `Chain` trait later => remove these consts at all.

svyatonik (commented):
Looks good. Left couple of suggestions + spellckech fix is required

relays/bin-substrate/src/cli/register_parachain.rs
fewensa: It fixed here https://github.com/paritytech/parity-bridges-common/pull/1248/commits/91c9d51a3ed08c448f31fdad3f4ff668120326b6 please check it

fewensa:
Done

fewensa:
I'm not sure what happened in CI

```
  --> /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/rsix-0.23.9/src/imp/linux_raw/arch/inline/x86_64.rs:73:5
   |
73 |     asm!(
   |     ^^^
   |
   = note: consider importing one of these items:
           std::arch::asm
           core::arch::asm
error: cannot find macro `asm` in this scope
```

It's ok for my local computer

svyatonik (approved):

svyatonik:
> I'm not sure what happened in CI
> 
> ```
>   --> /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/rsix-0.23.9/src/imp/linux_raw/arch/inline/x86_64.rs:73:5
>    |
> 73 |     asm!(
>    |     ^^^
>    |
>    = note: consider importing one of these items:
>            std::arch::asm
>            core::arch::asm
> error: cannot find macro `asm` in this scope
> ```
> 
> It's ok for my local computer

Should be something with todays nightly. Will look today

relays/bin-substrate/src/cli/mod.rs
svyatonik: Last q - is there any specific reason you have decided to use two `Option<u32>`arguments instead of single `Option<RuntimeVersion>` as I've suggested here? If not, we'll change this in near future

svyatonik:
CI issue should be fixed now. Could you, please, push some dummy commit (e.g. some change + immediate `git revert`) to restart CI (i have no rights for that :) ). There's also question from my side &&  it could be merged then

relays/bin-substrate/src/cli/mod.rs
fewensa: Sorry,  I missing here. Let me think about it. Your suggestions are better. I can change it. Please wait.

fewensa: https://github.com/paritytech/parity-bridges-common/pull/1248/commits/e607a372fb3c71e48e58306d47aa4e73216e0164

svyatonik (approved):
Perfect, thank you!

Issue #1246, opened 2021-12-09
Remove runtime APIs that are simply returning storage values


svyatonik:
When this project was started, I/we had an impression that we shall no access storage values directly and instead call runtime APIs for that. That's because reading from storage imples that the client (mostly relay, but now also pallets sometimes) needs to compute storage key itself. So now we have runtime APIs like `fn latest_generated_nonce()`, `fn latest_received_nonce()`, ... that are just read some storage value and return it. 

Today accessing storage directly is a common practice in our relays and pallets. And we have these storage-key computation functions in `bp-runtime`. The idea is to remove such simple runtime APIs - this will significantly decrease LOCs (both in relay and runtime code) and also will make implementing relay clients easier.

svyatonik:
Submitter warning: this is really a good first issue, but it would require a lot of code changes.

svyatonik:
There are 3 methods left in the runtime:
1) `<Chain>FinalityApi::best_finalized` - I'm leaving it, because it is a perfect solution for hiding finality pallet details. Relayer (or other apps) may used this method without knowledge of which pallet is used for finality - `pallet_bridge_grandpa`, `pallet_bridge_parachains` or `pallet_bridge_mmr`;
2) `To<Chain>OutboundLaneApi::estimate_message_delivery_and_dispatch_fee` - since logic of fee estimation is quite complex, it is better to use existing implementation, where fee is actually checked (in runtime), than do the same in every runtime client;
3) `To<Chain>OutboundLaneApi::message_details` - this is another potential client for removal. But given #1317, I think it is better to leave relayer (and possibly other apps) unaware of what is actually encoded as payload. So I left it as-is for now

PR #1260, opened 2021-12-21
Read latest_generated_nonce directly from storage

svyatonik:
part of #1246 

I've decided to split big PR, which replaces 5 (I'm unsure about 6th method yet) runtime methods with direct storage reads into separate PRs. This is the first one - instead of having `To<Chain>OutboundLaneApi::latest_generated_nonce` for every chain we'll be reading the value directly from chain storage.


relays/lib-substrate-relay/src/messages_source.rs
```diff
@@ -126,16 +129,19 @@ where
...
+		let outbound_lane_data: Option<OutboundLaneData> = self
 			.client
-			.state_call(
-				P::TargetChain::TO_CHAIN_LATEST_GENERATED_NONCE_METHOD.into(),
-				Bytes(self.lane_id.encode()),
+			.storage_value(
+				outbound_lane_data_key(
+					P::TargetChain::WITH_CHAIN_MESSAGES_PALLET_NAME,
+					&self.lane_id,
+				),
```
acatangiu, line 138: where is the code that _puts_ this data to storage?

svyatonik: We can only modify storage through transaction, so it's in the messages module itself (https://github.com/paritytech/parity-bridges-common/blob/master/modules/messages/src/lib.rs#L1062-L1081). Relay, otoh, has two options - one to use runtime API and query runtime storage directly. So this PR replaces 1st option with 2nd, because it is easier to maintain.

acatangiu: ok, makes sense

acatangiu (approved):
lgtm

PR #1261, opened 2021-12-21
Override conversion rate when computing message fee

### LINKED ISSUE #1259: Use actual conversion rate when submitting messages
follow up of #1189 . We need to change send-message + estimate-message relay subcommands to use actual conversion rate when submitting/estimating messages.

svyatonik:
closes #1259


relays/bin-substrate/src/cli/estimate_fee.rs
```diff
@@ -67,21 +68,68 @@ impl EstimateFee {
...
+	P: Clone + Encode,
+>(
+	client: &relay_substrate_client::Client<Source>,
 	estimate_fee_method: &str,
 	lane: bp_messages::LaneId,
 	payload: P,
-) -> anyhow::Result<Fee> {
-	let conversion_rate_override: Option<FixedU128> = None;
+) -> anyhow::Result<BalanceOf<Source>> {
+	// actual conversion rate CAN be lesser than the rate stored in the runtime. So we may try to
```
svyatonik, line 81: I was thinking of moving this to the runtime API implementation, but then the caller may wonder why he is providing `override` and this `override` isn't actually working.

svyatonik:
Marking this draft - will need to temporary revert #1189

svyatonik:
This needs a couple of fixes to be used in our test deployments - when a bunch of messages is being sent at once (this happens every hour), there are too much HTTP requests generated. That's because every `send-message` call, spawns new relay process, which computes conversion rate itself. This isn't good

Issue #1268, opened 2021-12-23
Refactor messages benchmarks


svyatonik:
I was going to add code for benchmarking our pallets to Rococo/Wococo runtime, but it'd be really helpful to refactor it a bit:
1) there are many benchmarks in messages pallet that are not used directly in weight formulae - they're here to prove some hypotheses. They (imo) still have some utility and value, but they'll slow down automatic benchmarks when called from Polkadot repo. So I suggest to put them under additional feature flag;
2) benchmarks support code is around 150 LOCs (https://github.com/paritytech/parity-bridges-common/blob/e7e83d8944db1cb850426a1e5a55061cae6af2f5/bin/rialto/runtime/src/lib.rs#L970-L1136), which isn't a super-useful addition to the Polkadot/Kusama/Rococo runtime code. So I suggest to explore options to reduce LOCs for this code.

svyatonik:
Additionally: Rialto benchmarks are broken in current `master`. I've been thinking that #1273 may help us with that, but no :/ The worst case (if I won't be able to solve that) is moving benchmarks to Millau runtime.

PR #1273, opened 2021-12-24
Reuse polkadot-service code

### LINKED ISSUE #1136: Try to reuse polkadot service code
https://github.com/paritytech/parity-bridges-common/pull/1053 brings introduces some code that is copypasted from polkadot service. Thanks to @ordian we now have this - https://github.com/paritytech/polkadot/pull/3820 && may try to reuse existing code. However, I prefer to do that until Rialto will be running as relay chain && there's (at least) dummy parachain registered there (basically - what we have in [this](https://github.com/paritytech/parity-bridges-common/tree/rialto-as-relay-chain) branch).

svyatonik:
closes #1136 

Unfortunately, `polkadot-runtime` is still requires `polkadot-service` (actually - one of P/K/W/R runtimes is now required - I've selected polkadot). If it isn't selected, no runtime-features of `polkadot-client` are activated and I'm getting [this error](https://github.com/paritytech/polkadot/blob/master/node/client/src/lib.rs#L49).

But at least I've been able to reuse `polkadot-service` and remove our custom `service.rs`, `overseer.rs` and `parachains_db.rs` (which is a lot of code).


svyatonik:
on hold, until I'll solve https://github.com/paritytech/parity-bridges-common/issues/1268#issuecomment-1000768074

acatangiu:
looks like this is unblocked now

svyatonik:
Yeah - it is ready for review :)

acatangiu (approved):
:100:

PR #1274, opened 2021-12-27
Polkadot staging update

svyatonik:
we need to have #1272 in release branch


svyatonik:
(squash/rebase leads to conflicts in the future, so this PR needs to be merged)

PR #1275, opened 2021-12-27
Revert "override conversion rate in estimate-message-fee RPC (#1189)"

svyatonik:
This reverts commit 50ffb5dd08d662d811e42f6a7cfd9cfb1a7f7de7.

So the situation is:
1) #1189 was marked as "breaksrelay" && I've missed the fact that it'll make our relay incompatible with what's currently in polkadot subtree;
2) #1189 has already landed in `polkadot-staging` => so relay we're going to 'release' will be incompatible with Rococo/Wococo => we'll either need to do changes directly in `polkadot-staging`, or revert in master and do `master -> polkadot-staging` again;
3) since I need to have #1272 in `polkadot-staging` && don't want to diverge these branches now, let's temporary revert #1189 in master.


svyatonik:
And the plan is:
1) merge this;
2) make `master -> polkadot-staging` PR;
3) release `sustrate-relay` 1.0 + make sure it is working;
4) restore #1189.

acatangiu (approved):

PR #1279, opened 2022-01-11
Removed extra messages benchmarks

svyatonik:
rationale: #1268 

I was trying to keep thee benchmarks, but this is hard to do, since they're declared within `benchmarks_instance_pallet! {}` macro itself && this macro has its own requirements (e.g. you can't hide the whole benchmark behind the feature flag, you can't hide the whole body behind feature flag, because first line must be `let i in ...` and so on). So let's remove it for now.


acatangiu (approved):
Code changes look good, but what's the plan for having benchmarks here in the future? Should we open an issue detailing what's needed?

svyatonik:
> what's the plan for having benchmarks here in the future? Should we open an issue detailing what's needed?

Let's just drop them for now. They were helpful when I've been working on weight formulae. Now they only serve as a proof of some assumptions I have made and these assumptions are only checked manually. Since this is very specific case imo (most of pallets don't need such extra benchmarks - see our other pallets for example) I don't think it worth to waste resources on  changing benchmarks macro or on maintaining this code. We may reconsider this later ofc - there's always git history :)

PR #1284, opened 2022-01-14
Support version mode CLI options in send-message subcommand

### LINKED ISSUE #1283: Support version mode cli option when message is being sent
follow-up of #1248

Right now we're using auto version mode in our Rococo<>Wococo test deployments for running relay. But when message is being sent (again - using `substrate-relay` binary), it always [uses](https://github.com/paritytech/parity-bridges-common/blob/8b9c4ec16d1c2bee4855cc1b18721e330dd3cee5/relays/bin-substrate/src/cli/send_message.rs#L124) [bundled](https://github.com/paritytech/parity-bridges-common/blob/8b9c4ec16d1c2bee4855cc1b18721e330dd3cee5/relays/bin-substrate/src/chains/wococo.rs#L73) version. The idea is to support `--version-mode` cli argument in `send-message` subcommand of the relay.

svyatonik:
closes #1283 

So right now Rococo has runtime version `9141` and Wococo is `9140`. "Release" relay thinks that both have `9140`. Since they're testnets, this is fine (given that headers+messages relay is running with `*-version-mode=Auto` CLI option). But relay still uses `9140` when encoding messages from Wococo to Rococo, meaning that they'll be delivered, but dispatch will fail with `MessageVersionSpecMismatch` error. This PR allows message submitter to provide custom version or to read actual version from the target node.

### DIFF
--- a/relays/bin-substrate/src/cli/send_message.rs
+++ b/relays/bin-substrate/src/cli/send_message.rs
@@ -19,7 +19,7 @@ use crate::cli::{
 	encode_call::{self, CliEncodeCall},
 	estimate_fee::estimate_message_delivery_and_dispatch_fee,
 	Balance, CliChain, ExplicitOrMaximal, HexBytes, HexLaneId, Origins, SourceConnectionParams,
-	SourceSigningParams, TargetSigningParams,
+	SourceSigningParams, TargetConnectionParams, TargetSigningParams,
 };
 use bp_message_dispatch::{CallOrigin, MessagePayload};
 use bp_runtime::{BalanceOf, Chain as _};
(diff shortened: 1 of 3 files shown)


relays/bin-substrate/src/cli/send_message.rs
```diff
@@ -110,6 +116,10 @@ impl SendMessage {
 
 			encode_call::preprocess_call::<Source, Target>(message, bridge.bridge_instance_index());
 			let target_call = Target::encode_call(message)?;
+			let target_spec_version = self
```
svyatonik, line 119: That's the main change - instead of always using `Target::RUNTIME_VERSION.spec_version`, I'm using version based on CLI flags

acatangiu (approved):

Issue #1290, opened 2022-01-19
Add missing metrics to Polkadot <> Kusama relay


svyatonik:
Full list of metrics:
- [x] storage proof overhead on Polkadot (also valid for R<>W bridge);
- [x] storage proof overhead on Kusama (also valid for R<>W bridge);
- [x] DOT->KSM conversion rate stored at Kusama;
- [x] actual DOT->KSM conversion rate;
- [x] KSM->DOT conversion rate stored at Polkadot;
- [x] actual KSM->DOT conversion rate;
- [ ] relay account balance on Polkadot (also valid for R<>W bridge);
- [ ] relay account balance on Kusama (also valid for R<>W bridge);
- [ ] Polkadot fee multiplier stored at Kusama (also valid for R<>W bridge);
- [ ] real Polkadot fee multiplier (also valid for R<>W bridge);
- [ ] Kusama fee multiplier stored at Polkadot (also valid for R<>W bridge);
- [ ] real Kusama fee multiplier (also valid for R<>W bridge).

svyatonik:
All remaining metrics are only valid for K<>P bridge (even though we may add some for R<>M or R<>W), so I'm moving this issue to P<>K milestone.

PR #1291, opened 2022-01-20
Relay balance metrics

svyatonik:
part of #1290

There's couple of things that need to be solved:
- [x] balance metric value must be nominated in chain tokens, not its minimal components. E.g. on Polkadot if relay has 100 DOTs on the balance, the metric must have value 100, not 100*10^10;
- [x] I'll need to add metrics (and probably alerts later) to Rialto+Millau maintenance dashboard just to show that it works.


svyatonik:
relay-account-balances

PR #1294, opened 2022-01-21
Polkadot staging update

svyatonik:
There are many changes in the runtime code (which normally should have security audited), but all these changes are just refactoring of benchmarks. So no audit is required.

Last 3 commits (#1288, #1289 and #1291) are not reviewed - please take a moment to look at these. They're about relay, so not super critical - I just wanted to have updated relay serving Rialto<>Millau bridge for couple of days before releasing next (I hope it may be tagged 1.0 finally) Rococo <> Wococo version.


acatangiu (approved):
LGTM!

PR #1310, opened 2022-02-04
Endow relayer account at target chain in message benchmarks

svyatonik:
another (see #1309 ) fix of benchmarks to prepare for P/K/R integration


bin/runtime-common/src/messages_benchmarking.rs
```diff
@@ -43,6 +43,29 @@ use sp_std::{fmt::Debug, prelude::*};
 use sp_trie::{record_all_keys, trie_types::TrieDBMutV1, LayoutV1, MemoryDB, Recorder, TrieMut};
 use sp_version::RuntimeVersion;
 
+/// Return this chain account, used to dispatch message.
```
svyatonik, line 46: This change isn't actually required, it's just a small refactoring I've made when working on this issue - to make it clear that this account is used to dispatch incoming messages

PR #1316, opened 2022-02-10
Add some tests to check integrity of chain constants + bridge configuration

### LINKED ISSUE #1313: Add constants integrity tests to runtimes with bridges (Millau/Rialto/Rococo/Kusama/Polkadot)
We have multiple runtime-related constants in the code. While some of them need to be checked manually on each upgrade, some (I assume) may be tested automatically if tests are located in the chain runtime. This shall help us with handling runtime upgrades. Examples of potentially testable constants:
- [ ] `WITH_<CHAIN>_GRANDPA_PALLET_NAME` - we can test it by checking that actual storage keys of runtime GRANDPA pallet are matching values computed using our `bp_runtime::*_storage_key()` functions;
- [ ] `WITH_<CHAIN>_MESSAGES_PALLET_NAME` - same as above;
- [ ] messages pallet parameter names - same as above;
- [ ] block limits (`bp_<chain>::BlockWeights` and `bp_<chain>::BlockLength`) most probably can be autotested by comparing values read from system pallet
- [ ] ...

We now already have similar tests for messages pallet weights, which fail if there's something wrong with pallet weights

svyatonik:
closes #1313

### DIFF
--- a/bin/runtime-common/src/integrity.rs
+++ b/bin/runtime-common/src/integrity.rs
@@ -31,7 +31,7 @@ use frame_system::limits;
 /// the same types (index, block number, hash, hasher, account id and header).
 #[macro_export]
 macro_rules! assert_chain_types(
-	( $r:path, $this:path ) => {
+	( runtime: $r:path, this_chain: $this:path ) => {
 		{
 			// if one of asserts fail, then either bridge isn't configured properly (or alternatively - non-standard
 			// configuration is used), or something has broke existing configuration (meaning that all bridged chains
@@ -53,7 +53,7 @@ macro_rules! assert_chain_types(
 /// the same types (hash, account id, ...).
 #[macro_export]
 macro_rules! assert_bridge_types(
-	( $bridge:path, $this:path, $bridged:path ) => {
+	( bridge: $bridge:path, this_chain: $this:path, bridged_chain: $bridged:path ) => {
 		{
 			// if one of this asserts fail, then all chains, bridged with this chain and bridge relays are now broken
 			//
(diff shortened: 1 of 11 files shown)


modules/grandpa/src/lib.rs
```diff
@@ -300,7 +300,7 @@ pub mod pallet {
 	/// runtime methods may still be used to do that (i.e. democracy::referendum to update halt
 	/// flag directly or call the `halt_operations`).
 	#[pallet::storage]
-	pub(super) type PalletOwner<T: Config<I>, I: 'static = ()> =
+	pub type PalletOwner<T: Config<I>, I: 'static = ()> =
```
svyatonik, line 303: This is a runtime change, but it is non-significant change. It only makes this field visible to other pallets.

bin/runtime-common/src/integrity.rs
```diff
@@ -0,0 +1,331 @@
...
+use bp_messages::MessageNonce;
+use bp_runtime::{Chain, ChainId};
+use codec::Encode;
+use frame_support::{storage::generator::StorageValue, traits::Get};
+use frame_system::limits;
+
+/// Macro that ensures that the runtime configuration and chain primitives crate are sharing
+/// the same types (index, block number, hash, hasher, account id and header).
+#[macro_export]
+macro_rules! assert_chain_types(
```
svyatonik, line 33: I can't use generic functions for testing types, because `assert_type_eq_all` doesn't support that. Something like "can't use generic argument of outer function"

PR #1322, opened 2022-02-18
Encode and estimate Rococo/Wococo/Kusama/Polkadot messages

### LINKED ISSUE #1319: An easy way to estimate Kusama <> Polkadot message fee
It seems that there's now no easy way to estimate `delivery_and_dispatch_fee` for Kusama <> Polkadot (and Rococo <> Wococo) messages. That's because all related subcommands (`encode-call`, `estimate-fee`, `encode-message`) now require runtime code (mostly to get call dispatch weight). And we have no access to K/P/R/W runtime from relay.

The idea is to pass all missing data using cli parameters, or read it directly from the node (reading weights is currently not possible for us iirc, but still may be a valid option for other parameters).

svyatonik:
closes #1319 

The main thing here is `EncodedOrDecodedCall`, which has the same encoding as on-chain calls. But it has two variants - the call may be decoded (that's what we have been using previously). Or it may be encoded and only needs to be decoded when it needs to. In terms of relay subcommands, it means that if you're using pre-encoded calls in cli options and specify their dispatch weight, we'll never need to decode this call (this is the main problem from #1319 - we can't decode Rococo/Wococo/Kusama/Polkadot calls, because we have no access to the runtime from the relay).

Examples:
```bash
svyatonik@xwnotebook:~/dev/rococo-wococo.test$ ./bin/substrate-relay encode-message rococo-to-wococo call --sender 5CUXMS6Ja5E8HGa8qq2jumboFKjqpjA8mrdzyANhk7P2mELp --dispatch-weight 10000000000 raw 00010442
INFO bridge Created Message Payload: MessagePayload {
    spec_version: 9140,
    weight: 10000000000,
    origin: CallOrigin::SourceAccount(
        122a87a73f92704a4556131f8c2d14a32dc1d624fc523ba4e528bff3f9dfe55f (5CUXMS6J...),
    ),
    dispatch_fee_payment: DispatchFeePayment::AtSourceChain,
    call: 0x00010442,
}
INFO bridge Encoded Message Payload: 0xb423000000e40b540200000002122a87a73f92704a4556131f8c2d14a32dc1d624fc523ba4e528bff3f9dfe55f001000010442
0xb423000000e40b540200000002122a87a73f92704a4556131f8c2d14a32dc1d624fc523ba4e528bff3f9dfe55f001000010442
svyatonik@xwnotebook:~/dev/rococo-wococo.test$ ./bin/substrate-relay estimate-fee rococo-to-wococo call --sender 5CUXMS6Ja5E8HGa8qq2jumboFKjqpjA8mrdzyANhk7P2mELp --dispatch-weight 0 raw 00010442
INFO bridge Created Message Payload: MessagePayload {
    spec_version: 9140,
    weight: 0,
    origin: CallOrigin::SourceAccount(
        122a87a73f92704a4556131f8c2d14a32dc1d624fc523ba4e528bff3f9dfe55f (5CUXMS6J...),
    ),
    dispatch_fee_payment: DispatchFeePayment::AtSourceChain,
    call: 0x00010442,
}
INFO bridge Encoded Message Payload: 0xb4230000000000000000000002122a87a73f92704a4556131f8c2d14a32dc1d624fc523ba4e528bff3f9dfe55f001000010442
INFO bridge Fee: Balance(293479234556)
293479234556
svyatonik@xwnotebook:~/dev/rococo-wococo.test$ ./bin/substrate-relay send-message rococo-to-wococo --source-signer //Alice --dispatch-weight 100000000000 raw 00010442
INFO bridge Created Message Payload: MessagePayload {
    spec_version: 9140,
    weight: 100000000000,
    origin: CallOrigin::SourceAccount(
        d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d (5GrwvaEF...),
    ),
    dispatch_fee_payment: DispatchFeePayment::AtSourceChain,
    call: 0x00010442,
}
INFO bridge Encoded Message Payload: 0xb423000000e876481700000002d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d001000010442
INFO bridge Sending message to Wococo. Lane: [0, 0, 0, 0]. Size: 177. Dispatch weight: 100000000000. Fee: 1,173,439,257,663
INFO bridge The source account (d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d (5GrwvaEF...)) balance will be reduced by (at most) 1173439257663 (message fee) + 31605393717 (tx fee	) = 1205044651380 Rococo tokens
INFO bridge Signed Rococo Call: 0xc502bd028400d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d0180c4412df47caab679790c7992755661d891bc73ed75b632a46372ab23e2200fafec0946b26eaa091bbd04671261c06c15911e0f4e1a421d6943c66fc9c3ca850008002c0300000000b423000000e876481700000002d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d0010000104423f1c6e36110100000000000000000000

```

### DIFF
--- a/primitives/polkadot-core/src/lib.rs
+++ b/primitives/polkadot-core/src/lib.rs
@@ -17,7 +17,7 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
 use bp_messages::MessageNonce;
-use bp_runtime::Chain;
+use bp_runtime::{Chain, EncodedOrDecodedCall};
 use frame_support::{
 	dispatch::Dispatchable,
 	parameter_types,
@@ -228,8 +228,12 @@ pub type SignedBlock = generic::SignedBlock<Block>;
 pub type Balance = u128;
 
 /// Unchecked Extrinsic type.
-pub type UncheckedExtrinsic<Call> =
-	generic::UncheckedExtrinsic<AccountAddress, Call, Signature, SignedExtensions<Call>>;
+pub type UncheckedExtrinsic<Call> = generic::UncheckedExtrinsic<
+	AccountAddress,
+	EncodedOrDecodedCall<Call>,
+	Signature,
+	SignedExtensions<Call>,
+>;
 
 /// Account address, used by the Polkadot-like chain.
 pub type Address = MultiAddress<AccountId, ()>;
@@ -336,12 +340,12 @@ where
 
 	fn pre_dispatch(
 		self,
-		who: &Self::AccountId,
-		call: &Self::Call,
-		info: &DispatchInfoOf<Self::Call>,
-		len: usize,
+		_who: &Self::AccountId,
+		_call: &Self::Call,
+		_info: &DispatchInfoOf<Self::Call>,
+		_len: usize,
 	) -> Result<Self::Pre, TransactionValidityError> {
-		Ok(self.validate(who, call, info, len).map(|_| ())?)
+		Ok(())
 	}
 }
 
--- a/relays/bin-substrate/src/chains/kusama.rs
+++ b/relays/bin-substrate/src/chains/kusama.rs
@@ -104,7 +104,7 @@ impl CliChain for Kusama {
 		match message {
 			encode_message::MessagePayload::Raw { data } => MessagePayload::decode(&mut &*data.0)
 				.map_err(|e| anyhow!("Failed to decode Kusama's MessagePayload: {:?}", e)),
-			encode_message::MessagePayload::Call { mut call, mut sender, weight } => {
+			encode_message::MessagePayload::Call { mut call, mut sender, dispatch_weight } => {
 				type Source = Kusama;
 				type Target = relay_polkadot_client::Polkadot;
 
@@ -116,13 +116,13 @@ impl CliChain for Kusama {
 					bridge::KUSAMA_TO_POLKADOT_INDEX,
 				);
 				let call = Target::encode_call(&call)?;
-				let weight = weight.map(Ok).unwrap_or_else(|| {
-					Err(anyhow::format_err!("Please specify weight of the encoded Polkadot call"))
+				let dispatch_weight = dispatch_weight.map(Ok).unwrap_or_else(|| {
+					Err(anyhow::format_err!("Please specify dispatch weight of the encoded Polkadot call"))
 				})?;
 
 				Ok(send_message::message_payload(
 					spec_version,
-					weight,
+					dispatch_weight,
 					origin,
 					&call,
 					DispatchFeePayment::AtSourceChain,
--- a/relays/bin-substrate/src/chains/kusama.rs
+++ b/relays/bin-substrate/src/chains/kusama.rs
@@ -14,6 +14,9 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
 
+use anyhow::anyhow;
+use bp_message_dispatch::{CallOrigin, MessagePayload};
+use bp_runtime::EncodedOrDecodedCall;
 use codec::Decode;
 use frame_support::weights::{DispatchClass, DispatchInfo, Pays, Weight};
 use relay_kusama_client::Kusama;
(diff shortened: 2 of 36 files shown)


svyatonik:
I'll merge this without approval for now - it needs to be tested on Rialto+Millau and we already have a bunch of unreviewed PRs. So maybe it'll be easier to review them in a `polkadot-staging-update` PR at once.

primitives/polkadot-core/src/lib.rs
```diff
@@ -336,12 +340,12 @@ where
...
-		who: &Self::AccountId,
-		call: &Self::Call,
-		info: &DispatchInfoOf<Self::Call>,
-		len: usize,
+		_who: &Self::AccountId,
+		_call: &Self::Call,
+		_info: &DispatchInfoOf<Self::Call>,
+		_len: usize,
 	) -> Result<Self::Pre, TransactionValidityError> {
-		Ok(self.validate(who, call, info, len).map(|_| ())?)
```
acatangiu, line 344: Do we forgo validation completely? Is this because we can't validate encoded call?

relays/bin-substrate/src/chains/kusama.rs
```diff
@@ -32,13 +37,15 @@ use crate::cli::{
 pub(crate) const SYSTEM_REMARK_CALL_WEIGHT: Weight = 2 * 1_345_000;
 
 impl CliEncodeCall for Kusama {
-	fn encode_call(call: &Call) -> anyhow::Result<Self::Call> {
+	fn encode_call(call: &Call) -> anyhow::Result<EncodedOrDecodedCall<Self::Call>> {
 		Ok(match call {
+			Call::Raw { data } => EncodedOrDecodedCall::Encoded(data.0.clone()),
```
acatangiu, line 42: so this fn only returns `EncodedOrDecodedCall::Encoded` variant for `Call::Raw`, right?

relays/bin-substrate/src/chains/millau.rs
```diff
@@ -25,24 +25,27 @@ use crate::cli::{
...
 use frame_support::weights::{DispatchInfo, GetDispatchInfo};
 use relay_millau_client::Millau;
 use sp_version::RuntimeVersion;
 
 impl CliEncodeCall for Millau {
-	fn encode_call(call: &Call) -> anyhow::Result<Self::Call> {
+	fn encode_call(call: &Call) -> anyhow::Result<EncodedOrDecodedCall<Self::Call>> {
 		Ok(match call {
-			Call::Raw { data } => Decode::decode(&mut &*data.0)?,
+			Call::Raw { data } => Self::Call::decode(&mut &*data.0)?.into(),
```
acatangiu, line 37: I'm confused, for Millau the same trait `fn encode_call()` actually decodes the call when raw data is provided?

Why not use `EncodedOrDecodedCall::Encoded(data.0.clone())` variant like in Kusama?

primitives/polkadot-core/src/lib.rs
svyatonik: Yeah - this function is only called within runtime and we're using `SignedExtensions` just to get the encoded data (so we don't need any validation at all). More over, the default `SignedExtension::validate` actually does nothing: https://github.com/paritytech/substrate/blob/master/primitives/runtime/src/traits.rs#L862-L879

relays/bin-substrate/src/chains/kusama.rs
svyatonik: That's correct. Why is it important for us - we can't maintain the whole copy of Polkadot/Kusama `enum Call` here, in substrate code. Instead, we allow to encode call externally and use this `raw` variant. We can't decode it here, because for most of calls, decode will fail (except for this `remark`). So yes - it only returns `EncodedOrDecodedCall::Encoded` for `Call::Raw`

relays/bin-substrate/src/chains/millau.rs
svyatonik: This can be changed. But:
1) imo earlier error = better error. Since `Call` of Rialto/Millau is the runtime call (not the mocked call as for the Polkadot/Kusama), decode only fail if the call is invalid. It is better to see decoding error here, than some strange error later (e.g. on the bridged chain during dispatch);
2) most of times (except when sending messages manually or using sending max-dispatch-weight messages) in Rialto<>Millau bridge, we'll be calling `get_dispatch_info()` on this call to get its dispatch weight. So it needs to be decoded anyways.

Feel free to insist, if you think it'll be better to change it to the `Encoded` here, though :)

PR #1323, opened 2022-02-18
Try fix duplicate header relay

fewensa:
In our tests, it was found that if multiple bridges are started, many meaningless repeated header relays will be sent. After making this modification, this problem will be alleviated.

Related:
- https://github.com/darwinia-network/parity-bridges-common/pull/12
- https://github.com/darwinia-network/parity-bridges-common/pull/10
- https://github.com/darwinia-network/bridger/issues/344

### DIFF
--- a/relays/messages/src/message_race_delivery.rs
+++ b/relays/messages/src/message_race_delivery.rs
@@ -72,6 +72,7 @@ pub async fn run<P: MessageLane, Strategy: RelayStrategy>(
 			max_messages_size_in_single_batch: params.max_messages_size_in_single_batch,
 			relay_strategy: params.relay_strategy,
 			latest_confirmed_nonces_at_source: VecDeque::new(),
+			latest_updated_confirm_nonce_at_source: None,
 			target_nonces: None,
 			strategy: BasicStrategy::new(),
 		},
@@ -251,6 +252,8 @@ struct MessageDeliveryStrategy<P: MessageLane, Strategy: RelayStrategy, SC, TC> 
 	/// Latest confirmed nonces at the source client + the header id where we have first met this
 	/// nonce.
 	latest_confirmed_nonces_at_source: VecDeque<(SourceHeaderIdOf<P>, MessageNonce)>,
+	/// latest updated confirm nonce at source chain
+	latest_updated_confirm_nonce_at_source: Option<MessageNonce>,
 	/// Target nonces from the source client.
 	target_nonces: Option<TargetClientNonces<DeliveryRaceTargetNoncesData>>,
 	/// Basic delivery strategy.
@@ -280,6 +283,10 @@ impl<P: MessageLane, Strategy: RelayStrategy, SC, TC> std::fmt::Debug
 			.field("max_messages_weight_in_single_batch", &self.max_messages_weight_in_single_batch)
 			.field("max_messages_size_in_single_batch", &self.max_messages_size_in_single_batch)
 			.field("latest_confirmed_nonces_at_source", &self.latest_confirmed_nonces_at_source)
+			.field(
+				"latest_updated_confirm_nonce_at_source",
+				&self.latest_updated_confirm_nonce_at_source,
+			)
 			.field("target_nonces", &self.target_nonces)
 			.field("strategy", &self.strategy)
 			.finish()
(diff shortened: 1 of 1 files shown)


svyatonik:
Hey! Thanks for the PR, but I have some questions:
1) what do you mean by "multiple bridges" - multiple relay processes that are serving the same lane? Is it possible to illustrate this case with test or at least with some text/diagram here? Is it only specific to the case when two relayers are running?
2) the only difference between `latest_updated_confirm_nonce_at_source` and `latest_confirmed_nonces_at_source.back().map(|_, n| n)` is that: (1) `latest_updated_confirm_nonce_at_source` is never set to `None` - am I right?
3) so far, this PR looks to me like revert of  #719, which has replaced `Option<MessageNonce>` with `VecDeque<HeaderId, MessageNonce>`. So imo it resurrects the issue, described in #719 (unless I'm missing something);
4) I've tried to read related issues and PRs, still not 100% sure, what's the issue is (hope you'll make it clear by introducing a test or detailed description/diagram, mentioned in (1)). But probably the 'problem' you see may be in [this code](https://github.com/paritytech/parity-bridges-common/blob/ddb4517e133c6d194d42733b659ff7624e22133e/relays/messages/src/message_race_delivery.rs#L321-L331). IIUC, it'll ask for more headers even if there are no new messages to deliver, but still some rewards are not confirmed to the target chain. Could you, please, give it a try? I mean - change code so that it'll always return `self.strategy.required_source_header_at_target(current_best)`? It may be incorrect and will need a double-check + possibly proper fix, just want to see if it solves your problem

fewensa:
1. Yes, "multiple bridges" is means "multiple relayers".
2. The charts you can click [darwinia bridgecrabgranpa submit_finality_proof extrinsics](https://darwinia.subscan.io/extrinsic?address=&module=bridgecrabgrandpa&call=submit_finality_proof&result=all&signedChecked=signed%20only&startDate=&endDate=&startBlock=&endBlock=&timeType=date&version=1180&page=132) to see our bridge header relay records. too more header relay at the same time sent by multiple bridges/relayers.
3. I guess the reason is:


You can read the log in this issue https://github.com/darwinia-network/bridger/issues/344

```
[2021-12-06T23:53:35Z TRACE finality_relay::finality_loop] Considering range of headers (8186014; 8186010]
[2021-12-06T23:53:35Z TRACE finality_relay::finality_loop] Can not improve selected Crab finality proof None. No unjustified headers and recent proofs
[2021-12-06T23:53:35Z DEBUG messages_relay::message_lane_loop] Received state from Crab node: ClientState { best_self: HeaderId(8186018, 0x7315a7e75dce411c14b8d5e3f128ac0da1630496507111dfbf2a259aae8bf713), best_finalized_self: HeaderId(8186015, 0x4727ea887eef10242d2701b1483ffe8d98a41e2e1530f13fc6183dd1fee58804), best_finalized_peer_at_best_self: HeaderId(6225408, 0xea1db9acef190c683e953e79186d20d5e72a725701c5d1bd5f15f0aedbb4f75b) }
[2021-12-06T23:53:35Z DEBUG messages_relay::message_race_loop] Asking Crab::MessagesDelivery about best message nonces
```

```
Considering range of headers (8186014; 8186010]
```

The source chain block number is great than source_confirmed_in_target_chain. so there https://github.com/paritytech/parity-bridges-common/blob/88d684d37ed044396b221b7d8f0661883e6f94b2/relays/messages/src/message_race_delivery.rs#L390 can not pass condition the result is `latest_confirmed_nonces_at_source` not pop front.
then https://github.com/paritytech/parity-bridges-common/blob/88d684d37ed044396b221b7d8f0661883e6f94b2/relays/messages/src/message_race_delivery.rs#L414-L420 will always return stored latest confirm block in `latest_confirmed_nonces_at_source`

svyatonik:
Thanks for the explanation. Still, imo this fix reverts #719, which is not good.

> 4. I've tried to read related issues and PRs, still not 100% sure, what's the issue is (hope you'll make it clear by introducing a test or detailed description/diagram, mentioned in (1)). But probably the 'problem' you see may be in [this code](https://github.com/paritytech/parity-bridges-common/blob/ddb4517e133c6d194d42733b659ff7624e22133e/relays/messages/src/message_race_delivery.rs#L321-L331). IIUC, it'll ask for more headers even if there are no new messages to deliver, but still some rewards are not confirmed to the target chain. Could you, please, give it a try? I mean - change code so that it'll always return `self.strategy.required_source_header_at_target(current_best)`? It may be incorrect and will need a double-check + possibly proper fix, just want to see if it solves your problem

I've made a quick and dirty fix for that - please see #1328. Is it possible for you to test if it is fixing your issue (I mean - instead of fix from this PR)? #1328 is not in final stage yet - it is definitely missing tests and I need to think whether it breaks something else, but I believe, it shall fix your issue.

fewensa:
Thanks, I'll test use your pull request.

fewensa:
Follow #1328

PR #1325, opened 2022-02-21
Implement Decode for Kusama/Polkadot SignedExtensions (otherwise transaction resubmitter panicks)

### LINKED ISSUE #1314: Potential panic in Kusama/Polkadot transactions resubmitter
Needs an additional check, but I have a doubts about [this line](https://github.com/paritytech/parity-bridges-common/blob/bdeedb7ab96558a96b8dfec2c7a65476116c2a03/primitives/polkadot-core/src/lib.rs#L263). Our tx resubmitter is decoding pool transactions, so iiuc we may see this expect failure.

svyatonik:
closes #1314


svyatonik:
// to be reviewed as a part of `polkadot-staging` branch update

Issue #1162, opened 2021-09-24
Prevent concurrent finality/messages submissions via `SignedExtension`


tomusdrw:
Related: #983 

Currently we allow anyone to submit a finality proof or deliver valid messages. This is nice, but in real world in case many relayers (per-lane) are present they are going to simply compete with each other to submit the information.

The winning relayer is going to be rewarded (also the submission costs might be returned `PaysFee::No` (see #983)), but loosing relayers are going to pay costs for failed dispatch.

To alleviate the issue we can easily prevent failing transactions from being included by adding another `SignedExtension` which would invalidate such submission transactions that are known to fail. We have to be careful to avoid too many storage reads (to prevent block authors DoS via validity checking), but overall it seems it should be possible.

svyatonik:
Imo we shall only enable this extension on dedicated bridges parachain. But since our parachain bridge is currently on hold, I'm going to try this extension on Millau and run several concurrent relayers to see it in action.

EmmanuellNorbertTulbure:
@svyatonik Do we have this already or something that still needs to be done?

svyatonik:
> Imo we shall only enable this extension on dedicated bridges parachain. But since our parachain bridge is currently on hold, I'm going to try this extension on Millau and run several concurrent relayers to see it in action.

Yes, it needs to be ported to Rococo/Wococo bridge (and to P/K later) + audited

bkontur:
@svyatonik 
I have TODO for `generate_bridge_reject_obsolete_headers_and_messages`,
so this issue is about it?

svyatonik:
Yes. I believe I wanted to add it myself, but there's a problem since it is in `brtidge-runtime-common` and back then you've been avoiding this crate because it had XCMv3 dependency. So yes - it still needs to be added to R<>W, but I haven't had a chance to look at its state today :)

We'll also have another extension related to fee payments (#1657).

bkontur:
actually, some time ago, I was updating to xcmv3 with (pub reexports HaulBlob,..), so I am using this `bridge-runtime-common` because of verify proof stuff + plus lots of other stuff,

[...]

PR #1352, opened 2022-03-11
CheckBridgedBlockNumber signed extension to reject duplicate header-submit transactions

svyatonik:
related to #1162 (we'll need a couple of other extensions to close the issue)

Things to do:
- [x] actually test it;
- [x] think about adding `BestBlockNumber` storage entry (or changing existing `BestFinalized` to store both block hash and number) to the grandpa pallet. This would allow us to use single storage read instead of two when validating. This needs to be done carefully (in a separate PR), because it may break existing deployments, which will need an upgrade in this case. UPD: extracted to #1368;
- [x] is there any other way than using macro to declare runtime-specific extension? Static type map or something like that? Maybe some existing macro. UPD: why do we need macro at all, instead of introducing multiple extensions - i.e. `CheckBridgedBlockNumber<Runtime, WestendGrandpaInstance>` and `CheckBridgedBlockNumber<Runtime, RialtoGrandpaInstance>`? That's because every extension has associated `IDENTIFIER` const, which has to be unique. The outcome of not being unique, iirc is invalid metadata. We could use `adt_const_params` rust feature if we were on nightly, but we are not. So the only solution I see now, is to have this single, multi-instance extension that can only be created using macro (as I said before - worth exploring).

Update related to last checkbox - I was thinking that possibly using `impl_for_tuples` would help. So something like:
```rust
trait FilterCall<Call> {
  fn is_valid(call: &Call) -> Option<bool>;
}

// implement `FilterCall` for Pallet<T, I> and for tuples, so we could use something like `(WithWestendGrandpa, WithRialtoGrandpa)`

type SignedExtension = SignedExtension<(WithWestendGrandpa, WithRialtoGrandpa)>;
```

But I've found no way to match (and deconstruct) runtime call when you're at pallet level. So given runtime call, but no knowledge of runtime, I can't do match inside `FilterCall::is_valid`. So I see no other option, but to use macro for now.


svyatonik:
Merging with deferred review

serban300:
Related to:
>     * [x]  is there any other way than using macro to declare runtime-specific extension? Static type map or something like that? Maybe some existing macro. UPD: why do we need macro at all, instead of introducing multiple extensions - i.e. `CheckBridgedBlockNumber<Runtime, WestendGrandpaInstance>` and `CheckBridgedBlockNumber<Runtime, RialtoGrandpaInstance>`? That's because every extension has associated `IDENTIFIER` const, which has to be unique. The outcome of not being unique, iirc is invalid metadata. We could use `adt_const_params` rust feature if we were on nightly, but we are not. So the only solution I see now, is to have this single, multi-instance extension that can only be created using macro (as I said before - worth exploring).
> 
> 
> Update related to last checkbox - I was thinking that possibly using `impl_for_tuples` would help. So something like:
> 
> ```rust
> trait FilterCall<Call> {
>   fn is_valid(call: &Call) -> Option<bool>;
> }
> 
> // implement `FilterCall` for Pallet<T, I> and for tuples, so we could use something like `(WithWestendGrandpa, WithRialtoGrandpa)`
> 
> type SignedExtension = SignedExtension<(WithWestendGrandpa, WithRialtoGrandpa)>;
> ```
> 
> But I've found no way to match (and deconstruct) runtime call when you're at pallet level. So given runtime call, but no knowledge of runtime, I can't do match inside `FilterCall::is_valid`. So I see no other option, but to use macro for now.

We might be able to match the call with something like:
```
fn validate<T: Config<I>, I: 'static + Callable<T, Call = crate::Call<T, I>>>(call: T::Call)
where
	T::Call: IsSubType<CallableCallFor<I, T>>,
	<T as pallet_bridge_grandpa::Config<T::BridgesGrandpaPalletInstance>>::BridgedChain:
		bp_runtime::Chain<
			BlockNumber = RelayBlockNumber,
			Hash = RelayBlockHash,
			Hasher = RelayBlockHasher,
		>,
{
	if let Some(crate::Call::<T, I>::submit_parachain_heads {
		ref at_relay_block,
		ref parachains,
		..
	}) = call.is_sub_type()
	{}
}
```

This example is for parachains but it can be adapted for the grandpa pallet. I will do some experiments these days to see if it works.

serban300:
The PR looks good and the comment related to reducing the number of macros has been addressed as part of #1352 . Marking the PR as reviewed.

serban300 (commented):
The PR looks good and the comment related to reducing the number of macros has been addressed as part of https://github.com/paritytech/parity-bridges-common/pull/1519 . Marking the PR as reviewed.

PR #1373, opened 2022-04-13
add api data() for inbound_lane

xiaoch05:
in outbound lane we have opened the data() api
but in the inbound lane this api is not opened, it's very useful for application using the bridge.

### DIFF
--- a/modules/messages/src/inbound_lane.rs
+++ b/modules/messages/src/inbound_lane.rs
@@ -71,6 +71,11 @@ impl<S: InboundLaneStorage> InboundLane<S> {
 		InboundLane { storage }
 	}
 
+	/// Get this lane data
+	pub fn data(&self) -> InboundLaneData<S::Relayer> {
+		self.storage.data()
+	}
+
 	/// Receive state of the corresponding outbound lane.
 	pub fn receive_state_update(
 		&mut self,


svyatonik (approved):

svyatonik:
> it's very useful for application using the bridge.

@xiaoch05 Is there an example of such application? `InboundLane` is a private struct, so you application must be somewhere in the pallet code, right? Can you provide us with details of you app and how it uses the `InboundLane`? The fact is that this commit generates annoying compilation warning and unless we'll have a strong evidence of how/why it is used, I'd like this commit to be reverted.

Thank you!

xiaoch05:
> > it's very useful for application using the bridge.
> 
> @xiaoch05 Is there an example of such application? `InboundLane` is a private struct, so you application must be somewhere in the pallet code, right? Can you provide us with details of you app and how it uses the `InboundLane`? The fact is that this commit generates annoying compilation warning and unless we'll have a strong evidence of how/why it is used, I'd like this commit to be reverted.
> 
> Thank you!

Hi @svyatonik , the previous version supported getting this interface for `InboundLane`, and we were using it [here](https://github.com/darwinia-network/darwinia-common/blob/main/frame/wormhole/backing/s2s/src/lib.rs#L391), emitting an event to correspond with the message in the source chain. But I find that we are able to get all the information we need from these events https://github.com/paritytech/parity-bridges-common/blob/master/modules/dispatch/src/lib.rs#L114-L135, so I changed my previous thought that this interface is not that important for the application. If this warning is more annoying, I think it would be better to revert this modification too.

svyatonik:
Thank you. Yep, I'll revert it then to avoid unnecessary warnings :)

PR #1374, opened 2022-04-13
FinalityEngine in substrate relay

svyatonik:
extracted from #1372

Since we'll have two finality "engines" now (GRANDPA and BEEFY), I've introduced `FinalityEngine` trait in this PR. This would allow us to reuse existing finality relay code for both GRANDPA and BEEFY.  `Beefy` engine will be added in #1372 

Important note - there's actually no new code in this PR. It is just about moving code around + adding generic/associated `Engine` argument to some structs.


relays/lib-substrate-relay/src/finality/source.rs
```diff
@@ -38,13 +37,19 @@ pub type RequiredHeaderNumberRef<C> = Arc<Mutex<<C as bp_runtime::Chain>::BlockN
...
-					HeaderOf<<P as SubstrateFinalitySyncPipeline>::SourceChain>,
-				>,
+				Item = <<P as SubstrateFinalitySyncPipeline>::FinalityEngine as Engine<
+					<P as SubstrateFinalitySyncPipeline>::SourceChain,
+				>>::FinalityProof,
 			> + Send,
 	>,
 >;
 
+/// TODO
```
acatangiu, line 47:
```suggestion
/// Substrate finality proof. Specific to the used `FinalityEngine`.
```

acatangiu (approved):
Awesome!

PR #1376, opened 2022-04-22
Remove support for encoded-call messaging from relay and runtime integration code

svyatonik:
I'm opening this PR early (it'll be merged not earlier than on monday evening) to give those who are building using our codebase a chance to prepare. So we'll concentrate on supporting XCM dispatch in our bridge in next month(s). What it means:
- we've decided to drop support for encoded-call messaging from relay and runtime integration code, because it'll be too hard now to support the stuff that we're not going to use in the near future;
- we'll remove `pallet-bridge-dispatch` and `pallet-token-swap` (which depends on the former pallet) from our codebase;
- core pallets of our bridge (`pallet-bridge-grandpa` and `pallet-bridge-messages`) are staying and will likely keep the same API (this may be changed in next XCM-related milestones though). It means that if you're using encoded-calls messaging, nothing changes for your runtime, except that you won't be able to use following runtime-related crates from our repo: `bp-message-dispatch`, `pallet-bridge-dispatch` and `bridge-runtime-common` which will be changed to support XCM delivery. We'll make a tag before merging this PR and a comment in top-level readme, referencing this tag;
- headers/messages/headers+messages relay will likely keep supporting both options - XCM messaging and encoded-call messaging. However all subcommands that we were using for tests (`send-message`, `encode-message`, `estimate-fee`) will be changed to support XCM messaging.

Pinging guys from Darwinia, as I know you're using some of our crates: @AurevoirXavier @fewensa @HackFisher


HackFisher:
@svyatonik We are replacing the offchain relay part using subquery indexing, including optimization for mandatory blocks and Justification.

For on-chain part, as I understand, the changes will be related to the format of the bridge message, from dispatch call to uenitfied XCM. We'll keep an eye on this, and see how to upgrade the protocols.

Currently using production bridges, Darwinia <> Crab, and Crab <> Crab Parachain is on the way.

svyatonik:
@acatangiu It should be ready for review. The result of this PR:
- we have dropped encoded-call messaging. There may be some leftovers though, but in general there's no "official" runtime+relay support for that from now on; 
- on our test deployments we are now generating + delivering simple blobs;
- dispatch of blob = simple logging for now. Later on, when we'll plugin XCM, we'll be using XCM dispatcher instead

acatangiu (approved):

PR #1378, opened 2022-04-27
Switch to gav-xcm-v3 branch to be able to test bridges + XCMv3 integration

svyatonik:
I've been trying first to jsut change all our references to `gav-xcm-v3` Polkadot/Cumulus branch, but both Substrate and Polkadot crates are duplicated in this case, because Polkadot/Cumulus code still references `master` branch. So solution is to patch references in the workspace' `Cargo.toml` temporarily, until XCMv3 PR is merged.

The goal of this PR is not to have working XCM on our chains, but to have code which we may start using for our experiments.

### DIFF
--- a/bin/rialto-parachain/node/src/command.rs
+++ b/bin/rialto-parachain/node/src/command.rs
@@ -256,7 +256,6 @@ pub fn run() -> Result<()> {
 
 			Ok(())
 		},
-
 		Some(Subcommand::Benchmark(cmd)) => {
 			let runner = cli.create_runner(cmd)?;
 			match cmd {


.gitlab-ci.yml
```diff
@@ -144,7 +144,8 @@ test:
     - time cargo fetch
     - time cargo fetch --manifest-path=`cargo metadata --format-version=1 | jq --compact-output --raw-output  ".packages[] | select(.name == \"polkadot-test-runtime\").manifest_path"`
     - time cargo fetch --manifest-path=`cargo metadata --format-version=1 | jq --compact-output --raw-output  ".packages[] | select(.name == \"polkadot-runtime\").manifest_path"`
-    - CARGO_NET_OFFLINE=true time cargo test --verbose --workspace
+    - time cargo fetch --manifest-path=`cargo metadata --format-version=1 | jq --compact-output --raw-output  ".packages[] | select(.name == \"kusama-runtime\").manifest_path"`
```
svyatonik, line 147: Originally this cargo fetch has been added because we had problems with concurrent runtime builds (we build 6 now :/) and `cargo metdata` was called during build, which doesn't work well with concurrent builds. I've added `SKIP_POLKADOT_RUNTIME_WASM_BUILD` and etc because we don't need these runtimes in `cargo test`.

acatangiu (approved):

PR #1379, opened 2022-04-27
Xcm in Rialto<>Millau bridge

svyatonik:
What does this PR brings:
1) XCM pallet is added to both Rialto and Millau;
2) we're now able to send XCM between chains using `send` call of the XCM pallet. Hint: dest is `MultiLocation { parents: 1, interior: X1(GlobalConsensus(Kusama)) }` for Rialto -> Millau messages and `MultiLocation { parents: 1, interior: X1(GlobalConsensus(Polkadot)) }` for Millau -> Rialto;
3) I haven't tried to send meaningful XCMs - just `Trap(42)`, which just throws an error when dispatched. I'm more interested in seeing that the bridge delivers exactly what is sent. Even more, current routers will probably be dropped when we'll start wirking on bridge hub;
4) I've removed `InstantCurrencyPayments`, since we're not going to use it anymore. When sending XCM using XCM pallet, the XCM engine is responsible for taking fees. The plan is to pay relayer rewards from the treasury (https://github.com/paritytech/polkadot/pull/4097#discussion_r866263870). But right now it is not handled, because we need to deal with that on bridge-hub parachain, which is missing now;
5) fee for messages that are sent using messages pallet is not withdrawn from the sender account (so they're free). In the future, we'll close this option. It doesn't mean removal of the `send_message` call from the pallet - just blocking it in our test deployments (and on real deployments later).


acatangiu (commented):
Looks good afaict :+1:

PR #1380, opened 2022-04-28
Update Substrate/Polkadot/Cumulus references (for BEEFY on Rialto)

acatangiu:
new refs:
  cumulus: 423a7f5a3777b21bd8193cd3e8bd54afe088c633
  polkadot: d3e67bb264f866215d9897764094b616117ba69f
  substrate: 4d3d480d8149fbca2834d05cc050a01f9a25afbf

use correct MMR primitives crate

update lockfile to compatible crate versions

integrate upstream changes

Signed-off-by: acatangiu <adrian@parity.io>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -327,9 +327,9 @@ checksum = "e91831deabf0d6d7ec49552e489aed63b7456a7a3c46cff62adad428110b0af0"
 
 [[package]]
 name = "async-trait"
-version = "0.1.52"
+version = "0.1.53"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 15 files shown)


svyatonik:
@acatangiu I've actually created similar #1378, which also updates refs. But we're switching to `gav-xcm-v3` branch (Polkadot and Cumulus) there. It also has this switch-to-sp-mmr-primitives crate commit. Let me know what do you think of using that PR instead - without that switch we'll be unable to work with XCMv3 features.

If you need something else, that is missing from #1378 we can just branch off from `gav-xcm-v3` and add required commits there.

acatangiu:
> If you need something else, that is missing from #1378 we can just branch off from `gav-xcm-v3` and add required commits there.

yes, i think we'll do that then

PR #1388, opened 2022-05-04
Generalize error text

pepyakin:
Right now, each time there is an error while executing `substrate-relay`
it will be reported as:

    ERROR bridge Failed to start relay: <Actual cause of error>

This is the case even if the invoked command did not have anything to do
with starting a relayer. Thus this removes this text. Now something like
this would be written:

    ERROR bridge <Actual cause of error>

### DIFF
--- a/relays/bin-substrate/src/main.rs
+++ b/relays/bin-substrate/src/main.rs
@@ -26,6 +26,6 @@ fn main() {
 	let run = command.run();
 	let result = async_std::task::block_on(run);
 	if let Err(error) = result {
-		log::error!(target: "bridge", "{}", error);
+		log::error!(target: "bridge", "substrate-relay: {}", error);
 	}
 }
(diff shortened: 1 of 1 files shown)


svyatonik (approved):
Imo it's more about binary (which is called `substrate-relay`), not the subcommand that actually perform relay-related tasks. I'll approve - feel free to merge if you think it's better than what we have now

pepyakin:
ah, it's the same as in

```
find not-exists
find: ‘not-exists’: No such file or directory
```

well, that was not obvious. Also, I think it still does not have anything to do with "starting".

pepyakin:
Perhaps, https://github.com/paritytech/parity-bridges-common/pull/1388/commits/1260f5db3088d781c6eeb43497e46a363e267328 makes more sense?

PR #1397, opened 2022-05-16
Fix matrix notifier startup

svyatonik:
(currently it fails to start with "`pick': Couldn't find handler for: thin, puma, reel, HTTP, webrick. (LoadError)" message)


acatangiu (commented):
:+1:

PR #1405, opened 2022-05-20
Complex RialtoParachain <> Millau relay

svyatonik:
This PR introduces new complex RialtoParachain <> Millau relay. It is not yet in deployments - I'd like to continue in a follow up PR


serban300:
#1535 was merged and the parachains code looks good. Marking this PR as reviewed.

PR #1423, opened 2022-05-30
Ensure that the bridge GRANDPA pallet is initialized in the finality relay

svyatonik:
There's one inconvenient thing in GRANDPA pallet initialization:
1) the pallet has `IsHalted` flag, which, if set to `true`, means that the pallet will reject (almost) all transactions. If the value of this flag is missing from the runtime storage, it takes default value, which is `false` => pallet is in operational mode;
2) there's another procedure in the pallet, called "initialization", which saves initial header and authorities set in the runtime storage. If pallet is not initialized, it'll reject all incoming headers, waiting for initialization first;
3) finally, there's pallet genesis config, which contains `init_data: Option<InitializationData { is_halted: bool, initial_header: H }>`. So I can't set `IsHalted` to `true` in genesis, without providing initial header.

What it means for the finality relay, is that it shall check both flags during startup - `IsInitialized` and `IsHalted`.

### DIFF
--- a/relays/lib-substrate-relay/src/finality/target.rs
+++ b/relays/lib-substrate-relay/src/finality/target.rs
@@ -53,10 +53,20 @@ impl<P: SubstrateFinalitySyncPipeline> SubstrateFinalityTarget<P> {
 	pub async fn ensure_pallet_active(&self) -> Result<(), Error> {
 		let is_halted = self.client.storage_value(P::FinalityEngine::is_halted_key(), None).await?;
 		if is_halted.unwrap_or(false) {
-			Err(Error::BridgePalletIsHalted)
-		} else {
-			Ok(())
+			return Err(Error::BridgePalletIsHalted)
 		}
+
+		let is_initialized =
+			super::initialize::is_initialized::<P::FinalityEngine, P::SourceChain, P::TargetChain>(
+				&self.client,
+			)
+			.await
+			.map_err(|e| Error::Custom(e.to_string()))?;
+		if !is_initialized {
+			return Err(Error::BridgePalletIsNotInitialized)
+		}
+
+		Ok(())
 	}
 }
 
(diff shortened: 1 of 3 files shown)


relays/lib-substrate-relay/src/finality/target.rs
```diff
@@ -53,10 +53,20 @@ impl<P: SubstrateFinalitySyncPipeline> SubstrateFinalityTarget<P> {
...
-			Ok(())
+			return Err(Error::BridgePalletIsHalted)
 		}
+
+		let is_initialized =
+			super::initialize::is_initialized::<P::FinalityEngine, P::SourceChain, P::TargetChain>(
+				&self.client,
+			)
+			.await
+			.map_err(|e| Error::Custom(e.to_string()))?;
```
serban300, line 64: Why return `Error::Custom` here and just propagate the initial error on `is_halted` check ?

serban300 (commented):
Looks good ! I only left a small question.

relays/lib-substrate-relay/src/finality/target.rs
svyatonik: Do you mean - why not return `BridgePalletIsHalted` here instead of `Error::Custom`? Because it may be any kind of issue, unrelated to the uninitialized pallet - e.g. transport error or decoding. We may add some additional info to the error though - e,g.`format!( "Failed to read pallet initialization state: {:?}", e)`

serban300: > Do you mean - why not return BridgePalletIsHalted here instead of Error::Custom?

I mean why not `await?` which would propagate the error returned by `P::FinalityEngine::is_initialized()` as we do for `P::FinalityEngine::is_halted(&self.client).await?`.

> Because it may be any kind of issue, unrelated to the uninitialized pallet - e.g. transport error or decoding. We may add some additional info to the error though - e,g.format!( "Failed to read pallet initialization state: {:?}", e)

True, but shouldn't this information be included in the error type ? For example if we're unable to read the initialization state we would get a `RpcError` I think. I'm not sure how detailed it would be though. Anyway adding some additional info might be safer, but I think we should do it both for `is_halted` and for `is_initialized` checks.

svyatonik: Ah, ok - got it now :) Yeah - seems like an unneeded conversion here.

serban300:
Marking the PR as reviewed.

PR #1432, opened 2022-06-01
Added tracked parachains filter

### LINKED ISSUE #1392: Think about introducing subset of tracked parachains in parachains finality pallet
Right now parachains finality pallet will track finality for all submitted parachains. Depending on its usage, it may be excessive - e.g. when building bridge hub parachains, we may only be interested in finality of the sibling bridge hub. My suggestion is to add an option to pallet configuration to configure subset of parachains we're interested in.

An alternate option is to provide an API (calls) for changing this subset.

### LINKED ISSUE #1393: Handle parachains offboarding in parachains finality pallet
#1068 

Parachains may be deregistered and then its head will be `None` (needs to be confirmed!!!) within `Paras::Heads` map. Right now we always expect it to be `Some(_)`. So it needs to be specially handled in the parachains finality pallet.

The other (less) important thing is that in theory the pararchain id may be reclaimed by the other pararchain. Even though it isn't planned, there's no guarantee that it'll never happens. So perhaps it's a good idea to have some API to completely purge previous pararchain from the pallet storage.

svyatonik:
closes #1392 
closes #1393 

The main change here is a new `TrackedParachains` configuration parameter, which allows restricting subset of tracked parachains from the bridged relaychain. It's pretty simple, so it closes #1392

I've also marked this PR to close #1393, since it is ab it related. A few words on that - after thinking about the issue, imo extending pallet to support parachain offboarding is too much (in terms of code complexity, performance, ...). Why:
1) the main usecase for this pallet would be to handle finality bridged bridgehub parachain. I.e. a single parachain, offboarding of which will cause a lot of (social) noise. So you can't miss that and use e.g. governance mechanisms to react;
2) if this pallet will be used to track multiple parachains finality (in other apps or in the future), the worst thing that we're risking is to store obsolete heads. This may be solved by using governance mechanism (or storage upgrade);
3) there's the possible situation when parachain-2 is registered using the same id that parachain-1 has been using before. This is unlikely to ever happen, but if it happen, imo it'll also cause a lot of noise and can't be missed.

### DIFF
--- a/bin/millau/runtime/src/lib.rs
+++ b/bin/millau/runtime/src/lib.rs
@@ -520,6 +520,7 @@ pub type WitRialtoParachainsInstance = ();
 impl pallet_bridge_parachains::Config<WitRialtoParachainsInstance> for Runtime {
 	type BridgesGrandpaPalletInstance = RialtoGrandpaInstance;
 	type ParasPalletName = RialtoParasPalletName;
+	type TrackedParachains = frame_support::traits::Everything;
 	type HeadsToKeep = HeadersToKeep;
 }
 


serban300 (commented):
Looks good. Marking the PR as reviewed.

PR #1436, opened 2022-06-06
Parachains pallet benchmarks

svyatonik:
TODOs:
- [ ] deduplicate `insert_relay_chain_header`;
- [ ] pruning benchmarks (refund if not pruned?);
- [ ] refund relay if some parachains are not updated.


serban300 (commented):
Looks good ! Marking it as reviewed.

PR #1446, opened 2022-06-08
Signed extension for rejecting obsolete messages pallet transactions

svyatonik:
TODOs:
- [x] reject obsolete message delivery transactions;
- [x] move tests to `bridges-runtime-common`?

This extension depends on types from the `bridges-runtime-common` crate, so it is also sitting there (unlike other extensions, which reside in the pallets code).

This PR breaks several runtime APIs (not storage). So it'll made existing relays incompatible with updated pallets and existing pallets incompatible with updated relays.


serban300 (commented):
Looks good ! Marking the PR as reviewed.

PR #1462, opened 2022-06-14
RialtoParachain<>Millau bridge maintenance dashboard

svyatonik:
TODOs:
- [ ] looks like some data from the new bridge has reached previous bridge dashboards (see e.g. "Millau relay balances" panel of the Rialto+Millau dashboard - there are 4 graphs instead of 2);
- [ ] something wrong with "Millau relay balances" at the new daashboard;
- [ ] "Whether with-RialtoParachain-parachains-pallet and RialtoParachain itself are on different forks" - this panel has permanent active alert (at least after ~5 minutes). Why? How do we compute this metric for parachains pallet?
- [ ] we need to adjust storage proof overhead for Millau (since hash is x2 there).


svyatonik:
Fixing 3rd issue isn't trivial - I'll open another PR for proper review. This PR shall be merged after that other one

PR #1465, opened 2022-06-15
Separate signers for different complex relay layers

### LINKED ISSUE #1463: Multiple stalls when both Rialto <> Millau and RialtoParachain <> Millau bridges are running
When both bridges are running at once, we have the following relayers running:
1) serving Rialto<>Millau bridge
  - Rialto -> Millau headers relay (on-demand);
  - Millau -> Rialto headers relay (on-demand);
  - messages relays;
2) serving RialtoParachain<>Millau bridge
  - Rialto -> Millau headers relay (on-demand);
  - Rialto -> Millau parachains relay (on-demand);
  - Millau -> Rialto headers relay (on-demand);
  - messages relays.

And now we have this `BridgeRejectObsoleteGrandpaHeader` (and similar) signed extensions at Millau, which makes transaction invalid if it provides obsolete Rialto header. From the above, you can see that we have two Rialto -> Millau headers relays running, so transactions invalidation happens sometimes. Apart from removing transaction itself, all dependent transactions are also invalidated. So if the same signer signs Rialto -> Millau header transaction and some other transaction right afterwards, that other one is also invalidated => relay stalls.

There are multiple ways to fix it:
1) add another instance of with-Rialto bridge GRANDPA pallet to Millau and use it in the RialtoParachain <> Millau bridge;
2) sign headers/parachains/messages transactions by different signers (cli option) in complex relay;
3) actually watch transactions and resubmit them if we see they're dropped. This would require massive code changes + it isn't super-compatible with stateless relay concept + we already have the similar mechanism (which is stall timeout);
4) make different layers of complex relay optional - e.g. disable headers relay in the RialtoParachain<>Millau bridge and rely on the parallel bridge running (which is bad, when it is running standalone);
5) ...

We need to evaluate risk of similar situation happening on the real deployments and fix it properly. Right now I'm thinking about implementing first option, but need to think a bit about that.

svyatonik:
closes #1463 

TODOs:
- [x] cleanup;
- [x] tests to illustrate how new cli options work;
- [x] expose additional relay balance metrics and use them in dashboards;
- [x] use in test deployments.

### DIFF
--- a/bin/millau/node/src/chain_spec.rs
+++ b/bin/millau/node/src/chain_spec.rs
@@ -143,6 +143,7 @@ fn endowed_accounts() -> Vec<AccountId> {
 		get_account_id_from_seed::<sr25519::Public>("George"),
 		get_account_id_from_seed::<sr25519::Public>("Harry"),
 		get_account_id_from_seed::<sr25519::Public>("Iden"),
+		get_account_id_from_seed::<sr25519::Public>("Ken"),
 		get_account_id_from_seed::<sr25519::Public>("Alice//stash"),
 		get_account_id_from_seed::<sr25519::Public>("Bob//stash"),
 		get_account_id_from_seed::<sr25519::Public>("Charlie//stash"),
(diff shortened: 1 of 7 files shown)


serban300 (commented):
Looks good and the idea behind this logic makes sense. Also some of the code was reviewed and adjusted as part of #1496 . Marking the PR as reviewed.

PR #1475, opened 2022-06-24
Store both block number and hash in best finalized storage value

### LINKED ISSUE #1368: Store best finalized header id (hash + number) instead of just hash in bridge-grandpa pallet
The signed extension from #1352 needs to know best finalized header number to validate transaction. Right now the only way to get it is to read best finalized header hash, read header by hash and then call `.number()` on it. So it two storage reads and one potentially complex decoding (e,g. on handoff headers). Changing type of `BestFinalized` from `BridgedBlockHash<T, I>` to `(BridgedBlockNumber<T, I>, BridgedBlockHash<T, I>)` would allow us to get best finalized header number with single db read without any complex decodings.

The fix should include function to handle storage upgrade + this upgrade must be called on all deployments,

svyatonik:
closes #1368 
needs some testing => draft

This breaks existing already deployed pallets.

### DIFF
--- a/modules/grandpa/src/benchmarking.rs
+++ b/modules/grandpa/src/benchmarking.rs
@@ -115,7 +115,7 @@ benchmarks_instance_pallet! {
 		let header: BridgedHeader<T, I> = bp_test_utils::test_header(header_number::<T, I, _>());
 		let expected_hash = header.hash();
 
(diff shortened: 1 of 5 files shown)


serban300 (commented):
Looks good. Marking it as reviewed.

Issue #1434, opened 2022-06-06
Add owner calls to the parachains pallet


svyatonik:
We have `set_owner` and `set_operating_mode`/`set_operational` (and linked owner role) in other bridge pallets (GRANDPA and messages). It'd be good to add similar methods to the parachains finality pallet as well.

PR #1477, opened 2022-06-27
Deduplicate owner calls for some pallets

serban300:
Some simplifications that should make it easier to implement #1434

Avoid duplicate function definitions for:
- ensure_owner_or_root()
- ensure_not_halted()
- set_owner()
- set_operating_mode() / set_operational()


primitives/runtime/src/lib.rs
```diff
@@ -257,6 +262,72 @@ pub fn storage_value_key(pallet_prefix: &str, value_name: &str) -> StorageKey {
...
 }
 
+#[derive(Encode, Decode, TypeInfo, PalletError)]
+pub enum OwnedBridgeModuleError {
+	/// All pallet operations are halted.
+	Halted,
+}
+
+pub trait OwnedBridgeModule<T: frame_system::Config> {
+	const NAME: &'static str;
```
svyatonik, line 272: Let's, please, rename this to something like `LOG_TARGET` or something like that. Otherwise it is unclear what it actually does :)

primitives/runtime/src/lib.rs
```diff
@@ -257,6 +262,72 @@ pub fn storage_value_key(pallet_prefix: &str, value_name: &str) -> StorageKey {
...
+	const OPERATING_MODE_KEY: &'static str;
```
svyatonik, line 273: I admit that without this key it'll be too confusing to read "Setting operating mode to false" instead of "Setting operating mode (IsHalted = false)." for the GRANDPA module :) WDYT about introducing something like `enum <GoodName> { Normal, Halted }` here and using it for both GRANDPA and parachains finality pallets? then we don't need this associated const at all.

I know that isn't a part of the issue, but I can recall at least two issues with improper use of `IsHalted`. And right now it is a good time to make this change, because we have no active deployments (apart from Rialto/Millau), so no runtime upgrades required. If you also think it is better, then let's probably file a new issue for that and impl in a separate PR? It must be marked with `breaksruntime` label, so that we (and other teams) won't miss it.

svyatonik (approved):
Seems super-cool, thank you! I've left some comments - feel free to merge whenever you consider PR ready

modules/grandpa/src/lib.rs
```diff
@@ -221,19 +233,9 @@ pub mod pallet {
 			origin: OriginFor<T>,
 			new_owner: Option<T::AccountId>,
 		) -> DispatchResultWithPostInfo {
```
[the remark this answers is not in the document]
serban300: Done. Seems to work.

primitives/runtime/src/lib.rs
```diff
@@ -257,6 +262,72 @@ pub fn storage_value_key(pallet_prefix: &str, value_name: &str) -> StorageKey {
 	StorageKey(final_key)
 }
 
+#[derive(Encode, Decode, TypeInfo, PalletError)]
```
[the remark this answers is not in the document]
serban300: Done.

primitives/runtime/src/lib.rs
serban300: Done.

primitives/runtime/src/lib.rs
serban300: Sounds good to me. Having an `enum` instead of `bool` would make things more clear. I oepend an issue for this: https://github.com/paritytech/parity-bridges-common/issues/1479

svyatonik (approved):
Perfect, thank you! Hit merge once it is green ;)

serban300:
> Perfect, thank you! Hit merge once it is green ;)

Ah awesome. Thanks !

Issue #1433, opened 2022-06-06
Remove without_storage_info on bridges pallets


svyatonik:
See https://github.com/paritytech/substrate/issues/8629

- [x] GRANDPA bridge pallet;
- [x] parachains bridge pallet;
- [x] messages pallet.

Not sure - how critical it is for us+Polkadot, since we'll be deploying as parachain now.

PR #1478, opened 2022-06-27
Remove pallet::without_storage_info from bridge GRANDPA pallet

svyatonik:
related to #1433 
it'll stay draft until the similar issue with system pallet is resolved (see explanation below)

I've tried to implement that, but looks like there are some more general problems that need to be solved first. Right now the main issue is that the [generic header](https://github.com/paritytech/substrate/blob/25417f5d3c2cf191771f33b738a96eebdcde23a7/primitives/runtime/src/generic/header.rs#L39) is not implementing the `MaxEncodedLen`. And it is what we're storing in the runtime storage. ~The similar case is with the system pallet, which is storing recent headers - so let's see how frame guys will solve it there. I assume that we'll have some upper bound on header size? Then we'll also need to do similar thing here.~ I've been wrong obviously - system pallet is only storing block hashes. But there's still problem with paras pallet (https://github.com/paritytech/polkadot/blob/master/runtime/parachains/src/paras/mod.rs), which doesn't generate storage info too.

With parachains finality pallet it is even more complex. because we deal with multiple parachains there. So we shall use `max(max-of-parachain-1, max-of-parachain-2, ...)`.

The fix for messages pallet looks easier - we already are limiting the message size. The only thing is that it is missing from the pallet configuration (it is configured through the message verifier trait).

UPD: I doubt that the parachains pallet will be fixed, because it is deployed at relay chain


modules/grandpa/src/storage_types.rs
```diff
@@ -0,0 +1,114 @@
...
+	}
+}
+
+impl<T: Config<I>, I: 'static> TryFrom<AuthoritySet> for StoredAuthoritySet<T, I> {
+	type Error = ();
+
+	fn try_from(t: AuthoritySet) -> Result<Self, Self::Error> {
+		Ok(StoredAuthoritySet { authorities: TryFrom::try_from(t.authorities)?, set_id: t.set_id })
+	}
+}
```
serban300, line 74: Is this needed ?

modules/grandpa/src/mock.rs
```diff
@@ -33,6 +33,9 @@ pub type TestNumber = crate::BridgedBlockNumber<TestRuntime, ()>;
 type Block = frame_system::mocking::MockBlock<TestRuntime>;
 type UncheckedExtrinsic = frame_system::mocking::MockUncheckedExtrinsic<TestRuntime>;
 
+pub const MAX_BRIDGED_AUTHORITIES: u32 = 256;
+pub const MAX_HEADER_SIZE: u32 = 65536;
```
acatangiu, line 37: nit: why not use same values as production?

modules/grandpa/src/storage_types.rs
```diff
@@ -0,0 +1,114 @@
...
+
+// Parity Bridges Common is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Wrappers for public types that are implementing Max
```
acatangiu, line 17:
```suggestion
//! Wrappers for public types that are implementing `MaxEncodedLen`
```

modules/grandpa/src/storage_types.rs
```diff
@@ -0,0 +1,114 @@
+// Copyright 2021 Parity Technologies (UK) Ltd.
```
acatangiu, line 1:
```suggestion
// Copyright 2022 Parity Technologies (UK) Ltd.
```

acatangiu (approved):

modules/grandpa/src/mock.rs
svyatonik: Changed `MAX_BRIDGED_AUTHORITIES`

bin/rialto-parachain/runtime/src/lib.rs
```diff
@@ -538,6 +538,8 @@ impl pallet_bridge_grandpa::Config for Runtime {
 	type BridgedChain = bp_millau::Millau;
 	type MaxRequests = MaxRequests;
 	type HeadersToKeep = HeadersToKeep;
+	type MaxBridgedAuthorities = frame_support::traits::ConstU32<5>;
+	type MaxBridgedHeaderSize = frame_support::traits::ConstU32<512>;
```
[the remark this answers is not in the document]
svyatonik: Sharing is a bad idea - we may change max number of authorities at any time for any chain, so it needs to be chain specific. But I think it worth moving constants to the primitives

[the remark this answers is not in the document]
svyatonik: Done

modules/grandpa/src/storage_types.rs
svyatonik: Seems like not

serban300 (approved):

modules/grandpa/src/storage_types.rs
```diff
@@ -0,0 +1,106 @@
...
+	pub authorities: StoredAuthorityList<<T as Config<I>>::MaxBridgedAuthorities>,
+	/// Monotonic identifier of the current GRANDPA authority set.
+	pub set_id: SetId,
+}
+
+impl<T: Config<I>, I: 'static> StoredAuthoritySet<T, I> {
+	/// Try to create a new bounded GRANDPA Authority Set from unbounded list.
+	///
+	/// Returns error if number of authorities in the provided list is too large.
+	pub fn new(authorities: AuthorityList, set_id: SetId) -> Result<Self, ()> {
```
bkontur, line 45:
```suggestion
	pub fn try_new(authorities: AuthorityList, set_id: SetId) -> Result<Self, ()> {
```

bkontur (approved):
lgtm

PR #1483, opened 2022-06-28
Unify the operating mode for bridge pallets

### LINKED ISSUE #1479: Change the IsHalted storage value for Grandpa from bool to enum
More details here: https://github.com/paritytech/parity-bridges-common/pull/1477#discussion_r907200966

serban300:
Fixes #1479 

It breaks runtime compatibility. I think it's also going to break relayers.

- define the OperationMode trait and BasicOperatingMode enum
- use the OperationMode trait in all the bridge pallets
- use BasicOperatingMode instead of IsHalted for the Grandpa pallet
- use BasicOperatingMode as part of MessagesOperatingMode

### DIFF
--- a/primitives/runtime/src/lib.rs
+++ b/primitives/runtime/src/lib.rs
@@ -271,6 +271,7 @@ pub enum OwnedBridgeModuleError {
 
 /// Operating mode for a bridge module.
 pub trait OperatingMode: Send + Copy + Debug + FullCodec {
+	// Returns true if the bridge module is halted.
 	fn is_halted(&self) -> bool;
 }
 
@@ -291,7 +292,6 @@ impl Default for BasicOperatingMode {
 }
 
 impl OperatingMode for BasicOperatingMode {
-	// Returns true if the bridge module is halted.
 	fn is_halted(&self) -> bool {
 		*self == BasicOperatingMode::Halted
 	}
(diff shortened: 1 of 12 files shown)


primitives/runtime/src/lib.rs
```diff
@@ -269,18 +269,46 @@ pub enum OwnedBridgeModuleError {
 	Halted,
 }
 
+/// Operating mode for a bridge module.
+pub trait OperatingMode: Send + Copy + Debug + FullCodec {
+	fn is_halted(&self) -> bool;
```
svyatonik, line 274: Please some docs for this public method

svyatonik (approved):
Looks great, thank you! Yep - it breaks both relay and existing runtimes, thanks for proper labeling! Fortunately, apart from Rialto/Millau, we only have those pallets deployed at Rococo/Wococo and we're not going to make any upgrades there - instead this deployment will be deprecated soon.

primitives/runtime/src/lib.rs
svyatonik: No worries! Thanks! :)

svyatonik: (btw - I still do not see any docs :) )

serban300: Yes, I placed it in the implementation by mistake. Fixed.

PR #1484, opened 2022-06-29
Parachain loop metrics

### LINKED ISSUE #1481: Provide metrics from parachains relay
Right now parachains relay provides no metrics at all && it'd be good to have some. We'd also need to add Westmint -> Millau dashboard to our Grafana (see #1443 )

svyatonik:
closes #1481 

Two metrics are provided: best parachain head numbers at the source relay chain and at the target chain.

### DIFF
--- a/primitives/parachains/src/lib.rs
+++ b/primitives/parachains/src/lib.rs
@@ -56,7 +56,7 @@ pub fn parachain_head_storage_key_at_source(
 /// Returns runtime storage key of best known parachain head at the target chain.
 ///
 /// The head is stored by the `pallet-bridge-parachains` pallet in the `BestParaHeads` map.
-pub fn parachain_head_storage_key_at_target(
+pub fn best_parachain_head_hash_storage_key_at_target(
 	bridge_parachains_pallet_name: &str,
 	para_id: ParaId,
 ) -> StorageKey {
@@ -64,5 +64,21 @@ pub fn parachain_head_storage_key_at_target(
 		bridge_parachains_pallet_name,
 		"BestParaHeads",
 		&para_id.encode(),
+	)
+}
+
+/// Returns runtime storage key of the parachain head with given hash at the target chain.
+///
+/// The head is stored by the `pallet-bridge-parachains` pallet in the `ImportedParaHeads` map.
+pub fn imported_parachain_head_storage_key_at_target(
+	bridge_parachains_pallet_name: &str,
+	para_id: ParaId,
+	head_hash: ParaHash,
+) -> StorageKey {
+	bp_runtime::storage_double_map_final_key::<Blake2_128Concat, Twox64Concat>(
+		bridge_parachains_pallet_name,
+		"ImportedParaHeads",
+		&para_id.encode(),
+		&head_hash.encode(),
 	)
 }
--- a/primitives/parachains/src/lib.rs
+++ b/primitives/parachains/src/lib.rs
@@ -75,7 +75,7 @@ pub fn imported_parachain_head_storage_key_at_target(
 	para_id: ParaId,
 	head_hash: ParaHash,
 ) -> StorageKey {
-	bp_runtime::storage_double_map_final_key::<Blake2_128Concat, Twox64Concat>(
+	bp_runtime::storage_double_map_final_key::<Blake2_128Concat, Blake2_128Concat>(
 		bridge_parachains_pallet_name,
 		"ImportedParaHeads",
 		&para_id.encode(),
--- a/relays/lib-substrate-relay/src/parachains/source.rs
+++ b/relays/lib-substrate-relay/src/parachains/source.rs
@@ -23,7 +23,10 @@ use async_trait::async_trait;
 use bp_parachains::parachain_head_storage_key_at_source;
 use bp_polkadot_core::parachains::{ParaHash, ParaHead, ParaHeadsProof, ParaId};
 use codec::Decode;
-use parachains_relay::parachains_loop::{ParaHashAtSource, SourceClient};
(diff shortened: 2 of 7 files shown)


primitives/parachains/src/lib.rs
```diff
@@ -75,7 +75,7 @@ pub fn imported_parachain_head_storage_key_at_target(
 	para_id: ParaId,
 	head_hash: ParaHash,
 ) -> StorageKey {
-	bp_runtime::storage_double_map_final_key::<Blake2_128Concat, Twox64Concat>(
+	bp_runtime::storage_double_map_final_key::<Blake2_128Concat, Blake2_128Concat>(
```
serban300, line 78: I'm not very familiar with storage values, but considering these fixes, I was wondering if we could use a safer approach here. I'm thinking of the following issues:
1. As seen here, we can use different hashers for the storage key vs the `StorageDoubleMap` that is defined in the pallet. This seems to be covered by unit tests however.
2. We can use different types when we decode the value. For example here the correct ones would be `ParaId`, `ParaHash`, but we might by mistake use something else when decoding. Not sure if this is covered by unit tests.

So I was wondering if we could use a safer approach to cover these issues. Maybe define a new structure to hold all these hashers and types and use if both when defining the `StorageDoubleMap` and when computing the key and reading the value. A scaled-down version of `StorageDoubleMap` maybe.

serban300 (commented):
Looks good ! I just left a few comments. I would be happy to address them if they make sense.

relays/lib-substrate-relay/src/parachains/source.rs
```diff
@@ -111,9 +115,11 @@ where
 			)))
 		}
 
-		Ok(match self.on_chain_parachain_header(at_block, para_id).await? {
+		let mut para_header_number_at_source = None;
+		let para_hash_at_source = match self.on_chain_parachain_header(at_block, para_id).await? {
 			Some(parachain_header) => {
 				let mut parachain_head = ParaHashAtSource::Some(parachain_header.hash());
+				para_header_number_at_source = Some(*parachain_header.number());
```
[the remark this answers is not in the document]
svyatonik: Yeah, but they have different scopes. Feel free to change the code if you think your version is better, though :)

relays/lib-substrate-relay/src/parachains/target.rs
```diff
@@ -115,15 +121,46 @@ where
...
+			let imported_para_head_key = imported_parachain_head_storage_key_at_target(
+				P::SourceRelayChain::PARACHAINS_FINALITY_PALLET_NAME,
+				para_id,
+				best_para_head_hash.head_hash,
+			);
+			let imported_para_header = self
+				.client
+				.storage_value::<ParaHead>(imported_para_head_key, Some(at_block.1))
+				.await?
+				.and_then(|h| match HeaderOf::<P::SourceParachain>::decode(&mut &h.0[..]) {
```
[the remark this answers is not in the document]
svyatonik: I don't mind - feel free to add :)

primitives/parachains/src/lib.rs
svyatonik: Yep, that's a good idea. I'm not sure, though that you'll able to do that - all storage delarations are handled by procedural macro and it may happen that you can't specify e,g, `MyMap::Hasher` there. But you can try to experiment of course :)

relays/lib-substrate-relay/src/parachains/target.rs
```diff
@@ -115,15 +121,46 @@ where
...
+		if let (Some(metrics), &Some(ref best_para_head_hash)) = (metrics, &best_para_head_hash) {
+			let imported_para_head_key = imported_parachain_head_storage_key_at_target(
+				P::SourceRelayChain::PARACHAINS_FINALITY_PALLET_NAME,
+				para_id,
+				best_para_head_hash.head_hash,
+			);
+			let imported_para_header = self
+				.client
+				.storage_value::<ParaHead>(imported_para_head_key, Some(at_block.1))
+				.await?
```
serban300, line 142: Sorry, one more thing. `await?` returns the error directly, if any, while if `HeaderOf::<P::SourceParachain>::decode()` returns an error, we also log a message. Is this desired? Or should we log a message in both error cases ?

svyatonik: If you're going to make a new method for double-encoded storage values, then let's log it here, where we know what the storage value actually mean and where we may log the meaningful error.

Ideally (imo) we'll need to migrate to some crate that would allow us to propagate error with context and log it once at the upper level (inside the relay loop). But that's not a part of your follow-up PR of course :)

serban300: Sure, I will log the error here, but I was wondering if we should log an error both when `storage_value()` returns an `Err` and when `decode` returns an `Err`. Or should we keep the `await?` and log an error only when `decode()` returns an `Err` ?

The idea with having a new method for double-encoded storage value is not going very well so far. I think I might have to give it up.

svyatonik: > Sure, I will log the error here, but I was wondering if we should log an error both when storage_value() returns an Err and when decode returns an Err. Or should we keep the await? and log an error only when decode() returns an Err ?

Imo more detailing logging = better :) The fact that logging occurs at the one branch only is probably the outcome of some bug that I've met. In the end the error is logged at the top level anyway, but this additional trace adds some context. So I'd add the second logging call here.

serban300:
The comments have been addressed in #1521 . Marking this PR as reviewed.

PR #1486, opened 2022-06-29
Add owner calls to the parachains pallet

### LINKED ISSUE #1434: Add owner calls to the parachains pallet

serban300:
Fixes #1434


svyatonik (approved):
Looks great, thank you! A couple of things I'd add:
1) test in parachains finality pallet to ensure that the `submit_parachain_heads` call is rejected when the pallet is halted;
2) an ability to set parachains finality pallet owner and operating mode at genesis.

Feel free to do that in a follow-up PRs though :)

serban300:
> 1. `submit_parachain_heads`

Thank you for the review and the suggestions ! I addressed both points in this PR but separate commits.

svyatonik (approved):
Great, thank you!

Issue #1415, opened 2022-05-25
Refactor on-demand relays startup


svyatonik:
Now our `substrate-relay` binary supports complex bridges between relay chains, between relay and parachain. In the near future the support for parachain <> parachain bridges will be added. This has lead to splitting single on-demand headers relay startup code to two separate functions (`start_on_demand_relay_to_relay` and `start_on_demand_relay_to_parachain`). It needs to be refactored.

Minimum:
- [ ] extract functions to a separate file;
- [ ] remove `allow(clippy::too_many_arguments)` (see #1414).

Ideally: we shall change it to some declarative form - i.e. if left chain is parachain and the right is relaychain, then we shall left-to-right-parachains-on-demand-relay and right-to-left-headers-relay. If both are relay chains, we shall start two on-demand header relays. And so on.

PR #1490, opened 2022-07-05
[CLI] Remove some macros

serban300:
Refactor the implementations for the following CLI methods in order to
avoid using macros:
- init_bridge
- relay_headers
- relay_parachains

This is an initial PR to gather feedback on the idea of using traits instead of macros for implementing the CLI methods. In my opinion this would bring 2 benefits:
- better code readability
- some code deduplication

If there are no concerns related to this approach we can move on and try to use it for the other commands as well. The final purpose is to try to use this for `relay_headers_and_messages`. If this works I think we can address #1415 more naturally, probably also using traits.

### DIFF
--- a/relays/bin-substrate/src/cli/bridge.rs
+++ b/relays/bin-substrate/src/cli/bridge.rs
@@ -349,9 +349,9 @@ impl CliBridgeBase for RialtoParachainToMillauCliBridge {
 }
 
 //// `WestendParachain` to `Millau` bridge definition.
-pub struct WestendParachainToMillauCliBridge {}
+pub struct WestmintToMillauCliBridge {}
 
-impl CliBridgeBase for WestendParachainToMillauCliBridge {
+impl CliBridgeBase for WestmintToMillauCliBridge {
 	type Source = relay_westend_client::Westmint;
 	type Target = relay_millau_client::Millau;
 }
--- a/relays/bin-substrate/src/cli/bridge.rs
+++ b/relays/bin-substrate/src/cli/bridge.rs
@@ -14,7 +14,10 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
 
+use crate::cli::CliChain;
+use relay_substrate_client::{AccountKeyPairOf, Chain, TransactionSignScheme};
 use strum::{EnumString, EnumVariantNames};
+use substrate_relay_helper::finality::SubstrateFinalitySyncPipeline;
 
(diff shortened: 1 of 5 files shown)


relays/bin-substrate/src/cli/init_bridge.rs
```diff
@@ -52,188 +61,42 @@ pub enum InitBridgeName {
...
+
+	/// Get the encoded call to init the bridge.
+	fn encode_init_bridge(
+		init_data: <Self::Engine as Engine<Self::Source>>::InitializationData,
+	) -> <Self::Target as Chain>::Call;
+
+	/// Initialize the bridge.
+	fn init_bridge(
+		data: InitBridge,
+	) -> Pin<Box<dyn Future<Output = anyhow::Result<()>> + Send + 'static>> {
```
svyatonik, line 78: Imo it would be better to use `async_trait` here instead of all this `Pin<Box<...>>` stuff :)

svyatonik: (there's also couple of other places where imo it shall be replaced)

relays/bin-substrate/src/cli/bridge.rs
```diff
@@ -218,3 +221,137 @@ macro_rules! select_full_bridge {
...
+//// `RialtoParachain` to `Millau` bridge definition.
+pub struct RialtoParachainToMillauCliBridge {}
+
+impl CliBridgeBase for RialtoParachainToMillauCliBridge {
+	type Source = relay_rialto_parachain_client::RialtoParachain;
+	type Target = relay_millau_client::Millau;
+}
+
+//// `WestendParachain` to `Millau` bridge definition.
+pub struct WestendParachainToMillauCliBridge {}
```
svyatonik, line 352: Let's rename it to `WestmintToMillauCliBridge` maybe? There are many parachains on Westend and iiuc this struct is dedicated to Westmint.

svyatonik (approved):
Changes seems good to me :) I hope you have a vision for the final solution, which will actually make the code easier to read :) Couple of suggestions.

relays/bin-substrate/src/cli/bridge.rs
serban300: Done.

relays/bin-substrate/src/cli/init_bridge.rs
serban300: Indeed, it looks better. Done.

svyatonik (approved):
Cool, thank you!

PR #1491, opened 2022-07-06
Remove obsolete bridges

svyatonik:
After https://github.com/paritytech/polkadot/pull/5752, both persistent relay<>relay bridges (Rococo<>Wococo and Kusama<>Polkadot) are now obsolete. I'm removing support for these bridges from substrate relay - hope this will make a refactoring a bit easier for @serban300 :) 

I've also (temporarily) removed `reinit-bridge` subcommand, since it has only been supported for Kusama<>Polkadot. Logged in #1492


serban300 (approved):
Nice ! Looks good to me.

PR #1493, opened 2022-07-06
[CLI] remove select_full_bridge!() macro

serban300:
This is a follow-up for #1490  . Removing the select_full_bridge!() macro and using the `CliBridge` classes that we previously defined. 

Next I'll focus on `relay_headers_and_messages` and #1415

### DIFF
--- a/relays/bin-substrate/src/cli/bridge.rs
+++ b/relays/bin-substrate/src/cli/bridge.rs
@@ -60,7 +60,7 @@ pub trait CliBridgeBase: Sized {
 		+ CliChain<KeyPair = AccountKeyPairOf<Self::Target>>;
 }
 
-/// Bridge representation that can be used from the CLI.
+/// Bridge representation that can be used from the CLI for relaying headers.
 pub trait HeadersCliBridge: CliBridgeBase {
 	/// Finality proofs synchronization pipeline.
 	type Finality: SubstrateFinalitySyncPipeline<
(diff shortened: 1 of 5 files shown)


relays/bin-substrate/src/cli/bridge.rs
```diff
@@ -155,6 +70,20 @@ pub trait CliBridge: CliBridgeBase {
 	>;
 }
 
+pub trait MessagesCliBridge: CliBridgeBase {
+	const ESTIMATE_MESSAGE_FEE_METHOD: &'static str;
```
svyatonik, line 74: Could you, please, add some docs for both trait and its associated consts/types + methods? Thanks! :)

relays/bin-substrate/src/cli/bridge.rs
```diff
@@ -155,6 +70,20 @@ pub trait CliBridge: CliBridgeBase {
...
+	type MessagesLane: SubstrateMessageLane<
+		SourceChain = Self::Source,
+		TargetChain = Self::Target,
+		SourceTransactionSignScheme = Self::Source,
+		TargetTransactionSignScheme = Self::Target,
+	>;
+
+	fn derive_account_from_id(
```
svyatonik, line 82: This actually shall be removed in the future (there's an issue for that somewhere in the repo). But of course, not it this PR - just sharing the info :)

svyatonik (approved):
Looks good, thank you! :)

relays/bin-substrate/src/cli/bridge.rs
serban300: Sorry, I forgot. Done.

svyatonik: That's fine :)

PR #1496, opened 2022-07-08
CLI refactoring

### LINKED ISSUE #1415: Refactor on-demand relays startup

serban300:
Fixes #1415 

Posting this draft in order to get feedback on the general approach first, since there are a lot of changes. If the feedback is positive, I can polish this, add comments, use the new backing structures everywhere, remove the old generated structures, rename, etc.

The main ideas behind this refactoring are the following:
1. Defining backing structures for the generated CLI structures and a way to convert the generated structures into the backing structure (e.g. having `ConnectionParams` for `MillauConnectionParams`, `RialtoConnectionParams`, etc and `MillauConnectionParams.into() -> ConnectionParams`, `RialtoConnectionParams.into() -> ConnectionParams`). This way we can have a uniform representation of the generate CLI structures, which makes it easier to operate with them.
3. The same strategy as the one at point 1. above for the generated CLI structures specific to `relay_headers_and_messages`
4. Using the strategy introduced in #1490 in order to define specific bridges for RelayChainToRelayChain and RelayChainToParachain scenarios. With the added complexity of the fact that now we have to also store a state.

I realize that the PR introduces some very significant changes for a really small issue. What I personally don't like is the verbosity of specifying the trait constraints for each structure in more places, for example :

```
pub struct RelayChainToParachainEnv<
	Left: Chain + TransactionSignScheme<Chain = Left> + CliChain<KeyPair = AccountKeyPairOf<Left>>,
	Right: Chain<Hash = ParaHash>
		+ TransactionSignScheme<Chain = Right>
		+ CliChain<KeyPair = AccountKeyPairOf<Right>>,
	RightRelay: Chain<BlockNumber = RelayBlockNumber, Hash = RelayBlockHash, Hasher = RelayBlockHasher>
		+ TransactionSignScheme<Chain = RightRelay>
		+ CliChain<KeyPair = AccountKeyPairOf<RightRelay>>,
	LeftFinalityToRight: SubstrateFinalitySyncPipeline<
		SourceChain = Left,
		TargetChain = Right,
		TransactionSignScheme = Right,
	>,
	RightFinalityToLeft: SubstrateParachainsPipeline<
		SourceRelayChain = RightRelay,
		SourceParachain = Right,
		TargetChain = Left,
		TransactionSignScheme = Left,
	>,
	RightRelayFinalityToLeft: SubstrateFinalitySyncPipeline<
		SourceChain = RightRelay,
		TargetChain = Left,
		TransactionSignScheme = Left,
	>,
>
```

I'm not sure if we can simplify this. But overall personally I prefer having these strongly defined structures for implementing the relaying logic instead of having a macro and I find the code more readable like this. However, I realize that these aspects may come down to personal preference and that there are pros and cons for both approaches. So I'm waiting for feedback :)


svyatonik (commented):
> I'm not sure if we can simplify this. But overall personally I prefer having these strongly defined structures for implementing the relaying logic instead of having a macro and I find the code more readable like this. However, I realize that these aspects may come down to personal preference and that there are pros and cons for both approaches. So I'm waiting for feedback :)

Yeah - that was the exact reason why these `select_bridge`, `select_full_bridge` appeared :) If you think the generics are more readable here than the macro - please continue! 

Now point-by-point:
1. Ok, sounds good to me :) But is it really needed? I mean - all generated structures already have methods like `to_client()`. Is there a reason why we need `MillauConnectionParams` -> `ConnectionParams` -> `Client` instead of simply `MillauConnectionParams` -> `Client`?
2. Feel free to change any structs/code there - I admit that the current version of headers-and-messages is a mess even with macro :) It really needs some love
3. Can you, please, elaborate - what do you mean by "now we have to also store a state"? Isn't the headers+messages bridge is also start-and-forget relay? Why do we need to store some state there?

Thank you for working on that! :)

serban300:
Thank you for the feedback !

> 1. Ok, sounds good to me :) But is it really needed? I mean - all generated structures already have methods like to_client(). Is there a reason why we need MillauConnectionParams -> ConnectionParams -> Client instead of simply MillauConnectionParams -> Client?

Actually when I started I tried extending the existing traits for the generated structures and adding new ones where needed and it got really complicated. Because we can't access fields directly since the naming is different, and sometimes there are "trait cannot be made into an object" errors. Probably it is possible to use just the generated structures, but it seemed much harder to me than just adding these uniform backing structures.

> 3. Can you, please, elaborate - what do you mean by "now we have to also store a state"? Isn't the headers+messages bridge is also start-and-forget relay? Why do we need to store some state there?

Sorry, state is not the right name. We just have to store some data. In order to avoid passing many arguments to the `start_on_demand_headers_relayers()` function, I created a new structure that holds this data. For example:

```
{
	params: RelayChainToParachainBridgeParams,

	left_client: Client<Left>,
	right_client: Client<Right>,

	at_left_accounts: Vec<TaggedAccount<AccountIdOf<Left>>>,
	at_right_accounts: Vec<TaggedAccount<AccountIdOf<Right>>>,
}
```

And now the `start_on_demand_headers_relayers(&self)` is defined on that structure and it can access the data on `self`.

> If you think the generics are more readable here than the macro - please continue!

Awesome. Thanks ! I will try to come up with a more polished version the following days.

serban300:
I polished the code and added some comments. Also I managed to make the trait constraints a bit smaller at places. Marking this PR as ready for review.

I think some further cosmetic improvements and code deduplication are possible, but I would leave that for a different PR.

relays/bin-substrate/src/cli/chain_schema.rs
```diff
@@ -0,0 +1,494 @@
...
+						spec_version: item.[<$chain_prefix _spec_version>],
+						transaction_version: item.[<$chain_prefix _transaction_version>],
+					}
+				}
+			}
+		}
+	};
+}
+
+#[derive(Debug, PartialEq, Eq, Clone)]
```
svyatonik, line 114: I'm still not keen on introducing new structures here. I've looked at all usages and it seems that everywhere (I've been looking at `ConnectionParams` only) we have following code:
```rust
let source_client = ConnectionParams::from(data.source).to_client::<Self::Source>().await?;
```

So it still seems to me that an redundant layer - why not to keep calling `.to_client()` directly on generated structures? :) 
```rust
let source_client = data.source.to_client::<Self::Source>().await?;
```

Don't get me wrong - I'm fine with moving code outside of macro, but this particular thing looks unneeded to me. Like we already had these generated structs before and then we have had `relay_substrate_client::ConnectionParams` with exactly same fields and now we have 3rd version :) Imo it is too much.

The similar situation is with other structs here - e.g. instead of calling `to_keypair` directly on generated struct, we call it like here:
```rust
let target_sign = SigningParams::from(data.target_sign).to_keypair::<Self::Target>()?;
```

TLDR: if there's no need for duplicate structs, I'd avoid adding it. Please don't give up if you have different thoughts on that - I'm ready to change my opinion :)

relays/bin-substrate/src/cli/bridge.rs
```diff
@@ -70,6 +75,29 @@ pub trait HeadersCliBridge: CliBridgeBase {
 	>;
 }
 
+/// Bridge representation that can be used from the CLI for relaying headers
+/// from a parachain to a relay chain.
+pub trait ParachainHeadersCliBridge: CliBridgeBase {
```
svyatonik, line 80: We'll also have parachain <> parachain bridge soon (our final goal). So probably `RelayParachainHeadersCliBridge` instead? Could be postponed, though :)

relays/bin-substrate/src/cli/bridge.rs
```diff
@@ -97,7 +126,7 @@ impl CliBridgeBase for MillauToRialtoCliBridge {
 	type Target = relay_rialto_client::Rialto;
 }
 
-impl HeadersCliBridge for MillauToRialtoCliBridge {
+impl RelayHeadersCliBridge for MillauToRialtoCliBridge {
```
svyatonik, line 129: Id also consider those chain-specific structs to the `chains` folder - imo there are too much structs here and the number will grow with a time. WDYT? If you agree, we may do that in a follow-up PRs

relays/bin-substrate/src/cli/relay_headers_and_messages/chain_to_chain.rs
```diff
@@ -0,0 +1,211 @@
+// Copyright 2019-2022 Parity Technologies (UK) Ltd.
```
svyatonik, line 1: Probably rename this file to `relay_to_relay,rs` (and all nested structs), since both parachain and relay chains are *chains*, so it isn't clear from the name - what this file is about :)

svyatonik (commented):
I'm still not convinced about introducing new structs - let's have another comment round. Otherwise - looks good, perfect! Thank you!

relays/bin-substrate/src/cli/chain_schema.rs
serban300: It's true that in these use cases the new structures only add a layer a complexity. Where I did find them more helpful was in the case of the `relay_headers_and_messages`, where they made it easier to work with the CLI supplied params. But also adding some traits there instead of the backing structures might work. I'll give it a try.

relays/bin-substrate/src/cli/bridge.rs
serban300: Sounds very good. At first sight it doesn't seem to imply many changes so I can try to do it in this PR.

relays/bin-substrate/src/cli/chain_schema.rs
svyatonik: Ok, if they're really necessary, can we then use `relay_substrate_client::ConnectionParams` instead of this new struct?

serban300: I removed the backing structures. PTAL on the last commit ! 

Now the problem is that we need a new trait that has to be implemented both by Relay<>Relay bridges and Relay<>Parachains bridges and there is duplication there.

Personally I still prefer the version with the backin structures but both have pros and cons. WDYT ?

relays/bin-substrate/src/cli/bridge.rs
serban300: Done

relays/bin-substrate/src/cli/relay_headers_and_messages/chain_to_chain.rs
serban300: Done

relays/bin-substrate/src/cli/bridge.rs
serban300: Actually I'll do this in a separate PR. This PR is getting too complex.

relays/bin-substrate/src/cli/chain_schema.rs
svyatonik: Yeah - I agree that it looks worse than when structs are used :) Sorry for grumbling about that :) So let's just try to reuse existing structs if possible - `relay_substrate_client::ConnectionParams`, `substrate_relay_helper::TransactionParams`, ...

serban300: Done. I reverted the chain schema to how it was before and I used backing structures only for the Full2WayBridge . I think it looks better now.

svyatonik (approved):
Thank you! Looks perfect to me. One thing - `create_relayers_fund_accounts` has been recently removed (unless I'm missing someting), so probably let's not reintroduce it yet?

relays/bin-substrate/src/cli/relay_headers_and_messages/mod.rs
```diff
@@ -0,0 +1,699 @@
...
+#[derive(Debug, PartialEq, StructOpt)]
+pub struct HeadersAndMessagesSharedParams {
+	/// Hex-encoded lane identifiers that should be served by the complex relay.
+	#[structopt(long, default_value = "00000000")]
+	pub lane: Vec<HexLaneId>,
+	#[structopt(long, possible_values = RelayerMode::VARIANTS, case_insensitive = true, default_value = "rational")]
+	pub relayer_mode: RelayerMode,
+	/// Create relayers fund accounts on both chains, if it does not exists yet.
+	#[structopt(long)]
+	pub create_relayers_fund_accounts: bool,
```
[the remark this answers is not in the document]
serban300: Ups ! Yes, there were some conflicts when merging from upstream/master and I missed this one. Thanks !

svyatonik (approved):
Cool, thank you!

PR #1497, opened 2022-07-08
Change account names to something meaningful

### LINKED ISSUE #1428: Change account names for our testnets
There are "standard" Rialto/Millau/RialtoParachain accounts (Alice, Bob, ...) that are used by different relays and message generators and it isn't always clear which account and where is used. As a result we see issues like that one #1427 . I suggest to add separate accounts to be used by relays/message generators and give them a meaningful names. There already were some attempts to fix that, but they're inactive now (see e.g. `MillauMessagesOwner`, `RialtoMessagesOwner`, `WithMillauTokenSwap`, ...).

BTW - `WithMillauTokenSwap` may be removed now.

svyatonik:
closes #1428

This PR changes account "names", used on our test deployments from generic (e.g. `//Harry`) to meaningful (`//Millau.MessagesSender`). We had too many issues related to generic names usage - it isn't always clear, which account is already submitting transactions and which is not. Documentation isn't of big help here. This change allows us to understand what account is doing just by looking at its name.

*SIDE INFO*: one account may be used to sign multiple transactions, but every Substrate transaction (in our chains) has so called `nonce`, which is just a "serial number" of transaction. You can't have two transactions with the same number and transactions are always mined in order. If we're using e.g. `//Alice` to sign transactions from different processes (relays), then some transactions may be dropped, because they'll be using the same nonce. We have a inter-process solution for that, but not for multiple processes (and we don't need that).

@serban300 Im summoning you here, because you're also working with codebase. So please let me know what you're thinking about this change and whether names I've choose are understandable. If you think there are better "names" - please suggest.

### DIFF
--- a/bin/millau/node/src/chain_spec.rs
+++ b/bin/millau/node/src/chain_spec.rs
@@ -15,7 +15,6 @@
 // along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
 
 use beefy_primitives::crypto::AuthorityId as BeefyId;
-use bp_millau::derive_account_from_rialto_id;
 use millau_runtime::{
 	AccountId, AuraConfig, BalancesConfig, BeefyConfig, BridgeRialtoMessagesConfig,
 	BridgeRialtoParachainMessagesConfig, BridgeWestendGrandpaConfig, GenesisConfig, GrandpaConfig,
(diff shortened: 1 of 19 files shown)


bin/millau/node/src/chain_spec.rs
```diff
@@ -134,54 +133,38 @@ impl Alternative {
 /// purposes), are all available on these chains.
 fn endowed_accounts() -> Vec<AccountId> {
 	vec![
```
serban300, line 135: Nit: Can we define these names in an enum (or consts) in order to be able to reuse them ?

serban300 (approved):
Looks good !

serban300 (approved):

Issue #1318, opened 2022-02-15
Support XCM delivery fees in bridges


svyatonik:
There's some support for XCM+bridges in https://github.com/paritytech/polkadot/tree/gav-xcm-v3 branch. Just making this issue to not to lose gained knowledge.

So in bridges+XCM:
1) dispatch fee is paid using `BuyExecution` XCVM instruction, so we don't need to pay it at the source chain. However I have some doubts about dispatch - if it'll be immediate dispatch, then we need to declare dispatch weight in advance (and relayer pays for it in advance). If it'll be just queueing - we'll need to pay for the queueing operation (single db write?). It must be decided before any changes are made;
2) XCM+bridge connector shall return message delivery fee from its [`SendXcm::validate()`](https://github.com/paritytech/polkadot/blob/6185cf32dc1c55759fc9481691a8409554aa0754/xcm/src/v3/traits.rs#L423) implementation; 
2) right now the delivery fee is not actually paid when the XCM message is sent - `Ok(price)` needs to be handled properly [here](https://github.com/paritytech/polkadot/blob/6185cf32dc1c55759fc9481691a8409554aa0754/xcm/pallet-xcm/src/lib.rs#L559-L562). But when it'll be fixed, we need to decide whether we need a separate relayer strategy? Like what happens with an XCM message when its pre-paid fee becomes lower than the actual cost of all involved transactions? For encoded calls, we have this `increase_message_fee` call. But for XCM messages, the fee is paid at the upper level (so I assume on messages pallet it'll be zero) => this method doesn't work. Do we need ordered delivery in this case at all? Can we just skip underpaid XCM? This is a bit larger than just fee-related question, but it needs an answer before this work starts. https://github.com/paritytech/polkadot/pull/4097/files#r854844093

svyatonik:
So we need to pay rewards from treasury https://github.com/paritytech/polkadot/pull/4097#discussion_r866263870

svyatonik:
Since we'll be deploying messaging pallet at the parachain, I suggest to defer this until we'll have some bridge-hub parachain template/prototype. Handling this at relay chain is just a waste of time now

svyatonik:
Current state of this issue: we are not taking any fees (and are not rewarding relayers) at Rialto<>Millau and at RialtoParachain<>Millau bridges. The plan is to discuss it (by opening a PR) when we'll have dedicate chain(s) for RococoBridgeHub/WococoBridgeHub.

[...]

PR #1498, opened 2022-07-12
Remove account derivation functions

### LINKED ISSUE #1426: Remove account derivation function and types
There are still some account-derivation functions, that were used by the encoded-call messaging (e.g. `bp_millau::derive_account_from_rialto_id` and `bp_millau::AccountIdConverter`). Imo most of functions/types may safely be removed now.

svyatonik:
closes #1426


svyatonik:
Depending on how #1318 will be solved, we may or may not need this `create_relayers_fund_accounts` flag in the headers-and-messages relay, but now it isn't required. So I've removed it

serban300 (approved):

PR #1505, opened 2022-07-13
Add another condition to the reject-obsolete-parachain-heads extension

svyatonik:
So currently this extension only rejects transaction if bundled at-relay-chain-block-number is lesser or equal than the existing one. But there's another condition - when parachain head remains the same (i.e. is not updated) in some relay blocks, then the other relay may use different relay chain block to prove it. The extension won't reject the transaction, even though the pallet will ignore the head (since it is the same). This PR handles this situation.

TODOs:
- [x] test it;
- [x] add unit tests;
- [x] cleanup;
- [x] more traces?

### DIFF
--- a/modules/parachains/src/extension.rs
+++ b/modules/parachains/src/extension.rs
@@ -146,7 +146,10 @@ mod tests {
 		Call::Parachains => ()
 	}
 
-	fn validate_submit_parachain_heads(num: RelayBlockNumber, parachains: Vec<(ParaId, ParaHash)>) -> bool {
+	fn validate_submit_parachain_heads(
+		num: RelayBlockNumber,
+		parachains: Vec<(ParaId, ParaHash)>,
+	) -> bool {
 		BridgeRejectObsoleteParachainHeader
 			.validate(
 				&42,
@@ -218,7 +221,10 @@ mod tests {
 			// when current best finalized is #10 and we're trying to import header#5, but another
 			// parachain head is also supplied => tx is accepted
 			sync_to_relay_header_10();
-			assert!(validate_submit_parachain_heads(5, vec![(ParaId(1), [1u8; 32].into()), (ParaId(2), [1u8; 32].into())]));
+			assert!(validate_submit_parachain_heads(
+				5,
+				vec![(ParaId(1), [1u8; 32].into()), (ParaId(2), [1u8; 32].into())]
+			));
 		});
 	}
 }
--- a/modules/parachains/src/extension.rs
+++ b/modules/parachains/src/extension.rs
@@ -72,15 +72,18 @@ macro_rules! declare_bridge_reject_obsolete_parachain_header {
 							ref parachains,
 							..
 						}) if parachains.len() == 1 => {
-							let parachain = parachains.get(0).expect("verified by match condition; qed");
+							let (parachain, parachain_head_hash) = parachains.get(0).expect("verified by match condition; qed");
 
 							let bundled_relay_block_number = at_relay_block.0;
 
 							let best_parachain_head = $crate::BestParaHeads::<$runtime, $instance>::get(parachain);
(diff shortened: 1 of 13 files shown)


modules/parachains/src/extension.rs
```diff
@@ -72,15 +72,34 @@ macro_rules! declare_bridge_reject_obsolete_parachain_header {
...
+								}
+								Some(best_parachain_head) if best_parachain_head.head_hash == *parachain_head_hash => {
+									log::trace!(
+										target: $crate::LOG_TARGET,
+										"Rejecting obsolete parachain-head {:?} transaction: head hash {:?}",
+										parachain,
+										best_parachain_head.head_hash,
+									);
+									sp_runtime::transaction_validity::InvalidTransaction::Stale.into()
+								}
```
serban300, line 102: This validation looks very similar with the one [here](https://github.com/paritytech/parity-bridges-common/blob/master/modules/parachains/src/lib.rs#L386-L420). 2 questions:
1. The equality case for `at_relay_block_number` seems to be treated differently. Here it is treated as invalid, in `update_parachain_head()` it's treated as valid. Is this expected ?
2. Would it make sense to do the same validation in both places ? In this case would it make sense to deduplicate the code ?

relays/lib-substrate-relay/src/parachains/source.rs
```diff
@@ -160,23 +160,46 @@ where
...
+	) -> Result<(ParaHeadsProof, Vec<ParaHash>), Self::Error> {
+		if parachains.len() != 1 || parachains[0].0 != P::SOURCE_PARACHAIN_PARA_ID {
+			return Err(SubstrateError::Custom(format!(
+				"Trying to prove unexpected parachains {:?}. Expected {:?}",
+				parachains,
+				P::SOURCE_PARACHAIN_PARA_ID,
+			)))
+		}
+
+		let parachain = parachains[0];
```
serban300, line 172: Nit: I would do something like:

```
		let parachain = ParaId(P::SOURCE_PARACHAIN_PARA_ID);
		if parachains != &[parachain] {
			return Err(SubstrateError::Custom(format!(
				"Trying to prove unexpected parachains {:?}. Expected {:?}",
				parachains, parachain,
			)))
		}
```

serban300 (commented):
Looks good ! I just have 2 comments. If they make sense I would be happy to implement them.

modules/parachains/src/extension.rs
svyatonik: 1. Imo in both cases `at_relay_block_number` leads to error, no (I maybe missing something - sorry)?

```rust
fn update_parachain_head() {
    ...
    match stored_best_head {
    	Some(stored_best_head) if stored_best_head.at_relay_block_number <= updated_at_relay_block_number => {
            if updated_head_hash == stored_best_head.head_hash {
                return Err(());
            }

            // else: pass
        }
    ...
}

fn validate() {
	...
	match best_parachain_head {
		Some(best_parachain_head) if best_parachain_head.at_relay_block_number
			>= bundled_relay_block_number => Err(()),
		Some(best_parachain_head) if best_parachain_head.head_hash == *parachain_head_hash => Err(())
	...
}
```

A few notes on that: the other condition (matching hashes) is actually a separate condition, so maybe in the extension it is expressed better, with a dedicated match branch. In general, we shall ignore tx if it supplies head updated at `<= stored_best_head.at_relay_block_number` and we shall ignore if it has matching hashes. In practice we can't have two blocks with the same hash (crypto-safe hashing) - there's a lot that will break if we'll see it, so second condition may be checked separately or not.

2. Yeah - if there's a better way to do that, go ahead! Thank you!

serban300: > 1. Imo in both cases `at_relay_block_number` leads to error, no (I maybe missing something - sorry)?

From what I understand, if `stored_best_head.at_relay_block_number == updated_at_relay_block_number` and the hashes are different, for `update_parachain_head()` it will go to ` // else: pass`, while the extension will trigger an error. So if I understand correctly this case is treated differently now, even if in practice we shouldn't have to blocks with the same hash.

Anyway, I just wanted to make sure if we should have the same check in both places.

serban300:
#1520 was merged. Marking this PR as reviewed.

Issue #1409, opened 2022-05-24
Add id and parent_id methods for all frame headers


svyatonik:
We have this `HeaderId(number, hash)` structure, which is used across relay and runtime code. Most of times it is filled like that: `HeaderId(*best_header.number(), best_header.hash())`. We can instead add some trait `trait HeaderIdProvider { fn id(&self) -> HeaderId; fn parent_id(&self) -> Option<HeaderId>;  }` and impl it for all implementers of `frame_support::traits::Header`. Then we can call `best_header.id()` instead.

PR #1509, opened 2022-07-14
Define HeaderIdProvider trait

serban300:
Fixes: #1409

Defined `HeaderIdProvider` trait and used it where possible instead of `HeaderId(*best_header.number(), best_header.hash())`

### DIFF
--- a/primitives/runtime/src/lib.rs
+++ b/primitives/runtime/src/lib.rs
@@ -18,21 +18,15 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
-use codec::{Codec, Decode, Encode, FullCodec, MaxEncodedLen};
+use codec::{Decode, Encode, FullCodec, MaxEncodedLen};
 use frame_support::{
 	log, pallet_prelude::DispatchResult, PalletError, RuntimeDebug, StorageHasher, StorageValue,
 };
 use frame_system::RawOrigin;
 use scale_info::TypeInfo;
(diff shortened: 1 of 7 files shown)


primitives/runtime/src/lib.rs
```diff
@@ -83,6 +90,53 @@ pub const ROOT_ACCOUNT_DERIVATION_PREFIX: &[u8] = b"pallet-bridge/account-deriva
...
+/// Generic header id provider.
+pub trait HeaderIdProvider<Number, Hash> {
+	// Get the header id.
+	fn id(&self) -> HeaderId<Hash, Number>;
+
+	// Get the header id for the parent block.
+	fn parent_id(&self) -> Option<HeaderId<Hash, Number>>;
+}
+
+impl<Number, Hash> HeaderIdProvider<Number, Hash::Output> for Header<Number, Hash>
```
svyatonik, line 102: I was thinking of something like `impl<T> HeaderIdProvider for T where T: sp_runtime::traits::Header` instead of relying on `generic::Header`. Then we don't need this extra `+ HeaderIdProvider<Self::BlockNumber, Self::Hash>` requirement.

Can you, please, check if it works? I may be missing something :) Thanks!

serban300: Oh sorry, total miss on my side. Done. Looks much better.

svyatonik: That's fine, thank you :)

PR #1520, opened 2022-07-25
Deduplicate parachains validation

serban300:
Follow-up on #1505

Deduplicate the logic that validates the updated parachain heads.

### DIFF
--- a/modules/parachains/src/extension.rs
+++ b/modules/parachains/src/extension.rs
@@ -17,7 +17,7 @@
 use crate::{Config, Pallet, RelayBlockHash, RelayBlockHasher, RelayBlockNumber};
 use bp_runtime::FilterCall;
 use frame_support::{dispatch::CallableCallFor, traits::IsSubType};
-use sp_runtime::transaction_validity::{InvalidTransaction, TransactionValidity, ValidTransaction};
+use sp_runtime::transaction_validity::{TransactionValidity, ValidTransaction};
 
 /// Validate parachain heads in order to avoid "mining" transactions that provide
 /// outdated bridged parachain heads. Without this validation, even honest relayers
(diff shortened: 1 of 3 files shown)


modules/parachains/src/lib.rs
```diff
@@ -383,40 +422,16 @@ pub mod pallet {
 		) -> Result<UpdateParachainHeadArtifacts, ()> {
 			// check if head has been already updated at better relay chain block. Without this
 			// check, we may import heads in random order
+			Self::validate_updated_parachain_head(
+				&stored_best_head,
+				updated_at_relay_block_number,
+				updated_head_hash,
+				format!("The head of parachain {:?} can't be updated", parachain),
```
svyatonik, line 429: Let's avoid any formatting outside of log calls (I wonder if it compiles actually) in the runtime code

modules/parachains/src/extension.rs
```diff
@@ -43,45 +43,26 @@ where
...
-				InvalidTransaction::Stale.into()
-			},
-			_ => Ok(ValidTransaction::default()),
-		}
+		let maybe_stored_best_head = crate::BestParaHeads::<T, I>::get(parachain);
+		Self::validate_updated_parachain_head(
+			&maybe_stored_best_head,
+			updated_at_relay_block_number,
+			parachain_head_hash,
+			format!("Rejecting obsolete parachain-head {:?} transaction", parachain),
```
svyatonik, line 64: Ditto :)

svyatonik (approved):
Looks good, thank you! Couple of fixes and it is good to merge!

modules/parachains/src/lib.rs
serban300: Yes, doesn't seem to compile. Done.

svyatonik (approved):
Great, thank you!

PR #1521, opened 2022-07-27
Follow-up on #1484

serban300:
Addressed the comments in #1484


svyatonik (approved):
Looks super cool, thank you! :)

PR #1528, opened 2022-08-01
Refund relayer if parachain head has not been updated

### LINKED ISSUE #1440: Refund relayer if parachain head has not been updated
extracted from #1436, since it probably requires more granular benchmarks

The idea is that if relayer by some reason is submitting obsolete (or missing) parachain head, some operations (most important: db writes) are not happened. So we can refund relayer and let other transactions to be included into the block.

svyatonik:
closes #1440 

I was going to use different approach for that initially, but the logic either becomes too complex (parachain head may be rejected at a different levels) or it is almost impossible to do that (e.g. when we have failed to read parachain head from the proof - shall we even try to refund?). So refunding is done similarly to pruning here - we're only refunding if proof is correct, but someone else has already provided better head.

### DIFF
--- a/modules/parachains/src/lib.rs
+++ b/modules/parachains/src/lib.rs
@@ -238,6 +238,7 @@ pub mod pallet {
 				&parachain_heads_proof,
 				parachains.len() as _,
 			);
+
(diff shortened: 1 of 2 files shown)


serban300 (commented):
Looks good ! Marking the PR as reviewed.

PR #1531, opened 2022-08-01
Follow-up on #1419

serban300:
Addressing the comments in #1419

I spent a lot of time thinking about naming. `NoopOption` was the best name I could find so far to make `ParaHashAtSource` more generic although it doesn't seem ideal.

Also it would be nice to define `max_head_id` as a `NoopOption` from the start, but I didn't do it since this way we would have to create an `Arc<Mutex<NoopOption>>` and it would be less efficient (because we would have to lock the mutex even if `max_head_id == Noop::None`, which is not needed now). We could have something more efficient using atomic variables, but that would increase the complexity of the code.

### DIFF
--- a/primitives/runtime/src/lib.rs
+++ b/primitives/runtime/src/lib.rs
@@ -27,7 +27,7 @@ use scale_info::TypeInfo;
 use sp_core::{hash::H256, storage::StorageKey};
 use sp_io::hashing::blake2_256;
 use sp_runtime::traits::{BadOrigin, Header as HeaderT};
-use sp_std::{cmp::Ordering, convert::TryFrom, fmt::Debug, vec, vec::Vec};
+use sp_std::{convert::TryFrom, fmt::Debug, vec, vec::Vec};
 
 pub use chain::{
 	AccountIdOf, AccountPublicOf, BalanceOf, BlockNumberOf, Chain, EncodedOrDecodedCall, HashOf,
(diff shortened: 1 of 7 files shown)


serban300:
> Also it would be nice to define `max_head_id` as a `NoopOption` from the start, but I didn't do it since this way we would have to create an `Arc<Mutex<NoopOption>>` and it would be less efficient (because we would have to lock the mutex even if `max_head_id == Noop::None`, which is not needed now). We could have something more efficient using atomic variables, but that would increase the complexity of the code.

Actually this mutex doesn't seem to be heavily used and changing it to `Arc<Mutex<NoopOption>>` doesn't seem to imply a significant penalty. I added a commit that does this.

primitives/runtime/src/lib.rs
```diff
@@ -85,6 +85,18 @@ pub const ROOT_ACCOUNT_DERIVATION_PREFIX: &[u8] = b"pallet-bridge/account-deriva
 #[derive(RuntimeDebug, Default, Clone, Encode, Decode, Copy, Eq, Hash, PartialEq)]
 pub struct HeaderId<Hash, Number>(pub Number, pub Hash);
 
+impl<Hash: PartialEq, Number: PartialOrd> PartialOrd for HeaderId<Hash, Number> {
+	fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+		self.0.partial_cmp(&other.0)
+	}
+}
+
+impl<Hash: Eq, Number: Ord> Ord for HeaderId<Hash, Number> {
```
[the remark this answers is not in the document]
svyatonik: (same for `PartialOrd`)

relays/lib-substrate-relay/src/parachains/source.rs
```diff
@@ -22,35 +22,34 @@ use async_std::sync::{Arc, Mutex};
...
 	Chain, Client, Error as SubstrateError, HeaderIdOf, HeaderOf, RelayChain,
 };
-use relay_utils::relay_loop::Client as RelayClient;
-use sp_runtime::traits::Header as HeaderT;
+use relay_utils::{relay_loop::Client as RelayClient, NoopOption};
 
 /// Shared updatable reference to the maximal parachain header id that we want to sync from the
 /// source.
-pub type RequiredHeaderIdRef<C> = Arc<Mutex<Option<HeaderIdOf<C>>>>;
+pub type RequiredHeaderIdRef<C> = Arc<Mutex<NoopOption<HeaderIdOf<C>>>>;
```
svyatonik, line 37: This actually seems like a redundant change - we'll be only using `Some` or `None` here. right? :) So we don't need this `Noop` at all. But I'm fine with leaving it as is :)

relays/utils/src/lib.rs
```diff
@@ -271,3 +271,28 @@ where
 		},
 	}
 }
+
+/// An extension over rust's `Option` that adds an extra possibility: `Noop`.
```
svyatonik, line 275: I would rather made it less generic - like something you have suggested [originally](https://github.com/paritytech/parity-bridges-common/pull/1419#discussion_r933856145) (or even `enum AvailableHeader<T> { Available(T), Missing, Unavailable }`, where `Missing` means `None`). It will (imo) add some context to the code + we may add more useful documentation to each option.

But please don't consider my opinion as the requirement - I'm fine with either solution, so please make your own decision here :)

svyatonik (approved):
Please fix the impl of Ord/PartialOrd and then it's fine to merge! Thank you!

svyatonik (approved):

PR #1533, opened 2022-08-02
Use jsonrpsee subscriptions

### LINKED ISSUE #1355: Some fixes for new jsonrpsee version
- [x] check if we need to migrate to new proc-macro based subscriptions mechanism: https://github.com/paritytech/jsonrpsee/blob/master/proc-macros/src/lib.rs#L231-L262 for subscriptions we use;
- [x] new proc macro requires that all its generic params need to implement `Serialize` + `Deserialize`, which isn't true for `Chain`, hence long-lines like [this one](https://github.com/paritytech/parity-bridges-common/pull/1353/files#diff-4c85e463b9a0cbf82ee5f1ee88fe1dcfce081107a76ce2b421c45980e95c247bR271-R279). Find a better solution for this

svyatonik:
closes #1355

I've also splitted this monstrous `Substrate` trait into multiple, grouped by RPC method namespaces.

### DIFF
--- a/relays/client-substrate/src/client.rs
+++ b/relays/client-substrate/src/client.rs
@@ -449,12 +449,6 @@ impl<C: Chain> Client<C> {
 				let tx_hash = C::Hasher::hash(&extrinsic.0);
 				let subscription =
 					SubstrateAuthorClient::<C>::submit_and_watch_extrinsic(&*client, extrinsic)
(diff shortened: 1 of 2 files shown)


serban300 (commented):
Very nice ! Looks good ! Marking this as reviewed.

PR #1535, opened 2022-08-03
[Complex Relayers] Deduplicate l2r and r2l logic

serban300:


svyatonik (approved):
Looks great, thank you! One thing that is confusing me is `l2r` and `r2l` :) When I've got PR notification I didn't even understand what this PR is about (`relayer-to-?`). IMO `left_to_right` and `right_to_left` gives more context when you look at the code, even though it is longer. But again - it may be my personal preference, so don't consider it as a requirement

serban300:
> Looks great, thank you! One thing that is confusing me is `l2r` and `r2l` :) When I've got PR notification I didn't even understand what this PR is about (`relayer-to-?`). IMO `left_to_right` and `right_to_left` gives more context when you look at the code, even though it is longer. But again - it may be my personal preference, so don't consider it as a requirement

Makes sense. I wanted to shorten the names a bit, but readability is more important. Done.

PR #1537, opened 2022-08-04
[Deployments] Deduplicate message sending code

### LINKED ISSUE #1526: Deduplicate some of the code in the deployment scripts
There are big similarities between:

- https://github.com/paritytech/parity-bridges-common/blob/master/deployments/bridges/rialto-millau/entrypoints/relay-messages-to-millau-generator-entrypoint.sh
- https://github.com/paritytech/parity-bridges-common/blob/master/deployments/bridges/rialto-millau/entrypoints/relay-messages-to-rialto-generator-entrypoint.sh
- https://github.com/paritytech/parity-bridges-common/blob/master/deployments/bridges/rialto-parachain-millau/entrypoints/relay-messages-to-millau-generator-entrypoint.sh
- https://github.com/paritytech/parity-bridges-common/blob/master/deployments/bridges/rialto-parachain-millau/entrypoints/relay-messages-to-rialto-parachain-generator-entrypoint.sh

We could try to move the common parts in function and call it from all the scripts

serban300:
Fixes #1526

I was playing with the deployments while reviewing #1412 and since I got here I thought it would be an useful exercise for me to also play with some code and address #1526. The approach might not be ideal (it might be nicer to have a script that receives CLI parameters, but it's harder to do from bash). But anyway, it deduplicates some code.

### DIFF
--- a/deployments/README.md
+++ b/deployments/README.md
@@ -43,14 +43,15 @@ and Grafana. We cover these in more details in the [Monitoring](#monitoring) sec
 the monitoring Compose file is _not_ optional, and must be included for bridge deployments.
 
 ### Running and Updating Deployments
-We currently support two bridge deployments
+We currently support three bridge deployments
 1. Rialto Substrate to Millau Substrate
+2. Rialto Parachain Substrate to Millau Substrate
 2. Westend Substrate to Millau Substrate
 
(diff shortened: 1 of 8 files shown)


svyatonik (approved):
Looks great, thank you!

PR #1545, opened 2022-08-09
Call To*InboundLaneApi::message_details() with batched messages

### LINKED ISSUE #1538: Change `To*InboundLaneApi::message_details` to accept single messages
Follow-up on: #1430

See: https://github.com/paritytech/parity-bridges-common/pull/1430#discussion_r938683806

`To*InboundLaneApi::message_details` receives multiple messages + payloads. If there are a lot of large messages WASM may oom. So it would be safer to change it to process single messages only

serban300:
Closes #1538

### DIFF
--- a/relays/lib-substrate-relay/src/messages_source.rs
+++ b/relays/lib-substrate-relay/src/messages_source.rs
@@ -31,8 +31,8 @@ use async_std::sync::Arc;
 use async_trait::async_trait;
 use bp_messages::{
 	storage_keys::{operating_mode_key, outbound_lane_data_key},
-	InboundMessageDetails, LaneId, MessageData, MessageNonce, MessagesOperatingMode,
-	OutboundLaneData, OutboundMessageDetails, UnrewardedRelayersState,
+	InboundMessageDetails, LaneId, MessageData, MessageNonce, MessagePayload,
+	MessagesOperatingMode, OutboundLaneData, OutboundMessageDetails, UnrewardedRelayersState,
 };
 use bp_runtime::{messages::DispatchFeePayment, BasicOperatingMode, HeaderIdProvider};
 use bridge_runtime_common::messages::{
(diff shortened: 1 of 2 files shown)


relays/lib-substrate-relay/src/messages_source.rs
```diff
@@ -571,71 +592,46 @@ where
...
+			return make_missing_nonce_error(nonce)
+		}
 	}
 
-	// check if last nonce is missing - loop below is not checking this
-	let last_nonce_is_missing =
-		weights.last().map(|details| details.nonce != *nonces.end()).unwrap_or(true);
-	if last_nonce_is_missing {
+	// Check if last nonce is missing. The loop above is not checking this.
+	if !nonces.is_empty() && nonces_iter.peek() == Some(nonces.end()) {
```
svyatonik, line 618: Please verify me here: `nonces_iter` is `noces.rev()`, so we're moving from end to begin and here we are comparing with the `end()` here. Is that intended?

svyatonik (commented):
Just couple of questions, then it is good to merge

relays/lib-substrate-relay/src/messages_source.rs
serban300: Yes, if `nonces_iter.peek() == Some(nonces.end())` at this point, it means that the last nonce was not found in the messages list (and that the messages list is empty actually), and we need to return an error.

relays/lib-substrate-relay/src/messages_source.rs
```diff
@@ -112,6 +112,73 @@ impl<P: SubstrateMessageLane> SubstrateMessagesSource<P> {
...
+					P::TargetChain::NAME,
+					self.lane_id,
+					out_msg_details.nonce,
+					out_msg_details.dispatch_weight,
+					in_msg_details.dispatch_weight,
+				);
+				out_msg_details.dispatch_weight = in_msg_details.dispatch_weight;
+			}
+
+			current_msgs_batch = next_msgs_batch;
```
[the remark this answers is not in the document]
serban300: > Actually imo would be cool if we could have a function that can be easily tested (i.e. without RPC calls). The function shall just split single vector (`msgs_to_refine`) to a vector of vectors. t is optional, but WDYT?

Yes, that's a very good idea. I'll try to move this logic to a separate function.

relays/lib-substrate-relay/src/messages_source.rs
svyatonik: Ok, can we then check "that the messages list is empty"? :) Imo it should be simpler :)

serban300: Yes, I think it should work.

relays/lib-substrate-relay/src/messages_source.rs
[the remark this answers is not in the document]
serban300: Done

relays/lib-substrate-relay/src/messages_source.rs
serban300: Done. Also added another check for the case where there are more messages than nonces.

svyatonik (approved):
Thank you!

PR #1548, opened 2022-08-17
Macro for auto-generating bridge runtime APIs declarations

### LINKED ISSUE #1431: Macro for auto-generating bridge runtime APIs declarations
Right now chain-specific primitives crates (`primitives/chain-kusama`, `primitives/chain-millau`, ...) mostly consist of 3 sections:
1) chain-specific types and numeric constants;
2) declaration of chain-specific bridge runtime APIs (`KusamaFinalityApi`, `ToKusamaOutboundLaneApi`, `FromKusamaInboundLaneApi`, ...);
3) constants that are stringified names of runtime API methods (`TO_KUSAMA_ESTIMATE_MESSAGE_FEE_METHOD`, `TO_KUSAMA_MESSAGE_DETAILS_METHOD`, `FROM_KUSAMA_MESSAGE_DETAILS_METHOD`, ...).

There must be a way to write something like `declare_bridge_finality_api` and `declare_bridge_messages_api` that will generate all the stuff from (2) and (3) above.

I've tried it long time ago, but the problem was that I was unable to call `decl_runtime_apis` macro from my macro back then. Probably there's a way to do that? Or something has changed.

serban300:
Fixes #1431


acatangiu (approved):
Looks good to me :+1:

svyatonik (approved):
Seems cool, thank you! One thing - it probably worth to split finality + message API declarations. I.e. there's still `bp-westend` crate, which is untouched by this PR, because (iiuc) it only declares finality APIs. So imo there may be two or three macro: one for declaring finality APIs, one for declaring message APIs and (optionally) one that declares both APIs. We'll even start with finality-only APIs in brand new RococoBridgeHub <> WococoBrtidgeHub bridge, this way we can use this macro there, WDYT?

svyatonik:
(it may be done in a separate PR though)

serban300:
> Seems cool, thank you! One thing - it probably worth to split finality + message API declarations. I.e. there's still `bp-westend` crate, which is untouched by this PR, because (iiuc) it only declares finality APIs. So imo there may be two or three macro: one for declaring finality APIs, one for declaring message APIs and (optionally) one that declares both APIs. We'll even start with finality-only APIs in brand new RococoBridgeHub <> WococoBrtidgeHub bridge, this way we can use this macro there, WDYT?

Makes sense. I can do it in this PR. I'll add a commit later.

serban300:
> > Seems cool, thank you! One thing - it probably worth to split finality + message API declarations. I.e. there's still `bp-westend` crate, which is untouched by this PR, because (iiuc) it only declares finality APIs. So imo there may be two or three macro: one for declaring finality APIs, one for declaring message APIs and (optionally) one that declares both APIs. We'll even start with finality-only APIs in brand new RococoBridgeHub <> WococoBrtidgeHub bridge, this way we can use this macro there, WDYT?
> 
> Makes sense. I can do it in this PR. I'll add a commit later.

Hmmm this is not trivial, because we can't use `decl_runtime_apis!` twice in a module. I will merge this PR as it is, keep the issue opened, and try to address this in a future PR.

svyatonik:
Thanks!

PR #1549, opened 2022-08-18
Follow-up on #1536

serban300:
Addressing the comments in #1536

I thought quite a bit about this. In the end I moved a lot of the logic to `RelayReference`. I hope it makes sense.


svyatonik (approved):
Looks perfect, thank you!

Issue #1369, opened 2022-04-05
Change grafana alert conditions on test deployments


svyatonik:
There are still some alerts that are starting with "[No Data]" state when deployments restarts:
```
[No Data] Rewards are not being confirmed
[No Data] Messages generated at Millau are not detected by relay
[No Data] Too many unconfirmed messages
[No Data] At-Rialto relay balances are too low
```

There are imo two solutions:
1) to change alert settings to not to fire if there's no data;
2) change alert formula so that it'll not fire if data is unavailable.

svyatonik:
shall be fixed now

serban300:
I've been experimenting with this. I think we could set
`If no data or all values are null - set state to - keep last state`
for all these alerts. 

And I'm trying to see if we can add a metric + alert on uptime in order to receive a notification when the deployments are restarted.

serban300:
A couple of updates:

- I managed to create an alarm on nodes uptime which would trigger on each restart. This seems reliable
- I tried to set `"noDataState": "keep_state"` for all the current alerts. It doesn't seem to generate any issues. The problem is that if we do this, in case we have a problem with a metric and it doesn't log anymore, we won't get any notification. So it doesn;t seem very safe.
- It would be nice if we could delay the alerts evaluation on no data somehow, but I couldn't find any way to do this in a generic way for all the alerts. Still researching.

serban300:
Options so far:

1. Setting `"noDataState": "keep_state"` for all the alerts. But if we do this, in case we have a problem with a metric and it doesn't log anymore, we won't get any notification. So it doesn't seem safe.
2. Changing the queries or adding another query which has a default value which would trigger the alarm. For example for the `Rewards are not being confirmed (Millau -> RialtoParachain messages)` alarm, we have the following query:
```
  (
      scalar(
          max_over_time(
            Millau_to_RialtoParachain_MessageLane_00000000_lane_state_nonces{instance="relay-millau-rialto-parachain-1:9616",type="source_latest_confirmed"}[2m]
          )
        or on ()
          vector(0)
      )
    -
      scalar(
          max_over_time(
            Millau_to_RialtoParachain_MessageLane_00000000_lane_state_nonces{instance="relay-millau-rialto-parachain-1:9616",type="target_latest_confirmed"}[2m]
          )
        or on ()
          vector(0)
      )
  )
*
  (
      max_over_time(
[...]

PR #1555, opened 2022-08-26
[RFC] Avoid receiving [No Data] messages when restarting deployments

serban300:
fixes: #1369

- Delays the startup of the grafana container by 3min in order to avoid receiving [No Data] messages
- Adds a metric and alert on running nodes, which will trigger when restarting deployment

Publishing this as RFC. This is the solution that sounds best to me, but I'm not sure about it and I would be happy to get some feedback on it.


svyatonik (approved):
That's a cool solution imo!

serban300:
> That's a cool solution imo!

Thanks for the review ! I'll merge the PR then.

PR #1561, opened 2022-08-30
Extract unsigned tx from the SignParam structure

svyatonik:
I've extracted this commit from the #1544, since imo it deserves a separate review. The issue is that the `submit_and_watch_signed_extrinsic` will be accepting a callback that crafts the `UnsignedTransaction` instead of `Bytes` - we need it to get proper transaction era to use in `TransactionTracker` (see #1544).

There may be small changes that are unrelated to this PR - they're a part of the #1544. I can remove or leave - up to reviewers.


serban300 (approved):
Looks good to me !

relays/bin-substrate/src/cli/register_parachain.rs
```diff
@@ -231,22 +223,18 @@ impl RegisterParachain {
...
-							unsigned: UnsignedTransaction::new(
-								force_lease_call.into(),
-								transaction_nonce,
-							),
-						})?
-						.encode(),
-					))
-				})
+				.submit_signed_extrinsic(
+					relay_sudo_account.clone(),
```
bkontur, line 227: @svyatonik 
I guess this last `relay_sudo_account.clone()` could be just `relay_sudo_account`

bkontur (approved):
lgtm

serban300 (approved):

PR #1567, opened 2022-09-02
Follow-up on #1068

serban300:
Addresses some comments from #1068

Also defines `StorageMapKeyProvider`

### DIFF
--- a/primitives/runtime/src/lib.rs
+++ b/primitives/runtime/src/lib.rs
@@ -286,6 +286,28 @@ pub fn storage_value_key(pallet_prefix: &str, value_name: &str) -> StorageKey {
 }
 
 /// Can be use to access the runtime storage key of a `StorageDoubleMap`.
+pub trait StorageMapKeyProvider {
+	// The name of the variable that holds the `StorageDoubleMap`
+	const MAP_NAME: &'static str;
+
+	// The same as `StorageMap::Hasher1`
+	type Hasher: StorageHasher;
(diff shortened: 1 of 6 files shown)


primitives/runtime/src/lib.rs
```diff
@@ -285,6 +285,28 @@ pub fn storage_value_key(pallet_prefix: &str, value_name: &str) -> StorageKey {
 	StorageKey(final_key)
 }
 
+/// Can be use to access the runtime storage key of a `StorageDoubleMap`.
```
svyatonik, line 288: Iiuc it isn't about `StorageDoubleMap`, but about `StorageMap`

primitives/runtime/src/lib.rs
```diff
@@ -285,6 +285,28 @@ pub fn storage_value_key(pallet_prefix: &str, value_name: &str) -> StorageKey {
...
+pub trait StorageMapKeyProvider {
+	// The name of the variable that holds the `StorageDoubleMap`
```
svyatonik, line 290: Let's change it to `///` maybe (and below as well)

svyatonik (approved):
LGTM

primitives/runtime/src/lib.rs
serban300: True. Done.

primitives/runtime/src/lib.rs
serban300: Done.

PR #1568, opened 2022-09-06
Add events to the bridge parachains pallet

### LINKED ISSUE #1566: Bridge parachain pallet error handling improvments
I am calling extrinsic `bridgeWococoParachain.submitParachainHeads`, everything seems to work, because it is marked as `system.ExtrinsicSuccess`

image

I was expecting for change of `chainstate -> bridgeWococoParachain -> bestParaHeads`, but nothing happens,

so I started parachain collators with `RUST_LOG=runtime::bridge-parachains`, so then I saw just this in logs:
```
2022-09-02 12:49:36.032 TRACE tokio-runtime-worker runtime::bridge-parachains: [Parachain] The read of head of parachain <wasm:stripped> has failed: <wasm:stripped>  
```

https://github.com/paritytech/parity-bridges-common/blob/master/modules/parachains/src/lib.rs#L242-L323
there are lots of `continue`, because we could iterate several parachains and we dont want to finish on the first error,
but now, there is hard to identify if call passed or not

---
**suggested improvements:**
1. at least change trace to error and would be cool the see real error in logs
```
                                                        Err(e) => {
								log::trace!(
									target: LOG_TARGET,
									"The read of head of parachain {:?} has failed: {:?}",
									parachain,
									e,
								);
								continue;
							},
```

2. maybe we could trigger for every parachain `for (parachain, parachain_head_hash) in parachains` a custom Event (Error or Success) on every `continue`:
```
Self::deposit_event(Event::ParaIdProcessed(1013);
...
Self::deposit_event(Event::ParaIdErrored("some message);
...
Self::deposit_event(Event::ParaIdNotProcessed("some message reason");
```

3. other?

svyatonik:
closes #1566 

cc @bkontur

### DIFF
--- a/bin/millau/runtime/src/lib.rs
+++ b/bin/millau/runtime/src/lib.rs
@@ -530,6 +530,7 @@ parameter_types! {
 pub type WithRialtoParachainsInstance = ();
 
 impl pallet_bridge_parachains::Config<WithRialtoParachainsInstance> for Runtime {
+	type Event = Event;
 	type WeightInfo = pallet_bridge_parachains::weights::MillauWeight<Runtime>;
 	type BridgesGrandpaPalletInstance = RialtoGrandpaInstance;
 	type ParasPalletName = RialtoParasPalletName;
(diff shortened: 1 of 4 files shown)


serban300 (approved):
LGTM !

PR #1582, opened 2022-09-23
Read extrinsic dispatch result for mined transaction

### LINKED ISSUE #1464: Watch for submitted relay transactions invaldiation
Extracted from #1463. There was similar suggestion in the past and I've left my comment there (can't find it now). We can't absolutely rely on the transaction watch mechanism, but in some cases it may be helpful to us to know that the transaction has been invalidated. In this case we may consider "manually" trigger stall condition and let relay restart.

Fixing it isn't trivial, so I'm opening this separate issue.

svyatonik:
closes #1464 

TODOs:
- [x] actually implement `fn extrinsic_dispatch_result` for `Client<C>`;
- [x] add tests (in messages relay);
- [x] test it locally.

~I'm currently unsure - what's the best way to fix (1), because it requires more runtime knowledge than we have in our mocked runtimes. And I don't want to extend these requirements. So probably we'll need some hack here - like assume that the system pallet always comes first in `construct_runtime!( .. )`, that it always have the "System" name and so on. Investigation in progress => WIP~ Unless I'm missing something, we can't decode events without knowing correct runtime `Event` type (and we can't know it, since we're using mocked events) or without decoding metadata (which is the end goal of relayer development, but it is too far away in the future). Is there any other way to do that? Can we handle it without reading events in realyer? Need some time to explore alternate solutions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8536,6 +8536,7 @@ dependencies = [
  "bp-header-chain",
  "bp-messages",
  "bp-runtime",
+ "env_logger 0.8.4",
  "finality-relay",
  "frame-support",
  "frame-system",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8536,7 +8536,6 @@ dependencies = [
  "bp-header-chain",
  "bp-messages",
  "bp-runtime",
- "env_logger 0.8.4",
  "finality-relay",
  "frame-support",
  "frame-system",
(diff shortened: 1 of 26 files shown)


svyatonik:
So the final idea is to change relay loops so that they will read required on-chain data when they see their transaction is finalized and if required data is not updated, they'll consider themselves stalled. Meaning restart and resubmit.

serban300 (approved):

PR #1586, opened 2022-09-30
Remove unnecessary consts

svyatonik:
There are some consts that are not needed anymore, since we're not going to submit transactions to some chains anymore. this PR removes such consts.


bkontur (approved):
lgtm

PR #1587, opened 2022-09-30
Cleaning deps + satisfy `cargo build --release --all-targets --all-fe…

bkontur:
…atures`


Cargo.toml
```diff
@@ -98,7 +98,7 @@ sc-consensus-babe = { git = "https://github.com/paritytech//substrate", branch =
 sc-consensus-babe-rpc = { git = "https://github.com/paritytech//substrate", branch = "sv-locked-for-gav-xcm-v3-and-bridges" }
 sc-consensus-epochs = { git = "https://github.com/paritytech//substrate", branch = "sv-locked-for-gav-xcm-v3-and-bridges" }
 sc-consensus-slots = { git = "https://github.com/paritytech//substrate", branch = "sv-locked-for-gav-xcm-v3-and-bridges" }
-sc-consensus-uncles = { git = "https://github.com/paritytech//substrate", branch = "sv-locked-for-gav-xcm-v3-and-bridges" }
+#sc-consensus-uncles = { git = "https://github.com/paritytech//substrate", branch = "sv-locked-for-gav-xcm-v3-and-bridges" }
```
svyatonik, line 101: It'd better to remove it if it isn't used anymore. Same for lines below

bin/rialto-parachain/runtime/Cargo.toml
```diff
@@ -11,9 +11,9 @@ substrate-wasm-builder = { git = "https://github.com/paritytech/substrate", bran
 
 [dependencies]
 codec = { package = 'parity-scale-codec', version = '3.1.5', default-features = false, features = ['derive']}
+hex-literal = "0.3"
```
svyatonik, line 14: Is it really required - can't see any added usages of this crate?

bin/rialto-parachain/runtime/src/lib.rs
```diff
@@ -777,6 +777,13 @@ impl_runtime_apis! {
 
 	#[cfg(feature = "runtime-benchmarks")]
 	impl frame_benchmarking::Benchmark<Block> for Runtime {
+		fn benchmark_metadata(_extra: bool) -> (
+			Vec<frame_benchmarking::BenchmarkList>,
+			Vec<frame_support::traits::StorageInfo>,
+		) {
+			todo!("TODO: fix or remove")
```
svyatonik, line 784: We don't have any tests for benchmarking on CI, so this could break existing benchmarks (I know nothing about this method - so not sure). Can you, please, file an issue so that we (I) can test and fix it in the near future? Thanks!

svyatonik (approved):
Looks good - left a couple of questions/suggestions

bin/rialto-parachain/runtime/Cargo.toml
bkontur: older code doesnt compile here for `#[cfg(feature = "runtime-benchmarks")]`
https://github.com/paritytech/parity-bridges-common/blob/master/bin/rialto-parachain/runtime/src/lib.rs#L790-L798

bin/rialto-parachain/runtime/src/lib.rs
bkontur: @svyatonik created https://github.com/paritytech/parity-bridges-common/issues/1588,
could you please also check and update it?

probably for rialto-parachain we dont need any benchmarks, but for bridges pallets used by BridgeHub, we will need to add some benchmarks for weights caclculation (I guess, I know how, because I was fixing similar stuff for XCMv3 branch)

svyatonik: Sure, thank you!

PR #1591, opened 2022-10-04
Use proper account types

### LINKED ISSUE #1590: Change account type, used in TargetHeaderChain implementation
thanks for report to @bkontur 

https://github.com/paritytech/parity-bridges-common/blob/master/primitives/messages/src/source_chain.rs#L64
https://github.com/paritytech/parity-bridges-common/blob/master/primitives/messages/src/source_chain.rs#L88
there is a AccountId which is used for InboundLaneData<AccountId>

but when I check implementation, e.g.:
https://github.com/paritytech/parity-bridges-common/blob/master/bin/millau/runtime/src/rialto_messages.rs#L239
```rust
impl TargetHeaderChain<ToRialtoMessagePayload, bp_rialto::AccountId> for Rialto {
```
AccountId is bp_rialto::AccountId but InboundLaneData<bp_millau::AccountId>

=> this needs to be fixed and we need better documentation for the trait

svyatonik:
closes #1590

### DIFF
--- a/primitives/messages/src/source_chain.rs
+++ b/primitives/messages/src/source_chain.rs
@@ -61,6 +61,10 @@ pub struct RelayerRewards<Balance> {
 /// All implementations of this trait should only work with finalized data that
 /// can't change. Wrong implementation may lead to invalid lane states (i.e. lane
 /// that's stuck) and/or processing messages without paying fees.
+///
+/// The `Payload` type here means the payload of the mesage that is sent from the
+/// source chain to the target chain. The `AccountId` type here means the account
+/// type used by the source chain.
 pub trait TargetHeaderChain<Payload, AccountId> {
 	/// Error type.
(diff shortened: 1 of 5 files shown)


primitives/messages/src/source_chain.rs
```diff
@@ -61,6 +61,10 @@ pub struct RelayerRewards<Balance> {
 /// All implementations of this trait should only work with finalized data that
 /// can't change. Wrong implementation may lead to invalid lane states (i.e. lane
 /// that's stuck) and/or processing messages without paying fees.
+///
+/// The `Payload` type here means the payload of the mesage that is sent from the
```
bkontur, line 65:
```suggestion
/// The `Payload` type here means the payload of the message that is sent from the
```

bkontur (approved):
lgtm

PR #1593, opened 2022-10-04
Rename MillauWeight to BridgeWeight

svyatonik:
To avoid chain names in abstract pallets as @bkontur has suggested


bkontur (approved):
lgtm :)

PR #1603, opened 2022-10-18
Increase max header size of Rialto and Millau (512 -> 1024)

### LINKED ISSUE #1602: Test deployments are broken
Rialto<>Millau<>RialtoParachain bridges are currently broken - relay fails with `BridgePalletIsNotInitialized` error.

svyatonik:
closes #1602 

Normally we have `3` authorities in the digest, but there's first session with `5` authorities, so the sync has been broken


serban300 (approved):

bkontur (approved):
lgtm

PR #1597, opened 2022-10-10
Bump Substrate/Polkadot/Cumulus refs (aka Weights v1.5)

svyatonik:
~WIP, just to show some progress~
I'm trying to use `Weight::from_ref_time()`, `Weight::ref_time()` everywhere instead of expressions like `Weight::zero()` to make it easier to upgrade to Weights v2 after they'll be available

For the future me (or whoever will be working on upgrading from weights 1.5 to 2.0):
1) grep for `ref_time` - all code needs to be double checked;
2) there's a lot of things happening with weights in messages pallet and in messages relay. There are couple of places where we need to div weights - this op is currently unavailable (?). So all these things need to be rechecked and fixed;
3) I guess there'll be a way to refund unused bandwidth component - we'll also need to do that in all our pallets

PR #1606, opened 2022-10-18
Add BEEFY finality pallet with basic functionality

serban300:
## General context ##

This PR is a continuation of: https://github.com/paritytech/parity-bridges-common/pull/1367

This PR introduces a new pallet that can handle BEEFY commitments. For the beginning I would just like to push an initial version of this pallet with basic functionality and then introducing more functionality, improvements and addressing the known shortcomings incrementally in future PRs. More details below:

## Prerequisites ##

Before merging this we need https://github.com/paritytech/parity-bridges-common/pull/1597 to be merged. And also to add https://github.com/paritytech/substrate/pull/12516 to our substrate branch.

### Leaf version ###

For the moment we're not checking the MMR Leaf version. I'm not sure if we have to, or if it will be handled automatically by the encode/decode logic.

### The stored data ###

Right now, the pallet stores a map of `block_number => { parent_number_and_hash, mmr_root }` which has one entry for each of the latest `CommitmentsToKeep` imported commitments.

There are 2 issues with this: 
1. The `FinalityApi::best_finalized()` method returns a full header (it calls the `best_finalized()` method exposed by the GRANDPA finality pallet). This method is used by the relayers
2. When relaying messages, the code ends up calling the `parse_finalized_storage_proof(hash: BridgedBlockHash<T, I>, storage_proof: sp_trie::StorageProof, ...)` method exposed by the GRANDPA finality pallet which:
   - retrieves a header by its hash
   - calls `header.state_root()` in order to be able to check storage proofs

So we might want to:
- store full headers too
- use the header hash as key for the map
In order to stay compatible with this logic. Otherwise we'll need significant changes to support both GRANDPA and BEEFY.

On the other hand, with BEEFY we can prove any previous block if we have the MMR root and the caller provides a valid proof. So this would be a valid alternative to `parse_finalized_storage_proof(hash: BridgedBlockHash<T, I>, storage_proof: sp_trie::StorageProof, ...)` . We could have a similar method that provides the header, a mmr proof and a storage proof. Not sure what's the right trade-off here.

Probably storing header would be the easiest solution for the beginning, but I would like to address this in a future PR.

## Unimplemented ##

Related to points 2. and 4. [here](https://github.com/paritytech/parity-bridges-common/pull/1367#issuecomment-1103859334):
```
2) do not store validator public keys - instead for every incoming signature the caller must provide merkle proof of validator membership. We know merkle root from the `BeefyNextAuthoritySet` structure;
4) the single signature verification must be isolated so that we can experiment with 1/3 random signatures in the future.
```

The pallet is not storing validator public keys, but right now it's receiving the entire list of validators in `submit_commitment()`. 

I'm not sure if providing a merkle proof for each validator is efficient, even if we could work only with 1/3 signatures. I think a proof would require log(validators.len()) hashes. So for example for 1000 validators:

Current approach (sending all the validators and signatures): 1000 validators + 1000 signatures = 1000 * 33 bytes + 1000 * 64 bytes = 97k bytes
Sending only 1/3 validators and signatures + proofs = 333 validators + 333 signatures + 333 proofs = 333 * 33 bytes + 333 * 64 bytes + 333 * 10 * 32 bytes = 139k bytes

Please correct me if I'm wrong or if I'm missing something.


svyatonik:
> So we might want to:

> - store full headers too
> - use the header hash as key for the map

> In order to stay compatible with this logic. Otherwise we'll need significant changes to support both GRANDPA and BEEFY.

Please keep in mind that messaging pallet has no proof verification logic - that is done outside of the pallet. So you can simply make another adapter for this new pallet and configure messaging pallet to use it instead of GRANDPA pallet adapter. I don't see any significant changes here.

---

> I'm not sure if providing a merkle proof for each validator is efficient, even if we could work only with 1/3 signatures. I think a proof would require log(validators.len()) hashes. So for example for 1000 validators:

> Current approach (sending all the validators and signatures): 1000 validators + 1000 signatures = 1000 * 33 bytes + 1000 * 64 bytes = 97k bytes
> Sending only 1/3 validators and signatures + proofs = 333 validators + 333 signatures + 333 proofs = 333 * 33 bytes + 333 * 64 bytes + 333 * 10 * 32 bytes = 139k bytes

> Please correct me if I'm wrong or if I'm missing something.

Re proof size (`333 * 10 * 32 bytes`) iiuc this formula is not correct - most of these `* 10` hashes will be the same. I.e. there's always an intersection of trie nodes used to prove `validator-A-membership` and `validator-B-membership`. In the worst case they'll only share the merkle root, but in average scenario more nodes will be shared. So your "total proof" will be lesser than `333 * 10 * 32 bytes` - don't know how much exactly :) Proof verification takes some additional time, though.

I'm not insisting here - just sharing my thoughts about the proof. You should make your own decision here :)

modules/beefy/src/lib.rs
```diff
@@ -0,0 +1,652 @@
+#![feature(assert_matches)]
```
svyatonik, line 1: Is it required?

modules/beefy/src/utils.rs
```diff
@@ -0,0 +1,361 @@
...
+	commitment
+		.commitment
+		.payload
+		.get_decoded(&bp_beefy::MMR_ROOT_PAYLOAD_ID)
+		.ok_or(Error::MmrRootMissingFromCommitment)
+}
+
+pub(crate) fn verify_commitment<T: Config<I>, I: 'static>(
+	commitment: &BridgedBeefySignedCommitment<T, I>,
+	authority_set: &BridgedBeefyAuthoritySet<T, I>,
```
svyatonik, line 109: Is there a difference between authorities and validators here? If not, then what `authority_set` means and how it is different from `validator_set`? Let's probably stick to same term everywhere - e.g. use authorities or validators everywhere?

modules/beefy/src/lib.rs
```diff
@@ -0,0 +1,652 @@
...
+		/// authority set and best finalized block number in order to verify that the commitment
+		/// is valid.
+		///
+		/// If successful in verification, it will update the underlying storage with the data
+		/// provided in the newly submitted commitment.
+		#[pallet::weight(0)]
+		pub fn submit_commitment(
+			origin: OriginFor<T>,
+			commitment: BridgedBeefySignedCommitment<T, I>,
+			validator_set: BridgedBeefyValidatorSet<T, I>,
```
svyatonik, line 208: IIUC, according to the `verify_commitment`, this should be the current authorities set, right? Why do we need it if we're already storing it under the `CurrentAuthoritySet` key?

svyatonik: (I don't see any places where `CurrentAuthoritySet` is updated actually - do you plan to fix it in follow up PRs?)

svyatonik (commented):
Couple of questions

modules/beefy/src/lib.rs
serban300: No. Forgot about it. I'll remove it.

modules/beefy/src/lib.rs
serban300: Sorry, I must admit that the names `authority_set` and `validator_set` are confusing. I need to find better ones. Maybe `AuthoritySetInfo`/`AuthoritySetMetadata` and `AuthoritySet`/`FullAuthoritySet` would be better.

The authority set contains only some metadata about the validators + the MMR root. It's defined [here](https://github.com/paritytech/substrate/blob/49734dd1d72a00b9d3b87ba397661a63e0e17af3/primitives/beefy/src/mmr.rs#L105-L124).

The validator set is the full list of authority ids (public keys). It's defined [here](https://github.com/paritytech/substrate/blob/49734dd1d72a00b9d3b87ba397661a63e0e17af3/primitives/beefy/src/lib.rs#L124-L129).

We're only storing the `AuthoritySet` (metadata + MMR root) and we receive the `ValidatorSet` (full list of authority public keys). Then:
1. we verify the provided validators are correct by checking `ValidatorSet` against the stored `AuthoritySet::root`
2. we use the provided validators public keys to verify the commitment signatures

The `CurrentAuthoritySet` is updated [here](https://github.com/serban300/parity-bridges-common/blob/64cae305e2a79d5b070927995cf6646e93e3f032/modules/beefy/src/lib.rs#L240-L242)

modules/beefy/src/utils.rs
serban300: Sorry, this names are confusing. Added more context about this in a different comment.

modules/beefy/src/lib.rs
svyatonik: Ok, thanks for the explanation :) Yeah - maybe renaming to something else is a good idea :)

serban300: Done.

svyatonik: I also suggest to think about optimizing usage of this argument a bit. If you have chose not to submit proof-of-membership and instead submit the whole set, then there are many low-hanging optimizations that will save a lot of tx/block size (thinking of bridge hub - would allow other transactions to be included into block). E.g. we may only submit whole set and verify merkle proof when authorities set is changed (that'ss what has been implemented in #1367 initially) or something like that (e.g. cache some public keys when they're first verified and then relayer may only submit public keys that are missing from the runtime storage). This is an optimization for a follow-up PRs of course but worth adding TODO or filing an issue for that.

bin/rialto/runtime/src/lib.rs
```diff
@@ -991,10 +1002,11 @@ mod tests {
...
 		);
-		// Largest inner Call is `pallet_session::Call` with a size of 224 bytes. This size is a
-		// result of large `SessionKeys` struct.
-		// Total size of Rialto runtime Call is 232.
-		const MAX_CALL_SIZE: usize = 232;
+		// Largest inner Call is `pallet_bridge_beefy::Call` with a size of 240 bytes. This size
+		// is a result of the large number and complexity of the structs received by the
+		// `submit_commitment` call.
+		// Total size of Rialto runtime Call is 280.
+		const MAX_CALL_SIZE: usize = 280;
```
svyatonik, line 1009: It is better to avoid increasing this constant when possible, especially when it is increased by 20%. Would be better to make your large args be `Box<T>` instead of `T`. Don't know why this particular value has been chosen (it is used by Polkadot/Kusama/..), but I guess there's some research+assumptions behind it.

modules/beefy/src/lib.rs
serban300: Makes sense. Thanks for the suggestions ! I created #1608 to track this.

svyatonik (approved):
lgtm

bin/rialto/runtime/src/lib.rs
serban300: Ok, sorry I didn't know that. For the moment I boxed `MmrLeaf`. It was taking 96 bytes.

modules/beefy/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+		/// If successful in verification, it will update the underlying storage with the data
+		/// provided in the newly submitted commitment.
+		#[pallet::weight(0)]
+		pub fn submit_commitment(
+			origin: OriginFor<T>,
+			commitment: BridgedBeefySignedCommitment<T, I>,
+			validator_set: BridgedBeefyAuthoritySet<T, I>,
+			mmr_leaf: Box<BridgedBeefyMmrLeaf<T, I>>,
+			mmr_proof: BridgedMmrProof<T, I>,
+		) -> DispatchResult
```
acatangiu, line 210: I think in the first version we deploy we should just save the current `validator_set` (full set of BEEFY validator keys) in storage instead of using block transaction space here on each call.

We know from MMR Leaf contents on which block the set changes and have its "metadata" (validator set keys mmr root) to validate the input.

This call will then only accept a `commitment` if either:
- it's for a BEEFY mandatory block (session-boundary block) _and_ full new validator set is also passed as argument (after validating it against known metadata, we update storage with new set),
- or, it's for a non-mandatory BEEFY block, _and_ `validator_set` param is `None` and the one from storage is used (we also verify the one from storage matches current commitment metadata).

```suggestion
		pub fn submit_commitment(
			origin: OriginFor<T>,
			commitment: BridgedBeefySignedCommitment<T, I>,
			validator_set: Option<BridgedBeefyAuthoritySet<T, I>>,
			mmr_leaf: Box<BridgedBeefyMmrLeaf<T, I>>,
			mmr_proof: BridgedMmrProof<T, I>,
		) -> DispatchResult
```

or we could expose different `submit_mandatory_commitment()` call that takes `validator_set` while `submit_commitment()` doesn't.

acatangiu (commented):
Review still in progress, but leaving some ideas

acatangiu (approved):

modules/beefy/src/lib.rs
serban300: We can do this. The current approach is not efficient at all so we'll definitely need to change it before deploying. Would it be ok to track this as part of #1608 ?

acatangiu:
Green light to merge from me :+1:

PR #1609, opened 2022-10-20
Finality loop - cosmetic changes

serban300:
Just moving some code around in order to simplify some methods and make the logic easier to follow (in my opinion).


acatangiu (approved):
nice, does look better imo as well 👍

svyatonik (approved):
LGTM

PR #1612, opened 2022-10-21
Fix parachains_loop::tests::minimal_working_case

### LINKED ISSUE #1610: Fix `parachains_loop::tests::minimal_working_case`
This test never passes if we run it individually or if we run all the tests with `--test-threads=1`

In our CI it probably succeeds because it runs in parallel with other tests which affects it somehow.

serban300:
Fixes #1610 

The parachains loop waits for new block or exit signal before performing the actual relaying logic:

```
		select! {
			_ = async_std::task::sleep(min_block_interval).fuse() => {},
			_ = exit_signal => return Ok(()),
		}
```

The problem is that when the `min_block_interval` is 0 and an an exit signal was also received, both branches will complete at the same time. In this case `select!` doesn't guarantee which branch will be chosen. When the new block branch is chosen, the test fails.

I changed this to use `select_biased` which chooses the first defined branch of the completed ones, and defined the branch for the exit signal before the branch for new block in order to prioritize it.

### DIFF
--- a/relays/parachains/src/parachains_loop.rs
+++ b/relays/parachains/src/parachains_loop.rs
@@ -24,7 +24,7 @@ use bp_polkadot_core::{
 };
 use futures::{
 	future::{FutureExt, Shared},
-	poll, select,
+	poll, select_biased,
 };
 use relay_substrate_client::{BlockNumberOf, Chain, HeaderIdOf};
 use relay_utils::{
@@ -217,9 +217,9 @@ where
 
 	loop {
 		// either wait for new block, or exit signal
-		select! {
-			_ = async_std::task::sleep(min_block_interval).fuse() => {},
+		select_biased! {
 			_ = exit_signal => return Ok(()),
+			_ = async_std::task::sleep(min_block_interval).fuse() => {},
 		}
 
 		// if source client is not yet synced, we'll need to sleep. Otherwise we risk submitting too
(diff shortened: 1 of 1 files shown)


svyatonik (approved):
The code changes looks good, but probably worth leaving a code comment to avoid accidental revert in the future :) There's also an alternative to this fix - to change test a bit - share `TestClientData` between source and target clients + update `target_heads` in `submit_parachain_heads_proof`. But this fix also works. Thank you!

serban300:
Thanks ! I added a comment.

> There's also an alternative to this fix - to change test a bit - share TestClientData between source and target clients + update target_heads in submit_parachain_heads_proof

Yes, this can also work. Personally I would prioritize the exit signal anyway since it doesn't make sense to do any further processing after we receive it, and testing the exit signal branch is a good thing also. But we could add another test that does what you suggest maybe. I'm not sure if this is covered already. Should I create an issue for adding such a test ?

svyatonik:
> Thanks ! I added a comment.
> 
> > There's also an alternative to this fix - to change test a bit - share TestClientData between source and target clients + update target_heads in submit_parachain_heads_proof
> 
> Yes, this can also work. Personally I would prioritize the exit signal anyway since it doesn't make sense to do any further processing after we receive it, and testing the exit signal branch is a good thing also. But we could add another test that does what you suggest maybe. I'm not sure if this is covered already. Should I create an issue for adding such a test ?

Please don't - your solution works as well :)

PR #1622, opened 2022-11-04
Reintroduce header chain trait

svyatonik:
There's a lot of changes, but it is mostly moving code around and removing some duplicate functions. The main change is the new `HeaderChain` which represents a storage of `Chain` headers. This trait is implemented for both bridge GRANDPA and parachains pallets (for latter the adapter struct `ParachainHeaders` is used). This way we may write code (e.g. glue code for bridge-messages pallet)  in abstracted way - e.g. there's now single `verify_messages_proof` function vs previous `verify_messages_proof` (for chains with GRANDPA) `verify_messages_proof_from_parachain` (for parachains).

I've also refactored + removed some obsolete concepts from the `messages.rs` file + changed how tests are working (we no longer need type tests, since we're using `Chain` trait).

~Draft until I'll test it locally~Tested

### DIFF
--- a/bin/millau/runtime/src/rialto_messages.rs
+++ b/bin/millau/runtime/src/rialto_messages.rs
@@ -16,7 +16,7 @@
 
 //! Everything required to serve Millau <-> Rialto messages.
 
-use crate::{OriginCaller, Runtime, RuntimeCall, RuntimeOrigin};
+use crate::{OriginCaller, RialtoGrandpaInstance, Runtime, RuntimeCall, RuntimeOrigin};
 
 use bp_messages::{
 	source_chain::TargetHeaderChain,
@@ -98,6 +98,8 @@ impl MessageBridge for WithRialtoMessageBridge {
 
 	type ThisChain = Millau;
 	type BridgedChain = Rialto;
+	type BridgedHeaderChain =
+		pallet_bridge_grandpa::GrandpaChainHeaders<Runtime, RialtoGrandpaInstance>;
 
 	fn bridged_balance_to_this_balance(
 		bridged_balance: bp_rialto::Balance,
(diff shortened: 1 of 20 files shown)


bin/runtime-common/src/messages.rs
```diff
@@ -134,7 +115,13 @@ impl<
 	}
 }
 
-/// This chain that has `pallet-bridge-messages` and `dispatch` modules.
+/// Chain that has `pallet-bridge-messages` module.
+pub trait ChainWithMessages {
```
bkontur, line 119: @svyatonik 
I started to work on the same thing yesterday :D - to simplify this ChainWithMessages, because the name is not very clear and also we have two different traits with the same name `trait ChainWithMessages` and it causes some unnecessery small duplications

https://github.com/paritytech/parity-bridges-common/compare/bko-nits?expand=1

but I am reverting it as you work on the same :)

svyatonik: Sorry for that :)

bin/millau/runtime/src/rialto_messages.rs
```diff
@@ -177,11 +175,7 @@ impl messages::ThisChainWithMessages for Millau {
 pub struct Rialto;
 
 impl messages::ChainWithMessages for Rialto {
-	type Hash = bp_rialto::Hash;
-	type AccountId = bp_rialto::AccountId;
-	type Signer = bp_rialto::AccountSigner;
-	type Signature = bp_rialto::Signature;
-	type Balance = bp_rialto::Balance;
+	type Chain = bp_rialto::Rialto;
```
bkontur, line 178: @svyatonik 
now, we could also remove all specific implementations for `messages::BridgedChainWithMessages::fn maximal_extrinsic_size()`

and we can have just one default generic implementation in `messages::BridgedChainWithMessages` which can use `Chain::max_extrinsic_size()` e.g.:

```
pub trait BridgedChainWithMessages: ChainWithMessages {

	fn maximal_extrinsic_size() -> u32 {
		Self::Chain::max_extrinsic_size()
	}
```

bkontur: @svyatonik 
I see `BridgedChainWithMessages::maximal_extrinsic_size` is used in one place:
```
    pub fn maximal_message_size<B: MessageBridge>() -> u32 {
		super::target::maximal_incoming_message_size(BridgedChain::<B>::maximal_extrinsic_size())
	}
```

is there any scenario, when somebody wants to override in `BridgedChainWithMessages::maximal_extrinsic_size()` what is configured on `Chain::max_extrinsic_size()` ??
if so, maybe we dont even need `BridgedChainWithMessages::maximal_extrinsic_size()`

svyatonik: No, I don't think so - `Chain::max_extrinsic_size` should work here. I'll add a fix

serban300 (approved):

bin/runtime-common/src/messages.rs
```diff
@@ -186,16 +170,16 @@ pub trait BridgedChainWithMessages: ChainWithMessages {
 pub type ThisChain<B> = <B as MessageBridge>::ThisChain;
 /// Bridged chain in context of message bridge.
 pub type BridgedChain<B> = <B as MessageBridge>::BridgedChain;
+/// Underlying chain type.
+pub type UnderlyingChainOf<C> = <C as ChainWithMessages>::Chain;
```
bkontur, line 174: @svyatonik 
just very nit: what about renaming `ChainWithMessages` -> `UnderlyingChain`,
`ChainWithMessages` here is not very descriptive and also we have two `trait ChainWithMessages`

PR #1623, opened 2022-11-07
Bump xcm v3

bkontur:


bin/rialto-parachain/node/Cargo.toml
```diff
@@ -66,15 +66,15 @@ sp-timestamp = { git = "https://github.com/paritytech/substrate", branch = "mast
...
-cumulus-client-consensus-aura = { git = "https://github.com/paritytech/cumulus", branch = "gav-xcm-v3" }
-cumulus-client-consensus-common = { git = "https://github.com/paritytech/cumulus", branch = "gav-xcm-v3" }
-cumulus-client-cli = { git = "https://github.com/paritytech/cumulus", branch = "gav-xcm-v3" }
-cumulus-client-network = { git = "https://github.com/paritytech/cumulus", branch = "gav-xcm-v3" }
-cumulus-client-service = { git = "https://github.com/paritytech/cumulus", branch = "gav-xcm-v3" }
-cumulus-primitives-core = { git = "https://github.com/paritytech/cumulus", branch = "gav-xcm-v3" }
-cumulus-primitives-parachain-inherent = { git = "https://github.com/paritytech/cumulus", branch = "gav-xcm-v3" }
-cumulus-relay-chain-interface = { git = "https://github.com/paritytech/cumulus", branch = "gav-xcm-v3" }
-cumulus-relay-chain-inprocess-interface = { git = "https://github.com/paritytech/cumulus", branch = "gav-xcm-v3" }
+cumulus-client-consensus-aura = { git = "https://github.com/paritytech/cumulus", branch = "master" }
```
acatangiu, line 69: So there isn't any xcmv3 specific stuff in `cumulus`?

Cargo.toml
```diff
@@ -17,226 +17,244 @@ members = [
...
+cumulus-client-service = { git = "https://github.com/paritytech//cumulus", branch = "gav-xcm-v3" }
+cumulus-primitives-core = { git = "https://github.com/paritytech//cumulus", branch = "gav-xcm-v3" }
+cumulus-primitives-parachain-inherent = { git = "https://github.com/paritytech//cumulus", branch = "gav-xcm-v3" }
+cumulus-relay-chain-interface = { git = "https://github.com/paritytech//cumulus", branch = "gav-xcm-v3" }
+cumulus-relay-chain-inprocess-interface = { git = "https://github.com/paritytech//cumulus", branch = "gav-xcm-v3" }
+cumulus-pallet-aura-ext = { git = "https://github.com/paritytech//cumulus", branch = "gav-xcm-v3" }
+cumulus-pallet-parachain-system = { git = "https://github.com/paritytech//cumulus", branch = "gav-xcm-v3" }
+cumulus-pallet-dmp-queue = { git = "https://github.com/paritytech//cumulus", branch = "gav-xcm-v3" }
+cumulus-pallet-xcm = { git = "https://github.com/paritytech//cumulus", branch = "gav-xcm-v3" }
+cumulus-pallet-xcmp-queue = { git = "https://github.com/paritytech//cumulus", branch = "gav-xcm-v3" }
```
acatangiu, line 258: `rialto-parachain` uses `cumulus/master` and rest use `cumulus/gav-xcm-v3`, why is that?

bin/rialto-parachain/node/Cargo.toml
bkontur: we need to have the same version of xcm/substrate/polkadot dependencies everywhere,
because bridges repo are based on origin/gav-xcm-v3 + substrate-at-some-commit

Cargo.toml
bkontur: everywhere should be just master, but this is the way that to tell cargo to use `branch = "gav-xcm-v3" instead of `master`,

at the end, everywhere should be master, and this patch part will be removed

acatangiu (approved):

bin/rialto-parachain/node/Cargo.toml
svyatonik: I'm also not sure why you've changed from `gav-xcm-v3` to `master` here :) But if it works for you and for Rialto/Millau, then it is fine :)

svyatonik (approved):

bin/rialto-parachain/node/Cargo.toml
bkontur: @svyatonik 
well, I just wanted to align it with **polkadot's** `locked-for-gav-xcm-v3-and-bridges`  and **substrate's** `sv-locked-for-gav-xcm-v3-and-bridges`, everywhere we use `branch=master` and patched branch version is just in main Cargo.toml,
so I just wanted to do the same for Cumulus repo deps

svyatonik: ahhh, ok - I see :) thank you for the explanation! :)

PR #1624, opened 2022-11-08
Change back branches

bkontur:
we use our branches now for patching:
**polkadot**: `locked-for-gav-xcm-v3-and-bridges`
**substrate:** `sv-locked-for-gav-xcm-v3-and-bridges`
**cumulus:** `cumulus-locked-for-gav-xcm-v3-and-bridges`

so if anybody, will move `gav-xcm-v3` (polkadot or cumulus), we should be still able to build and have more "deterministic" build


acatangiu (approved):

acatangiu:
I swear we'll soon need a spreasheet with all the branches we're juggling 😆 

Any news on XCMv3 merge?

svyatonik (approved):

PR #1625, opened 2022-11-08
Statemine primitives

svyatonik:
related to https://github.com/paritytech/parity-bridges-common/issues/1318#issuecomment-1307221390

### DIFF
--- a/relays/bin-substrate/src/chains/statemine.rs
+++ b/relays/bin-substrate/src/chains/statemine.rs
@@ -14,17 +14,21 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
 
-#![cfg_attr(not(feature = "std"), no_std)]
-// RuntimeApi generated functions
-#![allow(clippy::too_many_arguments)]
+//! Statemine chain specification for CLI.
 
-pub use bp_polkadot_core::*;
-use bp_runtime::decl_bridge_finality_runtime_apis;
+use crate::cli::CliChain;
+use relay_statemine_client::Statemine;
+use sp_version::RuntimeVersion;
 
-/// Polkadot Chain
-pub type Polkadot = PolkadotLike;
+impl CliChain for Statemine {
+	const RUNTIME_VERSION: Option<RuntimeVersion> = Some(bp_statemine::VERSION);
 
-/// Name of the With-Polkadot GRANDPA pallet instance that is deployed at bridged chains.
-pub const WITH_POLKADOT_GRANDPA_PALLET_NAME: &str = "BridgePolkadotGrandpa";
+	type KeyPair = sp_core::sr25519::Pair;
 
-decl_bridge_finality_runtime_apis!(polkadot);
+	fn ss58_format() -> u16 {
+		sp_core::crypto::Ss58AddressFormat::from(
+			sp_core::crypto::Ss58AddressFormatRegistry::SubstrateAccount,
+		)
+		.into() // TODO
+	}
+}
--- a/relays/bin-substrate/src/chains/statemine.rs
+++ b/relays/bin-substrate/src/chains/statemine.rs
@@ -27,8 +27,8 @@ impl CliChain for Statemine {
 
 	fn ss58_format() -> u16 {
 		sp_core::crypto::Ss58AddressFormat::from(
-			sp_core::crypto::Ss58AddressFormatRegistry::SubstrateAccount,
+			sp_core::crypto::Ss58AddressFormatRegistry::KusamaAccount,
 		)
-		.into() // TODO
+		.into()
 	}
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -783,7 +783,6 @@ version = "0.1.0"
 dependencies = [
  "bp-polkadot-core",
  "bp-runtime",
- "smallvec",
  "sp-api",
 ]
 
@@ -875,6 +874,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "bp-statemine"
+version = "0.1.0"
+dependencies = [
+ "bp-polkadot-core",
+ "sp-version",
+]
+
+[[package]]
 name = "bp-test-utils"
 version = "0.1.0"
 dependencies = [
@@ -894,10 +901,7 @@ version = "0.1.0"
 dependencies = [
  "bp-polkadot-core",
  "bp-runtime",
- "parity-scale-codec",
- "scale-info",
  "sp-api",
- "sp-runtime",
 ]
 
 [[package]]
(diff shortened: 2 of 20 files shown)


relays/bin-substrate/src/chains/statemine.rs
```diff
@@ -0,0 +1,34 @@
...
+impl CliChain for Statemine {
+	const RUNTIME_VERSION: Option<RuntimeVersion> = Some(bp_statemine::VERSION);
+
+	type KeyPair = sp_core::sr25519::Pair;
+
+	fn ss58_format() -> u16 {
+		sp_core::crypto::Ss58AddressFormat::from(
+			sp_core::crypto::Ss58AddressFormatRegistry::SubstrateAccount,
+		)
+		.into() // TODO
```
acatangiu, line 32: what is there "TO DO"? :)

relays/bin-substrate/src/chains/statemine.rs
```diff
@@ -0,0 +1,34 @@
...
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Statemine chain specification for CLI.
+
+use crate::cli::CliChain;
+use relay_statemine_client::Statemine;
+use sp_version::RuntimeVersion;
+
+impl CliChain for Statemine {
+	const RUNTIME_VERSION: Option<RuntimeVersion> = Some(bp_statemine::VERSION);
```
bkontur, line 24: @svyatonik 
I guess, we dont really need `bp_statemine::VERSION`, right?
for BridgeHubs we set this to `None` as `const RUNTIME_VERSION: Option<RuntimeVersion> = None;`

bkontur (approved):

relays/bin-substrate/src/chains/statemine.rs
svyatonik: No, we need it because we're going to submit transactions to Statemine. Each transaction has so called "signed extension" and to craft it you need to know actual chain version. We have three ways to sepcify this version: one is to use version that is bundled with the `substrate-relay` binary (you're looking at it now). Other option is to specify it using CLI. And the last one is to read it from the node.

The latter is ok for test chains, but for real chains we don't want to use it because call encoding can change and since we're using mocked runtime our transaction may just be rejected (and we'll probably waste our funds on that). The first option is for the production - we're going to check that our mocked runtime is not broken on every Statemine upgrade and redeploy our relay with updated version (at least that was the initial plan - feel free to suggest improvements). The 2nd option (specify version using CLI) is a temporary measure - e.g. to run relay if we sure that our call encoding is matching upgraded runtime version, but we still have not built updated relay.

bkontur: so lets say for BridgeHubRococo/Wococo we are ok with None (read from node), but for Kusama/Polkadot we will need to also setup VERSION: RuntimeVersion?

bkontur: I am just thinking loud and what to understand in deep:
- lets say we are running real Kusama/Polkadot bridges and there is Statemint/e in some versions (STv123 and SEv456),
- afaik when `transaction_version` is changed, then we need to probably check/update encoding for our mocked runtime calls
- Statemint can be upgraded just throught proposal/governance (same for bridge-hubs) and it takes "some time" (iirc cca 2 weeks) and after this period runtime upgrade is approved and new runtime is changed in running collators (lets call it TIME_OF_REAL_UPGRADE_IN_NETWORK)

---

as you said:
> we're going to check that our mocked runtime is not broken on every **Statemine upgrade** and **redeploy**

1. we can check **Statemine upgrade** - there is a Cumulus release process of runtimes, and we can check manually (e.g. if transaction_version is changed or not...), there is no problem, I guess
2. after release process, there is a proposal/governance phase for upgrading runtime
3. after governance is done, at some time (TIME_OF_REAL_UPGRADE_IN_NETWORK) **Statemine redeploy** occurrs (afaik it is triggered by user action, any user can do enact of approved runtime upgrade)
4. substrate-relay have two possibilities:
4.1. **no need to update** mocked runtime (e.g. transaction_version was not change)
4.2. **need to update** mocked runtime, at some point we need to redeploy new version (TIME_OF_REAL_UPGRADE_OF_SUBSTRATE_RELAYER) (lets say, redeploy new version, means just to push new tag)

this all also will apply for BridgeHub upgrades.

---
**Questions/open points:**
1. does initial plan count with some monitoring or automatization?
- do we need anything, e.g. for checking that Statemin/BridgeHub runtimes are redeployed?

2. **what about potential transactions, which could be rejected between TIME_OF_REAL_UPGRADE_IN_NETWORK and TIME_OF_REAL_UPGRADE_OF_SUBSTRATE_RELAYER?**
- even if we push new tag immediatelly after TIME_OF_REAL_UPGRADE_IN_NETWORK, there is still some gap till new relayer is redeployed and again start to submit extrinsics
- iiuc messages from **outbouned_queue** are removed on `receive_messages_delivery_proof`
  - could there be any problems, that messages stay too long (few hours/days) in **outbouned_queue**?
    - OutboundLane is fulfilled? I guess, now it is unbouned? also for InboundQueue?
  - what if we restart relayer, which continues to sync headers (there is a `HeadersToKeep`), could we miss/fail to execute some outbound messages on the other side because of too low `HeadersToKeep`?

are there any known issues with redeploy?

svyatonik:
> so lets say for BridgeHubRococo/Wococo we are ok with None (read from node), but for Kusama/Polkadot we will need to also setup VERSION: RuntimeVersion?

We need version for chains that we are "changing" (submitting transactions to). In our current bridge we're not going to submit transactions to Polkadot and Kusama. The only chains we are changing are bridge hubs. For Statemine we only need this transaction to update conversion rate

> does initial plan count with some monitoring or automatization?

The only thing we have now is version guards - they just kill the relay when actual runtime verison is different from what we are thinking of. The rest (rebuilding + republishing + restarting relay) is a manual process.

> what about potential transactions, which could be rejected between TIME_OF_REAL_UPGRADE_IN_NETWORK and TIME_OF_REAL_UPGRADE_OF_SUBSTRATE_RELAYER?

Relay shall be killed by version guard => no transactions.

> is primitives a good place for pub const VERSION: RuntimeVersion?

Actually the difference between `primitives/*` and `relays/*` is that `primitives` contains types/constants/methods/structs that we need to access from the runtime (i.e. they may be compiled with `no_std`). Relays is for our offchain `std` actors. Right now version is not used from the runtimes, so I suppose it is ok to move it to the `relays/*` dir.

bkontur:
@svyatonik sorry for confusion, with
`but for Kusama/Polkadot we will need to also setup VERSION: RuntimeVersion?`
I meant:
`but for BridgeHubKusama/BridgeHubPolkadot we will need to also setup VERSION: RuntimeVersion?` like for Statemint/e, right?

svyatonik:
Right!

Issue #1626, opened 2022-11-08
Fix Statemine runtime after with-Statemint bridge pallet is ready


svyatonik:
See TODOs in this file: https://github.com/paritytech/parity-bridges-common/pull/1625/files#diff-2e5d0fe8d548b2e7230cf55913c88f1a1c591e16c57ba3be6fd65221cdcdcffd

bkontur:
@svyatonik 
I guess, we dont need this issue, because we dont need conversion-rate, right?

PR #1635, opened 2022-11-11
Separate conversion rate updater for Statemine -> Statemint

### LINKED ISSUE #1630: At-Statemine conversion rate updater
We'll be computing and withdwaring message fee for Statemine -> Statemint (and their Rococo/Wococo analogues) on Statemine (https://github.com/paritytech/parity-bridges-common/issues/1318#issuecomment-1307221390). For that we'll need to know the Statemint -> Statemine tokens (i.e.DOT -> KSM) conversion rate on Statemine. So we need a separate relay mode for that

svyatonik:
closes #1630 

So far it is just a dummy code that may not actually work - we'll test it later after #1626 is fixed.

Small note on how it works - it allows on-chain rate to differ from the actual rate by at most 5% (`(abs(real - stored) / stored)`) in any direction. So we may store rate that is lower than actual => have some loses. This should have been covered by the relayer interest (10% of message fee). In our case imo it'll be covered either by `BridgeFee` (constant iiuc) or we may consider that it is included in the relayer reward on the source chain.

### DIFF
--- a/relays/bin-substrate/src/cli/conversion_rate_update.rs
+++ b/relays/bin-substrate/src/cli/conversion_rate_update.rs
@@ -1,0 +1,224 @@
+// Copyright 2019-2021 Parity Technologies (UK) Ltd.
+// This file is part of Parity Bridges Common.
+
+// Parity Bridges Common is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Parity Bridges Common is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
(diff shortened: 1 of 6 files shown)


relays/bin-substrate/src/cli/conversion_rate_update.rs
```diff
@@ -0,0 +1,224 @@
...
+}
+
+/// Statemine -> Statemint conversion rate update loop.
+struct StatemineToStatemintLoop;
+
+impl ConversionRateUpdateLoop for StatemineToStatemintLoop {
+	type SourceChain = Statemine;
+	// this is a bit hacky - we are not going (?) to connect to the Statemint now, so we don't need
+	// those primitives and client crates. We're only using the `TargetChain` to get its `TOKEN_ID`.
+	// Statemint is the common god parachain of the Polkadot, so let's use Polkadot here
```
acatangiu, line 121:
```suggestion
	// Statemint is the common good parachain of the Polkadot, so let's use Polkadot here
```

acatangiu (approved):

relays/bin-substrate/src/cli/conversion_rate_update.rs
bkontur: I guess, we should use better "common good" -> "system"

svyatonik:
Let's made it onice until we'll have an approval

svyatonik:
there'll be just some const fee - no formula, so we don't need this functionality

Issue #1629, opened 2022-11-11
Complex bridge hub relay


svyatonik:
#1565 is introducing standalone messages, GRANDPA and parachains relay. But what we'll need in our actual deployment is the complex relay. It needs to be implemented once #1565 is merged

UPD: also version guard is currently missing from the finality relays - it needs to be added

bkontur:
complex relay means something like to join commands (loops): RelayHeaders/RelayParachains/RelayMessages to the one command (loop) ComplexRelay? 

so that we need to run just one substrate-relayer command?

svyatonik:
Exactly. Apart from that it is decreasing costs of running relayers - i.e. standalone GRANDPA relayer will submit every header it is seeing (unless `--only-mandatory-headers` flag is passed). Standalone parachains relay will submit every updated head it sees. In complex relay messages relay is orchestrating everything - it'll say to GRANDPA and parachains relay that it needs specific relay header/parachain head at the bridged chain => most of time it is idle and transactions are only submitted when we need them

bkontur:
@svyatonik 
perfect, that would be very cool, to have just one configurable `substrate-relay` at the end,
it will very simplify deployment and usage and other stuff

svyatonik:
Yeah, I agree. But we also have this conversion-rate-updater at Statemine, which (I believe) will be running as a standalone process (it doesn't fit the current complex relay architecture at all :) ). Following the initial deployment, we may also want to deploy transaction resubmitters - also a separate processes, but they may be embedded into complex relay (possibly good idea to explore that too).

bkontur:
@svyatonik 
Slava, do we need to do here anything else?
complex relay for Rococo/Wococo is done and works: https://github.com/paritytech/parity-bridges-common/issues/1662

svyatonik:
> UPD: also version guard is currently missing from the finality relays - it needs to be added

I think this part is missing

PR #1565, opened 2022-09-02
Rococo/Wococo bridge hub relay integration

bkontur:

PR #1637, opened 2022-11-14
Complex parachain <> parachain relay

svyatonik:
Related to #1629 . To close it we'll need to define at least Rococo <> Wococo bridge hubs relay (requires #1565 )


serban300 (approved):
Looks good !

Issue #1638, opened 2022-11-14
Remove messaging code that doesn't fit our current deployment scenario


svyatonik:
There's a lot of code that we are still supporting, but which is not needed (and sometimes even causes some slowdowns/troubles/mistakes/misunderstandings) in our planned deployment. Everything below needs a final confirmation, which is expected yet => onice.

1. ~We are not going to accept XCM messages from end-users, hence we don't need `send_message` and `increase_message_fee` calls of the messaging pallet~ Done in #1642

2. ~Since we are not going to reward relayers initially, we don't need relayer account ID in `receive_messages_proof` and `InboundLaneData::relayers`. But we may need it in the future, so probably worth keeping it for now?~ We'll be rewarding relayers - see #1648

3. ~We are no longer interested in computing exact message fee on bridge hubs, so we don't need conversion rate, fee multipliers messaging parameters in the runtime. Also we don't need this whole estimate-fee functionality at all (requires changes in both runtime and relay).~ Done in #1642

4. ~An effect of all previous points - we don't need conversion rate updater in complex relay. We don't need a rational relay mode anymore.~ Done in #1642

Imo if we won't use this code soon (and apparently we won't), then it is better to remove it asap.

UPD: most of this is done #1642, but I'll leave it open for some time - pretty sure there are some leftovers that need to be wiped.

svyatonik:
We no loner need `DeliveredMessages::dispatch_results`

svyatonik:
`ESTIMATE_MESSAGE_FEE_METHOD`

svyatonik:
`BlockWeights` from primitives - do we need it?

svyatonik:
`WeigthToFee`? https://github.com/paritytech/parity-bridges-common/pull/1712/files#diff-8518c3485aec64d8472c504c979776c33c3cc7c5603f401db705bd5230fdf54cR86

serban300:
@svyatonik I can check if `WeigthToFee` is used and remove it if needed. If that's ok with you.

svyatonik:
> @svyatonik I can check if `WeigthToFee` is used and remove it if needed. If that's ok with you.

Sure, thank you! :)

serban300:
Yes, `WeightToFee` doesn't seem to be needed. It's used in the definitions of the Millau and Rialto chains, but it's not used by the bridges. I posted #1713 for removing it.

svyatonik:
Messages pallet owner?

[...]

Issue #1576, opened 2022-09-20
Switch to HaulBlobExporter and BridgeBlobDispatcher from polkadot repo


svyatonik:
The fact that these types can't be used from outside is a bug. Right now we have a copy in the `bridge-runtime-common` crate. This copy shall be removed and the original XCM v3 PR shall be fixed so that the "all pub types in universal_exports.rs should be reexported from lib.rs".

bkontur:
@svyatonik 
Slava, I would love to take this as a good start to dive into xcm bridge messaging, ok?

svyatonik:
@bkontur 
Sure, imo it involves three steps:
1) opening separate PR `some-branch` -> `gav-xcm-v3` with "all pub types in universal_exports.rs should be reexported from lib.rs". We may also ask if an isolated commit directly to the `gav-xcm-v3` would work here;
2) updating our refs to the latest `gav-xcm-v3`
3) actually using those types. TBH I'm not sure if they are currently used in Rialto/RialtoParachain<>Milau - I've had too many troubles with using them back when we've been switching to the XCM messaging. This may take a while and require many changes. In any case - if you'll have troubles here, let's chat/discuss later

EmmanuellNorbertTulbure:
@bkontur status on this one?

EmmanuellNorbertTulbure:
When closing, the questions from the PR should be transferred into issue(s).

bkontur:
@EmmanuellNorbertTulbure PR is closed and I guess, the only thing, which is not in any issue, is just this test: https://github.com/paritytech/parity-bridges-common/pull/1813#discussion_r1133927238

PR #1642, opened 2022-11-15
Remove message fee + message send calls

svyatonik:
resolves some parts of #1638 
very wip => draft

Splitting it into separate issues turned out to be hard - removed parts are tightly tied together, so single PR. I'll explicitly add comments where I'd like to have an approval.

UPD: this PR is not touching everything that is related to relayer rewards

UPD2: what this PR removes:
- `send_message` and `increase_message_fee` calls of the messages pallet + their corresponding benchmarks;
- everything that is related to message delivery and dispatch fee - since messages fee is now supposed to be taken at bridge "appication level";
- rational relayer mode;
- conversion rates - storage entries, computations, metrics + some linked parameters that have been involved in fee computation;
- `estimate-fee` subcommand of relayer


bin/runtime-common/src/messages.rs
```diff
@@ -499,39 +328,22 @@ pub mod source {
...
-						T::MessageBridge::BRIDGED_CHAIN_ID,
-						e,
-					);
-					*dest = Some(d);
-					return Err(SendError::Transport(e))
-				},
-			};
-			let fee_assets = MultiAssets::from((Here, fee));
+			// let's jsut take fixed (out of thin air) fee per message in our test bridges
+			// (this code won't be used in production anyway)
```
svyatonik, line 332: @bkontur Am I right here? Do you have some branch where you're using those blob dispatcher and exporter? So in the end we won't need this custom `SendXcm` implementation, right?

svyatonik: (I mean #1576)

svyatonik: That's super cool, thank you!

modules/relayers/src/payment_adapter.rs
```diff
@@ -20,65 +20,61 @@
...
-			lane_id,
 			messages_relayers,
 			received_range,
 		);
 
 		register_relayers_rewards::<T>(
 			confirmation_relayer,
 			relayers_rewards,
-			GetConfirmationFee::get(),
+			// TODO (https://github.com/paritytech/parity-bridges-common/issues/1318): this shall be fixed
```
svyatonik, line 56: Shall be fixed either in this PR, or later

svyatonik: Left it for later. New relayer rewards scheme will be introduced later && it deserves a separate review

svyatonik:
Need to test it locally => back to draft

relays/bin-substrate/src/cli/encode_message.rs
```diff
@@ -63,15 +55,36 @@ pub(crate) fn encode_message<Source: Chain, Target: Chain>(
...
+				ExplicitOrMaximal::Explicit(size) => size,
+				ExplicitOrMaximal::Maximal => compute_maximal_message_size(
 					Source::max_extrinsic_size(),
 					Target::max_extrinsic_size(),
+				),
+			};
+
+			// there's no way to craft XCM of the given size - we'll be using `ExpectPallet`
+			// instruction, which has byte vector inside
+			let mut current_vec_size = expected_xcm_size;
```
svyatonik, line 69: That's the only new code in this PR - since we are now only sending messages using XCM pallet, then the message shall be valid xcm and `vec[42; 1024]` won't work here.

acatangiu (approved):

bkontur (commented):
@svyatonik 
wow, lots of code removed,
very nice, it will also remove lots of my "0/None TODOs" from bridge-hubs :)

Issue #1647, opened 2022-11-17
Move delivered messages pruning and callbacks outside of send_message method


svyatonik:
Follow up for #1642

Since we are no longer going to send messages directly using messages pallet, we need to move all heavy unstable operations outside of `send_message` method. There are two of them - `OnMessageAccepted` callback and messages pruning. Callbacks are no longer used so they both can be removed (`OnMessageAccepted` and `OnDeliveryConfirmed`) and pruning must be moved to the `on_idle`.

Then the `send_message` will have the constant weight, that is close to 3 db writes (outbound lane state update, message insertion, event).

PR #1650, opened 2022-11-18
Prune messages from on-idle callback

svyatonik:
part of #1647

Pruning from on-idle isn't that easy, since we support multiple lanes and (I believe) we may actually use other lanes in the future - e.g. to prioritize some bridge over others or to increase bridge bandwidth. But then to prune messages we need to iterate all lanes and check if those have unpruned messages. Iterating = db read and if we'll be reading lane state it at every on-idle call, we may run out of remaining weight without doing any actual work. So instead I've added `ActiveOutboundLanes` to the messages pallet configuration and we're pruning only one lane per block (`block_number % lanes_len`). This still may cause troubles in the future, but for our start it is fine

### DIFF
--- a/modules/messages/src/lib.rs
+++ b/modules/messages/src/lib.rs
@@ -66,6 +66,7 @@ use bp_runtime::{BasicOperatingMode, ChainId, OwnedBridgeModule, Size};
 use codec::{Decode, Encode, MaxEncodedLen};
 use frame_support::{dispatch::PostDispatchInfo, ensure, fail, traits::Get};
 use num_traits::Zero;
+use sp_runtime::traits::UniqueSaturatedFrom;
 use sp_std::{
 	cell::RefCell, collections::vec_deque::VecDeque, marker::PhantomData, ops::RangeInclusive,
 	prelude::*,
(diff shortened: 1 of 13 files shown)


modules/messages/src/lib.rs
```diff
@@ -191,6 +190,42 @@ pub mod pallet {
...
+			// first db read - outbound lane state
+			let mut active_lane = outbound_lane::<T, I>(active_lane_id);
+			let original_remaining_weight = remaining_weight;
+			remaining_weight -= db_weight.reads(1);
+			// and here we'll have writes. It is safe to do regular subtraction, since
+			// `prune_messages` respects the `remaining_weight`
+			remaining_weight -= active_lane.prune_messages(db_weight, remaining_weight);
+
+			// safe to use regular subtraction, since we respect the `remaining_weight`
+			original_remaining_weight - remaining_weight
```
acatangiu, line 225: nit:
could simplify this to be smth like

```suggestion
			// first db read - outbound lane state
			let mut active_lane = outbound_lane::<T, I>(active_lane_id);
			let mut used_weight = db_weight.reads(1);
			// and here we'll have writes
			used_weight -= active_lane.prune_messages(db_weight, remaining_weight);

			// we already checked we have enough `remaining_weight` to cover this `used_weight`
			used_weight
```

acatangiu (approved):

modules/messages/src/lib.rs
svyatonik: Thanks! I've also have been brave enough on Friday to leave this function untested && I admit I was wrong. So some of typos/issues are also fixed here. Additionally, I've moved lane check to the `send_message` function - now it check that the message is sent using one of lanes from `ActiveOutboundLanes` set.

Issue #1648, opened 2022-11-17
Pay partial relayer reward at the target chain


svyatonik:
https://github.com/paritytech/parity-bridges-common/issues/1318#issuecomment-1318494287

Steps:
- [x] introduce callback that must either call `balances::transfer(who: sovereign-account-of-bridged-chain, to: relayer)` or just return `Pays::No` in appropriate cases (https://github.com/paritytech/parity-bridges-common/pull/1657);
- [ ] add relayer mode that will use `utility.batchAll(relay-chain-finality, parachain-finality, message-delivery)` to submit message delivery transactions. Or maybe a separate call/message proof - yet to be decided;
- [x]  find some way to include compensation for finality transactions in the at-target-chain-callback (https://github.com/paritytech/parity-bridges-common/pull/1657);
- [x] fix our signed extensions - they'll be affected by this change (https://github.com/paritytech/parity-bridges-common/pull/1657).

PR #1652, opened 2022-11-21
New relayer rewards scheme integration

svyatonik:
part of #1648

This PR will be the final of PR series that are changing our current reward scheme. Since new code will be introduced (=> needs proper review) plan for changes is to:
- [x] PR1-with-new-code: introduce trait `payment-for-delivery-transaction-at-target-chain` and `pallet-bridge-relayers` adapter for it (#1653);
- [x] PR2-with-new-code: introduce trait `payment-for-confirmation-transaction-at-source-chain` and `pallet-bridge-relayers` adapter for it (#1657);
- [x] this PR without new code, which will just reconfigure messages pallet and all runtimes to work with new adapters. The end goal is to have on Rialto<>Millau something that we will have for other bridges.


svyatonik:
Should be ready for review now, but I want to test it locally

acatangiu (approved):
Looking good!

PR #1653, opened 2022-11-21
DeliveryConfirmationPayments trait for paying relayer rewards at the source chain

svyatonik:
extracted part from #1652 

This PR introduces `DeliveryConfirmationPayments` trait, which is executed at the source chain during delivery confirmation transaction. The trait itself is a (partial) copy of the existing `MessageDeliveryAndDispatchPayment`, which will be removed in favor of that new trait later in #1652 .

Please pay attention to the `DeliveryConfirmationPayments` implementation for the `DeliveryConfirmationPaymentsAdapter` - I assume it'll be reused on Rococo<>Wococo and later Kusama<>Polkadot bridge. The idea is simple - since we have dropped the message delivery + dispatch fee computation, we can't refund relayer its actual spendings => some constant is used. E.g. every delivered message will have some fixed reward + every message delivery confirmation has some reward. I have no idea now how we'll compute the consts and how they'll be updated. For our testnets I've just picked a "random" `100_000` value.

Please also keep in mind that the `pallet-bridge-relayers` pallet assuming manual reward claiming. So we are not actually paying reward during the confirmation transaction - instead we're registering the payment and the actual payment will happen when the relayer will call the `claim_rewards` transaction (also see #1654 ).

@acatangiu @bkontur Please review - imo we all must explicitly conclude on this (and follow-up PR about reward-at-target-chain), since this is the last major design change before deploying the bridge. Thank you!


modules/relayers/src/payment_adapter.rs
```diff
@@ -19,22 +19,25 @@
...
 /// for the messages pallet.
-pub struct MessageDeliveryAndDispatchPaymentAdapter<T, MessagesInstance>(
-	PhantomData<(T, MessagesInstance)>,
+pub struct DeliveryConfirmationPaymentsAdapter<T, DeliveryReward, ConfirmationReward>(
+	PhantomData<(T, DeliveryReward, ConfirmationReward)>,
 );
 
-impl<T, MessagesInstance> MessageDeliveryAndDispatchPayment<T::RuntimeOrigin, T::AccountId>
-	for MessageDeliveryAndDispatchPaymentAdapter<T, MessagesInstance>
+// TODO (https://github.com/paritytech/parity-bridges-common/issues/1318): this impl must be removed
```
acatangiu, line 35: #1318 is closed.. should we open another dedicated one?

Also sorry for dumb q, this comment is tracking the removal of `MessageDeliveryAndDispatchPayment` trait not just this impl, right? Since we don't have message dispatch anymore. Or will we still use this and rename to `trait MessageDeliveryPayment`?

svyatonik: I'm gonna remove impl (`DeliveryConfirmationPaymentsAdapter`) and will keep the other impl (`DeliveryConfirmationPaymentsAdapter`). The old trait also needs to be removed

acatangiu: Sounds good! It's clear now, thanks!

PR #1655, opened 2022-11-21
Added event `MessagesReceived` for better visibility on receiving side

bkontur:
We generate (messages) events only on source chain, so this adds event on target chain, it could help better understand and see and anaylze what is going on, better visibility, e.g. at least, we could see easily in polkadot.js in what block messages were received.
If this will be too much overhead, later we could remove that or just add there only some counters.

plus trying to fix https://substrate.stackexchange.com/questions/5901/failed-to-run-parity-bridges-common-message-realy-demo/5906?noredirect=1#comment5781_5906

### DIFF
--- a/modules/messages/src/lib.rs
+++ b/modules/messages/src/lib.rs
@@ -46,7 +46,7 @@ pub use weights_ext::{
 };
 
 use crate::{
-	inbound_lane::{InboundLane, InboundLaneStorage, ReceivalResult},
+	inbound_lane::{InboundLane, InboundLaneStorage},
 	outbound_lane::{OutboundLane, OutboundLaneStorage, ReceivalConfirmationResult},
 };
 
@@ -91,7 +91,7 @@ pub const LOG_TARGET: &str = "runtime::bridge-messages";
 #[frame_support::pallet]
 pub mod pallet {
 	use super::*;
-	use bp_messages::{ReceivedMessageResult, ReceivedMessages};
+	use bp_messages::{ReceivalResult, ReceivedMessages};
 	use frame_support::pallet_prelude::*;
 	use frame_system::pallet_prelude::*;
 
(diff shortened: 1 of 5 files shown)


scripts/send-message-from-millau-rialto.sh
```diff
@@ -15,20 +15,14 @@ case "$1" in
 			--source-host localhost \
 			--source-port $MILLAU_PORT \
 			--source-signer //Alice \
-			--target-signer //Bob \
-			--lane 00000000 \
-			--origin Target \
-			remark \
+			raw 020419ac
```
bkontur, line 18: @svyatonik 
Slava, how was this value `020419ac` generated? 
I can see, that it is kind of "trap message for Kusama",

we also should fix this send-message script,
do we want to leave here remark/transfer? how to fix them?
or what do you suggest?


https://substrate.stackexchange.com/questions/5901/failed-to-run-parity-bridges-common-message-realy-demo/5906?noredirect=1#comment5781_5906

primitives/messages/src/lib.rs
```diff
@@ -218,6 +218,50 @@ pub struct UnrewardedRelayer<RelayerId> {
 	pub messages: DeliveredMessages,
 }
 
+/// Received messages with their dispatch result.
+#[derive(Clone, Default, Encode, Decode, RuntimeDebug, PartialEq, Eq, TypeInfo)]
+pub struct ReceivedMessages<Result> {
+	pub lane: LaneId,
```
svyatonik, line 224: Please leave some comments for public items

modules/messages/src/lib.rs
```diff
@@ -314,24 +316,41 @@ pub mod pallet {
...
+				let mut is_lane_processing_stopped_no_weight_left = false;
+
 				for mut message in lane_data.messages {
 					debug_assert_eq!(message.key.lane_id, lane_id);
+					total_messages += 1;
+
+					if is_lane_processing_stopped_no_weight_left {
+						lane_messages_received_status.push((
+							message.key.nonce,
+							NotDispatchedReason::NotEnoughWeightForLane.into(),
```
svyatonik, line 330: Probably we can just leave single `NotDispatchedReason::NotEnoughWeightForLane` instead of duplicating it for every nonce? I mean - all remaining messages will be rejected, so probably single `NotEnoughWeightForLane` would be enough? I admit that the current version is better for understanding though :)

modules/messages/src/lib.rs
```diff
@@ -349,18 +368,38 @@ pub mod pallet {
...
+								.push((message.key.nonce, ReceivedMessageResult::Dispatched));
 							(
 								dispatch_result.unspent_weight,
 								!dispatch_result.dispatch_fee_paid_during_dispatch,
 							)
 						},
-						ReceivalResult::InvalidNonce |
-						ReceivalResult::TooManyUnrewardedRelayers |
-						ReceivalResult::TooManyUnconfirmedMessages => (dispatch_weight, true),
+						ReceivalResult::InvalidNonce => {
```
svyatonik, line 378: Maybe we can have `impl From<ReceivalResult>` for `NotDispatchedReason` => less code

svyatonik (approved):
We've had dispatch level events before (in bridge dispatch pallet), but now they are gone. So yeah - imo it is a good idea, thank you!

scripts/send-message-from-millau-rialto.sh
svyatonik: IIRC it is just a `Trap(42)` - just to see that it is dispatched in logs (instead of some "Unable to decode XCM"). Send message script is in abandoned state currently (as well as its complicated version - https://github.com/svyatonik/parity-bridges-common.test) - imo we'll need fix it after stabilizing everything + updating project-level documentation.

modules/messages/src/lib.rs
bkontur: I was not 100% sure about this part, but I wanted to have all messages from request in the result :)
so I removed value `NotEnoughWeightForLane` and collect all skipped messages to the `skipped_for_not_enough_weight: Vec<MessageNonce>`  so there will be everything in the result without duplicating anything

svyatonik: Cool, thank you! :)

modules/messages/src/lib.rs
bkontur: thats what I had at first :D and made `pub struct ReceivedMessages<Result> {`  generic,
but then I added `NotEnoughWeightForLane`  so I better created new enum/struct,
but now when I removed `NotEnoughWeightForLane`  I can use `ReceivalResult` :D

ok, no problem, I will clear that :)

scripts/send-message-from-millau-rialto.sh
bkontur: @svyatonik Slava, so you dont remember how was that `Trap(42)` encoded to `020419ac`?

svyatonik: Not sure, but probably I've just printed it in [this test](https://github.com/paritytech/parity-bridges-common/blob/master/bin/millau/runtime/src/xcm_config.rs#L302).

PR #1657, opened 2022-11-22
Signed extension to refund relayer at the target chain

svyatonik:
extracted part from #1652
just an initial TODO-level prototype - without any actual code => draft

A brief intro: in our new concept the cost of at-target-chain relayer transactions will be refunded by sovereign account of the bridged chain at bridge hub (e.g. Statemine on Bridge-Hub-Polkadot). The main point is that it'll be paid in DOTs (for Statemine -> Statemint example), not in KSMs. There's other part of relayer "reward" that is paid in KSMs, but this PR is not about that.

---

I was planning to just add another trait that will be called during delivery transaction and that will compensate costs of finality and messages delivery. But it is hard to organize cross-call communication in the runtime - i.e. if relayer has submitted a batch of `(relay-chain-finality, parachain-finality, messages-delivery)` calls, then how would we say to the message delivery transaction (that will do refund) that we shall also compensate other two costs? The obvious (but very hacky) solution would be to save the cost in some ephemeral runtime storage value. But then - who'll remove this value from the storage? What if the removing code won't be called by some reason?

The other thing is that the cost of transaction isn't just a cost of weight. And during the call execution we only know the weight part (there's also ephemeral encoded exrtrinsic somewhere in the `frame-system` storage, but we don't know what's the extrinsic is, so we need to decode => even more problems). So to compensate full transaction cost, we need to resolve complex issues.

So I'm suggesting to do refund in a signed extension. It has access to decoded extrinsic, to its len, there's a callback that is called after dispatch (with post-dispatch info) and so on. There are some caveats, though. E.g. we'll need to check that transactions are dependent, that they have brought new headers/heads/messages, ... But it seems easier to handle it at that level.

Another important thing is that we won't use `Pays::No` and will have explicit payments (deferred, using `pallet-bridge-relayers`). That's because in `Pays::No` means that the "chain" is not receiving/losing any funds. And IIUC we wanted them to pay for each other - e.g. Statemine sovereign account will pay for transactions on BHP and vice versa. With `Pays::No` noone pays for anything.

@acatangiu @bkchr Are you ok with this plan. Please check code for detailed TODOs on what I'm going to implement here. Maybe you have some better ideas.

@serban300 Maybe you have other ideas too? I'll definitely need you later to review this code, since you've already reviewed our other extensions :)


svyatonik:
https://github.com/paritytech/parity-bridges-common/pull/1657/commits/dc35853887d0e8c9f6f73a38ce6407d8c9dd0a39 implements post_dispatch method of the signed extension. The main change with original vision is that we never refund parts of the batch transaction - i.e. if tx is `(relay-finality-update, parachain-finality-update, message-delivery)` and e.g. `parachain-finality-update` is not updating parachain state, then we won't refund the relayer, even though he may have actually have delivered some messages. That's because it is too hard to deal with partial weights of the tx even from signed extension,

bin/runtime-common/src/messages_batch_extension.rs
```diff
@@ -0,0 +1,297 @@
...
+
+		// check if messages have been delivered
+		let actual_messages_state = messages_state::<Runtime, MessagesInstance>();
+		let pre_dispatch_messages_state = call_type.pre_dispatch_messages_state();
+		if actual_messages_state == pre_dispatch_messages_state {
+			// we only refund relayer if all calls have updated chain state
+			return Ok(())
+		}
+
+		// regarding the tip - refund that happens here (at this side of the bridge) isn't the whole
```
svyatonik, line 265: That's an important thing - we don't refund the whole transaction cost here. Tip is not included in refund. Reasoning is in the code

bin/runtime-common/src/messages_batch_extension.rs
```diff
@@ -0,0 +1,138 @@
...
+	fn post_dispatch(
+		_pre: Option<Self::Pre>,
+		_info: &DispatchInfoOf<Self::Call>,
+		_post_info: &PostDispatchInfoOf<Self::Call>,
+		_len: usize,
+		_result: &DispatchResult,
+	) -> Result<(), TransactionValidityError> {
+		// TODO:
+		// if batch matches!(relay-chain-finality-call, parachain-finality-call,
+		// deliver-messages-call) then:
```
acatangiu: The plan laid out below looks good to me.

acatangiu (commented):
Sorry for the delay, this sounds good to me Slava. I am now also looking at the actual implementation, not just the TODOs description.

bin/runtime-common/src/messages_batch_extension.rs
```diff
@@ -0,0 +1,297 @@
...
+/// Data that is crafted in `pre_dispatch` method and used at `post_dispatch`.
+#[derive(PartialEq)]
+pub struct PreDispatchData<AccountId> {
+	/// Transaction submitter (relayer) account.
+	pub relayer: AccountId,
+	/// Type of the call.
+	pub call_type: CallType,
+}
+
+/// Type of the call that the extension recognizing.
```
acatangiu, line 100: nit

```suggestion
/// Type of the call that the extension recognizes.
```

bin/runtime-common/src/messages_batch_extension.rs
```diff
@@ -0,0 +1,385 @@
...
+	}
+
+	fn post_dispatch(
+		pre: Option<Self::Pre>,
+		info: &DispatchInfoOf<Self::Call>,
+		post_info: &PostDispatchInfoOf<Self::Call>,
+		len: usize,
+		result: &DispatchResult,
+	) -> Result<(), TransactionValidityError> {
+		// we never refund anything if that is not bridge transaction or if it is a bridge
```
acatangiu, line 249: nit

```suggestion
		// we never refund anything if it's not bridge transaction or if it is a bridge
```

bin/runtime-common/src/messages_batch_extension.rs
```diff
@@ -0,0 +1,385 @@
...
+		// check if relay chain state has been updated
+		if let CallType::AllFinalityAndDelivery(expected_relay_chain_state, _, _) = call_type {
+			let actual_relay_chain_state = relay_chain_state::<Runtime, GrandpaInstance>();
+			if actual_relay_chain_state != expected_relay_chain_state {
+				// we only refund relayer if all calls have updated chain state
+				return Ok(())
+			}
+
+			// there's a conflict between how bridge GRANDPA pallet works and the
+			// `AllFinalityAndDelivery` transaction. If relay cahin header is mandatory, the GRANDPA
```
acatangiu, line 270:
```suggestion
			// `AllFinalityAndDelivery` transaction. If relay chain header is mandatory, the GRANDPA
```

bin/runtime-common/src/messages_batch_extension.rs
```diff
@@ -69,70 +200,98 @@ where
...
+		len: usize,
+		result: &DispatchResult,
 	) -> Result<(), TransactionValidityError> {
-		// TODO:
-		// if batch matches!(relay-chain-finality-call, parachain-finality-call,
-		// deliver-messages-call) then:
-		//
-		// 0) ensure that the `result` is `Ok(_)`
-		//
-		// 1) ensure that all calls are dependent (relaychain -> parachain of this relaychain and
```
acatangiu: I don't see an explicit check that all inner batched calls are dependent (`relay_best`, `new_parahead` using `relay_best`, `msg_proof` using `new_parahead`).
Do you plan on checking this using yet unimplemented `fn relay_chain_state()` and `fn parachain_state()`?

Or do you want to drop this dependency check and allow batching unrelated/non-dependent relay/para/msgs? I mean, as long as they are valid, we don't have to force them dependent (but might end up more complex than actually expecting and checking they're dependent).

bin/runtime-common/src/messages_batch_extension.rs
```diff
@@ -0,0 +1,385 @@
...
+		// malicious relayer may use huge tips, effectively depleting account that pay rewards. The
+		// cost of this attack is nothing. Hence we use zero as tip here.
+		let tip = Zero::zero();
+
+		// compute the relayer reward
+		let reward = pallet_transaction_payment::Pallet::<Runtime>::compute_actual_fee(
+			len as _, info, post_info, tip,
+		);
+
+		// finally - regiater reward in relayers pallet
```
acatangiu, line 310:
```suggestion
		// finally - register reward in relayers pallet
```

bin/runtime-common/src/messages_batch_extension.rs
```diff
@@ -281,6 +314,61 @@ impl<Runtime, GrandpaInstance, ParachainsInstance, MessagesInstance> SignedExten
...
+
+/// Extracts expected parachain state from the call.
+fn extract_expected_parachain_state<Runtime, GrandpaInstance, ParachainsInstance>(call: &CallOf<Runtime>) -> Option<ExpectedParachainState> where
+	Runtime: GrandpaConfig<GrandpaInstance> + ParachainsConfig<ParachainsInstance, BridgesGrandpaPalletInstance = GrandpaInstance>,
+	GrandpaInstance: 'static,
+	ParachainsInstance: 'static,
+	<Runtime as GrandpaConfig<GrandpaInstance>>::BridgedChain: Chain<BlockNumber = RelayBlockNumber, Hash = RelayBlockHash, Hasher = RelayBlockHasher>,
+	CallOf<Runtime>: IsSubType<CallableCallFor<ParachainsPallet<Runtime, ParachainsInstance>, Runtime>>,
+{
+	// TODO: check para id (we'll need to bound instance of messages pallet with some ParaId)
```
acatangiu, line 341: so we'll have a messages pallet instance for each Para we bridge, right? does this mean we have different `pallet_utility::Call::<Runtime>::batch_all` calls one for each ParaId?

bin/runtime-common/src/messages_batch_extension.rs
svyatonik: Yep - it isn't implemented yet

bin/runtime-common/src/messages_batch_extension.rs
svyatonik: > so we'll have a messages pallet instance for each Para we bridge, right?

In theory we may serve different chains with the same instance of messages pallet, but it'll be too complex. So yes - 1 messages instance per 1 bridged chain. Please mind that for us this bridged chain is the other bridge hub, not the Statemin?. So we may serve XCM messages from any Polkadot/Kusama parachain using single messages pallet.

> does this mean we have different pallet_utility::Call::<Runtime>::batch_all calls one for each ParaId?

I'm not sure I got this part. In this extension we'll support 2 types of batch calls (relay+parachain+messages and parachain+messages). And we expect this parachain transaction to bring single parachain header - yes. If relayer wants to update the other parachain (which we don't plan to do at least initially) then yeah - he can craft the separate batch or submit a dedicated transaction.

acatangiu: Ok, thanks, it's clear now. I misinterpreted/mismapped-in-my-brain `parahead` to be head of msg source parachain (Statemine), and that's why I was confused and thought about how will we transfer messages coming from other parachains :see_no_evil: 

But yes, that was a brainfart. `parahead` here is other BridgeHub and messages pallet instance is same regardless of message source (at most we might have different lanes later on, but still single pallet instance).

bkchr:
> Another important thing is that we won't use `Pays::No` and will have explicit payments (deferred, using `pallet-bridge-relayers`). That's because in `Pays::No` means that the "chain" is not receiving/losing any funds. And IIUC we wanted them to pay for each other - e.g. Statemine sovereign account will pay for transactions on BHP and vice versa. With `Pays::No` noone pays for anything.

I did not have looked into the pr that deeply. However, why not just refund at the end of your transactions? Like we do here: https://github.com/paritytech/substrate/blob/a1dc9d8ff3a62580491d4c38c06251d5a3ca99f6/frame/sudo/src/lib.rs#L165-L166

svyatonik:
> > Another important thing is that we won't use `Pays::No` and will have explicit payments (deferred, using `pallet-bridge-relayers`). That's because in `Pays::No` means that the "chain" is not receiving/losing any funds. And IIUC we wanted them to pay for each other - e.g. Statemine sovereign account will pay for transactions on BHP and vice versa. With `Pays::No` noone pays for anything.
> 
> I did not have looked into the pr that deeply. However, why not just refund at the end of your transactions? Like we do here: https://github.com/paritytech/substrate/blob/a1dc9d8ff3a62580491d4c38c06251d5a3ca99f6/frame/sudo/src/lib.rs#L165-L166

That's answered in the quote :) In Statemint <> Statemine bridge we may allow that. But if other parachains start using our bridge hub, we don't want transactions to be free for relayers (means our collators/"treasury" receives no payments). So we're using that schema with parachain sovereign accounts on bridge hubs, where they pay for each other on different sides of the bridge. @acatangiu may share the doc we have also please correct me if I'm wrong :)

bkchr:
> That's answered in the quote :) In Statemint <> Statemine bridge we may allow that. But if other parachains start using our bridge hub, we don't want transactions to be free for relayers (means our collators/"treasury" receives no payments). So we're using that schema with parachain sovereign accounts on bridge hubs, where they pay for each other on different sides of the bridge. @acatangiu may share the doc we have also please correct me if I'm wrong :)

Sorry! I now looked into the code and it made more sense what you are doing here. If you would miss some call to include in your signed extension, the worst outcome would be that someone needs to pay for its call? I think it would make sense to abstract the fee calculation/fee refund, so that not only transaction-payment pallet is supported.

svyatonik:
> If you would miss some call to include in your signed extension, the worst outcome would be that someone needs to pay for its call?

Yep, that's correct! We'll be refunding players which are following our rules.

> I think it would make sense to abstract the fee calculation/fee refund, so that not only transaction-payment pallet is supported.

I've abstracted fee calculation. Fore the refund we have our own bridge-specific pallet, so probably not worth it

bkchr:
> I've abstracted fee calculation. Fore the refund we have our own bridge-specific pallet, so probably not worth it

Yeah that sounds completely fine :)

acatangiu (approved):
:100:

bin/runtime-common/src/refund_relayer_extension.rs
```diff
@@ -0,0 +1,844 @@
...
+					return Some(CallType::ParachainFinalityAndDelivery(
+						extract_expected_parachain_state::<R, GI, PI, PID>(&calls[0])?,
+						extract_messages_state::<R, GI, MI, LID>(&calls[1])?,
+					))
+				}
+				return None
+			}
+
+			Some(CallType::Delivery(extract_messages_state::<R, GI, MI, LID>(call)?))
+		};
```
serban300, line 259: Can we have this logic defined on the `CallType` struct in a constructor ? Something like `from_batch_all_call()`. Also could we define `extract_expected_relay_chain_state` on `ExpectedRelayChainState` as `from_call()` or `from_grandpa_call()` ? `extract_expected_parachain_state` on `ExpectedParachainState`, `extract_messages_state`  on `MessagesState` ?

serban300 (commented):
Overall the idea looks good. I just left a comment.

bin/runtime-common/src/refund_relayer_extension.rs
svyatonik: That's a good idea, but then I have to duplicate this all 30-line `where` in this `from_batch_all_call`, which I don't like :) Perhaps get back to this when we'll make it a bit easier to configure (decrease `where` LOCs)?

serban300: Sure, sounds good.

serban300 (approved):

bin/runtime-common/src/refund_relayer_extension.rs
svyatonik: Added note to #1666

PR #1658, opened 2022-11-22
Utility to generate hex bytes for send-message command

bkontur:


svyatonik (approved):
I'm not quite sure what is the point of having it in the codebase? But if you think it'll be helpful - let's keep it.

bkontur:
> I'm not quite sure what is the point of having it in the codebase? But if you think it'll be helpful - let's keep it.

well, at least, we will now how was that const `020419ac` generated,
and also can be useful for anybody else to play with bridges: https://substrate.stackexchange.com/questions/5901/failed-to-run-parity-bridges-common-message-realy-demo/5906#5906

bkontur:
> I'm not quite sure what is the point of having it in the codebase? But if you think it'll be helpful - let's keep it.

@svyatonik no problem, we can also remove that, is there any utility for simple generation XCM messages?

svyatonik:
My point is that this test isn't actually testing something :) Also it does prints, so it looks more like a good candidate to became a subcommand of `substrate-relay` (we already had the `encode-message` subcommand before). But for XCM - what you gonna pass using CLI? You'll need some other language to describe what you actually want, so imo it doesn't worth it.

As for the encoding tool, you can use https://polkadot.js.org/apps/. E.g.:
Screenshot from 2022-11-22 15-17-15

bkontur:
> As for the encoding tool, you can use https://polkadot.js.org/apps/. E.g.: Screenshot from 2022-11-22 15-17-15



hmm, `polkadot.js.org` thats where I started yesterday, but it didnt work,
I will try that later again, and when it will work as I expect, I will remove that test

PR #1659, opened 2022-11-22
More accurate dispatch_result: true/false

bkontur:


svyatonik (approved):

bin/runtime-common/src/messages.rs
```diff
@@ -511,14 +511,40 @@ pub mod target {
...
+				Ok(outcome) => {
+					log::trace!(
+						target: "runtime::bridge-dispatch",
+						"Incoming message {:?} dispatched with result: {:?}",
+						message_id,
+						outcome,
+					);
+					match outcome.ensure_execution() {
+						Ok(_weight) => true,
+						Err(e) => {
```
bkontur, line 524: @svyatonik 
demo scripts end here `Err(Barrier)`,
but all the time we return `dispatch_result: true,`, which is a little bit confusing with `true if dispatch has succeeded and `false` otherwise`

> pub struct MessageDispatchResult {
	/// Dispatch result flag. This flag is relayed back to the source chain and, generally
	/// speaking, may bring any (that fits in single bit) information from the dispatcher at
	/// the target chain to the message submitter at the source chain. If you're using immediate
	/// call dispatcher, then it'll be result of the dispatch - `true` if dispatch has succeeded
	/// and `false` otherwise.
	pub dispatch_result: bool,


wdyt?

svyatonik: Sorry, I've never used demo scripts - can't say what happens there. We'll fix them after completing all our major changes. But on our test deployments I can see following in Millau logs:
```
2022-11-22 08:19:36.005 TRACE tokio-runtime-worker runtime::bridge-dispatch: Incoming message ([0, 0, 0, 0], 790) dispatched with result: Ok(Incomplete(2000000000, Trap(43)))    
```

I'm not sure - if that's what you also getting in demo scripts? That's at least what I wanted to see in logs :) That was the main point - to see that the incoming message is decoded properly. Feel free to suggest changes - probably you'll need some other tooling.

svyatonik: Also - right now I'm removing this `dispatch_result` field (https://github.com/paritytech/parity-bridges-common/issues/1638#issuecomment-1319825292) :) It is no longer valid in XCM environment && it won't be used

svyatonik: (but your log changes are still ok && helpful)

bkontur: I have something like: 
```
tokio-runtime-worker runtime::bridge-dispatch: Incoming message ([0, 0, 0, 0], 790) dispatched with result: Ok(Error(Barrier))
```

svyatonik: > I have something like:
> 
> ```
> tokio-runtime-worker runtime::bridge-dispatch: Incoming message ([0, 0, 0, 0], 790) dispatched with result: Ok(Error(Barrier))
> ```

Hmmm. That's strange. Maybe some misconfiguration in `dev` configuration of involved chains? On test deployments we're using `local` configuration - there may be some differences (though I'm not sure)?

PR #1660, opened 2022-11-22
Remove dispatch_result field

svyatonik:
#1638

### DIFF
--- a/bin/runtime-common/src/messages.rs
+++ b/bin/runtime-common/src/messages.rs
@@ -520,7 +520,7 @@ pub mod target {
 						outcome,
 					);
 					match outcome.ensure_execution() {
-						Ok(_weight) => true,
+						Ok(_weight) => (),
 						Err(e) => {
 							log::error!(
 								target: "runtime::bridge-dispatch",
@@ -528,7 +528,6 @@ pub mod target {
 								message_id,
 								e,
 							);
-							false
 						},
 					}
 				},
@@ -539,7 +538,6 @@ pub mod target {
 						message_id,
 						e,
 					);
-					false
 				},
 			}
 


primitives/runtime/src/messages.rs
```diff
@@ -23,12 +23,6 @@ use scale_info::TypeInfo;
 /// Message dispatch result.
 #[derive(Encode, Decode, RuntimeDebug, Clone, PartialEq, Eq, TypeInfo)]
 pub struct MessageDispatchResult {
-	/// Dispatch result flag. This flag is relayed back to the source chain and, generally
-	/// speaking, may bring any (that fits in single bit) information from the dispatcher at
-	/// the target chain to the message submitter at the source chain. If you're using immediate
-	/// call dispatcher, then it'll be result of the dispatch - `true` if dispatch has succeeded
-	/// and `false` otherwise.
-	pub dispatch_result: bool,
```
bkontur, line 31: @svyatonik 
so this is now kind of "fire and forget"?
so, we dont want to add some information if dispatch was ok/failed to the `Event::MessagesDelivered`  ?

svyatonik: Yes. In our previous API, the messaging pallet was the entrypoint for all messages, so we were building applications around that palet (e.g. atomic token swap - https://github.com/paritytech/parity-bridges-common/tree/c7c6f0ce5e88b3ba9cdcd817f4286e76221b8222/modules/token-swap). There we were actually using this (one-bit) `dispatch_result` to see if our call has succeeded or not.

Now the XCM infra is the central point, so applications are built around XCM. We are just transport, so no extra logic is required. E.g. XCM has its own mechanism to send responses back to the caller (see `ReportHolding`, `ReportTransactStatus`, `ReportError` XCM instructions). Hence we don't need our own mechanism anymore

bkontur: ok, got it, thank you :)

bkontur (approved):
lgtm

PR #1670, opened 2022-11-30
Added generic DispatchLevelResult to the MessageDispatchResult

bkontur:


bin/runtime-common/src/messages.rs
```diff
@@ -449,6 +449,7 @@ pub mod target {
 		WeightCredit: Get<Weight>,
 	{
 		type DispatchPayload = FromBridgedChainMessagePayload<CallOf<ThisChain<B>>>;
+		type DispatchLevelResult = ();
```
bkontur, line 452: @svyatonik 
do you want to change this to `type DispatchLevelResult = Result<Outcome, Error>` (I didnt try if compiles) or `()` is ok as default for `FromBridgedChainMessageDispatch`?

svyatonik: IMHO `()` is enough for our testnets - dispatch result is not important here, only delivery

primitives/messages/src/lib.rs
```diff
@@ -237,12 +237,12 @@ impl<Result> ReceivedMessages<Result> {
 
 /// Result of single message receival.
 #[derive(RuntimeDebug, Encode, Decode, PartialEq, Eq, Clone, TypeInfo)]
-pub enum ReceivalResult {
+pub enum ReceivalResult<DispatchLevelResult: Clone + Decode + sp_std::fmt::Debug + Eq> {
```
svyatonik, line 240: Can this be avoided by replacing with `#[derive(RuntimeDebugNoBound, Encode, Decode, PartialEqNoBound, EqNoBound, CloneNoBound, TypeInfo]`?

primitives/runtime/src/messages.rs
```diff
@@ -22,7 +22,7 @@ use scale_info::TypeInfo;
 
 /// Message dispatch result.
 #[derive(Encode, Decode, RuntimeDebug, Clone, PartialEq, Eq, TypeInfo)]
-pub struct MessageDispatchResult {
+pub struct MessageDispatchResult<DispatchLevelResult: Clone + Decode + sp_std::fmt::Debug + Eq> {
```
svyatonik, line 25: Can you also try `NoBound` versions here? :) I've found them recently && iiuc it should help with removing these constraints (maybe I'm wrong though)

svyatonik (approved):
LGTM

primitives/messages/src/lib.rs
bkontur: hmm, I just removed those `Clone + Decode + sp_std::fmt::Debug + Eq` from ReceivalResult and MessageDispatchResult and they are not needed here.
I have no idea why I added them there :)

primitives/runtime/src/messages.rs
bkontur: when I tried to change `RuntimeDebug` to `RuntimeDebugNoBound` I get this error:
image

PR #1691, opened 2022-12-01
Replace const parameters types

### LINKED ISSUE #1449: Remove explicit const parameters declarations
So Rust now has const generics and we may use
```rust
type MaxConsumers = frame_support::traits::ConstU32<16>;
```
instead of
```rust
parameter_types! {
	pub const MaxConsumers: u32 = 16;
}
...
type MaxConsumers = frame_support::traits::ConstU32<16>;
```

Changing that would save us dozens of lines, so it is worth fixing imo :)

binsta:
Replace const parameters_types! with const generics this resolve #1449


svyatonik (approved):
LGTM. Thank you!

Issue #1692, opened 2022-12-02
Support pallet instances in bridge-relayers pallet


svyatonik:
After #1654 we are identifying "bridge" in bridge-relayers pallet with `LaneId`. But it isn't consistent with out other pallets. The current approach is 1 grandpa + messages pallet for single bridged chain (parachains is a bit different - it supports all parachains of the bridged relay chain). And since we don't support instances in the relayers pallet, it would mean that it may only be used by single bridge (unless we make `LaneId` values unique across all bridges).

The solutions are:
- add instances support to the bridge-relayers pallet;
- use pair `(ChainId, LaneId)` to identify payment source.
It isn't actual now, since we're planning to have only one instance of every pallet initially

serban300:
Resolved by #1907

PR #1693, opened 2022-12-02
removed MintReward and added PayLaneRewardFromAccount

### LINKED ISSUE #1654: Remove MintReward and introduce PayRewardFromAccount
Right now there's a single implementation of `PaymentProcedure` for the `pallet-bridge-relayers` - it is the `MintReward`, which just mints relayer reward. Minting is not actual anymore. Instead the reward will be paid from some account (sovereign account of the parachain at bridge hub), so having `PayRewardFromAccount` would be great

svyatonik:
closes #1654

See #1694 and #1692

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -865,6 +865,7 @@ name = "bp-relayers"
 version = "0.1.0"
 dependencies = [
  "bp-messages",
+ "bp-rialto",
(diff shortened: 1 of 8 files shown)


acatangiu (approved):
👍🏻

Issue #1627, opened 2022-11-11
Separate header type and storage type


svyatonik:
GRANDPA and parachains finality pallets are accepting the chain header/head and storing it directly in the runtime. For most of cases (including ours) we may only store the storage root, thus decreasing the occupied runtime storage + message transactions bandwidth (?).

The implementation may look like:
```rust
trait HeaderArtifacts {
  type Artifacts;
  fn into(header: Header) -> Self::Artifacts;
}

struct StorageRootArtifact;
impl HeaderArtifacts for StorageRootArtifact {
  type Artifacts = H256;
  fn into(header: Header) -> H256 {
    *header.storage_root()  
  }
}

...

impl pallet_grandpa::Config for Runtime {
 ...
 type Artifacts = StorageRootArtifact;
 ...
}
```

svyatonik:
I'm prioritizing this issue, since we're now changing the "scope" of our pallets to specific case. So would be better to have this asap.

The same thing may be done for the parachains pallet (would require head parser configuration for every parachain)

PR #1699, opened 2022-12-05
Only store header state root in pallet-bridge-grandpa

svyatonik:
part of #1627

This is an important change. Before that PR we've been storing the full header of the bridged chain in the runtime storage. Headers may be big (headers that are changing GRANDPA authorities contain all authorities public keys). That is bad - we're increasing runtime storage + max PoV size is larger. Why we're storing full header - that's an artifact of our previous design, where apps were built on top of bridge. But now everything has changed && apps will be built on top of XCM - I doubt that we ever need more header fields in runtime storage.

If this PR will be accepted, I'll introduce similar changes to the `pallet-bridge-parachains` pallet in follow-up PR.

### DIFF
--- a/modules/grandpa/src/lib.rs
+++ b/modules/grandpa/src/lib.rs
@@ -38,10 +38,11 @@
 
 use storage_types::StoredAuthoritySet;
 
-use bp_header_chain::{justification::GrandpaJustification, HeaderChain, InitializationData};
-use bp_runtime::{
-	BlockNumberOf, BoundedStorageValue, Chain, HashOf, HasherOf, HeaderOf, OwnedBridgeModule,
+use bp_header_chain::{
+	justification::GrandpaJustification, HeaderChain, InitializationData, StoredHeaderData,
+	StoredHeaderDataBuilder,
 };
+use bp_runtime::{BlockNumberOf, Chain, HashOf, HasherOf, HeaderId, HeaderOf, OwnedBridgeModule};
 use finality_grandpa::voter_set::VoterSet;
 use frame_support::{ensure, fail};
 use frame_system::ensure_signed;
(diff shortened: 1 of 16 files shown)


acatangiu (approved):

modules/grandpa/src/lib.rs
```diff
@@ -315,10 +295,10 @@ pub mod pallet {
 	pub(super) type ImportedHashesPointer<T: Config<I>, I: 'static = ()> =
 		StorageValue<_, u32, ValueQuery>;
 
-	/// Headers which have been imported into the pallet.
+	/// Relevant fields of imported headers.
 	#[pallet::storage]
 	pub type ImportedHeaders<T: Config<I>, I: 'static = ()> =
-		StorageMap<_, Identity, BridgedBlockHash<T, I>, StoredBridgedHeader<T, I>>;
+		StorageMap<_, Identity, BridgedBlockHash<T, I>, BridgedStoredHeaderData<T, I>>;
```
bkontur, line 301: @svyatonik 
I guess, this is not backwards compatible, right?
and there is no migration.

so, for live deployed Rococo/Wococo BridgeHubs, next time I will update runtime which contains this commit, I should clean database and/or reonboard parachains?

It is not a problem, I just want to know.

svyatonik: Yes - it isn't backwards-compatible, but iiuc the upgrade will be smooth - headers import is not touching any previous headers, only current authorities set. And we are not touching set here. So the issue will only be with verifying storage proofs using old headers, which we are not doing currently. We'll leave some garbage in Rococo/Wococo runtime storage though. I'm not sure if it is critical, since they're testnets.

bkontur: ok, cool, thank you, so just decoding will fail for old stored headers, this is no problem,
yes, we will have some garbage, but anyway later or sooner, I am sure, we will do clean reonboard :)

PR #1701, opened 2022-12-05
Only store header state root (pallet-bridge-parachains)

### LINKED ISSUE #1627: Separate header type and storage type

svyatonik:
follow up of #1699
closes #1627 

Some things yet to be done + testing required


relays/utils/src/lib.rs
```diff
@@ -51,7 +52,7 @@ pub mod relay_loop;
 pub trait BlockNumberBase:
 	'static
 	+ From<u32>
-	+ Into<u64>
+	+ UniqueSaturatedInto<u64>
```
svyatonik, line 55: This is now used instead of simple `.into()` in two places: (1) for metrics (2) when we talk about transaction mortality. But we talk about block number here and it is limited by 2^64, which is imo enough. In worst case we'll have: (1) an alert when there are more than 2^64 blocks and (2) transaction which dies after 2^64 blocks (but if you're crafting such tx, you probably mean immportal tx)

modules/parachains/src/lib.rs
```diff
@@ -502,27 +535,28 @@ pub mod pallet {
 			}
 
 			// verify that the parachain head size is <= `MaxParaHeadSize`
```
acatangiu, line 537:
```suggestion
			// verify that the parachain head data size is <= `MaxParaHeadDataSize`
```

acatangiu, line 537:
```suggestion
			// verify that the parachain head data size is <= `MaxParaHeadDataSize`
```

modules/parachains/src/lib.rs
```diff
@@ -502,27 +535,28 @@ pub mod pallet {
...
-					return Err(())
-				},
-			};
+			let updated_head_data =
+				match StoredParaHeadDataOf::<T, I>::try_from_inner(updated_head_data) {
+					Ok(updated_head_data) => updated_head_data,
+					Err(e) => {
+						log::trace!(
+							target: LOG_TARGET,
+							"{}. The parachain head size for {:?} is {}. It exceeds maximal configured size {}.",
```
acatangiu, line 544:
```suggestion
							"{}. The parachain head data size for {:?} is {}. It exceeds maximal configured size {}.",
```

modules/parachains/src/mock.rs
```diff
@@ -40,7 +36,108 @@ type UncheckedExtrinsic = frame_system::mocking::MockUncheckedExtrinsic<TestRunt
 
 pub const PARAS_PALLET_NAME: &str = "Paras";
 pub const UNTRACKED_PARACHAIN_ID: u32 = 10;
-pub const MAXIMAL_PARACHAIN_HEAD_SIZE: u32 = 512;
+pub const MAXIMAL_PARACHAIN_HEAD_DATA_SIZE: u32 = 1 + 8 + 32;
```
acatangiu, line 39: nit:

```suggestion
// use exact expected encoded size: `vec_len_size + header_number_size + state_root_hash_size`
pub const MAXIMAL_PARACHAIN_HEAD_DATA_SIZE: u32 = 1 + 8 + 32;
```

also maybe I'm misunderstanding the `1 + 8 + 32` here, but isn't vec.len() supposed to be an usize (8 bytes)?

primitives/parachains/src/lib.rs
```diff
@@ -86,5 +93,59 @@ impl StorageDoubleMapKeyProvider for ImportedParaHeadsKeyProvider {
...
+					.encode(),
+			))
+		}
+		None
+	}
+}
+
+#[impl_trait_for_tuples::impl_for_tuples(1, 30)]
+#[tuple_types_custom_trait_bound(Parachain)]
+impl ParaStoredHeaderDataBuilder for C {
```
acatangiu, line 140: nit:

```suggestion
/// Tries to build header data from each tuple member, short-circuiting on first successful one.
impl ParaStoredHeaderDataBuilder for C {
```

acatangiu (approved):

modules/parachains/src/mock.rs
svyatonik: Len is encoded as compact int, so for small values it is 1 byte

Issue #1711, opened 2022-12-12
Remove SubstrateParachainsPipeline::SOURCE_PARACHAIN_PARA_ID


svyatonik:
Instead restrict `SubstrateParachainsPipeline::SourceParachain` to `Parachain` and use `SubstrateParachainsPipeline::SourceParachain::PARACHAIN_ID` everywhere.

PR #1716, opened 2022-12-13
Remove SOURCE_PARACHAIN_PARA_ID

serban300:
Fixes: #1711 

Remove SubstrateParachainsPipeline::SOURCE_PARACHAIN_PARA_ID

### DIFF
--- a/primitives/runtime/src/chain.rs
+++ b/primitives/runtime/src/chain.rs
@@ -192,10 +192,44 @@ pub trait Chain: Send + Sync + 'static {
 	fn max_extrinsic_weight() -> Weight;
 }
 
+pub trait ChainShadow {
+	type Chain: Chain;
+}
+
+impl<T> Chain for T
+where
+	T: Send + Sync + 'static + ChainShadow,
+{
+	type BlockNumber = <T::Chain as Chain>::BlockNumber;
+	type Hash = <T::Chain as Chain>::Hash;
+	type Hasher = <T::Chain as Chain>::Hasher;
+	type Header = <T::Chain as Chain>::Header;
+	type AccountId = <T::Chain as Chain>::AccountId;
+	type Balance = <T::Chain as Chain>::Balance;
+	type Index = <T::Chain as Chain>::Index;
+	type Signature = <T::Chain as Chain>::Signature;
+
+	fn max_extrinsic_size() -> u32 {
+		<T::Chain as Chain>::max_extrinsic_size()
+	}
+
+	fn max_extrinsic_weight() -> Weight {
+		<T::Chain as Chain>::max_extrinsic_weight()
+	}
+}
+
 /// Minimal parachain representation that may be used from no_std environment.
 pub trait Parachain: Chain {
 	/// Parachain identifier.
 	const PARACHAIN_ID: u32;
+}
+
+impl<T> Parachain for T
+where
+	T: Chain + ChainShadow,
+	<T as ChainShadow>::Chain: Parachain,
+{
+	const PARACHAIN_ID: u32 = <<T as ChainShadow>::Chain as Parachain>::PARACHAIN_ID;
 }
 
 /// Block number used by the chain.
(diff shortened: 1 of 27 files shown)


relays/client-substrate/src/chain.rs
```diff
@@ -72,6 +73,9 @@ pub trait RelayChain: Chain {
 	const PARACHAINS_FINALITY_PALLET_NAME: &'static str;
 }
 
+/// Substrate-based parachain from minimal relay-client point of view.
+pub trait Parachain: Chain + ParachainBase {}
```
svyatonik, line 77: Is it really required? If so, can we do a blanket impl here (`impl<T: Chain + ParachainBase> Parachain for T {}`)?

svyatonik (approved):
LGTM, modulo this new trait (e.g. reuse "base" `Parachain` trait?). If it isn't really required, maybe remove it

relays/client-bridge-hub-wococo/src/lib.rs
```diff
@@ -140,6 +140,12 @@ impl ChainWithMessages for BridgeHubWococo {
 	type WeightInfo = ();
 }
 
+impl ParachainBase for BridgeHubWococo {
+	const PARACHAIN_ID: u32 = bp_bridge_hub_wococo::BRIDGE_HUB_WOCOCO_PARACHAIN_ID;
```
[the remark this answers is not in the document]
bkontur: here, maybe replace:
`impl ChainBase for BridgeHubWococo {`
with:
```
impl UnderlyingChainProvider for BridgeHubWococo {
	type Chain = bp_bridge_hub_wococo::BridgeHubWococo; // primitives has this already - see below
}
```

```
impl Chain for BridgeHubWococo {
	type BlockNumber = BlockNumber;
	type Hash = Hash;
	type Hasher = Hasher;
	type Header = Header;

	type AccountId = AccountId;
	type Balance = Balance;
	type Index = Index;
	type Signature = Signature;

	fn max_extrinsic_size() -> u32 {
		*BlockLength::get().max.get(DispatchClass::Normal)
	}

	fn max_extrinsic_weight() -> Weight {
		BlockWeights::get()
			.get(DispatchClass::Normal)
			.max_extrinsic
			.unwrap_or(Weight::MAX)
	}
}
```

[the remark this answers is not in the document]
serban300: Good catch ! I didn't see this. Seems redundant. Trying to remove the duplication.

relays/client-substrate/src/chain.rs
serban300: It's not required, for the moment it's just a convenience trait (we can specify only `Parachain` instead of `Chain + ParachainBase`). I think this helps. Also maybe it would be better to keep it in case in the future we need to add more properties to the `Parachain` trait on the relayer side. But I don't have strong preferences here.

relays/client-bridge-hub-wococo/src/lib.rs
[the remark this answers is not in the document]
serban300: Done

primitives/runtime/src/chain.rs
```diff
@@ -192,12 +192,46 @@ pub trait Chain: Send + Sync + 'static {
 	fn max_extrinsic_weight() -> Weight;
 }
 
+pub trait ChainShadow {
```
bkontur, line 195: @serban300 
cool, and now we can merge/replace `ChainShadow` with `UnderlyingChainProvider` from runtime-common - it is exactly the same trait, so I would just move UnderlyingChainProvider here instead of ChainShadow

```
/// A trait that provides the type of the underlying chain.
pub trait UnderlyingChainProvider {
	/// Underlying chain type.
	type Chain: Chain;
}
```

bkontur: and also this:
```
/// Underlying chain type.
pub type UnderlyingChainOf<C> = <C as UnderlyingChainProvider>::Chain;
```

can be moved to the primites here also

PR #1718, opened 2022-12-14
Check origin?

bkontur:


svyatonik (commented):
Some issue you're fixing with this change? I don't have anything against it, but would like to know :)

bkontur:
> Some issue you're fixing with this change? I don't have anything against it, but would like to know :)

no issue, I was just checking something else and I accidentally found that it is missing here,
I thought that origin needs to be checked somehow everytime.

I am not sure why it is missing here, so the question "Check origin?" :)

svyatonik (approved):
I'd rather remove the similar check from the grandpa pallet (because - why do we need it), but this option is fine too.

bkontur:
ok, I think I understand now, 
so it should be correct/better `ensure_owner_or_root_or_signed`, but we dont use origin here, so there is no need to check a anything, right?

svyatonik:
> ok, I think I understand now, so it should be correct/better `ensure_owner_or_root_or_signed`, but we dont use origin here, so there is no need to check a anything, right?

I don't think it should be `ensure_owner_or_root_or_signed` - because we don't care who's submitting headers. The only thing we care is that the transaction is valid (we're checking this inside the call) and paid (that must be checked outside). In messages e.g. we care about signed txs because we pay rewards to the signers. But here - I do not see why we need this check at all. But may be I miss something - that's why I have asked you :)

bkontur:
> > ok, I think I understand now, so it should be correct/better `ensure_owner_or_root_or_signed`, but we dont use origin here, so there is no need to check a anything, right?
> 
> I don't think it should be `ensure_owner_or_root_or_signed` - because we don't care who's submitting headers. The only thing we care is that the transaction is valid (we're checking this inside the call) and paid (that must be checked outside). In messages e.g. we care about signed txs because we pay rewards to the signers. But here - I do not see why we need this check at all. But may be I miss something - that's why I have asked you :)

ok, cool, thank you Slava for explanation, so now I understand :),
I removed those

PR #1725, opened 2022-12-16
Do not require new headers if lane is empty

### LINKED ISSUE #1723: Parachain heads are synced without demand in parachain <> parachain bridge
It seems that when we bridge two parachains and run a complex relayer, parachains heads are relayed even when this isn't required by the message relay. It needs to be investigated and fixed.

svyatonik:
closes #1723

If latest confirmed nonce is `0`, it means that nothing has been actually confirmed. So we do not need to ask for new headers just because we have nothing to deliver

### DIFF
--- a/relays/messages/src/message_race_delivery.rs
+++ b/relays/messages/src/message_race_delivery.rs
@@ -316,8 +316,11 @@ where
 	) -> Option<SourceHeaderIdOf<P>> {
 		let header_required_for_messages_delivery =
 			self.strategy.required_source_header_at_target(current_best);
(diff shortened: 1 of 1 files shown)


acatangiu (approved):

svyatonik:
Sorry, the fix is not complete (on a second thought). Will come back with proper fix

svyatonik:
Now it should be fine - I've remembered the similar #1328, so the fix should be general. But let's not merge this PR until next week - if something will break, I want to be sure that it is the #1669, not this PR

PR #1728, opened 2022-12-19
Remove messages pallet owner relay argument

svyatonik:
It was used to update conversion rate and we don't need it anymore


bkontur (approved):
lgtm

acatangiu (approved):

PR #1733, opened 2022-12-26
Replace `BATCH_CALL_SUPPORTED`

### LINKED ISSUE #1724: Replace BATCH_CALL_SUPPORTED with compile-time traits check
Details: https://github.com/paritytech/parity-bridges-common/pull/1669#discussion_r1050572677

serban300:
Fixes #1724 

More details in the commit messages

### DIFF
--- a/relays/client-substrate/src/client.rs
+++ b/relays/client-substrate/src/client.rs
@@ -426,6 +426,19 @@ impl<C: Chain> Client<C> {
 		.await
 	}
 
+	async fn build_sign_params(&self, signer: AccountKeyPairOf<C>) -> Result<SignParam<C>>
+	where
+		C: ChainWithTransactions,
+	{
+		let (spec_version, transaction_version) = self.simple_runtime_version().await?;
+		Ok(SignParam::<C> {
+			spec_version,
+			transaction_version,
+			genesis_hash: self.genesis_hash,
+			signer,
+		})
+	}
+
 	/// Submit an extrinsic signed by given account.
 	///
 	/// All calls of this method are synchronized, so there can't be more than one active
@@ -447,13 +460,7 @@ impl<C: Chain> Client<C> {
 		let _guard = self.submit_signed_extrinsic_lock.lock().await;
 		let transaction_nonce = self.next_account_index(signer.public().into()).await?;
 		let best_header = self.best_header().await?;
-		let (spec_version, transaction_version) = self.simple_runtime_version().await?;
-		let signing_data = SignParam::<C> {
-			spec_version,
-			transaction_version,
-			genesis_hash: self.genesis_hash,
-			signer: signer.clone(),
-		};
+		let signing_data = self.build_sign_params(signer.clone()).await?;
 
 		// By using parent of best block here, we are protecing again best-block reorganizations.
 		// E.g. transaction may have been submitted when the best block was `A[num=100]`. Then it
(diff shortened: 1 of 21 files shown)


relays/lib-substrate-relay/src/messages_source.rs
```diff
@@ -366,14 +364,10 @@ where
...
 		if let Some(ref target_to_source_headers_relay) = self.target_to_source_headers_relay {
-			if P::SourceBatchCallBuilder::BATCH_CALL_SUPPORTED {
-				let (proved_header, prove_calls) =
-					target_to_source_headers_relay.prove_header(id.0).await?;
-				return Ok(Some(BatchConfirmationTransaction {
-					messages_source: self.clone(),
-					proved_header,
-					prove_calls,
-				}))
+			let (proved_header, prove_calls) =
```
svyatonik, line 367: This changes logic a bit - you always generate a header proof (which is a heavy operation) and then, if batch calls are not supported, you simply drop it. I'd keep the old behavior, where it is only generated when strictly required. OTOH it isn't that critical - we don't care that much about relay performance. If you'd like to keep it, please at least file an issue to fix it

relays/client-substrate/src/client.rs
```diff
@@ -435,18 +435,25 @@ impl<C: Chain> Client<C> {
 	/// Note: The given transaction needs to be SCALE encoded beforehand.
 	pub async fn submit_signed_extrinsic(
 		&self,
-		extrinsic_signer: C::AccountId,
-		signing_data: SignParam<C>,
```
svyatonik, line 439: :+1:

relays/lib-substrate-relay/src/lib.rs
```diff
@@ -90,50 +91,41 @@ impl<AccountId> TaggedAccount<AccountId> {
...
 	/// Create batch call from given calls vector.
-	fn build_batch_call(_calls: Vec<Call>) -> Result<Call, Self::Error>;
+	fn build_batch_call(&self, _calls: Vec<Call>) -> Call;
 }
 
-impl<Call> BatchCallBuilder<Call> for () {
-	type Error = SubstrateError;
-	const BATCH_CALL_SUPPORTED: bool = false;
+/// Batch call builder constructor.
+pub trait BatchCallBuilderConstructor<Call> {
```
svyatonik, line 100: I won't insist on changing that, but I'd like to note that imo 4 traits (`BatchCallBuilderConstructor`, `BatchCallBuilder`, `ChainWithUtilityPallet`, `UtilityPallet`) instead of single one (`BatchCallBuilder`) is too much :) E.g. which other implementations of the `UtilityPallet` we may have? :) But as I said - if it solves the problem, let's keep it as is :)

relays/client-substrate/src/client.rs
svyatonik: We shall probably do the same with the `submit_and_watch_signed_extrinsic` (not a part of this PR, of course).

svyatonik (approved):
Had a couple of suggestions, but they aren't critical - feel free to merge as is

relays/client-substrate/src/client.rs
serban300: I did these changes for `submit_and_watch_signed_extrinsic` also. Also added a function for creating the signing params in order to avoid duplicating this logic.

relays/lib-substrate-relay/src/lib.rs
serban300: Yes, I don't like to have so many traits either, but I couldn't think of another solution. I think that `ChainWithUtilityPallet` makes sense since this way we can associate the batching functionality with the chain. And also we can add constraints like implementing `UtilityPalletBatchCallBuilder` only for chains with utility pallet. And adding `BatchCallBuilderConstructor` is a small compromise for safety. About the `UtilityPallet` trait I'm not sure, but from what I understand we will have 2 implementations for it. One for the chains that are defined in `parity-bridges-common` and for which we have the full runtime (Rialto, Millau). And one for the chains defined outside (the RococoBridgeHub, Wococo bridgeHub for example) where we only have a RuntimeWrapper. I might be wrong. It would be nice if we can avoid it.

relays/lib-substrate-relay/src/messages_source.rs
serban300: Fixed. Thanks ! I completely missed this.

PR #1739, opened 2022-12-30
Simplify read_client_state()

serban300:


svyatonik (approved):
Looks much better now, thank you! :)

PR #1740, opened 2022-12-30
Make debug display of LaneId compatible with its previous version

svyatonik:
follow up of #1738

We're using formatted version in logs, would be good to keep it shorter (as before)

### DIFF
--- a/primitives/messages/src/lib.rs
+++ b/primitives/messages/src/lib.rs
@@ -78,11 +78,30 @@ impl OperatingMode for MessagesOperatingMode {
 	Ord,
 	PartialOrd,
 	PartialEq,
-	RuntimeDebug,
 	TypeInfo,
 	MaxEncodedLen,
 )]
 pub struct LaneId(pub [u8; 4]);
+
+pub use lane_id_debug_impl::*;
+
+#[cfg(not(feature = "std"))]
+mod lane_id_debug_impl {
+	impl core::fmt::Debug for super::LaneId {
+		fn fmt(&self, fmt: &mut core::fmt::Formatter) -> core::fmt::Result {
+			fmt.write_str("<stripped>")
+		}
+	}
+}
+
+#[cfg(feature = "std")]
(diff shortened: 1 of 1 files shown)


acatangiu (approved):

primitives/messages/src/lib.rs
```diff
@@ -69,21 +69,30 @@ impl OperatingMode for MessagesOperatingMode {
...
+
+#[cfg(feature = "std")]
+mod lane_id_debug_impl {
+	impl std::fmt::Debug for super::LaneId {
+		fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
+			self.0.fmt(fmt)
+		}
+	}
+}
+
```
serban300, line 95: Do we want to print `<stripped>` for non-std environments ? Was this the previous behaviour ?

Could we just do the following ?

```
impl core::fmt::Debug for LaneId {
	fn fmt(&self, fmt: &mut core::fmt::Formatter) -> core::fmt::Result {
		self.0.fmt(fmt)
	}
}
```

Would this work for both std and non-std environments ?

primitives/messages/src/lib.rs
```diff
@@ -458,4 +467,9 @@ mod tests {
 		assert!(delivered_messages.contains_message(150));
 		assert!(!delivered_messages.contains_message(151));
 	}
+
+	#[test]
+	fn lane_id_debug_format_works_as_before() {
```
serban300, line 472: Nit: I would rename this to something like: `lane_id_debug_prints_short_string`

primitives/messages/src/lib.rs
svyatonik: No, "<stripped>" is how `RuntimeDebug` works. I think you're right and in previous version, it was printing the same from the no-std env too, because we weren't using some custom type. Let's change it

serban300 (approved):

PR #1741, opened 2022-12-30
Remove `CliChain::KeyPair`

### LINKED ISSUE #1631: Remove KeyPair associated type from the CliChain trait
Since we're now using `ChainWithTransactions::AccountKeyPair`, we may remove it.

serban300:
Fixes #1631


svyatonik (approved):
Great, thank you!

acatangiu (approved):

PR #1743, opened 2023-01-04
Improve relayer initialization behaviour

### LINKED ISSUE #1003: Improve Relayer Initialization Behaviour
Right now the relayer claims to have succesfully initialized the bridge pallet when it
fires off a transaction to a node. However, if the node rejects the transactions (if for
example the init request was submitted with the wrong `owner`) the relayer has no way of
knowing that its initialization request failed.

We should do two things:
1. Tweak the wording of the initialization log message to indicate that the bridge might
   not actually be initialized yet (just that the tx was submitted)
2. Wait for a couple of seconds and follow up with the node to see if the transaction was
   succesful
   - We may need to have the runtime emit an `initialized` event which the relayer can
     listen for

serban300:
Fixes #1003 

Wait until the initialization transaction is finalized/lost.

### DIFF
--- a/relays/lib-substrate-relay/src/finality/initialize.rs
+++ b/relays/lib-substrate-relay/src/finality/initialize.rs
@@ -24,10 +24,12 @@
 use crate::{error::Error, finality::engine::Engine};
 use sp_core::Pair;
 
+use bp_runtime::HeaderIdOf;
 use relay_substrate_client::{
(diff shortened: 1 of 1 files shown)


acatangiu (approved):
looks good to me

bkontur (commented):
lgtm

Issue #1675, opened 2022-12-01
[R-W Bridge] Redeploy to start relaying/synching


EmmanuellNorbertTulbure:
- [x] add version guards to RBH<>WBH GRANDPA finality 
https://github.com/paritytech/parity-bridges-common/pull/1697 
- [x] update relayer to complex one:
https://gitlab.parity.io/parity/infrastructure/parity-testnet/-/blob/master/parity-bridge-common/wococo-relay-header.yaml#L4
https://gitlab.parity.io/parity/infrastructure/parity-testnet/-/blob/master/parity-bridge-common/rococo-relay-header.yaml#L4
- [x] set fixed defined RuntimeVersion for RBH and WBH 
https://github.com/paritytech/parity-bridges-common/pull/1756

bkontur:
@serban300 
you could also take this one, now there is no version guard for Ro/Wo,
we are waiting for deployment of `bridge-hub-rococo-9302` https://github.com/paritytech/devops/issues/1934#issuecomment-1375554017

so I would like to setup `bridge-hub-rococo-9302` as starting version guard for Ro/Wo,
it is not really needed for testnet, but I would like to set it up to be as-close-as-possible to the later Kusama/Polkadot bridge - just to avoid any unexpected surprieses in production :)

serban300:
Thanks ! I'll assign it to me then and I think I'll be able to do it these days.

svyatonik:
Version guard isn't about monitoring - it is about stopping the process when the relay detects it is connected to node that is using unexpected version. E.g. in Kusama<>Polkadot bridge when one of chains is upgraded, the relay stops (because of version guard) and we shall check that the upgrade isn't breaking our bridge and then we have following options:
- just recompile relay with upgraded bundled chain version and restart it;
- wait until the other chain is upgraded (e.g. we may need to wait for Kusama upgrade if Polkadot has been upgraded) - during that time, we shall not start relay;
- do some coding, republish and restart relay;
- something else that I can't think of now.

In any way - let me prepare some doc (before Thursday) - we need to discuss how we'll be handling runtime upgrades (have some process, some tools, ...). Maybe we can optimize/reduce number of manual actions. Let's discuss it on Thursday.

bkontur:
yes, yes, ok, cool,
I was thinking more about, how do we receive information that relayer was stopped by guard? Do we receive some notification on matrix or how?

svyatonik:
[...]

PR #1744, opened 2023-01-09
Deduplicate pallet call structs used for indirect runtime calls

serban300:
Part of #446


primitives/runtime/src/calls.rs
```diff
@@ -0,0 +1,30 @@
...
+//! Basic runtime calls.
+
+use codec::{Decode, Encode};
+use scale_info::TypeInfo;
+use sp_std::vec::Vec;
+
+/// A minimized version of `frame-system::Call` that can be used without a runtime.
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone, TypeInfo)]
+#[allow(non_camel_case_types)]
+pub enum SystemCall {
```
svyatonik, line 26: I don't think we are using remarks now (with XCM) - can you, please, check that?

relays/client-bridge-hub-wococo/src/runtime_wrapper.rs
```diff
@@ -16,11 +16,91 @@
...
-pub type Call = relay_bridge_hub_rococo_client::runtime::Call;
+use codec::{Decode, Encode};
+use scale_info::TypeInfo;
+
+pub use bp_header_chain::BridgeGrandpaCallOf;
+pub use bp_parachains::BridgeParachainCall;
+pub use bp_runtime::calls::SystemCall;
+pub use bridge_runtime_common::messages::BridgeMessagesCallOf;
+
+// TODO:check-parameter - check SignedExtension
```
svyatonik, line 27: Can you, please, file an issue for this TODO and leave the reference here. Same for other similar TODOs - this way we won't forget to actually check it when the PR will be merged :)

relays/client-bridge-hub-rococo/src/runtime_wrapper.rs
```diff
@@ -18,201 +18,45 @@
...
+pub use bp_parachains::BridgeParachainCall;
+pub use bp_runtime::calls::SystemCall;
+pub use bridge_runtime_common::messages::BridgeMessagesCallOf;
 
 // TODO:check-parameter - check SignedExtension
 /// Unchecked BridgeHubRococo extrinsic.
 pub type UncheckedExtrinsic = bp_bridge_hub_rococo::UncheckedExtrinsic<Call>;
 
-/// Rococo Runtime `Call` enum.
+pub type BridgeWococoGrandpaCall = BridgeGrandpaCallOf<bp_wococo::Wococo>;
```
svyatonik, line 33: Maybe add some tiny doc for public exported items.

svyatonik (approved):
LGTM, thank you! :)

svyatonik:
Also - how about adding tests to every pallet, where we'll be verifying that the `pallet::Call` are SCALE-compatible with their mocked versions at least for mock runtime? This way we'll be sure that we won't accidentally change the call without changing the mocked call.

relays/client-bridge-hub-wococo/src/runtime_wrapper.rs
bkontur: I guess, this TODO was about those `generate_bridge_reject_obsolete_headers_and_messages / BridgeRejectObsoleteHeadersAndMessages`, so this can be removed here

bkontur (approved):
lgtm

I also vote for that test to check BridgeGrandpaCall/BridgeMessagesCall/BridgeParachainCall, that Slava mentioned

serban300:
> Also - how about adding tests to every pallet, where we'll be verifying that the `pallet::Call` are SCALE-compatible with their mocked versions at least for mock runtime? This way we'll be sure that we won't accidentally change the call without changing the mocked call.

Good idea ! I added tests. Also In the future I want to see if it's possible to check this via the `TypeInfo` or some other method. But for the moment it's good to have tests.

relays/client-bridge-hub-wococo/src/runtime_wrapper.rs
serban300: I opened #1748 for this

svyatonik (approved):
Thank you! All is perfect now :)

relays/client-bridge-hub-wococo/src/runtime_wrapper.rs
```diff
@@ -16,11 +16,91 @@
...
+/// `BridgeHubWococo` Runtime `Call` enum.
+///
+/// The enum represents a subset of possible `Call`s we can send to `BridgeHubWococo` chain.
+/// Ideally this code would be auto-generated from metadata, because we want to
+/// avoid depending directly on the ENTIRE runtime just to get the encoding of `Dispatchable`s.
+///
+/// All entries here (like pretty much in the entire file) must be kept in sync with
+/// `BridgeHubWococo` `construct_runtime`, so that we maintain SCALE-compatibility.
+///
+/// TODO:check-parameter -> change bridge-hub-rococo-wococo when merged to master in cumulus
```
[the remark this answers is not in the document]
bkontur: @serban300 just remove that TODO pls., on my way I left lots of TODOs for final check, but Ro/Wo stuff is deployed, so we have this under control,
and also I would like to set "version guard": here https://github.com/paritytech/parity-bridges-common/issues/1675

PR #1753, opened 2023-01-10
Use indirect runtime calls for RialtoParachain

### LINKED ISSUE #1745: Consider using indirect runtime calls for `RialtoParachain`
At the moment we use direct runtime calls for all the runtimes that are defined in this repo. The problem with this is that we don't have any way to test that the indirect runtime calls logic works within our test deployments. So for example if there is any issue with the indirect runtime logic calls we will find out only after we deploy it to rococo/wococo.

If we used the indirect call logic for one of the runtimes that are defined in this repo, we would catch many potential issues as part of our test deployments. For example I think it would be the easiest to use this for `RialtoParachain`.

Context:
By direct runtime call I'm referring to using the runtime calls defined by `construct_runtime!()`. We can do this only for the runtimes defined in our repo. For example [for rialto parachain](https://github.com/paritytech/parity-bridges-common/blob/209cba353ce09a6262f3b518011b794bc5990726/relays/client-rialto-parachain/src/lib.rs#L50). While for runtimes that are not defined in our repo we use "indirect runtime calls", which is are basically a mock of the runtime. For example [bridge hub rococo](https://github.com/paritytech/parity-bridges-common/blob/209cba353ce09a6262f3b518011b794bc5990726/relays/client-bridge-hub-rococo/src/runtime_wrapper.rs#L46).

@svyatonik do you think this would make sense ?

serban300:
Fixes #1745


bkontur:
@serban300 
if we use now indirect calls for rialto-parachain, wouldnt it possible to remove dependency?

`relays/bin-substrate/Cargo.toml`
```
rialto-parachain-runtime = { path = "../../bin/rialto-parachain/runtime" }
```

I think (maybe later) we should think to remove dependencies for other runtimes (rialto, millau) from `substrate-relay` or think about some configured builds/features maybe like:
```
#[cfg(feature = "rialto-millau")]
or
#[cfg(feature = "rococo-wococo")]
or 
#[cfg(feature = "kusama-polkadot")]
```

for example: production-like Kusama/Polkadot substrate-relay binary does not need all rialto/millau/rococo dependecies and stuff

primitives/runtime/src/calls.rs
```diff
@@ -28,3 +30,21 @@ pub enum SystemCall {
 	#[codec(index = 1)]
 	remark(Vec<u8>),
 }
+
+/// A minimized version of `pallet-sudo::Call` that can be used without a runtime.
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone, TypeInfo)]
+#[allow(non_camel_case_types)]
+pub enum SudoCall<Call> {
+	/// `pallet-sudo::Call::sudo`
+	#[codec(index = 0)]
```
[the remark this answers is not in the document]
serban300: Yes, `0` is the right value. Functionally it's not useful. `0` would be used by default here. But I would keep it, just to have it explicitly specified.

primitives/runtime/src/calls.rs
```diff
@@ -28,3 +30,21 @@ pub enum SystemCall {
...
+	sudo(Box<Call>),
+}
+
+/// A minimized version of `pallet-xcm::Call`, that can be used without a runtime.
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone, TypeInfo)]
+#[allow(non_camel_case_types)]
+pub enum XcmCall {
```
bkontur, line 46: I think `primites/runtime` is more like support for building runtimes,
and this XcmCall/SudoCall is used by relay level stuff,

what about moving it from here under some `relay` directory/package?
probably, this is related to `relay/client-*` layer

serban300: Yes, I wasn't sure what would be the best place where to define these calls. Moved them, to `client-substrate`.

serban300:
> @serban300 if we use now indirect calls for rialto-parachain, wouldnt it possible to remove dependency?
> 
> `relays/bin-substrate/Cargo.toml`
> 
> ```
> rialto-parachain-runtime = { path = "../../bin/rialto-parachain/runtime" }
> ```
> 
> for example: production-like Kusama/Polkadot substrate-relay binary does not need all rialto/millau/rococo dependecies and stuff

Nice ! Done. And the cool effect of this is that now the build time for the substrate-relay (on my laptop) is down from ~8m30s to ~5m10 .

> I think (maybe later) we should think to remove dependencies for other runtimes (rialto, millau) from `substrate-relay` or think about some configured builds/features maybe like:
> 
> ```
> #[cfg(feature = "rialto-millau")]
> or
> #[cfg(feature = "rococo-wococo")]
> or 
> #[cfg(feature = "kusama-polkadot")]
> ```

Yes, I think it would make sense to have this. I will create an issue.

primitives/runtime/src/calls.rs
bkontur: very nice now :)

and also that issue with that "version guard", which means you will need to add `RuntimeVersion` cfg for BridgeHubs, those also should go the the `relay/client-*` level as we discussed "somewhere" with Slava

bkontur:
> > @serban300 if we use now indirect calls for rialto-parachain, wouldnt it possible to remove dependency?
> > `relays/bin-substrate/Cargo.toml`
> > ```
> > rialto-parachain-runtime = { path = "../../bin/rialto-parachain/runtime" }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > for example: production-like Kusama/Polkadot substrate-relay binary does not need all rialto/millau/rococo dependecies and stuff
> 
> Nice ! Done. And the cool effect of this is that now the build time for the substrate-relay (on my laptop) is down from ~8m30s to ~5m10 because the runtime for RialtoParachain is not built together with the relay anymore.
> 
> > I think (maybe later) we should think to remove dependencies for other runtimes (rialto, millau) from `substrate-relay` or think about some configured builds/features maybe like:
> > ```
> > #[cfg(feature = "rialto-millau")]
> > or
> > #[cfg(feature = "rococo-wococo")]
> > or 
> > #[cfg(feature = "kusama-polkadot")]
> > ```
> 
> Yes, I think it would make sense to have this. I will create an issue.



wow, very cool, if we could remove also deps for `rialto-runtime` and `millau-runtime` from `substrate-relay`, maybe it will be ~1m :D

bkontur (approved):
lgtm

serban300:
> wow, very cool, if we could remove also deps for rialto-runtime and millau-runtime from substrate-relay, maybe it will be ~1m :D

Yes :) I'm not sure if we would lose anything if we used indirect calls for rialto and millau as well. But for the compilation time it would be very good.

PR #1756, opened 2023-01-11
Set `R/WococoBridgeHub` bundle runtime version

serban300:
Part of #1675

### DIFF
--- a/relays/bin-substrate/src/chains/wococo.rs
+++ b/relays/bin-substrate/src/chains/wococo.rs
@@ -26,5 +26,6 @@ impl CliChain for Wococo {
 }
 
 impl CliChain for BridgeHubWococo {
-	const RUNTIME_VERSION: Option<SimpleRuntimeVersion> = None;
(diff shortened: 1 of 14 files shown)


svyatonik (approved):

relays/bin-substrate/src/chains/wococo.rs
```diff
@@ -18,13 +18,14 @@
 
 use crate::cli::CliChain;
 use relay_bridge_hub_wococo_client::BridgeHubWococo;
+use relay_substrate_client::SimpleRuntimeVersion;
 use relay_wococo_client::Wococo;
-use sp_version::RuntimeVersion;
```
bkontur, line 22: @serban300 
I guess, we can also remove dependency `sp-version` at least from `bin-substrate` and maybe from other modules,
could you check it please?

bkontur (commented):
nice :)

relays/bin-substrate/src/chains/wococo.rs
serban300: You're right. Opened #1758 for this.

PR #1765, opened 2023-01-12
increase MAX_UNREWARDED_RELAYERS_IN_CONFIRMATION_TX and MAX_UNCONFIRMED_MESSAGES_IN_CONFIRMATION_TX for RBH/WBH

### LINKED ISSUE #1757: Increase MAX_UNREWARDED_RELAYERS_IN_CONFIRMATION_TX constant in RBH/WBH
Related to https://github.com/paritytech/srlabs_findings/issues/49

For that we need to:
- [x] check why weights of `receive_delivery_proof_for_two_messages_by_two_relayers` and `receive_delivery_proof_for_two_messages_by_single_relayer` are currently the same (almost). My guess that there must be at least 1 db read and 1 db write per relayer;
- [x] increase `MAX_UNREWARDED_RELAYERS_IN_CONFIRMATION_TX`, respecting weights from previous point and inbound lane state size.

svyatonik:
closes #1757

I've left a big code comment as a justification. There's also a test `ensure_able_to_receive_confirmation` (which is now missing from the RBH/WBH code - see https://github.com/paritytech/cumulus/pull/1975#issuecomment-1380116501) that will detect if I was wrong or if something will change.

### DIFF
--- a/primitives/chain-bridge-hub-cumulus/src/lib.rs
+++ b/primitives/chain-bridge-hub-cumulus/src/lib.rs
@@ -85,7 +85,7 @@ pub type AccountSigner = MultiSigner;
 /// The address format for describing accounts.
 pub type Address = MultiAddress<AccountId, ()>;
 
-// Not about selecting values of two following constants:
+// Note about selecting values of two following constants:
 //
 // Normal transactions have limit of 75% of 1/2 second weight for Cumulus parachains. Let's keep
 // some reserve for the rest of stuff there => let's select values that fit in 50% of maximal limit.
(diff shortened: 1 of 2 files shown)


primitives/chain-bridge-hub-cumulus/src/lib.rs
```diff
@@ -85,3 +84,32 @@ pub type AccountSigner = MultiSigner;
 
 /// The address format for describing accounts.
 pub type Address = MultiAddress<AccountId, ()>;
+
+// Not about selecting values of two following constants:
```
serban300, line 88: Nit: Do you mean `Note about ...` ?

serban300 (approved):

acatangiu (approved):

Issue #1672, opened 2022-12-01
[BridgeHub] Add and run Basic unit tests 


EmmanuellNorbertTulbure:
For Rococo - Wococo

- [ ] add integrity tests to the bridge-hub-rococo runtime (for both sides R->W and W->R)
```
     fn ensure_bridge_integrity() {
		assert_complete_bridge_types!(
```
- [ ] ...

serban300:
Leaving the integrity test for Rialto -> Millau as an example: https://github.com/paritytech/parity-bridges-common/blob/ec6bafaf0190539644291fc544563c34619647e7/bin/rialto/runtime/src/millau_messages.rs#L206

serban300:
Started adding integrity tests for Rococo -> Wococo and Wococo -> Rococo. There are 2 issues so far:

1. [Checking the type of the InboundPayload](https://github.com/paritytech/parity-bridges-common/blob/8649d12afb92863ed4cda1b13114315a46fa4c1a/bin/runtime-common/src/integrity.rs#L92). The tests expects a [`FromBridgedChainMessagePayload`](https://github.com/paritytech/parity-bridges-common/blob/8649d12afb92863ed4cda1b13114315a46fa4c1a/bin/runtime-common/src/messages.rs#L377), but we use [`XcmAsPlainPayload`](https://github.com/paritytech/cumulus/blob/d42acfdd35db3500cb51945b882795d1d71e1f33/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/bridge_common_config.rs#L31) for [`BridgeHubRococo` and `BridgeHubWococo`](https://github.com/paritytech/cumulus/blob/d42acfdd35db3500cb51945b882795d1d71e1f33/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs#L530) 

@bkontur @svyatonik do we plan to use `FromBridgedChainMessagePayload` here in the future ? 

[...]

PR #1766, opened 2023-01-12
Define method for checking message lane weights

serban300:
Related to #1672

### DIFF
--- a/bin/runtime-common/src/integrity.rs
+++ b/bin/runtime-common/src/integrity.rs
@@ -292,15 +292,15 @@ where
 
 /// Check that the message lane weights are correct.
 pub fn check_message_lane_weights<C: Chain, T: frame_system::Config>(
-	extra_storage_proof_size: u32,
-	max_unrewarded_relayers: MessageNonce,
-	max_unconfirmed_messages: MessageNonce,
+	bridged_chain_extra_storage_proof_size: u32,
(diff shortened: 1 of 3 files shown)


bin/runtime-common/src/integrity.rs
```diff
@@ -289,3 +289,32 @@ where
 	assert_bridge_messages_pallet_constants::<R, MI>(params.messages_pallet_constants);
 	assert_bridge_pallet_names::<B, R, GI, MI>(params.pallet_names);
 }
+
+pub fn check_message_lane_weights<C: Chain, T: frame_system::Config>(
```
svyatonik, line 293: Please leave some doc for public item :)

svyatonik (approved):
Perfect, thank you!

bin/runtime-common/src/integrity.rs
```diff
@@ -289,3 +289,32 @@ where
...
+	extra_storage_proof_size: u32,
```
svyatonik, line 294: Also maybe give some prefixes to arguments - like `bridged_` or `this_`. Because e.g. `extra_storage_proof_size` is for bridged chain and other args are about this chain. WDYT?

serban300: Good point. Done.

PR #1770, opened 2023-01-16
docs: add Security Policy doc

acatangiu:


serban300 (commented):
Looks good, but I see that it's different than the security docs in [substrate](https://github.com/paritytech/substrate/blob/master/docs/SECURITY.md) and [polkadot](https://github.com/paritytech/polkadot/blob/master/SECURITY.md) repos. I was just wondering if we should use something closer to those ?

acatangiu:
> Looks good, but I see that it's different than the security docs in [substrate](https://github.com/paritytech/substrate/blob/master/docs/SECURITY.md) and [polkadot](https://github.com/paritytech/polkadot/blob/master/SECURITY.md) repos. I was just wondering if we should use something closer to those ?

This is the updated version agreed upon with security team. The substrate and polkadot docs will also be updated soon.

PR #1776, opened 2023-01-17
Define a separate signed extension for BHR/BHW

### LINKED ISSUE #1748: Add the `BridgeRejectObsoleteHeadersAndMessages` `SignedExtra` for BridgeHubRococo/Wococo
We added the bridge_reject_obsolete_headers_and_messages extension to BridgeHubRococo/BridgeHubWococo, but in the relayer we use [this](https://github.com/paritytech/parity-bridges-common/blob/e575269e5f42624ea2e144974933d0dcc2a7eb05/primitives/polkadot-core/src/lib.rs#L205) which doesn't include it. 

We should add the `BridgeRejectObsoleteHeadersAndMessages` `SignedExtra` for BridgeHubRococo/Wococo

serban300:
Fixes #1748 

- Make signed extensions for indirect runtimes more extensible
- Define a separate signed extension for BHR/BHW

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -997,6 +997,7 @@ dependencies = [
  "frame-system",
  "hash-db",
  "hex-literal",
+ "impl-trait-for-tuples",
  "num-traits",
  "parity-scale-codec",
  "scale-info",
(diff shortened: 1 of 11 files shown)


svyatonik (approved):
:+1:

PR #1780, opened 2023-01-18
Add integrity check for signed extensions

### LINKED ISSUE #1768: Ensure indirect runtime signed extra is correct
Related to #1748 but I think it's better to have a separate issue.

It would be nice if we could add an integrity test to check that the `SignedExtension` defined in the actual runtime is compatible with the one defined in the bridge's indirect runtime. We should check that `AdditionalSigned` and the constructor parameters are the same / compatible.

serban300:
Fixes #1768


svyatonik (approved):
Cool, thank you!

Issue #1805, opened 2023-01-20
Use wss on test deployments


svyatonik:
It seems that the #1785 happens only if we connect to RPC node using `wss`, not the `ws`. We need to change our test deployments to use `wss` to catch such issues before they break our other bridges.

Issue #1785, opened 2023-01-19
Rococo <> Wococo bridge hub relay disconnects from our RBH node


svyatonik:
Some logs:
```
2023-01-19 14:46:49 | [Wococo_to_BridgeHubRococo_Parachains] 2023-01-19 11:46:49 +00 WARN bridge Wococo client has failed to return its sync status: RpcError(RestartNeeded("Networking or low-level protocol error: WebSocket connection error: connection closed")) |  
-- | -- | --
  |   | 2023-01-19 14:46:48 | [Rococo_to_BridgeHubWococo_Parachains] 2023-01-19 11:46:48 +00 WARN bridge Rococo client has failed to return its sync status: RpcError(RestartNeeded("Networking or low-level protocol error: WebSocket connection error: connection closed")) |  
  |   | 2023-01-19 14:46:44 | [Wococo_to_BridgeHubRococo_Sync] 2023-01-19 11:46:44 +00 INFO bridge Synced 4030136 of 4030136 headers |  
  |   | 2023-01-19 14:46:43 | 2023-01-19 11:46:43 +00 WARN rustls::check Received a ServerHelloDone handshake message while expecting [CertificateRequest] |  
...
  |   | 2023-01-19 14:46:42 | 2023-01-19 11:46:42 +00 WARN rustls::check Received a ServerHelloDone handshake message while expecting [CertificateRequest]
```

svyatonik:
Seems like `Received a ServerHelloDone handshake message while expecting [CertificateRequest]` isn't a problem. And a real problem is the "connection closed" thing. The RPC node we connect to has following in its logs:
```
2023-01-19 16:48:06 | 2023-01-19 13:48:06 ✨ Imported #4031561 (0xc632…9264) |  
-- | -- | --
  |   | 2023-01-19 16:48:04 | unexpected end of file) |  
  |   | 2023-01-19 16:48:04 | Caused by: |  
  |   | 2023-01-19 16:48:04 |   |  
  |   | 2023-01-19 16:48:04 | 2023-01-19 13:48:04 Rejected connection: Transport(i/o error: unexpected end of file |  
  |   | 2023-01-19 16:48:04 | 2023-01-19 13:48:04 Accepting new connection 8/100
```

Is that the RPC connection? Why there's unexpected end of file (some decoding fails?)?

svyatonik:
When it's all started (relay logs):
```
[...]

PR #1809, opened 2023-01-20
Use wss to connect to the Rialto node in test deployments

svyatonik:
prototype of #1805 
related to #1785 

Millau<>Rialto relay now connects to Rialto node using wss. I've also enabled some additional logs for relay and for the wss RPC node. We'll revert this PR after problem is solved.


svyatonik:
> Can this be ran locally instead of merging it into master and then reverting ?

It took me ~4 hours before I got this issue, so I'd prefer to test it on dedicated machine

serban300 (approved):

PR #1816, opened 2023-01-24
Remove InboundPayload check from integrity tests

serban300:
I would remove this check if the InboundPayload doesn't have to be `FromBridgedChainMessagePayload`.


svyatonik (approved):
In general it shouldn't be (from pallets PoV). It's just to check that the runtime is compatible with `bridge-runtime-common` expectations. We may drop it now until we'll fully migrate to Brabislav structs

PR #1817, opened 2023-01-25
Functions to benchmark messages pallet with linked to parachain

svyatonik:
Required for https://github.com/paritytech/cumulus/pull/2107

All current helpers are inserting headers to the `pallet_bridge_grandpa`, which isn't used by messages pallet when it is bridged with the parachain. Also some cleanup


bkontur (approved):
nice, removed lots of trait bound :)

PR #1823, opened 2023-01-27
Use named parameters for indirect calls

serban300:
We will need this in order to use the following macros with the calls in the autogenerated runtimes:

`generate_mocked_submit_finality_proof_call_builder`
`generate_mocked_receive_message_proof_call_builder`
`generate_mocked_receive_message_delivery_proof_call_builder`

### DIFF
--- a/modules/grandpa/src/lib.rs
+++ b/modules/grandpa/src/lib.rs
@@ -1178,7 +1178,7 @@ mod tests {
 
 		let direct_initialize_call =
 			Call::<TestRuntime>::initialize { init_data: init_data.clone() };
-		let indirect_initialize_call = BridgeGrandpaCall::<TestHeader>::initialize(init_data);
+		let indirect_initialize_call = BridgeGrandpaCall::<TestHeader>::initialize { init_data };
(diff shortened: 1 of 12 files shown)


svyatonik (approved):

bkontur:
@serban300 @svyatonik 
I know this is not important, but when I see the word "mock/mocked", it feels like "something about testing",
what about renaming those macros to something little but more descriptive like:
`generate_submit_finality_proof_call_builder`
`decl_submit_finality_proof_call_builder`
`submit_finality_proof_call_builder`
`submit_finality_proof_autogenerated_call_builder`
`...`
I dont know, wdyt?

serban300:
> @serban300 @svyatonik I know this is not important, but when I see the word "mock/mocked", it feels like "something about testing", what about renaming those macros to something little but more descriptive like: `generate_submit_finality_proof_call_builder` `decl_submit_finality_proof_call_builder` `submit_finality_proof_call_builder` `submit_finality_proof_autogenerated_call_builder` `...` I dont know, wdyt?

Totally agree. I will open a PR for chaning the names of these macros.

PR #1829, opened 2023-01-30
Use specific error for case when para head is missing from the bridge pallet

svyatonik:
It is strange to see `BridgePalletIsNotInitialized` in relay logs, even though the parachains pallet do not have this "initialization" step.

### DIFF
--- a/relays/lib-substrate-relay/src/on_demand/parachains.rs
+++ b/relays/lib-substrate-relay/src/on_demand/parachains.rs
@@ -473,12 +473,12 @@ where
 		P::SourceParachain,
 	>(target.client(), best_target_block_hash)
 	.await;
-	// if there are no parachain heads at the target (`BridgePalletIsNotInitialized`), we'll need
-	// to submit at least one. Otherwise the pallet will be treated as uninitialized and messages
+	// if there are no parachain heads at the target (`NoParachainHeadAtTarget`), we'll need to
+	// submit at least one. Otherwise the pallet will be treated as uninitialized and messages
 	// sync will stall.
 	let para_header_at_target = match para_header_at_target {
 		Ok(para_header_at_target) => Some(para_header_at_target.0),
-		Err(SubstrateError::BridgePalletIsNotInitialized) => None,
+		Err(SubstrateError::NoParachainHeadAtTarget(_, _)) => None,
 		Err(e) => return Err(map_target_err(e)),
 	};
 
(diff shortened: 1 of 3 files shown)


relays/lib-substrate-relay/src/on_demand/parachains.rs
```diff
@@ -473,12 +473,12 @@ where
...
 	.await;
-	// if there are no parachain heads at the target (`BridgePalletIsNotInitialized`), we'll need
-	// to submit at least one. Otherwise the pallet will be treated as uninitialized and messages
+	// if there are no parachain heads at the target (`NoParachainHeadAtTarget`), we'll need to
+	// submit at least one. Otherwise the pallet will be treated as uninitialized and messages
 	// sync will stall.
 	let para_header_at_target = match para_header_at_target {
 		Ok(para_header_at_target) => Some(para_header_at_target.0),
-		Err(SubstrateError::BridgePalletIsNotInitialized) => None,
+		Err(SubstrateError::NoParachainHeadAtTarget(_, _)) => None,
```
bkontur, line 481: @svyatonik 
`para_header_at_target` is returned from `best_finalized_peer_header_at_self`
so what about `.ok_or(SubstrateError::BridgePalletIsNotInitialized)`  here?
```
/// Reads best `PeerChain` header known to the `SelfChain` using provided runtime API method.
///
/// Method is supposed to be the `<PeerChain>FinalityApi::best_finalized()` method.
pub async fn best_finalized_peer_header_at_self<SelfChain, PeerChain>(
	self_client: &Client<SelfChain>,
	at_self_hash: HashOf<SelfChain>,
) -> Result<HeaderIdOf<PeerChain>, SubstrateError>
where
	SelfChain: Chain,
	PeerChain: Chain,
{
	// now let's read id of best finalized peer header at our best finalized block
	self_client
		.typed_state_call::<_, Option<_>>(
			PeerChain::BEST_FINALIZED_HEADER_ID_METHOD.into(),
			(),
			Some(at_self_hash),
		)
		.await?
		.ok_or(SubstrateError::BridgePalletIsNotInitialized)
}
```

svyatonik: This code works with both standalone and parachains, so we can't use parachain-specific error there

svyatonik: NVM, I see the issue now, thanks

bkontur (approved):

PR #1834, opened 2023-01-31
Fix on demand parachains relay when no parachain head at target

svyatonik:
follow up on #1829 (actually I was thinking that the #1829) would fix the issue, but it has not

The main issue is that the messages relay needs to read "client state" first to understand if he wants to deliver something or not. But if there were no parachain head in the bridge parachains pallet at the target chain, it was treated as an error => the messages relay was unable to decide that it needs some parachain header to be relayed => it wasn't relayed "normally".

### DIFF
--- a/relays/lib-substrate-relay/src/on_demand/parachains.rs
+++ b/relays/lib-substrate-relay/src/on_demand/parachains.rs
@@ -422,11 +422,14 @@ struct RelayData<ParaHash, ParaNumber, RelayNumber> {
 	pub para_header_at_source: Option<HeaderId<ParaHash, ParaNumber>>,
 	/// Parachain header, that is available at the source relay chain at `relay_header_at_target`
 	/// block.
+	///
+	/// May be `None` if there's no `relay_header_at_target` yet, or if the
+	/// `relay_header_at_target` is too old and we think its state has been pruned.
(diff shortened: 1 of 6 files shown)


relays/lib-substrate-relay/src/on_demand/parachains.rs
```diff
@@ -502,51 +504,70 @@ where
...
 			// The required relay header hasn't yet been relayed. Ask / wait for it.
 			return state
 		}
 
 		// We may switch to `RelayingParaHeader` if parachain head is available.
-		state = data
-			.para_header_at_relay_header_at_target
-			.clone()
-			.map_or(RelayState::Idle, RelayState::RelayingParaHeader);
+		if let Some(para_header_at_relay_header_at_target) =
```
bkontur, line 564: very nit, what about do clone on return?
I dont know, if this is even worth optimization :)

```
        if let Some(ref para_header_at_relay_header_at_target) = data.para_header_at_relay_header_at_target
		{
			return RelayState::RelayingParaHeader(para_header_at_relay_header_at_target.clone())
		}
```

bkontur (approved):
I am not very familiar yet with those races/loops, but lgtm

Issue #1543, opened 2022-08-08
Migrate to weights V2


svyatonik:
We'll need to migrate to this someday (my guess - before Kusama<>Polkadot bridge hub? or even for Rococo<>Wococo bridge hub?).

Right now we can't do anything because: (1) the weights PR is not merged yet (2) we are still referencing XCM v3 branch from our repo - so even if weights will be merged, we can't migrate until XCM is also merged.

https://github.com/paritytech/substrate/pull/10918
https://github.com/paritytech/substrate/issues/12176

svyatonik:
There's also a limit on max POV size for parachains, which is `cumulus_primitives_core::relay_chain::v2::MAX_POV_SIZE` (`5 * 1024 * 1024`). We need to be sure that our size doesn't exceed that (verify in tests?)

bkontur:
- [ ] search in code
      ```TODO: https://github.com/paritytech/parity-bridges-common/issues/1543 - check `set_proof_size` 0 or 64*1024 or 1026?```

svyatonik:
Remaining things:
- [ ] remove hack with changing max number of GRANDPA authorities in benchmarks - let's always use `MaxBridgedAuthorities`;
- [ ] we need to respect PoV size when building message delivery transaction - see that huge `proof_size` in `select_delivery_transaction_limits_works` test;
- [ ] ensure that we are rejecting GRANDPA proofs with too-many authorities and decrease maximal number of authorities for Kusama/Polkadot/Rococo/Wococo from `100_000` (const from GRANDPA pallet config) to something much smaller;
- [ ] refund (if that is possible) unused `proof_size` in all calls where it is possible;
- [ ] maybe use `PovEstimationMode::Estimated` mode where appropriate?

svyatonik:
I'm not sure about the last point (using `PovEstimationMode::Estimated`). E.g. if we rely on call arguments to provide this `proof_size`, then what if actual `proof_size` will be larger? We may fail the call by returning error, but the proof size will be larger than "estimated" anyway. Let's keep using `MaxEncodedLen` as it seems safer. But we need to impl #1853 anyway

svyatonik:
Some TODOs are still left in code

PR #1850, opened 2023-02-06
Update Substrate/Polkadot/Cumulus dependencies + weights v2

svyatonik:
related to #1543 

Important things here:
- I found that we now (after #1812) have duplicate Substrate crates references in top-level `Cargo.lock`. Thought that it maybe an issue/typo, but actually it isn't. Some crates are from the `monthly-2022-07` (used by `runtime-codegen`) and the rest (used by the rest of bridge stuff) comes from Substrate `master`. IMO ideally we need to have two `Cargo.lock`s then - e.g. we already have additional `Cargo.lock` in `fuzz/storage-proof`. WDYT?
- all our benchmarks are using `MaxEncodedLen` to compute their proof size (the other weight component). I'm not closing #1543 because of that - IMO we may try to use `Estimated` in some places. And we definitely need to do refund if the actual read size is less than `MaxEncodedLen`. But I haven't the impact of the PoV size on the transaction fee anywhere yet.

Overall, PoV size of our pallets **on testnets** seems ok. The limit for Cumulus-based pallets is `5Mb`. And max estimated sizes for our pallets are:
- `46001` for GRANDPA finality proof submit. That's strange, because according to generated comments, the most comes from the `CurrentAuthoritySet` and the max number of authorities is set to `5` for Rialto/Millau. So what gonna happen when we'll switch to `100_000` for Polkadot/Kusama/Rococo/Wococo? It needs to be investigated yet. UPD: there's a hack that changes max number of authorities for benchmarks, so this value is for `1024` authorities. In any case - this hack must be removed imo + we need to think how to deal with `100_000` authorities from GRANDPA pallet configuration (estimated PoV size will be around `46001 * 1000 ~ 4.5Mb`). Maybe we want to use different constants in our config and reject proofs with larger number of authorities;
- `55198` for message delivery. Most comes from the `InboundLanes` map and this size is limited by two consts - `MaxUnrewardedRelayerEntriesAtInboundLane` and `MaxUnconfirmedMessagesAtInboundLane`. So we can play with it.

In any case - this PR looks self-sufficient, I'll keep investigating/fixing things in follow-up PRs


serban300:
> I found that we now (after https://github.com/paritytech/parity-bridges-common/pull/1812) have duplicate Substrate crates references in top-level Cargo.lock. Thought that it maybe an issue/typo, but actually it isn't. Some crates are from the monthly-2022-07 (used by runtime-codegen) and the rest (used by the rest of bridge stuff) comes from Substrate master. IMO ideally we need to have two Cargo.locks then - e.g. we already have additional Cargo.lock in fuzz/storage-proof. WDYT?

Thank you for taking a look on this ! Yes, sounds like the best solution. I'll try to add a different Cargo.lock for the codegen tool.

bkontur (approved):

acatangiu (commented):
> In any case - this hack must be removed imo + we need to think how to deal with 100_000 authorities from GRANDPA pallet configuration (estimated PoV size will be around 46001 * 1000 ~ 4.5Mb).

I am not worried while we're fine with `1000`. I expect moving bridges to beefy-finality _before_ Kusama or Polkadot get more than 1000 validators.

PR #1851, opened 2023-02-06
Fix `HeadersToKeep` and `MaxBridgedAuthorities` in Millau benchmarks

svyatonik:
1st point from https://github.com/paritytech/parity-bridges-common/issues/1543#issuecomment-1418814066

### DIFF
--- a/modules/grandpa/src/benchmarking.rs
+++ b/modules/grandpa/src/benchmarking.rs
@@ -53,25 +53,28 @@ use sp_finality_grandpa::AuthorityId;
 use sp_runtime::traits::Zero;
 use sp_std::vec::Vec;
 
-// The maximum number of vote ancestries to include in a justification.
-//
-// In practice this would be limited by the session length (number of blocks a single authority set
-// can produce) of a given chain.
+/// The maximum number of vote ancestries to include in a justification.
+///
+/// In practice this would be limited by the session length (number of blocks a single authority set
+/// can produce) of a given chain.
 const MAX_VOTE_ANCESTRIES: u32 = 1000;
 
-// The maximum number of pre-commits to include in a justification. In practice this scales with the
-// number of validators.
-pub const MAX_VALIDATOR_SET_SIZE: u32 = 1024;
-
-// `1..MAX_VALIDATOR_SET_SIZE` and `1..MAX_VOTE_ANCESTRIES` are too large && benchmarks are
-// running for almost 40m (steps=50, repeat=20) on a decent laptop, which is too much. Since
-// we're building linear function here, let's just select some limited subrange for benchmarking.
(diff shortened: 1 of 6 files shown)


modules/grandpa/src/benchmarking.rs
```diff
@@ -53,26 +53,29 @@ use sp_finality_grandpa::AuthorityId;
...
 	MAX_VOTE_ANCESTRIES_RANGE_BEGIN + MAX_VOTE_ANCESTRIES_RANGE_BEGIN;
 
+// the same with validators - if there are too much validators, let's run benchmarks on subrange
+fn validator_set_range_end<T: Config<I>, I: 'static>() -> u32 {
+	let max_bridged_authorities = T::MaxBridgedAuthorities::get();
+	if max_bridged_authorities > 128 {
+		max_bridged_authorities / 5
+	} else {
+		max_bridged_authorities
+	}
```
serban300, line 76: Would it make sense to do something like

```
	if max_bridged_authorities > 128 {
		std::cmp::max(128, max_bridged_authorities / 5)
	}
```

?

modules/grandpa/src/lib.rs
```diff
@@ -464,6 +467,20 @@ pub mod pallet {
...
+		T::HeadersToKeep::get()
+	}
+
+	/// Return number of headers to keep in the runtime storage.
+	#[cfg(feature = "runtime-benchmarks")]
+	pub(crate) fn headers_to_keep<T: Config<I>, I: 'static>() -> u32 {
+		// every db operation (significantly) slows down benchmarks, so let's keep as min as
+		// possible
+		2
+	}
```
serban300, line 482: Nit: Could we do this distinction in the mock runtime instead ? Is the mock runtime used for benchmarks ?

serban300 (approved):

modules/grandpa/src/benchmarking.rs
svyatonik: Yeah, seems like a good idea

modules/grandpa/src/lib.rs
svyatonik: No, benchmarks are using the actual runtimes. That's why the weights of our pallets may be (are) different on Rialto, on Millau and on WBH/RBH. So every runtime generates its own `weights.rs`  file (e.g. for `frame-system` ; https://github.com/paritytech/polkadot/blob/master/runtime/rococo/src/weights/frame_system.rs, https://github.com/paritytech/substrate/blob/master/frame/system/src/weights.rs, https://github.com/paritytech/cumulus/blob/master/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/weights/frame_system.rs)

acatangiu (approved):

PR #1854, opened 2023-02-06
Use separate Cargo.lock for `tools/runtime-codegen`

serban300:


svyatonik:
It seems that all `monthly-2022-07` are still in the top-level `Cargo.lock`. Maybe it works in other way than I thought :) Removing it from top-level toml doesn't help too:
```
svyatonik@covid:~/dev/parity-bridges-common$ cargo run -p runtime-codegen --manifest-path tools/runtime-codegen/Cargo.toml
error: current package believes it's in a workspace when it's not:
current:   /home/svyatonik/dev/parity-bridges-common/tools/runtime-codegen/Cargo.toml
workspace: /home/svyatonik/dev/parity-bridges-common/Cargo.toml

this may be fixable by adding `tools/runtime-codegen` to the `workspace.members` array of the manifest located at: /home/svyatonik/dev/parity-bridges-common/Cargo.toml
Alternatively, to keep it out of the workspace, add the package to the `workspace.exclude` array, or add an empty `[workspace]` table to the package's manifest.
```

So the only way may be to add this `[workspace]` to the utility toml. But maybe it is too much - feel free to ignore :) And sorry that I took your time

serban300:
> monthly-2022-07

Yes, the `?tag=monthly-2022-07` references remain. 

We could have a separate workspace for this tool. I don't know how it would be best.

acatangiu (commented):
I am leaning towards adding `[workspace]` rather than having polluted top-level `Cargo.lock`

serban300:
> I am leaning towards adding `[workspace]` rather than having polluted top-level `Cargo.lock`

Done.

svyatonik (approved):

PR #1859, opened 2023-02-07
Message delivery transaction is not free!!!

svyatonik:
Must be trace of one of experiments that I did in #1652. But in any way - the relayer shall pay fee for message delivery transaction. Rewards are paid from the sovereign accounts, not by the "chain" itself.

### DIFF
--- a/modules/messages/src/lib.rs
+++ b/modules/messages/src/lib.rs
@@ -899,6 +899,7 @@ mod tests {
 	use bp_test_utils::generate_owned_bridge_module_tests;
 	use frame_support::{
 		assert_noop, assert_ok,
+		dispatch::Pays,
 		storage::generator::{StorageMap, StorageValue},
 		traits::Hooks,
 		weights::Weight,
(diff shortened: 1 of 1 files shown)


bkontur (approved):

modules/messages/src/lib.rs
```diff
@@ -398,7 +398,7 @@ pub mod pallet {
 
 			Self::deposit_event(Event::MessagesReceived(messages_received_status));
 
-			Ok(PostDispatchInfo { actual_weight: Some(actual_weight), pays_fee: Pays::No })
+			Ok(PostDispatchInfo { actual_weight: Some(actual_weight), pays_fee: Pays::Yes })
```
bkontur, line 401: @svyatonik 
just thinking, is it worth to have some tests for extrinsics what is free and what not? on pallet level or runtime level?

svyatonik: Yeah,, let's add one here. Actually - messages transactions were never intended to be free, so it's just a typo. But in anyway - let's do this

serban300 (approved):

Issue #1853, opened 2023-02-06
Weights v2: optimizations and fixes


svyatonik:
Extracted from https://github.com/paritytech/parity-bridges-common/issues/1543#issuecomment-1418814066 - we need (?) to return actual `proof_size` from some of our calls. Especially - from calls that are declaring a large `proof_size` component - i.e. GRANDPA finality proof submit and message delivery.

Right now there are open questions:
1) it shall affect the transaction payment iiuc. But I can't find anything related to `proof_size` in `pallet-transaction-payment`;
2) how do we need to change this component? Shall we encode the value again and subtract the encoded value length from pre-dispatch-weight?

What has been done and what yet needs to be done:
- [x] add `MaxValues` to maps in `pallet-bridge-parachains` and in `pallet-bridge-grandpa`. By default benchmars assume that there are `1_000_000` entries in the map, so it assumes another `4` merkle trie nodes in the storage proof. Since our maps are limited, we are able to reduce this number;
- [x] we now refund unused `proof_size` in some `pallet-bridge-parachains` and `pallet-bridge-messages` calls. Actually - in all that we're actually interested (which have large `proof_size` weights);
- [x] check if above optimizations may be applied to our other pallets/calls;
- [x] messages pallet calls have complicated weight formulae. In particular, some are using weight subtractions. What worries me is that we may accidentally subtract `proof_size` where it isn't actually required. It needs to be checked and fixed

svyatonik:
Re 3rd point:
- there's no more things where we can make **significant** refund. So it seems unnecessary;
- we may add `MaxValues` to `2` remaining maps in the messages module:`InboundLanes` and `OutboundMessages`. For outbound lanes it is problematic to give its limit (e.g. on RBH/WBH it is limited to `u64::MAX/2`, which is much larger than 1m). For `InboundLanes` - it depends on #1651.

So I'm gonna close this issue when last of active PRs ^^^ will be closed

PR #1864, opened 2023-02-08
Refund extra proof bytes in message delivery transaction

svyatonik:
related to #1853 

Most of proof bytes in message delivery transaction comes from the inbound lane state. It is limited by the `MaxUnrewardedRelayerEntriesAtInboundLane` configuration constant. Every entry is an account id + two `u64`. For bridge hubs it is 4096, so the max size is `40*4096 = 160Kb`. If only one relayer will be serving our lane, it'll be decreased to (around) `40 bytes`, which is a big gain imo.


modules/messages/src/lib.rs
```diff
@@ -1583,14 +1609,92 @@ mod tests {
...
+				proof.clone(),
+				messages_count,
+				REGULAR_PAYLOAD.declared_weight,
+			)
+			.unwrap()
+			.actual_weight
+			.unwrap();
+			assert_eq!(post_dispatch_weight.proof_size(), pre_dispatch_weight.proof_size());
+
+			// if cound of unrewarded relayer entries is less than maximal, then some `proof_size`
```
acatangiu, line 1662:
```suggestion
			// if count of unrewarded relayer entries is less than maximal, then some `proof_size`
```

modules/messages/src/lib.rs
```diff
@@ -1962,4 +2066,37 @@ mod tests {
...
+		let max_entries = crate::mock::MaxUnrewardedRelayerEntriesAtInboundLane::get() as usize;
+
+		// when we have exactly `MaxUnrewardedRelayerEntriesAtInboundLane` unrewarded relayers
+		assert_eq!(storage(max_entries).extra_proof_size_bytes(), 0);
+
+		// when we have less than `MaxUnrewardedRelayerEntriesAtInboundLane` unrewarded relayers
+		assert_eq!(
+			storage(max_entries - 1).extra_proof_size_bytes(),
+			relayer_entry().encode().len() as u64
+		);
```
acatangiu, line 2096: nit1: I hope `encoded_size_hint()` used in `extra_proof_size_bytes()` is consistent in its return value and this test will be robust - worst case it will intermittently fail and we'll worry about that then.

nit2: I would add another one for extra safety (especially since previous test only verifies generic `post_dispatch_weight.proof_size() < pre_dispatch_weight.proof_size()` without checking/knowing how much smaller it should be.

```suggestion
		assert_eq!(
			storage(max_entries - 1).extra_proof_size_bytes(),
			relayer_entry().encode().len() as u64
		);
		assert_eq!(
			storage(max_entries - 2).extra_proof_size_bytes(),
			2 * relayer_entry().encode().len() as u64
		);
```
(above could also be a loop, but testing at least two values is good enough imo)

acatangiu (approved):

modules/messages/src/lib.rs
svyatonik: > nit1: I hope encoded_size_hint() used in extra_proof_size_bytes() is consistent in its return value and this test will be robust - worst case it will intermittently fail and we'll worry about that then.

I'm not sure exactly what you mean here, but both expressions in the `extra_proof_size_bytes` are using values, returned from the `max_encoded_len`. So iiuc they are somewhat consistent :)

> nit2: I would add another one for extra safety (especially since previous test only verifies generic post_dispatch_weight.proof_size() < pre_dispatch_weight.proof_size() without checking/knowing how much smaller it should be.

Done

Issue #1774, opened 2023-01-17
Ensure that we may remove extra folders from bridges subtree in Cumulus


svyatonik:
We may want to remove unneeded folders from the bridges subtree in Cumulus. We only need pallets, their primitives and `runtime-common`. But e.g. `runtime-common` has references to our testnets, so we can't simply remove testnet runtimes.

We need fix it (remove backreferences) and prepare some script that we may use when updating our subtree in Cumulus that will only leave required folders.

bkontur:
I started with this: https://github.com/paritytech/cumulus/blob/bridge-hub-rococo-wococo/scripts/bridges_update_subtree.sh

svyatonik:
Closed #1832: `./bridges/scripts/verify-pallets-build.sh --no-revert` && then `git commit -a -m "Remove extra folders"`.

bkontur:
`verify-pallets-build.sh` is included in cumulus update subtree notes/process:
https://github.com/paritytech/cumulus/blob/bridge-hub-rococo-wococo/scripts/bridges_update_subtree.sh#L48
https://github.com/paritytech/cumulus/blob/bridge-hub-rococo-wococo/parachains/runtimes/bridge-hubs/README.md#how-to-update-bridges-subtree

I was able to create subtree update with:
```
./scripts/bridges_update_subtree.sh fetch
./scripts/bridges_update_subtree.sh patch (runs `verify-pallets-build.sh`)
./scripts/bridges_update_subtree.sh merge

+ some manual conflict resolution
```

but it is not that straight because `git subtree pull` creates conflicts with previously deleted files, so `./scripts/bridges_update_subtree.sh merge` just deletes them and simplify conflict resolutions, 
but at the end of merge process `./scripts/bridges_update_subtree.sh patch` have to work and it does, so cool :)

anyway, merge conflicts will just depend on subtree diff

bkontur:
@svyatonik 
wdyt about runnning `partial-repo-build-test` on all PRs?

I am just thinking, we should be able to release subtree whenever, so maybe this check should run on all PRs, ideally, it should run, if PR touches relevant files, but that is very very nice to have :)

or is there any way how to trigger `partial-repo-build-test` step manually in CI?

PR #1867, opened 2023-02-09
Run partial-repo-build-test on every PR

svyatonik:
@bkontur That's for your request here: https://github.com/paritytech/parity-bridges-common/issues/1774#issuecomment-1422831767 . My thoughts were:
- it is just a helper to leave minimal set of code in the Cumuls repo;
- if something goes wrong, we may leave additional code in Cumuls repo, until it is fixed;
- running this on every PR slows down CI a bit;
- some PRs may be more important than leaving repo in partial-compilable-state, so we may defer this fix until better times.

But if you think we need it here, let's do that - just press Approve :)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -194,7 +194,7 @@ benchmarks-test:
 partial-repo-build-test:
   stage:                           test
   <<:                              *docker-env
-  <<:                              *nightly-test
+  <<:                              *test-refs
   script:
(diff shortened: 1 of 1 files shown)


bkontur:
@svyatonik 
ok, lets just close this without merge, we dont need it now,
maybe later when we have final release plan/design for pallets/runtimes, then we can come back to this or something similar

PR #1870, opened 2023-02-09
More tests for message pallet weights

svyatonik:
related to #1853

I have some concerns related to `proof_size` component of the weight and our message pallet weight formulae. In particular, we do some weight subtraction there. While this is fine for `ref_time`, it may be wrong for the `proof_size`, because to compute `proof_size`, our benchmarks use the `MaxEncodedLen` approach. So maybe in some formula we subtract weights and because of that are getting `proof_size = 0`. It needs to be checked + better to add runtime-level tests for that.


modules/messages/src/weights_ext.rs
```diff
@@ -122,6 +145,146 @@ pub fn ensure_able_to_receive_confirmation<W: WeightInfoExt>(
...
+		"Messages delivery proof size does not affect values that we read from our storage",
+	);
+	ensure_ref_time_lte(
+		weight_when_proof_size_is_8k,
+		weight_when_proof_size_is_16k,
+		"Larger messages delivery proofs mean more hashing and iterations => computation time",
+	);
+}
+
+/// Panics if `proof_size` of delivery confirmation call depends on the number of confirmed
```
svyatonik, line 224: This isn't true in general - there may be reward schemes, which read some stuff e.g. for every message or so. But our impl doesn't do that && other may just not use our test.

Feel free to argue

svyatonik:
Hmm: https://gitlab.parity.io/parity/mirrors/parity-bridges-common/-/jobs/2373568

serban300 (approved):

acatangiu (approved):

PR #1873, opened 2023-02-12
update messages relay sequence diagram

cuteolaf:


svyatonik (commented):
I don't see why it needs to be changed - that's for demonstrating that there may be lags in finality relay between two chains && we don't need to relay all headers. Do you have some other thoughts about it?

cuteolaf:
@svyatonik Are you on Telegram?
I'd like to have a chat with you.

svyatonik:
Better ask here, but if you really need a chat., you can reach me on [matrix](https://matrix.org/): `@svyatonik:parity.io`

svyatonik:
Please reopen if you have some thoughts about it

Issue #1838, opened 2023-02-01
Improve RefundRelayerForMessagesFromParachain readability


svyatonik:
It is hard to understand what all those generic arguments (see https://github.com/paritytech/parity-bridges-common/blob/33632313dfd588483593b390b31d84ed388e4952/bin/runtime-common/src/refund_relayer_extension.rs#L104) mean - we need document every parameter, or make them self-descriptive.

Issue #1667, opened 2022-11-29
Support multiple refund-message-delivery-cost-at-target-chain signed extensions


svyatonik:
#1657 introduces new signed extension for refunding message relayers at the target chain (`RefundRelayerForMessagesFromParachain`). The issue is that it may only work with single bridge (single set of `pallet_bridge_grandpa::Pallet` + `pallet_bridge_parachains::Pallet` + `pallet_bridge_messages::Pallet`). We need to fix it to support multiple on-chain bridges.

It may be done using approach similar to what @serban300 has done with our other extensions (https://github.com/paritytech/parity-bridges-common/blob/8b506304c879403c94e1992812bd1b5a7dadd9fc/bin/runtime-common/src/lib.rs#L79).

serban300:
Resolved via #1895

Issue #1666, opened 2022-11-28
Make pallets XCM compatible + less configuration overhead


svyatonik:
We've removed a lot of code (#1638) because the architecture has been changed significantly. And now we have some options to make pallets configuration easier. That's a task for the future (after initial deployment), but I'm gonna leave some ideas here:
- [ ] now we have 3 types with almost the same meaning: XCM `MiltuLocation`, `ChainId` and `LaneId`. The idea is that we'll have two bridge hubs, which will be used to build bridges between different parachains. Every bridge now (not at the code, but in plans) uses types in following matter: every parachain (bridge hubs, Statemin?, ...) all have XCM "addresses" (`MultiLocations`), bridge hubs also have `ChainId`s and every bridge will be assigned a single lane (e.g. Statimine <> Statemint will be using lane `0x00000000`). We may remove `ChainId` and `LaneId` types and instead use pair of `MultiLocation`s (or whatever is used to identify chains in global consensus);
- [ ] we may remove configurable proof types from the messaging pallet and move storage proof verification code to the messaging pallet;
- [ ] we may move most of `bridge-runtime-common` code to the messaging pallet and then we'll be able to replace most of configuration options and glue code by adding single `type BridgedChain = SomeBridgedChain;`. This would also allow us to implement things like #1657 easier;

svyatonik:
Also see this comment: https://github.com/paritytech/parity-bridges-common/pull/1657#discussion_r1042998528

svyatonik:
Future `pallet_bridge_messages::Config` (may be changed):
```rust
#[pallet::config]
pub trait Config<I: 'static = ()>: frame_system::Config {
    // General types

    /// The overarching event type.
    type RuntimeEvent: From<Event<Self, I>>
        + IsType<<Self as frame_system::Config>::RuntimeEvent>;
    /// Benchmarks results from runtime we're plugged into.
    type WeightInfo: WeightInfoExt;

    /// Bridged chain.
    type BridgedChain: Chain;
    /// Bridged header chain.
    type BridgedHeaderChain: HeaderChain<Self::BridgedChain>;

    /// Maximal number of unrewarded relayer entries at inbound lane. Unrewarded means that the
    /// relayer has delivered messages, but either confirmations haven't been delivered back to
    /// the source chain, or we haven't received reward confirmations yet.
    ///
    /// This constant limits maximal number of entries in the `InboundLaneData::relayers`. Keep
[...]

PR #1879, opened 2023-02-13
RefundRelayerForMessagesFromParachain improvements

serban300:
Working towards multiple issues: #1838 and #1667, https://github.com/paritytech/parity-bridges-common/pull/1657#discussion_r1042998528 . Publishing what I have so far. Will continue to work on these in future PRs.

What this PR attempts:
- addressing partially https://github.com/paritytech/parity-bridges-common/pull/1657#discussion_r1042998528 which is part of #1666 
- achieving some degree of reusability between `RefundRelayerForMessagesFromParachain` and the other signed extensions for Grandpa, Parachains and Messages
- simplifying a bit the bounds for the `RefundRelayerForMessagesFromParachain` generic parameters
- removing the `BE` generic parameter for `RefundRelayerForMessagesFromParachain`, by calling the `check_obsolete_*` function for each nested call individually instead of calling the `validate()` function on the `BridgeRejectObsoleteHeadersAndMessages`. In my opinion this has a couple of advantages:
  - removing one generic param and simplifying the bounds a bit. This is also enough to remove the `#[allow(clippy::type_complexity)]`
  - should be more efficient
  - it allows to create a `RefundRelayerForMessagesFromParachain` signed extension without having a `BridgeRejectObsoleteHeadersAndMessages` signed extension. It's not necessarily something that we'll ever need, but I think it's nice to have.


bin/runtime-common/src/refund_relayer_extension.rs
```diff
@@ -215,18 +190,23 @@ where
...
+		_info: &DispatchInfoOf<Self::Call>,
+		_len: usize,
 	) -> TransactionValidity {
-		// reject batch transactions with obsolete headers
-		if let Some(UtilityCall::<R>::batch_all { ref calls }) = call.is_sub_type() {
-			for nested_call in calls {
-				let reject_obsolete_transactions = BE::default();
-				reject_obsolete_transactions.pre_dispatch(who, nested_call, info, len)?;
-			}
+		let calls = match self.expand_call(call) {
```
svyatonik, line 198: This has changed behavior a bit - we don't support rejecting obsolete stuff e.g. in `BatchAll { MessagesDelivery, MessagesDelivery }` anymore. This isn't critical for us - just caught my attention

bin/runtime-common/src/refund_relayer_extension.rs
```diff
@@ -241,40 +221,40 @@ where
...
+
+			let mut tx_info = TxData {
+				submit_finality_proof: None,
+				submit_parachain_heads: None,
+				receive_messages_proof: calls
+					.receive_messages_proof
+					.receive_messages_proof_info_for(LID::get())?,
+			};
+			if let Some(submit_parachain_heads) = calls.submit_parachain_heads {
+				tx_info.submit_parachain_heads =
```
svyatonik, line 236: Not an issue - just sharing my thoughts :) IMO strongly typed `enum CallType` here is better than the struct with 2 `Option`s. With struct you may (accidentally) create it with `submit_finality_proof` set to `Some(_)` and `submit_parachain_heads` set to `None`, which is an issue. Enum forces you to have `submit_parachain_heads` call in the batch if you have the `submit_parachain_heads` call. I had to return to this method twice to ensure that that never happens :) But that's just my option - you're the author, so do what you think is best here.

svyatonik (approved):
Looks good, thank you!

bin/runtime-common/src/refund_relayer_extension.rs
serban300: Makes sense. Reverted to using the `enum`.

bin/runtime-common/src/refund_relayer_extension.rs
serban300: Oh, I didn't think of that. Changed it to do:
```
			call.check_obsolete_submit_finality_proof()?;
			call.check_obsolete_submit_parachain_heads()?;
			call.check_obsolete_receive_messages_proof()?;
```
for all nested calls in batches with at most 3 calls. 

But what would be the desired behavior here ? I see a couple of possibilities:
- checking for all possible obsolete headers and messages for any nested calls (as was done before with BE::validate()). But if in the future we'll support multiple `RefundRelayerForMessagesFromParachain` instances, I guess we don't want to check for all possible obsolete headers and messages in all the instances of `RefundRelayerForMessagesFromParachain`.  
- checking the batches supported by this extension (the batches contains at most 3 calls) with BE::validate()
- checking the batches supported by this extension (the batches contains at most 3 calls) with 
  ```
  call.check_obsolete_submit_finality_proof()?;
  call.check_obsolete_submit_parachain_heads()?;
  call.check_obsolete_receive_messages_proof()?;
  ```

Or maybe we can just extend `BridgeRejectObsoleteHeadersAndMessages` to support batch calls, which should cover everything. And remove this check from here.

Just trying to understand what the desire behavior is since this came up. But I would implement it in a future PR since it can get complicated.

svyatonik: TBH - I don't know if we need it or not - just spotted the difference :)

> checking for all possible obsolete headers and messages for any nested calls (as was done before with BE::validate()). But if in the future we'll support multiple RefundRelayerForMessagesFromParachain instances, I guess we don't want to check for all possible obsolete headers and messages in all the instances of RefundRelayerForMessagesFromParachain.

IDK if we'll ever need more than one instance of this extension. Now the plan is to bridge two bridge hubs only and then use HTMP to pass messages between parachains. So maybe we don't need it at all. But in any case - iiuc we don't do any heavy ops there - its merely a `match call { ... }` and only if we find a bridge call, we do some actual checks (db reads and comparisons). So maybe it isn't a bog issue?

IMO let's leave it in current state for now - we may revisit it later.

PR #1882, opened 2023-02-15
Weight+size limits for bridge GRANDPA pallet calls

svyatonik:
related to https://github.com/paritytech/srlabs_findings/issues/266

What this PR does:
- switches on-chain GRANDPA justification verification to "strict" mode. It'll reject all justifications with unknown, duplicate and redundant votes. Relay now "optimizes" justification before submitting it to the target chain (#1887), so it should be fine;
- we now have the `REASONABLE_HEADERS_IN_JUSTIFICATON_ANCESTRY` constant. If submitted GRANDPA justification has more headers in its votes ancestries, then: (1) if call is a simple `submit_finality_proof` transaction, then it'll never return `Pays::No` (meaning that relayer pays full transaction fee) (2) if call is a part of `Utility::batch_all` transaction, then we won't refund relayer for doing this particular call. It still may be refunded for other useful calls (parachains finality and messages delviery)
- we use the `REASONABLE_HEADERS_IN_JUSTIFICATON_ANCESTRY`, `AVERAGE_HEADER_SIZE_IN_JUSTIFICATION` and `MAX_HEADER_SIZE` constants to compute maximal expected size of `submit_finality_call` arguments. If arguments size is above this limit, then: (1) if call is a simple `submit_finality_proof` transaction, then it'll never return `Pays::No` (meaning that relayer pays full transaction fee) (2) if call is a part of `Utility::batch_all` transaction, then we will deduct "extra bytes" from the transaction size we use to compute refund amount.

After this PR is merged, relayers must be unable to submit large (in terms of weight and size) headers and justifications without paying a proper fee for that.

What is left here:
- [x] self review (after some rest :) )
- [x] check constant values. I'm not sure that `AVERAGE_HEADER_SIZE_IN_JUSTIFICATION` and `MAX_HEADER_SIZE` are correct for our testnets
- [x] add issue to fire alerts (?) when relayer submit larger-than-expected transactions: https://github.com/paritytech/parity-bridges-common/issues/1897

Apart from that ^^^ it is ready for review


bin/runtime-common/src/refund_relayer_extension.rs
```diff
@@ -587,7 +619,7 @@ mod tests {
 	fn dispatch_info() -> DispatchInfo {
 		DispatchInfo {
 			weight: Weight::from_ref_time(
```
acatangiu, line 621: is this now double so that we get `extra_weight` in tests?

bin/runtime-common/src/refund_relayer_extension.rs
```diff
@@ -823,6 +855,54 @@ mod tests {
...
+			run_post_dispatch(Some(pre_dispatch_data), Ok(()));
+			assert_eq!(
+				RelayersPallet::<TestRuntime>::relayer_reward(
+					relayer_account_at_this_chain(),
+					TestLaneId::get()
+				),
+				Some(regular_reward),
+			);
+
+			// no repeat the same with size+weight refund: we expect smaller reward
```
acatangiu, line 880: typo/nit

```suggestion
			// now repeat the same with size+weight refund: we expect smaller reward
```

primitives/polkadot-core/src/lib.rs
```diff
@@ -51,29 +51,36 @@ pub mod parachains;
 ///
 /// Right now both Kusama and Polkadot aim to have around 1000 validators. Let's be safe here and
 /// take twice as much here.
```
acatangiu, line 53: nit
```suggestion
/// take a bit more here.
```

primitives/polkadot-core/src/lib.rs
```diff
@@ -51,29 +51,36 @@ pub mod parachains;
...
-/// `AVERAGE_HEADER_SIZE_IN_JUSTIFICATION` bytes for other stuff.
-///
 /// See [`bp_header_chain::ChainWithGrandpa`] for more details.
-pub const MAX_HEADER_SIZE: u32 = MAX_AUTHORITIES_COUNT
-	.saturating_mul(3)
-	.saturating_add(AVERAGE_HEADER_SIZE_IN_JUSTIFICATION);
+///
+/// This value comes from recent (February, 2023) Kusama headers. Maximal header is a mandatory
+/// header. In its SCALE-encoded form it is `80348` bytes. Let's have some reserve here.
+pub const MAX_HEADER_SIZE: u32 = 90_000;
```
acatangiu, line 83: Could we add some tests that simply verify the actual maximal and average sizes against the constants here?

Just add a issue for it if you think it's worth it.

acatangiu (approved):
Looks good to me! :+1:

primitives/polkadot-core/src/lib.rs
svyatonik: In real world, there's no max header size - you could add an arbitrary blob to digest and it would be a valid header. So that's just an estimation for current chain state. I've filed this issue: https://github.com/paritytech/parity-bridges-common/issues/1897 to be able to monitor pallet transactions and know when we need to update constants. But overall, we can't have any rust tests for that :/

bin/runtime-common/src/refund_relayer_extension.rs
svyatonik: Yes, but actually it isn't required iiuc, because I'm doing the same change in that test that needs it

primitives/polkadot-core/src/lib.rs
```diff
@@ -51,29 +51,36 @@ pub mod parachains;
...
-pub const MAX_AUTHORITIES_COUNT: u32 = 2_048;
+pub const MAX_AUTHORITIES_COUNT: u32 = 1_256;
 
 /// Reasonable number of headers in the `votes_ancestries` on Polkadot-like chains.
 ///
 /// See [`bp_header_chain::ChainWithGrandpa`] for more details.
-pub const REASONABLE_HEADERS_IN_JUSTIFICATON_ANCESTRY: u32 = 8;
+///
+/// This value comes from recent (February, 2023) Kusama and Polkadot headers. There are no
+/// justifications with any additional headers in votes ancestry, so reasonable headers may
```
serban300, line 61: I would be interested in learning more about this. Do you have a link to the code where this value comes from ?

bin/runtime-common/src/refund_relayer_extension.rs
svyatonik: reverted

primitives/polkadot-core/src/lib.rs
```diff
@@ -51,29 +51,36 @@ pub mod parachains;
 ///
 /// Right now both Kusama and Polkadot aim to have around 1000 validators. Let's be safe here and
 /// take twice as much here.
-pub const MAX_AUTHORITIES_COUNT: u32 = 2_048;
+pub const MAX_AUTHORITIES_COUNT: u32 = 1_256;
```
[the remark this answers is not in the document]
svyatonik: Fixed by @acatangiu commit :)

primitives/polkadot-core/src/lib.rs
svyatonik: Sure. Here's the code (if you need the same for Polkadot - just change the host to `rpc.polkadot.io`):
```rust
async_std::task::block_on(async {
		use codec::{Decode, Encode};
		use relay_substrate_client::BlockWithJustification;
		use sp_finality_grandpa::{ConsensusLog, GrandpaJustification, GRANDPA_ENGINE_ID};
		use sp_runtime::traits::Header;

		let kusama = relay_substrate_client::Client::<relay_kusama_client::Kusama>::new(
			relay_substrate_client::ConnectionParams {
				host: "kusama-rpc.polkadot.io".into(),
				port: 443,
				secure: true,
				..Default::default()
			},
		)
		.await;

		let mut total_maximal_headers = 0;
		let mut total_maximal_header_size = 0;
		let mut maximal_header_size = 0;

		let mut total_average_headers = 0;
		let mut total_average_header_size = 0;
		let mut maximal_average_header_size = 0;

		let mut total_justifications = 0;
		let mut total_justifications_size = 0;
		let mut max_justification_ancestry = 0;

		let mut best_finalized = kusama.best_finalized_header().await.unwrap();
		while *best_finalized.number() > 0 {
			let block = kusama.get_block(Some(best_finalized.hash())).await.unwrap();
			let header = block.header();
			best_finalized = kusama.header_by_hash(*header.parent_hash()).await.unwrap();

			let is_maximal_header = header
				.digest()
				.convert_first(|log| log.consensus_try_to(&GRANDPA_ENGINE_ID))
				.and_then(|log: ConsensusLog<bp_polkadot_core::BlockNumber>| match log {
					ConsensusLog::ScheduledChange(change) => Some(change),
					_ => None,
				})
				.is_some();

			let header_size = header.encode().len();
			if is_maximal_header {
				total_maximal_headers += 1;
				total_maximal_header_size += header.encode().len();
				if header_size > maximal_header_size {
					maximal_header_size = header_size;
				}
			} else {
				total_average_headers += 1;
				total_average_header_size += header_size;
				if header_size > maximal_average_header_size {
					println!("{:?}", header);

					maximal_average_header_size = header_size;
				}
			}

			let justification = block.justification(GRANDPA_ENGINE_ID);
			if let Some(justification) = justification {
				let justification = GrandpaJustification::<bp_polkadot_core::Header>::decode(
					&mut &justification[..],
				)
				.unwrap();
				total_justifications += 1;
				total_justifications_size += justification.encode().len();

				let votes_ancestries_len = justification.votes_ancestries.len();
				if votes_ancestries_len > max_justification_ancestry {
					max_justification_ancestry = votes_ancestries_len;
				}
			}

			if header.number() % 1024 == 0 {
				println!(
					"MaxHeader: total={} maximal={} average={:?}",
					total_maximal_headers,
					maximal_header_size,
					total_maximal_header_size.checked_div(total_maximal_headers)
				);
				println!(
					"AvgHeader: total={} maximal={} average={:?}",
					total_average_headers,
					maximal_average_header_size,
					total_average_header_size.checked_div(total_average_headers)
				);
				println!(
					"Justifica: total={} average={:?} max_votes={}",
					total_justifications,
					total_justifications_size.checked_div(total_justifications),
					max_justification_ancestry
				);
			}
		}
	});
```

modules/grandpa/src/call_ext.rs
```diff
@@ -105,15 +142,90 @@ impl<T: Config<I>, I: 'static> CallSubType<T, I> for T::RuntimeCall where
...
+	finality_target: &BridgedHeader<T, I>,
+	justification: &GrandpaJustification<BridgedHeader<T, I>>,
+) -> SubmitFinalityProofInfo<BridgedBlockNumber<T, I>> {
+	let block_number = *finality_target.number();
+
+	// the `submit_finality_proof` call will reject justifications with invalid, duplicate,
+	// unknown and extra signatures. It'll also reject justifications with less than necessary
+	// signatures. So we do not care about extra weight because of additional signatures here.
+	let precommits_len = justification.commit.precommits.len().saturated_into();
+	let required_precommits = precommits_len;
```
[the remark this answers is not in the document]
svyatonik: I wanted to be explicit here, because e.g. week ago (before this "optimized justification" stuff) `required_precommits` wasn't equal to `precommits_len`. So they have different meaning

modules/grandpa/src/call_ext.rs
```diff
@@ -105,15 +142,90 @@ impl<T: Config<I>, I: 'static> CallSubType<T, I> for T::RuntimeCall where
...
+		.saturating_add(justification.encoded_size())
+		.saturated_into();
+	let max_expected_call_size = max_expected_call_size::<T, I>(required_precommits);
+	let extra_size = actual_call_size.saturating_sub(max_expected_call_size);
+
+	SubmitFinalityProofInfo { block_number, extra_weight, extra_size }
+}
+
+/// Returns maximal expected size of `submit_finality_proof` call arguments.
+fn max_expected_call_size<T: Config<I>, I: 'static>(required_precommits: u32) -> u32 {
```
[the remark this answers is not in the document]
svyatonik: done

serban300 (approved):

PR #1884, opened 2023-02-16
Fix clippy issues

svyatonik:
https://gitlab.parity.io/parity/mirrors/parity-bridges-common/-/jobs/2399714 + couple of others

I had to disable `extra_unused_type_parameters` clippy check, because it causes  false alerts in current version. Actually - we need to revisit all currently disabled checks later - maybe some are fixed already. But let's do that later :

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -92,7 +92,7 @@ clippy-nightly:
     RUSTFLAGS:                     "-D warnings"
   script:
     # clippy currently is raising `derive-partial-eq-without-eq` warning even if `Eq` is actually derived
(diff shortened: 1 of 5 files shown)


serban300 (approved):

svyatonik:
`clippy::extra-unused-type-parameters` should be fixed by https://github.com/rust-lang/rust-clippy/pull/10321, so hopefully this will reach our nightly images soon

mkrasnitski:
> `clippy::extra-unused-type-parameters` should be fixed by [rust-lang/rust-clippy#10321](https://github.com/rust-lang/rust-clippy/pull/10321), so hopefully this will reach our nightly images soon

The fix should now be shipping in nightly 2023-02-26.

PR #1887, opened 2023-02-16
optimize justification before submit

### LINKED ISSUE #1731: Optimize GRANDPA justifications before submitting them to the pallet-bridge-grandpa
Noticed one strange thing when testing our relay with Kusama/Polkadot - within the same session, GRANDPA justifications received from the node contain different number of signatures. E.g. given session with `1000` validators in Kusama, on justification has `672` precommits, other has `674` and the next one has `676`. Since on-chain signature verification is expensive, it'd be better for relayers to "optimize" justifications before submitting them. Iiuc in this (`1000` validators) case we only need `667` signatures, so extra ~10 signatures may be removed before sending it to the pallet.

svyatonik:
closes #1731
required for #1882
related to https://github.com/paritytech/srlabs_findings/issues/266

This PR introduces additional step to the finality relay - after justification is read from the source node, it is verified and optimized (offchain) and then submitted to the target node. Optimization = removal of unknown, duplicate and redundant votes. Now it isn't treated as an error, but after #1882 if will be an error and such call will fail.

~Draft, because I want to test it locally dirst~ Haven't seen any `InvalidJustification` errors for a while, so loos like all works fine


acatangiu (approved):
this should work well

PR #1893, opened 2023-02-20
Use complex transactions on RBH/WBH bridge hubs

svyatonik:


serban300 (approved):

relays/client-substrate/src/chain.rs
```diff
@@ -276,6 +278,21 @@ where
 	}
 }
 
+/// Structure that implements `UtilityPalletProvider` based on a call conversion.
+pub struct MockedRuntimeUtilityPallet<Call> {
```
bkontur, line 282: I dont know, I just cannot get used to `Mock` word for production code :D
Wdyt about `WrappedRuntimeUtilityPallet`  (we have runtime_wrapper.rs in clients) or just `RuntimeUtilityPallet` or some other equivalent to `FullRuntimeUtilityPallet`?
Also, maybe in near future we will have `CodegenRuntimeUtilityPallet` for Serban's subxt stuff

bkontur (approved):

PR #1894, opened 2023-02-20
Rewards refund for relaying BridgeHubRococo/BridgeHubWococo

bkontur:


primitives/chain-bridge-hub-cumulus/src/lib.rs
```diff
@@ -124,3 +124,75 @@ pub const MAX_UNREWARDED_RELAYERS_IN_CONFIRMATION_TX: MessageNonce = 1024;
...
 pub const MAX_UNCONFIRMED_MESSAGES_IN_CONFIRMATION_TX: MessageNonce = 4096;
+
+/// Module with rewarding bridge signed extension support
+pub mod rewarding_bridge_signed_extension {
+	use super::*;
+	use bp_polkadot_core::PolkadotLike;
+	use bp_runtime::extensions::*;
+
+	/// The `SignedExtensionSchema` for `RefundRelayerForMessagesFromParachain`.
+	type RefundRelayerForMessagesFromParachain = GenericSignedExtensionSchema<(), ()>;
```
serban300, line 135: Nit: I would move this definition together with the other signed extension schemas in `bp_runtime`.

primitives/chain-bridge-hub-cumulus/src/lib.rs
```diff
@@ -124,3 +124,75 @@ pub const MAX_UNREWARDED_RELAYERS_IN_CONFIRMATION_TX: MessageNonce = 1024;
 
 /// Maximal number of unconfirmed messages at inbound lane for Cumulus-based parachains.
 pub const MAX_UNCONFIRMED_MESSAGES_IN_CONFIRMATION_TX: MessageNonce = 4096;
+
+/// Module with rewarding bridge signed extension support
+pub mod rewarding_bridge_signed_extension {
```
svyatonik, line 129: I'd remove all signing extensions from `bp-polkadot-core`, because we are no longer submit transactions to polkadot-like chains. And move proper struct and stuff to this crate.

Also I've just found that there are two signed extensions there - `BridgeSignedExtension` and `DefaultSignedExtension`. And it seems to me that `DefaultSignedExtension` may be safely removed and `BridgeSignedExtension` may be safely moved here.

It might be done in a separate PR though. I just find this `mod` and existing structs a bit confusing :)

svyatonik (approved):
LGTM module single comment, which may be addressed in a separate PR (by anyone - e.g. by me :) )

serban300 (approved):

primitives/chain-bridge-hub-cumulus/src/lib.rs
bkontur: `DefaultSignedExtension` is used by chain-rialto-parachain now 
and `BridgeSignedExtension` I thought that it is like generic example, so I didnt want to touch it.

`BridgeSignedExtension` / `DefaultSignedExtension` both are types and implement `trait PolkadotSignedExtension`,
so I was not able to use `PolkadotSignedExtension` outside, so thats why I created this `mod`.

if you want to move this stuff out of `bp-polkadot-core`, I just agree and I can do it easily done now here,

but lets first merge Serban's PR: https://github.com/paritytech/parity-bridges-common/pull/1895
because `mod rewarding_bridge_signed_extension` as it is will be used for Kusama/Polkadot,
but for BridgeHubRococo/Wococo I probably will have two instance of `RefundBridgedParachainMessages`, so signed_extensions extra would have two `(), ()`, right?

svyatonik: I've submitted https://github.com/paritytech/parity-bridges-common/issues/1896

bkontur: yeah, friday I just added one `RefundBridgedParachainMessages` to the runtime and I didnt touch relayer side where was `BridgeSignedExtension` (with BridgeRejectObsoleteHeadersAndMessages at the end) and the reward was registered, so I was also thinking that it is about some encoding stuff with `_` vs `()` vs `(), ()` 

I just created copies to the `runtime_wrapper.rs` with multiple `RefundBridgedParachainMessages` only for client-bridge-hub-rococo/wococo + TODO when to remove this in future

relays/client-bridge-hub-rococo/src/runtime_wrapper.rs
```diff
@@ -69,3 +69,75 @@ impl From<UtilityCall<Call>> for Call {
...
+		CheckSpecVersion,
+		CheckTxVersion,
+		CheckGenesis<PolkadotLike>,
+		CheckEra<PolkadotLike>,
+		CheckNonce<Nonce>,
+		CheckWeight,
+		ChargeTransactionPayment<PolkadotLike>,
+		BridgeRejectObsoleteHeadersAndMessages,
+		RefundBridgedParachainMessages,
+		RefundBridgedParachainMessages,
```
serban300, line 90: Sorry for going back on this

I think it would make sense to group the 2 refund extensions together. At least in the actual runtime. Something like:

```
RewardingBridgeSignedExtra = (
		CheckNonZeroSender,
		CheckSpecVersion,
		CheckTxVersion,
		CheckGenesis<PolkadotLike>,
		CheckEra<PolkadotLike>,
		CheckNonce<Nonce>,
		CheckWeight,
		ChargeTransactionPayment<PolkadotLike>,
		BridgeRejectObsoleteHeadersAndMessages,
		(RefundBridgedParachainMessages, RefundBridgedParachainMessages)
	);
```

But, even more generic, thinking again about https://github.com/paritytech/parity-bridges-common/pull/1894#discussion_r1112041653 , here I think it would be better to rename `RefundBridgedParachainMessages` to `RefundBridgedParachainMessagesTuple` (or something clearer, maybe even add a comment) in order to be able to use it in any indirect runtime as a wildcard for either a single `RefundBridgedParachainMessages` or a tuple of `(RefundBridgedParachainMessages, RefundBridgedParachainMessages, ...)`

bkontur: and how should runtime's side look like?

```
pub type SignedExtra = (
	frame_system::CheckNonZeroSender<Runtime>,
	frame_system::CheckSpecVersion<Runtime>,
	frame_system::CheckTxVersion<Runtime>,
	frame_system::CheckGenesis<Runtime>,
	frame_system::CheckEra<Runtime>,
	frame_system::CheckNonce<Runtime>,
	frame_system::CheckWeight<Runtime>,
	pallet_transaction_payment::ChargeTransactionPayment<Runtime>,
	BridgeRejectObsoleteHeadersAndMessages,
	(BridgeRefundBridgeHubRococoRelayers, BridgeRefundBridgeHubRococoRelayers),
);
```

`SignedExtension` has some tuple impl:

```
impl<AccountId, Call: Dispatchable> SignedExtension for Tuple {
	for_tuples!( where #( Tuple: SignedExtension<AccountId=AccountId, Call=Call,> )* );
	type AccountId = AccountId;
	type Call = Call;
	const IDENTIFIER: &'static str = "You should call `identifier()`!";
	for_tuples!( type AdditionalSigned = ( #( Tuple::AdditionalSigned ),* ); );
	for_tuples!( type Pre = ( #( Tuple::Pre ),* ); );
```

serban300: > and how should runtime's side look like?

Yes, like that, exactly

```
pub type SignedExtra = (
	frame_system::CheckNonZeroSender<Runtime>,
	frame_system::CheckSpecVersion<Runtime>,
	frame_system::CheckTxVersion<Runtime>,
	frame_system::CheckGenesis<Runtime>,
	frame_system::CheckEra<Runtime>,
	frame_system::CheckNonce<Runtime>,
	frame_system::CheckWeight<Runtime>,
	pallet_transaction_payment::ChargeTransactionPayment<Runtime>,
	BridgeRejectObsoleteHeadersAndMessages,
	(BridgeRefundBridgeHubRococoRelayers, BridgeRefundBridgeHubWococoRelayers),
);
```

I tried something like this with 2 test refund signed extensions for millau and it worked.

> SignedExtension has some tuple impl:

Yes, it works for nested tuples.

bkontur: @serban300 
something like this: https://github.com/paritytech/parity-bridges-common/pull/1894/commits/83ad36081decd09fa229f41ba81e08597283ac67 ?

I havent tried it yet with runing runtime (tomorrow), but at least it compiles :)

serban300: No, no, sorry, I'm bad at explaining. Sorry for misleading you. This would work, but we don't need this level of complexity. What I meant is on the relayer side you can just leave it as it was before:

```
/// The `SignedExtensionSchema` for `RefundBridgedParachainMessages`.
pub type RefundBridgedParachainMessages = GenericSignedExtensionSchema<(), ()>;
```

```
type RewardingBridgeSignedExtra = (
		CheckNonZeroSender,
		CheckSpecVersion,
		CheckTxVersion,
		CheckGenesis<PolkadotLike>,
		CheckEra<PolkadotLike>,
		CheckNonce<Nonce>,
		CheckWeight,
		ChargeTransactionPayment<PolkadotLike>,
		BridgeRejectObsoleteHeadersAndMessages,
		RefundRelayerForMessagesFromParachain,
		RefundBridgedParachainMessages,
	);
```

and just rename `RefundBridgedParachainMessages` to something like `RefundBridgedParachainMessagesTuple` (or a better name) and add a comment to clarify that this signed extension schema can be used as a counterpart for a `RefundBridgedParachainMessages` or a tuple with as many `(RefundBridgedParachainMessages, RefundBridgedParachainMessages, ...)` in the actual runtime. Because on the relay side, `()` works even if in the actual runtime we have `()` or `((), ())`, or `((), (), ())` ... . The transaction will work because it will decode to the same thing anyway.

bkontur: @serban300 
hehe, ok, I got it now what you meant, it is like I said [here](https://github.com/paritytech/parity-bridges-common/pull/1894/files#r1112065075) that we dont even need to add anything, because after `ChargeTransactionPayment<PolkadotLike>` we have `BridgeRejectObsoleteHeadersAndMessages` which is `()` so it covers everything `()` or `((), ())`, or `((), (), ())`

now I understand what you meant by `RefundBridgedParachainMessages` as Wildcard/placeholder which relies on the scale encoding for `()` or `((), ())`, or `((), (), ())` is the same

yeah, my commit is opposite, and can be totally explicit: `rewarding_bridge_signed_extension::RewardingBridgeSignedExtension<((), ()), ((), ())>,`

now we now several ways, but for simplicity I am going with your placeholder-like way

serban300: Looks good now !

bkontur: well, [as we found out](https://github.com/paritytech/cumulus/pull/2194#discussion_r1113203831), downside of this wildcard/placeholder way is, that we cannot use this for testing `check_additional_signed` (with [complex way](https://github.com/paritytech/parity-bridges-common/pull/1894#discussion_r1112132383) it would be possible )

I had to remove `check_additional_signed` from cumulus side
https://github.com/paritytech/cumulus/pull/2194/files#diff-554a991cc27b681ef4c6491717f9d505b167ca7bf9e5490f88b8532754d70de9R1164-R1202

----------

anyway, regardless of wildcard/placeholder or complex way, this **Cumulus test** [ensure_signed_extension_definition_is_compatible_with_relay](https://github.com/paritytech/cumulus/pull/2194/files#diff-554a991cc27b681ef4c6491717f9d505b167ca7bf9e5490f88b8532754d70de9R1164) 
just checks Runtime's SignedExtension vs. definition in subtree `bridges/primitives/chain-bridge-hub-cumulus`

so there are several implications:
1. **Cumulus test** is kind of useful:
- it means, if we change Runtime's definition, then we need to also change subtree `bridges/primitives/chain-bridge-hub-cumulus`
- which means somebody has to backport/PR that change to the Bridges repo, but this can be released/merged only after new Runtime is onboarded onlive
2. `pub mod rewarding_bridge_signed_extension` does not belong to the  `primitives/chain-bridge-hub-cumulus` but should be moved to `relays/*` layer, e.g. `relays/client-cumulus` (because it is used just by relayer)
- if we do this, we lose ability to do that **Cumulus test**, because we dont propage `relays/*` layer to Cumulus subtree
-   but sure still we could make exception in `verify-pallets-build.sh` for `relays/client-cumulus` and use it as `[dev-dependencies]`
- it is just a question of what is more clear

serban300: Could you open a separate issue related to this points please ? In order to not lose track of them

bkontur: ok, I just put it as a part https://github.com/paritytech/parity-bridges-common/issues/1896

Issue #1760, opened 2023-01-11
Dynamic way of adding lanes (bridges between parachains)


svyatonik:
related to https://github.com/paritytech/srlabs_findings/issues/142

Our design is to have a dedicated lane for two bridged chains (see https://github.com/paritytech/parity-bridges-common/blob/master/docs/polkadot-kusama-bridge-overview.md#connecting-parachains). E.g. Polkadot' Statemint and Kusama' Statemine will use the lane `[0, 0, 0, 0]`.

Later we want to be able to connect other parachains. While we may do that by doing runtime upgrades and manually coding that, we may also have some calls for that, so that lanes may be opened without waiting for runtime upgrades.

Apart from that, we shall think of some deposit that must be paid for opening a lane. The amount of deposit may affect (or take into account) the maximal number of outbound messages for that lane. That's because parachains may just queue messages and don't run any relayer => bridge hub runtime storage will grow and storage proofs size will grow too. We may start to take some fee from deposit if messages are queued for too long and finally close the lane if deposit is depleted.

bkontur:
well, this kind of on-chain configuration, I mentioned here https://github.com/paritytech/parity-bridges-common/issues/1759#issuecomment-1378499404
could be really used here, and governance can manage deposit amount or what ever

Polkadot-Forum:
This issue has been mentioned on **Polkadot Forum**. There might be relevant details there:

https://forum.polkadot.network/t/bridge-hub-questions/2693/2

bkontur:
- maybe it looks like if we use xcm [`channel: u32`](https://github.com/paritytech/polkadot/pull/7244) and convert it to the `LaneId` in HaulBlob impl [here](https://github.com/paritytech/parity-bridges-common/blob/master/bin/runtime-common/src/messages_xcm_extension.rs#L130-L132) we dont need hard-coded LaneId,
- plus `ActiveOutboundLanes` can be probably set as `pub storage ActiveOutboundLanes` so it can be managed by governance to add new once
- the last thing could be fix `RefundBridgedParachainMessages`

svyatonik:
[...]

PR #1895, opened 2023-02-20
RefundRelayerForMessagesFromParachain improvements

serban300:
Fixes: #1838 and #1667

- some renamings and cosmetics
- removing the GI generic param since we can get it from the parachains config
- merging (Parachains Instance, Parachain Id) and (Messages Instance, Messages lane Id) generic params
- removing unnecessay derives
- renaming for the `TransactionFeeCalculation` and defining `ActualFeeRefund` concrete implementation
- adding support for multiple RefundRelayerForMessagesFromParachain (by adding a `StaticStrProvider` generic param)

### DIFF
--- a/bin/millau/runtime/src/lib.rs
+++ b/bin/millau/runtime/src/lib.rs
@@ -80,7 +80,13 @@ pub use pallet_sudo::Call as SudoCall;
 pub use pallet_timestamp::Call as TimestampCall;
 pub use pallet_xcm::Call as XcmCall;
 
-use bridge_runtime_common::generate_bridge_reject_obsolete_headers_and_messages;
+use bridge_runtime_common::{
+	generate_bridge_reject_obsolete_headers_and_messages,
+	refund_relayer_extension::{
+		ActualFeeRefund, RefundBridgedParachainMessages, RefundableMessagesLane,
+		RefundableParachain,
+	},
+};
 #[cfg(any(feature = "std", test))]
 pub use sp_runtime::BuildStorage;
 pub use sp_runtime::{Perbill, Permill};
@@ -582,17 +588,15 @@ generate_bridge_reject_obsolete_headers_and_messages! {
 	BridgeRialtoMessages, BridgeRialtoParachainMessages
 }
 
+bp_runtime::generate_static_str_provider!(BridgeRefundRialtoPara2000Lane0Msgs);
 /// Signed extension that refunds relayers that are delivering messages from the Rialto parachain.
-pub type BridgeRefundRialtoParachainRelayers =
-	bridge_runtime_common::refund_relayer_extension::RefundRelayerForMessagesFromParachain<
-		Runtime,
-		RialtoGrandpaInstance,
-		WithRialtoParachainsInstance,
-		WithRialtoParachainMessagesInstance,
-		RialtoParachainId,
-		RialtoParachainMessagesLane,
-		Runtime,
-	>;
+pub type BridgeRefundRialtoParachainMessages = RefundBridgedParachainMessages<
+	Runtime,
+	RefundableParachain<WithRialtoParachainsInstance, RialtoParachainId>,
+	RefundableMessagesLane<WithRialtoParachainMessagesInstance, RialtoParachainMessagesLane>,
+	ActualFeeRefund<Runtime>,
+	StrBridgeRefundRialtoPara2000Lane0Msgs,
+>;
 
 /// The address format for describing accounts.
 pub type Address = AccountId;
@@ -615,7 +619,7 @@ pub type SignedExtra = (
 	frame_system::CheckWeight<Runtime>,
 	pallet_transaction_payment::ChargeTransactionPayment<Runtime>,
 	BridgeRejectObsoleteHeadersAndMessages,
-	BridgeRefundRialtoParachainRelayers,
+	BridgeRefundRialtoParachainMessages,
 );
 /// The payload being signed in transactions.
 pub type SignedPayload = generic::SignedPayload<RuntimeCall, SignedExtra>;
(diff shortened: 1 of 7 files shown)


primitives/runtime/src/lib.rs
```diff
@@ -507,6 +507,24 @@ impl WeightExtraOps for Weight {
...
+/// Trait that provides a static `str`.
+pub trait StaticStrProvider {
+	const STR: &'static str;
+}
+
+#[macro_export]
+macro_rules! generate_static_str_provider {
+	($str:expr) => {
+		$crate::paste::item! {
+			pub struct [<Str $str>];
```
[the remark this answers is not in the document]
serban300: It's in case we want to use the same name for the signed extension. For example something like:

```
bp_runtime::generate_static_str_provider!(BridgeRefundRialtoPara2000Lane0Msgs);
/// Signed extension that refunds relayers that are delivering messages from the Rialto parachain.
pub type BridgeRefundRialtoPara2000Lane0Msgs = RefundBridgedParachainMessages<
	Runtime,
	RefundableParachain<WithRialtoParachainsInstance, RialtoParachainId>,
	RefundableMessagesLane<WithRialtoParachainMessagesInstance, RialtoParachainMessagesLane>,
	ActualFeeRefund<Runtime>,
	StrBridgeRefundRialtoPara2000Lane0Msgs,
>;
```

bin/runtime-common/src/refund_relayer_extension.rs
```diff
@@ -123,71 +221,30 @@ where
...
-	FEE: 'static + Send + Sync + TransactionFeeCalculation<<R as RelayersConfig>::Reward>,
-	CallOf<R>: Dispatchable<Info = DispatchInfo, PostInfo = PostDispatchInfo>
-		+ IsSubType<CallableCallFor<UtilityPallet<R>, R>>
-		+ GrandpaCallSubType<R, GI>
-		+ ParachainsCallSubType<R, PI>
-		+ MessagesCallSubType<R, MI>,
+	Para: RefundableParachainId,
+	Msgs: RefundableMessagesLaneId,
+	Refund: RefundCalculator<Balance = Runtime::Reward>,
+	Id: StaticStrProvider,
```
bkontur, line 236: @serban300 
can we use here just `Get<&'static str>` instead of StaticStrProvider?

serban300: No, it would have been nice, but it doesn't work :( Because Get<...>::get is not `const`.

primitives/runtime/src/lib.rs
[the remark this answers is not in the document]
serban300: Yes, but then the id of the signed extension would also be `StrBridgeRefundRialtoPara2000Lane0Msgs`. I think it would be better for the id to be just `BridgeRefundRialtoPara2000Lane0Msgs`.

Anyway another option would also be to change the macro in order to accept 2 parameters. The name of the struct and the string. I can do it if you think it's needed.

bin/runtime-common/src/messages_call_ext.rs
```diff
@@ -20,10 +20,10 @@ use crate::messages::{
 use bp_messages::{LaneId, MessageNonce};
 use frame_support::{dispatch::CallableCallFor, traits::IsSubType};
 use pallet_bridge_messages::{Config, Pallet};
-use sp_runtime::{transaction_validity::TransactionValidity, RuntimeDebug};
+use sp_runtime::transaction_validity::TransactionValidity;
 
 /// Info about a `ReceiveMessagesProof` call which tries to update a single lane.
-#[derive(Copy, Clone, PartialEq, RuntimeDebug)]
+#[derive(PartialEq, Debug)]
```
svyatonik, line 26: IIRC using `Debug` (and all string operations) wasn't desired in runtime. So maybe revert to `RuntimeDebug` here?

bin/millau/runtime/src/lib.rs
```diff
@@ -582,17 +588,15 @@ generate_bridge_reject_obsolete_headers_and_messages! {
...
-		WithRialtoParachainsInstance,
-		WithRialtoParachainMessagesInstance,
-		RialtoParachainId,
-		RialtoParachainMessagesLane,
-		Runtime,
-	>;
+pub type BridgeRefundRialtoParachainMessages = RefundBridgedParachainMessages<
+	Runtime,
+	RefundableParachain<WithRialtoParachainsInstance, RialtoParachainId>,
+	RefundableMessagesLane<WithRialtoParachainMessagesInstance, RialtoParachainMessagesLane>,
```
bkontur, line 596: @serban300 @svyatonik 
I think in the future https://github.com/paritytech/parity-bridges-common/issues/1760,
we will have to read Lanes dynamically (e.g. from onchain storage),
but I guess it is ok for now

svyatonik (approved):
LGTM

bin/runtime-common/src/refund_relayer_extension.rs
bkontur: I see one usage: `N: Get<&'static str>`  for some staking migrations:
```
pub struct StakingMigrationV11OldPallet;
impl Get<&'static str> for StakingMigrationV11OldPallet {
	fn get() -> &'static str {
		"VoterList"
	}
}
```

primitives/runtime/src/lib.rs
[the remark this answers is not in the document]
bkontur: ok, ok, not needed :), I didnt realize that, I got it now: `[<Str $str>]` vs `stringify!($str)`

bkontur (approved):
lgtm,

just one nit: `Get<&'static str>` instead of `StaticStrProvider`, but I am also ok with `StaticStrProvider`

bin/runtime-common/src/messages_call_ext.rs
serban300: You're right. I forgot. Done.

bin/runtime-common/src/refund_relayer_extension.rs
serban300: Yes, tried again. I'm getting:

```
cannot call non-const fn `<Id as frame_support::traits::Get<&str>>::get` in constants
```

We'll be able to use const generics in the future when they are stable. Until then we need this workaround.

Issue #1677, opened 2022-12-01
[R-W Bridge] (Fees) Redeploy to synch Rococo and Wococo with fees/weights/rewards


EmmanuellNorbertTulbure:
check and setup weight/fees everywhere
- [x] BridgeHub extrinsics (finality, messages)
- [x] BridgeHub xcm execution (ParentAsUmp for bridge-hub-kusama/rococo)
- [x] Rockmine/Statemine XCM send export_message 
- [x] Wockmint/Statemint XCM receive execution
- [x] hints from Slava
  - [ ] https://github.com/paritytech/parity-bridges-common/pull/1813#discussion_r1086312831
  - [ ] https://github.com/paritytech/parity-bridges-common/pull/1813#discussion_r1086297303
- [x] design some tests https://github.com/paritytech/parity-bridges-common/issues/2078
- [x] `type OnChargeTransaction` change `()` -> `DealWithFees<Runtime>`
- [ ] ~~setup reward constants~~ - checking constants will be a part of https://github.com/paritytech/parity-bridges-common/issues/2078
   ```
	pub storage DeliveryRewardInBalance: u64 = 1_000_000;
	pub storage RequiredStakeForStakeAndSlash: Balance = 1_000_000;
   ```

bkontur:
One delivered message (local setup) BridgeHubRococo -> BridgeHubWococo:

**on BridgeHubRococo:**
```
submitFinalityProof:		0 (Pays=No) (multiple)
submitParachainHeads: 		0.0000505074 	ROC (fee for extrinsic)
submitParachainHeads: 		0.000050478224	ROC (fee for extrinsic)
submitParachainHeads: 		0.000050507133	ROC (fee for extrinsic)
receiveMessagesDeliveryProof:	0.000033512269	ROC (fee for extrinsic)
---------------------------------------------------------------------
Relayer expenses:		0.000185005	ROC (fees)


RewardRegistered:		0.0000001 	ROC `by DeliveryConfirmationPaymentsAdapter<100_000, 10_000>`
claimRewards:			0.000005074699	ROC (fee for extrinsic)
```





**on BridgeHubWococo:**
```
submitFinalityProof:		0 (Pays=No) (multiple)
submitParachainHeads:		0.000050536633	ROC (fee for extrinsic)
submitParachainHeads:		0.0000505368	ROC (fee for extrinsic)
[...]

PR #1898, opened 2023-02-21
Wrap confirmation and finality transactions into batch_all in Millau -> RialtoParachain bridge

svyatonik:
Unless I miss something, that is all we need for https://github.com/paritytech/parity-bridges-common/issues/1677#issuecomment-1438347928. ~Testing right now -> draft~ Yep - seems like that is enough

### DIFF
--- a/relays/messages/src/message_race_loop.rs
+++ b/relays/messages/src/message_race_loop.rs
@@ -576,14 +576,10 @@ pub async fn run<P: MessageRace, SC: SourceClient<P>, TC: TargetClient<P>>(
 					"Going to submit proof of messages in range {:?} to {} node{}",
 					nonces_range,
 					P::target_name(),
-					if let Some(ref target_batch_transaction) = target_batch_transaction {
-						format!(
-							". This transaction is batched with sending the proof for header {:?}.",
-							target_batch_transaction.required_header_id(),
-						)
-					} else {
-						Default::default()
-					},
+					target_batch_transaction.as_ref().map(|tx| format!(
+						". This transaction is batched with sending the proof for header {:?}.",
+						tx.required_header_id())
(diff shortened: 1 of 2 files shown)


relays/messages/src/message_race_loop.rs
```diff
@@ -573,17 +573,18 @@ pub async fn run<P: MessageRace, SC: SourceClient<P>, TC: TargetClient<P>>(
...
 					nonces_range,
 					P::target_name(),
+					if let Some(ref target_batch_transaction) = target_batch_transaction {
+						format!(
+							". This transaction is batched with sending the proof for header {:?}.",
+							target_batch_transaction.required_header_id(),
+						)
+					} else {
+						Default::default()
+					},
```
serban300, line 586: Nit: I would use `map` here:

```
					target_batch_transaction.as_ref().map(|tx| format!(
						". This transaction is batched with sending the proof for header {:?}.",
						tx.required_header_id())
					).unwrap_or_default()
```

acatangiu (approved):

serban300 (approved):

serban300:
It seems that now the confirmation race is stalling:

image

svyatonik:
@serban300 Do you use some customized version of relay/nodes? Or they're just our latest images from docker hub?

serban300:
It should have been the latest version on master yesterday, compiled locally, but I might have done something wrong. I'll check again.

svyatonik:
Ok, now we have it live on test deployments + I've started it locally. So if something is wrong there, hope we'll find it

svyatonik:
@serban300 I also had ~3 minutes delay in confirmations delivery locally. Filed a separate issue #1901. Thanks for spotting that!

PR #1899, opened 2023-02-22
Get rid of ChainWithMessages::WeightInfo, because we can't have message pallet weights for "external chains"

### LINKED ISSUE #1640: Fix ChainWithMessages::WeightInfo with correct values according to BridgeHub parachains
From Slava PR comment:

> WeightInfo is used to compute message delivery transaction limits. I think it is fine to use () for initial estimations (relay leaves some reserve anyway && () weights are probably larger than actual weights), but later we'll need to use weights from the parachains.

svyatonik:
closes #1640

This is just a proposal for fixing #1640. TLDR: instead of storing "external" chains weights in the relay code (which is too hard to maintain), let's use RPC calls to estimate required call (right now - only delivery call) weights. 

Pros are: we no longer need to have `WeightInfo` in relayer.

Cons are: hard to test it (error prone), relay requires some knowledge of how call weight is computed, less precise.

What can be changed: we can add CLI arguments for delivery call limits and make this computation optional if arguments are provided. This gives us some flexibility + ability to counter issues that may happen during computations. Please leave comment if you thing it'll be helpful.

### DIFF
--- a/modules/messages/src/lib.rs
+++ b/modules/messages/src/lib.rs
@@ -43,6 +43,7 @@ pub use weights::WeightInfo;
 pub use weights_ext::{
 	ensure_able_to_receive_confirmation, ensure_able_to_receive_message,
 	ensure_weights_are_correct, WeightInfoExt, EXPECTED_DEFAULT_MESSAGE_LENGTH,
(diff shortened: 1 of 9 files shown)


bkontur (approved):
lgtm

PR #1904, opened 2023-02-23
Refund messages confirmation tx

serban300:
Related to #1677

I added this logic to the existing `RefundBridgedParachainMessages` since it seemed to fit quite well: it's for the same parachain, same lane, it doesn't require extra configurations, and almost the entire logic remains the same. But if you think it would be better to have a separate signed extension, I can do that.


bin/runtime-common/src/refund_relayer_extension.rs
```diff
@@ -373,10 +371,9 @@ where
 
 		// Check if the `ReceiveMessagesProof` call delivered at least some of the messages that
 		// it contained. If this happens, we consider the transaction "helpful" and refund it.
-		let msgs_proof_info = call_info.receive_messages_proof_info();
-		if !ReceiveMessagesProofHelper::<Runtime, Msgs::Instance>::was_partially_successful(
-			msgs_proof_info,
-		) {
+		let msgs_call_info = call_info.messages_call_info();
+		if !MessagesCallHelper::<Runtime, Msgs::Instance>::was_partially_successful(msgs_call_info)
```
svyatonik, line 375: The big question here is: now we gonna use the same extension instance for `lane_00000000` that is used to send `RBH` -> `WBH` messages and for the same lane `lane_00000000` used to send `WBH` -> `RBH`. Who is paying rewards in our design? My understanding that:
- sovereign `Wockmine` account is used to pay rewards to relayers at the `RBH` in ROC (refund for delivering messages) (I);
- sovereign `Wockmine` account is used to pay rewards to relayers at the `WBH` in WOC (refund+reward for confirming delivery) (II);
- sovereign `Rockmine` account is used to pay rewards to relayers at the `WBH` in WOC (refund for delivering messages) (III);
- sovereign `Rockmine` account is used to pay rewards to relayers at the `RBH` in ROC (refund+reward for confirming delivery) (IV).

Since we gonna use the same extension, it'll use the same instance of `pallet-bridge-relayers`. The `pallet-bridge-relayers` is using message lane id to derive account that is paying rewards to relayers. So in our case at `RBH` the `derive(lane_00000000)` will be paying for delivering messages from `WBH` to `RBH` and for confirming messages sent from `RBH` to `WBH`. So it kinda represent the sovereign `Wockimnie` account at `RBH` (see (I) above) and sovereign `Rockime` account at `RBH` (see (IV) above). So the question is - do we consider the bridge economics as a single entity (then it is fine), or do we want to have separate sovereign accounts for each side of the bridge (that is my understanding I've described above).

Please don't treat this as a request for change - that's just a question. So I guess we need to discuss it here with @acatangiu and @bkontur as well

serban300: Oh good point. I missed this. I have to think if we can extend the current signed extension to account for this, or if a new one would be better. But, yes, let's also discuss the payment strategy here.

svyatonik: Let's probably use separate accounts for that.

IMO the most correct way to handle that here is to change the account derivation scheme - e.g. include bridge direction and use `(SourceChainId, TargetChainId, LaneId)` instead of simple `LaneId` to compute account id. So the extension will have two parameters - `ThisChainId` and `BridgedChainId`. When it'll be registering rewards for incoming messages (message delivery transaction), it'll use `(BridgedChainId, ThisChainId, LaneId)` to compute account. When it'll be registering rewards for message confirmations, it'll use `(ThisChainId, BridgedChainId, LaneId)`. Since we also registering rewards from `DeliveryConfirmationPaymentsAdapter`, we'll need to use the same scheme there.

serban300: Sounds good ! Thanks for the idea ! Will try to implement it.

serban300: Merged upstream/master (including #1907) into this PR and did the necessary adjustments

svyatonik (approved):
Thank you!

acatangiu:
> So the question is - do we consider the bridge economics as a single entity (then it is fine), or do we want to have separate sovereign accounts for each side of the bridge (that is my understanding I've described above).

I see this is no longer a hanging question, but just wanted to confirm that I was also thinking of the latter - aka "we want to have separate sovereign accounts for each side of the bridge (that is my understanding I've described above)"

PR #1907, opened 2023-02-24
Relayers pallet: extend payment source id

serban300:
Implementing [this comment](https://github.com/paritytech/parity-bridges-common/pull/1904#discussion_r1116582216) and also addressing #1692 

2 questions however (for future PRs maybe):

1. This approach might make it more complicated for the relayer to claim rewards. It will have to execute a `claim_rewards` call for each `(lane_id, bridged_chain_id, lane_direction)`. Could this be an inconvenience ? Should we make it easier for the relayer ? Maybe modify `claim_rewards` in order to handle `LaneDirection::In` and `LaneDirection::Out` at once ?

2. Right now we have 2 steps for awarding rewards:
  - registering the reward
  - the relayer claims the reward

    What if the account that pays the reward doesn't have funds at step 2 ? Can this occur ? It may not have funds in the first place, or it may have funds when registering the rewards, and spend them until the relayer claims the reward. Should / could we handle this ? Maybe freeze some funds at step 1 ?


primitives/messages/src/lib.rs
```diff
@@ -89,6 +89,33 @@ impl TypeId for LaneId {
 	const TYPE_ID: [u8; 4] = *b"blan";
 }
 
+/// Lane direction.
+#[derive(Copy, Clone, Debug, Decode, Encode, Eq, PartialEq, TypeInfo, MaxEncodedLen)]
+pub enum LaneDirection {
```
svyatonik, line 94: Just sharing my thoughts - maybe some may be implemented:
1) we are only using it (`LaneDirection` and `FullLaneId`) in the rewards/refund contest, so maybe move it to the `bp-relayers` crate? We don't actually need to pass it to the `DeliveryConfirmationPayments::pay_rewards()`, because, as name says, it is called during delivery confirmation transaction, so the direction would always be the `Out`. And you can build this `FullLaneId` in the implementation (`DeliveryConfirmationPaymentsAdapter`) when calling `register_relayer_rewards`
2) (very questionable) implementing (1) would allow us to use some reward/refund specific names (and comments) for this enum and struct. Since `FullLaneId` is used to compute account id that is used to pay rewards, then maybe rename `FullLaneId` to `RewardsAccountParams` and `LaneDirection` to `RewardsSource { ThisChainPaysRewards, BridgedChainPaysRewards }`. Why: the same problem isn't trivial (with who's paying for what), so imo we need to use all means to make it visible to ourselves and our crate users (at least for relayers)
3) in any case, let's add some docs (maybe to this srtuct), explaining why we are using bridged chain id and lane direction to compute rewards-source-account. Also would be great to have some extended docs for the `claim_rewards`, so that relayers know exactly what they need to pass to the call from its docs. Like: "if you have been submitting messages proof to this chain - pass <this>, if you have been submitting messages proof to the other chain - pass <this>, if you have been submitting messages **delivery** proof to this chain - pass <this>, ..."

svyatonik (approved):
All looks good, but please review my suggestions - maybe some of those worth implementing

svyatonik:
@serban300 Sorry, I wanted to answer your questions before hitting approve, but had to switch contexts and totally forgot about that :facepalm: 

> This approach might make it more complicated for the relayer to claim rewards. It will have to execute a claim_rewards call for each (lane_id, bridged_chain_id, lane_direction). Could this be an inconvenience ? Should we make it easier for the relayer ? Maybe modify claim_rewards in order to handle LaneDirection::In and LaneDirection::Out at once ?

But that would mean two transfers in a single tx, right? Because for `LaneDirection::In` it'll be one reward-source-account and for `LaneDirection::Out` it'll be another. If so, then I'd prefer to keep it separated - it isn't that big overhead imo,.

> What if the account that pays the reward doesn't have funds at step 2 ? Can this occur ? It may not have funds in the first place, or it may have funds when registering the rewards, and spend them until the relayer claims the reward. Should / could we handle this ? Maybe freeze some funds at step 1 ?

Yes - we may implement some kind of freezing. But IMO what we have now is enough:
1) the only way to withdraw funds from `relayers-reward-source` is to call `claim_rewards` of `pallet-bridge-relayers` (apart from governance calls ofc);
2) so relayers may know in advance that the account don't have enough funds to pay its rewards => they can stop at any time;
3) bridge maintainers are interested in keeping bridge operational, so they'll try to keep the bridge operational by keeping enough funds at `relayers-reward-source`.

Once any player doesn't do what he need to do, the other side may stop its operations too. IMO that is enough.

We may add some monitoring tools for relayers - they should cause alert if there aren't enough funds at the source account. But i.e. for our Statemint <> Statemine bridge we are not going to halt relayers in this case.

That said, this all may be just my wrong impression, so feel free to argue

primitives/messages/src/lib.rs
serban300: All the suggestions look good. Thanks ! Tried to implement all of them.

relays/lib-substrate-relay/src/messages_metrics.rs
```diff
@@ -88,10 +88,23 @@ where
 					bp_relayers::RelayerRewardsKeyProvider::<AccountIdOf<C>, BalanceOf<C>>::final_key(
 						relayers_pallet_name,
 						account.id(),
-						lane,
+						&FullLaneId::new(*lane, BC::ID, LaneDirection::In),
 					),
-					format!("at_{}_relay_{}_reward_for_lane_{}_with_{}", C::NAME, account.tag(), hex::encode(lane.as_ref()), BC::NAME),
-					format!("Reward of the {} relay account for serving lane {:?} with {} at the {}", account.tag(), lane, BC::NAME, C::NAME),
+					format!("at_{}_relay_{}_reward_for_in_lane_{}_with_{}", C::NAME, account.tag(), hex::encode(lane.as_ref()), BC::NAME),
```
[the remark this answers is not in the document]
serban300: Done.

serban300:
> @serban300 Sorry, I wanted to answer your questions before hitting approve, but had to switch contexts and totally forgot about that facepalm
> 
> > This approach might make it more complicated for the relayer to claim rewards. It will have to execute a claim_rewards call for each (lane_id, bridged_chain_id, lane_direction). Could this be an inconvenience ? Should we make it easier for the relayer ? Maybe modify claim_rewards in order to handle LaneDirection::In and LaneDirection::Out at once ?
> 
> But that would mean two transfers in a single tx, right? Because for `LaneDirection::In` it'll be one reward-source-account and for `LaneDirection::Out` it'll be another. If so, then I'd prefer to keep it separated - it isn't that big overhead imo,.
> 
> > What if the account that pays the reward doesn't have funds at step 2 ? Can this occur ? It may not have funds in the first place, or it may have funds when registering the rewards, and spend them until the relayer claims the reward. Should / could we handle this ? Maybe freeze some funds at step 1 ?
> 
> Yes - we may implement some kind of freezing. But IMO what we have now is enough:
> 
>     1. the only way to withdraw funds from `relayers-reward-source` is to call `claim_rewards` of `pallet-bridge-relayers` (apart from governance calls ofc);
> 
>     2. so relayers may know in advance that the account don't have enough funds to pay its rewards => they can stop at any time;
> 
>     3. bridge maintainers are interested in keeping bridge operational, so they'll try to keep the bridge operational by keeping enough funds at `relayers-reward-source`.
> 
> 
> Once any player doesn't do what he need to do, the other side may stop its operations too. IMO that is enough.
> 
> We may add some monitoring tools for relayers - they should cause alert if there aren't enough funds at the source account. But i.e. for our Statemint <> Statemine bridge we are not going to halt relayers in this case.
> 
> That said, this all may be just my wrong impression, so feel free to argue

@svyatonik 

Thanks for the answers. Sounds good to me. If the relayers can know in advance if an account doesn't have funds and halt, this should be enough.

Also, could you PTAL when you have time ? There are a lot of changes from the previous version of this PR.

svyatonik (approved):
Looks great, thank you!

PR #1916, opened 2023-02-27
Fix multiple parachain headers submission for single message delivery

svyatonik:
See multiple `submitParachainHeads` transactions in this comment from Branislav: https://github.com/paritytech/parity-bridges-common/issues/1677#issuecomment-1435292502

The problem is that relay gets incorrect understanding of which source-parachain header is now available at the target-chain and thinks that there are headers that he needs to deliver. The reason is that we use by-relay-chain-block-number-ordering for parachain headers and there's a place where relay gets "mocked" state: `{ at_source_relay_block_number: 100, at_source_para_head_hash: <hash-of-para-head-50>, at_target_relay_block_number: 200, at_target_para_head_hash: <hash-of-para-head-60> }`, but since hashes are different, it thinks that it needs to submit more parachain heads to target chain.

There are two options how to fix that:
- https://github.com/paritytech/parity-bridges-common/pull/1916/commits/ef652f0648010ec61ca6c9878a4705291d7c76af - is a simple, hacky solution that fixes the problem. But still - there's some time where relay still may get into this wrong view of chains-state and submit extra transactions;
- all other commits are the new solution (which also reverts the previous one) - it is about removing support of multiple parachains from single parachains-relay instance. Because of that, we are no longer using the by-relay-chain-block-number-ordering for parachain headers - we can use natural ordering instead (because we have single parachain-block-number type and can decode parachain heads). So it fixes the issue naturally.

TODOs:
- [x] test it on Rialto<>Millau bridge;
- [x] test it on local RBH<>WBH bridge.

### DIFF
--- a/relays/lib-substrate-relay/src/parachains/target.rs
+++ b/relays/lib-substrate-relay/src/parachains/target.rs
@@ -24,16 +24,13 @@ use crate::{
 };
 
 use async_trait::async_trait;
-use bp_parachains::{BestParaHeadHash, ImportedParaHeadsKeyProvider, ParasInfoKeyProvider};
 use bp_polkadot_core::parachains::{ParaHash, ParaHeadsProof, ParaId};
 use bp_runtime::HeaderIdProvider;
 use codec::Decode;
-use parachains_relay::{
-	parachains_loop::TargetClient, parachains_loop_metrics::ParachainsLoopMetrics,
-};
+use parachains_relay::parachains_loop::TargetClient;
 use relay_substrate_client::{
 	AccountIdOf, AccountKeyPairOf, BlockNumberOf, Chain, Client, Error as SubstrateError, HashOf,
-	HeaderIdOf, ParachainBase, RelayChain, TransactionEra, TransactionTracker, UnsignedTransaction,
+	HeaderIdOf, ParachainBase, TransactionEra, TransactionTracker, UnsignedTransaction,
 };
 use relay_utils::{relay_loop::Client as RelayClient, HeaderId};
 use sp_core::{Bytes, Pair};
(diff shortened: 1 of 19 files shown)


relays/parachains/src/parachains_loop.rs
```diff
@@ -129,48 +96,47 @@ pub trait TargetClient<P: ParachainsPipeline>: RelayClient {
...
 		proof: ParaHeadsProof,
 	) -> Result<Self::TransactionTracker, Self::Error>;
 }
 
 /// Return prefix that will be used by default to expose Prometheus metrics of the parachains
 /// sync loop.
 pub fn metrics_prefix<P: ParachainsPipeline>() -> String {
-	format!("{}_to_{}_Parachains", P::SourceChain::NAME, P::TargetChain::NAME)
+	format!(
+		"{}_to_{}_Parachains_{}",
```
serban300, line 124: Nit: Would it make sense to use something like:
```
	format!(
		"{}_Parachain_{}_to_{}",
		P::SourceRelayChain::NAME,
		P::SourceParachain::PARACHAIN_ID,
		P::TargetChain::NAME,
	)
``` 
?

serban300 (approved):

relays/lib-substrate-relay/src/parachains/target.rs
```diff
@@ -92,11 +89,11 @@ where
 		Ok(best_id)
 	}
 
-	async fn best_finalized_source_block(
+	async fn best_finalized_source_relay_chain_block(
 		&self,
 		at_block: &HeaderIdOf<P::TargetChain>,
 	) -> Result<HeaderIdOf<P::SourceRelayChain>, Self::Error> {
-		let encoded_best_finalized_source_block = self
+		let encoded_best_finalized_source_relay_block = self
```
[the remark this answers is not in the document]
svyatonik: We should probably make regular`state_call` private :) I totally forgot that we have this typed call

relays/parachains/src/parachains_loop.rs
svyatonik: Actually I'd prefer similar relays (if we'll ever have it) to have similar prefix - i.e. `Rialto_to_Millau_Parachains_1000` and `Rialto_to_Millau_Parachains_2000` vs `Rialto_Parachain_1000_to_Millau` and `Rialto_Parachain_2000_to_Millau`. But let's revisit it later

PR #1922, opened 2023-03-01
Update BHR and BHW spec version

serban300:


scripts/verify-pallets-build.sh
```diff
@@ -66,6 +66,7 @@ rm -rf $BRIDGES_FOLDER/bin/rialto
 rm -rf $BRIDGES_FOLDER/bin/rialto-parachain
 rm -rf $BRIDGES_FOLDER/bin/.keep
 rm -rf $BRIDGES_FOLDER/deployments
+rm -rf $BRIDGES_FOLDER/docs
```
svyatonik, line 69: Are you sure about this? I thought that maybe it'll help someone to understand what is this strange `bridges` folder about :) But it is up to you :)

serban300: You're right. Maybe it's bnetter to kkep them. I chenged this to remove only `dockerhub-*`

bkontur (approved):

PR #1957, opened 2023-03-09
Fix invalid batch transaction

### LINKED ISSUE #1955: Invalid message delivery/confirmation proofs
Sometimes there's a situation when relay submits transaction with "invalid messages proof". This happens because messages proof contains invalid (parachain/relay) header reference...and this happens because of issue in futures synchronzation in the messages race. Fix is on the way

svyatonik:
closes #1955

Needs to be tested => draft

### DIFF
--- a/relays/lib-substrate-relay/src/lib.rs
+++ b/relays/lib-substrate-relay/src/lib.rs
@@ -91,18 +91,21 @@ impl<AccountId> TaggedAccount<AccountId> {
 }
 
 /// Batch call builder.
-pub trait BatchCallBuilder<Call>: Send {
+pub trait BatchCallBuilder<Call>: Clone + Send {
 	/// Create batch call from given calls vector.
 	fn build_batch_call(&self, _calls: Vec<Call>) -> Call;
 }
 
 /// Batch call builder constructor.
-pub trait BatchCallBuilderConstructor<Call> {
+pub trait BatchCallBuilderConstructor<Call>: Clone {
+	/// Call builder, used by this constructor.
+	type CallBuilder: BatchCallBuilder<Call>;
 	/// Create a new instance of a batch call builder.
-	fn new_builder() -> Option<Box<dyn BatchCallBuilder<Call>>>;
+	fn new_builder() -> Option<Self::CallBuilder>;
 }
 
 /// Batch call builder based on `pallet-utility`.
+#[derive(Clone)]
 pub struct UtilityPalletBatchCallBuilder<C: Chain>(PhantomData<C>);
 
 impl<C: Chain> BatchCallBuilder<C::Call> for UtilityPalletBatchCallBuilder<C>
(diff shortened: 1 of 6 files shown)


relays/messages/src/message_race_loop.rs
```diff
@@ -524,6 +607,12 @@ pub async fn run<P: MessageRace, SC: SourceClient<P>, TC: TargetClient<P>>(
 			source_client_is_online = false;
 
 			// if we've started to submit batch transaction, let's prioritize it
+			//
+			// we're using `take` here, because we don't need batch transaction (i.e. some
+			// underlying finality proof) anymore for our future calls - we were unable to
+			// use it for our current state, so why would we need to keep an obsolete proof
+			// for the future?
+			let target_batch_transaction = target_batch_transaction.take();
```
svyatonik, line 615: That's the key point of this PR - instead of just trying to use the prepared batch tx, we're "taking" it here. Because otherwise, it may be reused later to generate invalid proofs

relays/lib-substrate-relay/src/lib.rs
```diff
@@ -91,18 +91,21 @@ impl<AccountId> TaggedAccount<AccountId> {
...
+pub trait BatchCallBuilder<Call>: Clone + Send {
 	/// Create batch call from given calls vector.
 	fn build_batch_call(&self, _calls: Vec<Call>) -> Call;
 }
 
 /// Batch call builder constructor.
-pub trait BatchCallBuilderConstructor<Call> {
+pub trait BatchCallBuilderConstructor<Call>: Clone {
+	/// Call builder, used by this constructor.
+	type CallBuilder: BatchCallBuilder<Call>;
```
serban300, line 102: An alternative to this would be to add a method to `BatchCallBuilder`: 
`fn clone_box(&self) -> Box<dyn BatchCallBuilder<Call>>;`

And then:
```
impl<Call> Clone for Box<dyn BatchCallBuilder<Call>> {
	fn clone(&self) -> Box<dyn BatchCallBuilder<Call>> {
		self.clone_box()
	}
}
```

Also we would need to use `#[derive(CloneNoBound)]` instead of `#[derive(Clone)]` for `BatchProofTransaction`.

serban300 (approved):
Looks good ! I left just a small comment.

relays/lib-substrate-relay/src/lib.rs
svyatonik: Do you think it'd be better? I mean - I'm more in favor of associated types than `Box<dyn>` :) But if you think it'd be better - I could explore and fix it

serban300: Not sure which version is best. The associated type should be safe also since for the `()` implementation of `BatchCallBuilderConstructor`, `new_builder` will always return `None`. I feel like with the `Box<dyn>` it would be safest, but it's true that there are cons to this as well. 

Anyway, I'm ok with merging it like this (with associated types) and we can revisit the decision later.

PR #1962, opened 2023-03-13
Bump substrate/polkadot/cumulus

bkontur:


svyatonik (approved):

bin/rialto-parachain/runtime/src/lib.rs
```diff
@@ -588,7 +588,7 @@ construct_runtime!(
 		System: frame_system::{Pallet, Call, Storage, Config, Event<T>},
 		Timestamp: pallet_timestamp::{Pallet, Call, Storage, Inherent},
 		Sudo: pallet_sudo::{Pallet, Call, Storage, Config<T>, Event<T>},
-		RandomnessCollectiveFlip: pallet_randomness_collective_flip::{Pallet, Storage},
+		RandomnessCollectiveFlip: pallet_insecure_randomness_collective_flip::{Pallet, Storage},
```
bkontur, line 591: @svyatonik 
do you know what is this about and if we need this? I have no idea :)
I just saw that they removed it from statemine/t:  https://github.com/paritytech/cumulus/blob/master/parachains/runtimes/assets/statemint/src/lib.rs#L607

and leave it just for constracts-rococo: https://github.com/paritytech/cumulus/blob/master/parachains/runtimes/contracts/contracts-rococo/src/lib.rs#L335

bin/millau/runtime/src/lib.rs
```diff
@@ -234,23 +235,21 @@ impl pallet_aura::Config for Runtime {
 impl pallet_beefy::Config for Runtime {
```
bkontur, line 235: @acatangiu @svyatonik @serban300 
guys, can you please confirm this change for `pallet_beefy` pallet?

1. `MaxSetIdSessionEntries` ?
   0 or
   ```
   parameter_types! {
	pub const BeefySetIdSessionEntries: u32 = BondingDuration::get() * SessionsPerEra::get();
   }
   ```   
2. `EquivocationReportSystem = ()` ?

bin/millau/runtime/src/lib.rs
```diff
@@ -234,23 +235,21 @@ impl pallet_aura::Config for Runtime {
...
-		KeyTypeId,
-		GrandpaId,
-	)>>::IdentificationTuple;
-	type HandleEquivocation = ();
 	// TODO: update me (https://github.com/paritytech/parity-bridges-common/issues/78)
 	type WeightInfo = ();
 	type MaxAuthorities = ConstU32<10>;
 	type MaxSetIdSessionEntries = ConstU64<0>;
+	type KeyOwnerProof = <Historical as KeyOwnerProofSystem<(KeyTypeId, GrandpaId)>>::Proof;
+	type EquivocationReportSystem = ();
```
bkontur, line 252: @acatangiu @svyatonik @serban300 
guys, can you please confirm this change for `pallet_grandpa` pallet?

1. `EquivocationReportSystem = ()` ?

bin/millau/runtime/src/lib.rs
svyatonik: Sorry - my comment was for the `pallet_grandpa`

svyatonik: But I guess here it is the same. We'll probably even don't need this `pallet_session::historical` and may use `()` as `KeyOwnerProof` (if it compiles - at least it was previously)

bkontur: aha, ok, thank you, removing,
`()` does not work, but `sp_core::Void` does

acatangiu: yes, `sp_core::Void` is fine for millau and rialto, we don't expect or test equivocations there :) (both beefy and grandpa)

acatangiu (approved):

PR #1983, opened 2023-03-20
Remove deprecated code from bridge-runtime-common

### LINKED ISSUE #1946: Remove all unused code from Backport xcm bridging extensions to the bridge repo PR
Remove all unused code from #1813

svyatonik:
closes #1946 

The goal of this PR is to make our testnets use the same/similar XCM configuration as bridge hubs do. This allows us to remove deprecated messages pallet configuration code from the `bridge-runtime-common` crate.

Ready for review. I left some PR comments about important changes - ptal


bin/runtime-common/src/messages.rs
```diff
@@ -205,24 +172,16 @@ pub mod source {
...
-
-			// reject message if there are too many pending messages at this lane
-			let max_pending_messages = ThisChain::<B>::maximal_pending_messages_at_outbound_lane();
-			let pending_messages = lane_outbound_data
-				.latest_generated_nonce
-				.saturating_sub(lane_outbound_data.latest_received_nonce);
-			if pending_messages > max_pending_messages {
-				return Err(TOO_MANY_PENDING_MESSAGES)
-			}
+			// IMPORTANT: any error that is returned here is fatal for the bridge, because
```
svyatonik, line 180: I've removed max-pending-messages check here - it was set to `2^63` in RBH/WBH (iirc), so it doesn't work anyway.

relays/bin-substrate/src/cli/encode_message.rs
```diff
@@ -42,11 +43,16 @@ pub enum Message {
...
-	fn encode_send_xcm(
-		message: xcm::VersionedXcm<()>,
-		bridge_instance_index: u8,
+	/// Encode an `execute` XCM call of the XCM pallet.
+	fn encode_execute_xcm(
+		message: xcm::VersionedXcm<Self::Call>,
 	) -> anyhow::Result<EncodedOrDecodedCall<Self::Call>>;
+
+	/// Estimate value of `max_weight` argument for the `execute` XCM call of the XCM pallet.
+	fn estimate_execute_xcm_weight() -> Weight {
```
svyatonik, line 52: We're now using `execute` instead of `send` on testnets - it is easier in configuration and the XCM program is similar to what we actually have on bridge hubs.

bin/runtime-common/src/messages_xcm_extension.rs
```diff
@@ -28,13 +28,42 @@ use bp_messages::{
...
+	PartialEqNoBound,
+};
 use scale_info::TypeInfo;
+use sp_std::marker::PhantomData;
 use xcm_builder::{DispatchBlob, DispatchBlobError, HaulBlob, HaulBlobError};
 
 /// Plain "XCM" payload, which we transfer through bridge
 pub type XcmAsPlainPayload = sp_std::prelude::Vec<u8>;
 
+// TODO: below are just rough estimations. Other things also happen there (including hashing and so
```
svyatonik, line 42: @bkontur I've copypasted this `XcmRouterWeigher` from RBH/WBH code. I think we'll be using it for all bridge hubs, right? Could you please check my big comment below - am I right there?

Right now the biggest problem here is that it only returns the `ref_time` component and the `proof_size` is set to zero. Which is obviously wrong. I see two options to fix it:
- come up with some good estimation of the `proof_size` and return it here. Pros: easy and fast. Cons: inaccurate + we need some mechanism to verify that it is still actual (if something changes in queue implementation);
- remove this `XcmRouterWeigher` at all and add some benchmarks to the runtimes, which will measure the actual weight of the `queue.push()` (for us it'll be the `send_fragment` in the `cumulus-pallet-xcmp-queue`).

WDYT?

svyatonik:
The downside of this PR is that our XCM programs are no longer actually executed at the bridged chain, because we're using this new dispatcher (`BridgeBlobDispatcher`), which does `send`, not execute. But it isn't critical for our testnets - we only care about delivery here.

bin/runtime-common/src/messages_xcm_extension.rs
svyatonik: Submitted https://github.com/paritytech/parity-bridges-common/issues/1986 to track this

PR #1997, opened 2023-03-28
MaxRequests -> MaxFreeMandatoryHeadersPerBlock in pallet-bridge-grandpa

svyatonik:
related to https://github.com/paritytech/srlabs_findings/issues/265

This PR removes `MaxRequests` rate limiter from the `pallet-bridge-grandpa`. Instead the `MaxFreeMandatoryHeadersPerBlock` is added, which limits a number of free mandatory headers submit calls per block. Important change: this value doesn't span through multiple block - it is a limit for single block only. Why: I still not super-convinced that we need this value at all, and having per-block limit allows us to avoid storing any values in the runtime storage => avoid extra db reads and writes. The associated runtime storage value is getting killed from block `on_finalize`, so it never ends up in the storage trie. Feel free to argue if you think we need a limiter that spans through multiple blocks.

The open question is the constant value that we gonna set on our production bridge hubs. The minimal value is `parachain_block_time / bridged_relay_chain_block_time = 12s / 6s = 2`. If we want to account GRANDPA/parachain finality lags, it may be increased, say, to `4`, which is a good estimation imo. Please share your opinions if you have any other arguments

### DIFF
--- a/modules/grandpa/src/lib.rs
+++ b/modules/grandpa/src/lib.rs
@@ -189,7 +189,7 @@ pub mod pallet {
 			let is_authorities_change_enacted =
 				try_enact_authority_change::<T, I>(&finality_target, set_id)?;
 			let may_refund_call_fee = is_authorities_change_enacted &&
-				// if we are seeing too much mandatory headers, we don't want to refund
+				// if we have seen too many mandatory headers in this block, we don't want to refund
 				Self::free_mandatory_headers_remaining() > 0 &&
 				// if arguments out of expected bounds, we don't want to refund
 				submit_finality_proof_info_from_args::<T, I>(&finality_target, &justification)
--- a/modules/grandpa/src/lib.rs
+++ b/modules/grandpa/src/lib.rs
@@ -284,7 +284,7 @@ pub mod pallet {
 		}
 	}
 
-	/// Number mandatory headers that we have accepted in the current block for free (returning
+	/// Number mandatory headers that we may accept in the current block for free (returning
 	/// `Pays::No`).
 	///
 	/// If the `FreeMandatoryHeadersRemaining` hits zero, all following mandatory headers in the
--- a/modules/grandpa/src/lib.rs
+++ b/modules/grandpa/src/lib.rs
@@ -103,14 +103,17 @@ pub mod pallet {
 		/// The chain we are bridging to here.
 		type BridgedChain: ChainWithGrandpa;
 
-		/// The upper bound on the number of requests allowed by the pallet.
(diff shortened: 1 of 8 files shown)


acatangiu:
> it is a limit for single block only. Why: I still not super-convinced that we need this value at all, and having per-block limit allows us to avoid storing any values in the runtime storage

+1 on having this per-block and not saving to storage; limiting free mandatory headers per block is enough to prevent _free_ spam even if bridged chain validators are broken and are trying to spam authority changes.

> The minimal value is parachain_block_time / bridged_relay_chain_block_time = 12s / 6s = 2. If we want to account GRANDPA/parachain finality lags, it may be increased, say, to 4, which is a good estimation imo. Please share your opinions if you have any other arguments

4 is a perfectly reasonable value for it

modules/grandpa/src/lib.rs
```diff
@@ -273,16 +284,20 @@ pub mod pallet {
 		}
 	}
 
-	/// The current number of requests which have written to storage.
+	/// Number mandatory headers that we have accepted in the current block for free (returning
```
serban300, line 287: Nit: Isn't this the number of mandatory headers which we are still accepting in this block for free ?

modules/grandpa/src/lib.rs
```diff
@@ -273,16 +284,20 @@ pub mod pallet {
...
+	/// `Pays::No`).
 	///
-	/// If the `RequestCount` hits `MaxRequests`, no more calls will be allowed to the pallet until
-	/// the request capacity is increased.
+	/// If the `FreeMandatoryHeadersRemaining` hits zero, all following mandatory headers in the
+	/// current block are accepted with fee (`Pays::Yes` is returned).
 	///
-	/// The `RequestCount` is decreased by one at the beginning of every block. This is to ensure
-	/// that the pallet can always make progress.
+	/// The `RequestCount` is an ephemeral value that is set to `MaxFreeMandatoryHeadersPerBlock`
```
serban300, line 293: Leftover probably `RequestCount`

modules/grandpa/src/lib.rs
```diff
@@ -185,9 +189,16 @@ pub mod pallet {
 			let is_authorities_change_enacted =
 				try_enact_authority_change::<T, I>(&finality_target, set_id)?;
 			let may_refund_call_fee = is_authorities_change_enacted &&
+				// if we are seeing too much mandatory headers, we don't want to refund
```
serban300, line 192: Typo: too many.

serban300 (approved):
I think it's a very good idea to have this limit per block

PR #2003, opened 2023-03-30
Fix test step on CI

svyatonik:
https://gitlab.parity.io/parity/mirrors/parity-bridges-common/-/jobs/2615134

#1998 doesn't help (inside `paritytech/bridges-ci:staging`), so I'm gonna change the command to skip building  all WASM runtimes. I've checked locally with docker image and now test work.

There's this other `benchmark-test` issue (https://gitlab.parity.io/parity/mirrors/parity-bridges-common/-/jobs/2615138). According to the error message, it fails to decode WASM blob (unknown wasm opcode - 192). Have no idea - whether rust now actually emits this `i32.extend8_s` instruction and we haven't updated our compilators/interpreters refs, or it is something else. Needs investigation


serban300 (approved):

svyatonik:
Regarding benchmarks: https://github.com/paritytech/substrate/issues/13636

PR #2017, opened 2023-04-04
Remove unneeded error debug strings

serban300:
Small follow up on #1813. We can use an `Option<DispatchBlobError>` instead of `&'static str` and avoid defining hardcoded strings. 

However not sure if this is needed. @bkontur do we need `NotDispatched` to contain the `DispatchBlobError` ? Normally `XcmBlobMessageDispatchResult` can be saved in a runtime event. But since `DispatchBlobError` is marked with `#[codec(skip)]` I think the `DispatchBlobError` won't be saved anyway. And it doesn't seem to be checked anywhere.

### DIFF
--- a/bin/runtime-common/src/messages_xcm_extension.rs
+++ b/bin/runtime-common/src/messages_xcm_extension.rs
@@ -42,7 +42,7 @@ pub type XcmAsPlainPayload = sp_std::prelude::Vec<u8>;
 pub enum XcmBlobMessageDispatchResult {
 	InvalidPayload,
 	Dispatched,
-	NotDispatched(#[codec(skip)] &'static str),
+	NotDispatched(#[codec(skip)] Option<DispatchBlobError>),
 }
 
 /// [`XcmBlobMessageDispatch`] is responsible for dispatching received messages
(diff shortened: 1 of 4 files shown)


bin/runtime-common/src/messages_xcm_extension.rs
```diff
@@ -42,7 +42,7 @@ pub type XcmAsPlainPayload = sp_std::prelude::Vec<u8>;
 pub enum XcmBlobMessageDispatchResult {
 	InvalidPayload,
 	Dispatched,
-	NotDispatched(#[codec(skip)] &'static str),
+	NotDispatched(#[codec(skip)] Option<DispatchBlobError>),
```
bkontur, line 45: @serban300 
so, can we get rid of `skip` and `Option` here now? if so, lets clean it
or do we use `NotDispatched(None)` somewhere?

serban300: Yes, I double checked. The events are stored in an encoded form, so if we do `#[codec(skip)]`, the `DispatchBlobError` won't be saved anyway. And it doesn't seem to be used elsewhere. I removed it. PTAL !

bkontur: I meant to leave it as:
`NotDispatched(Option<DispatchBlobError>)`
or 
`NotDispatched(DispatchBlobError)`

serban300: Oh, sorry, I misunderstood. `NotDispatched(Option<DispatchBlobError>)` and `NotDispatched(DispatchBlobError)` don't work. Because `DispatchBlobError` doesn't implement `Encode/Decode`. The only thing we could do here is to derive `Encode, Decode` for `DispatchBlobError` in the `Polkadot` repo. I can create a PR for that and see if it is accepted. But until then, would it be ok to use `NotDispatched(#[codec(skip)] Option<DispatchBlobError>)` ? Functionally, it's the same as now, but this way we can remove the error strings.

bkontur: oki, lets remove &str stuff, and why do we need `Option`?

serban300: Ok. Done. Reverted the 2nd commit. PTAL.

> and why do we need Option

Deriving `Decode` needs a type that implements `Default`. Because for example when we derives `Decode` for `NotDispatched(#[codec(skip)] Option<DispatchBlobError>)`, since `Option<DispatchBlobError>` is skipped at encoding time, it won't be present at decoding time and we wouldn't know what kind of `DispatchBlobError` to create. So we creates a `NotDispatched(Default::default())`. and  `DispatchBlobError` doesn't implement `Default` either.

bkontur (approved):

PR #2019, opened 2023-04-04
Only refund relayer if all bundled messages have been delivered/confirmed

svyatonik:
Otherwise relayer may get refund for messages that are rejected. E.g. if target chain is ready to accept `10` messages (because of unrewarded relayers limits), the relayer can try to deliver `100` messages and fill block for free (because we'll do refund).


acatangiu (approved):

Polkadot-Forum:

https://forum.polkadot.network/t/bridge-hub-questions/2693/2

PR #2021, opened 2023-04-04
Reject delivery transactions with at least one obsolete message

svyatonik:
Given that we do refund message delivery transactions cost to relayers, it makes no sense to only accept transactions with obsolete message - they'll occupy portion of block size limits for free (because of refund).


serban300 (approved):

bin/runtime-common/src/messages_call_ext.rs
```diff
@@ -21,18 +21,24 @@ use bp_messages::{LaneId, MessageNonce};
...
+	/// Nonce of the best message, stored by this chain before the call is dispatched.
 	pub best_stored_nonce: MessageNonce,
 }
 
 impl BaseMessagesProofInfo {
+	/// Returns true if `bundled_range` cannot be directly appended to the `best_stored_nonce`
+	/// or if the `bundled_range` is empty.
 	fn is_obsolete(&self) -> bool {
-		self.best_bundled_nonce <= self.best_stored_nonce
+		*self.bundled_range.start() != self.best_stored_nonce + 1 || self.bundled_range.is_empty()
```
acatangiu, line 41: this may be dumb, I haven't followed **all** path, but afaict we allow `FromBridgedChainMessagesProof` with `nonces_end < nonces_start` => here we might "wrongly" report `is_obsolete == false` - which may also be fine in that case since no message will be included ; not sure, just wanted to bring it to your attention

acatangiu (approved):
Apart from condition with comment to check, everything looks good

bin/runtime-common/src/messages_call_ext.rs
svyatonik: You're right, thanks for spotting that. I've changed code a bit. Now we allow empty message delivery transactions but only if there's no room left in the unrewaded relayers vector. So we allow only when lane is blocked. Honest relayers don't want to block the lane, so they'll include reward confirmation in regular message delivery transactions.

svyatonik: Actually I think it needs to be implemented in other way: We need to reject delivery transaction with the message that occupies the last free entry in unrewarded relayers vector if this transaction doesn't include rewards confirmation. In current impl, other relayer may submit empty delivery tx with larger tip and the other relayer won't be refunded for the same information (lose his funds).

svyatonik: > In current impl, other relayer may submit empty delivery tx with larger tip and the other relayer won't be refunded for the same information (lose his funds).

This is wrong - the second transaction will be considered stale, because `is_obsolete` would return `true` in this way, so it won't be included into the block => no lose-his-funds. OTOH if we'll forbid empty delivery transactions, we'll allow single relayer to block the lane (that's the reason why we have such transactions). So current code should be fine. PTAL when you have a time @acatangiu

svyatonik: Filed #2026 to fix relayer code if this PR is accepted

svyatonik: Another fix for this PR: right now it would accept empty delivery transactions when there's no empty entries in the unrewarded relayers vec. But there's another constraint, which limits total number of unconfirmed messages (e.g. there may be 1 relayer that has delivered 4096 messages and the lane is blocked until rewards are confirmed. So the empty-delivery-transaction allowance condition needs to be fixed

svyatonik: Fixed, please take a look

acatangiu (approved):
Nice! :ok_hand:

serban300 (commented):
Sorry, I left a few more small comments. I missed them during the first iteration.

svyatonik:
@serban300 Thanks, I'll make a follow-up PR to implement your suggesions

Polkadot-Forum:

https://forum.polkadot.network/t/bridge-hub-questions/2693/2

PR #2027, opened 2023-04-07
Reintroduce msg dispatch status reporting

### LINKED ISSUE #2022: Report message dispatch status to source chain
Related to #1944

I think we might need to report the message dispatch status from target chain to source chain.

For example for the flow `Rockmine -> BridgeHub Rococo -> BridgeHubWococo -> Wockmint` we might have an XCM message that reached `BridgeHubWococo`, but the dispatch failed. In this case the assets won't be trapped anywhere (more details in [this comment](https://github.com/paritytech/parity-bridges-common/issues/1944#issuecomment-1495939380)) . I think it would be good if we could report the failed dispatch to `BridgeHub Rococo` so that `BridgeHub Rococo` could then send a `QueryResponse` back to `Rockmine` to somehow unlock the assets there.

@svyatonik do you think this would make sense ?

serban300:
Fixes #2022

Mainly reverting #1660 

Later we will also need to link this status to the original XCM message received by the source bridge hub and send some response based on it. But I'm not sure how for the moment. Will do it in a separate issue.

### DIFF
--- a/modules/messages/src/lib.rs
+++ b/modules/messages/src/lib.rs
@@ -527,7 +527,7 @@ pub mod pallet {
 		MessagesReceived(
 			Vec<
 				ReceivedMessages<
-					<T::MessageDispatch as MessageDispatch<T::AccountId>>::DispatchLevelResult,
+					<T::MessageDispatch as MessageDispatch<T::AccountId>>::DispatchError,
 				>,
 			>,
 		),
--- a/modules/messages/src/lib.rs
+++ b/modules/messages/src/lib.rs
@@ -66,7 +66,7 @@ use bp_messages::{
 use bp_runtime::{BasicOperatingMode, ChainId, OwnedBridgeModule, Size};
 use codec::{Decode, Encode, MaxEncodedLen};
 use frame_support::{dispatch::PostDispatchInfo, ensure, fail, traits::Get};
-use sp_runtime::traits::UniqueSaturatedFrom;
+use sp_runtime::{traits::UniqueSaturatedFrom, SaturatedConversion};
 use sp_std::{cell::RefCell, marker::PhantomData, prelude::*};
 
 mod inbound_lane;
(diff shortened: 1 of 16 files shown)


modules/messages/src/lib.rs
```diff
@@ -1700,8 +1706,8 @@ mod tests {
 			.actual_weight
 			.unwrap();
 			assert!(
-				post_dispatch_weight.proof_size() < pre_dispatch_weight.proof_size(),
-				"Expected post-dispatch PoV {} to be less than pre-dispatch PoV {}",
+				post_dispatch_weight.proof_size() <= pre_dispatch_weight.proof_size(),
```
svyatonik, line 1709: Why this is changed? The test is called `proof_size_refund_from_receive_messages_proof_works` - it should test that the `post_dispatch_weight.proof_size()` is less than the `pre_dispatch_weight.proof_size()`. Otherwise, we should remove it :)

modules/messages/src/lib.rs
```diff
@@ -807,16 +807,14 @@ impl<T: Config<I>, I: 'static> RuntimeInboundLaneStorage<T, I> {
...
+	/// constants from the pallet configuration. The PoV of the call includes the maximal size
+	/// of the inbound lane state. If the actual size is smaller, we may subtract extra bytes
+	/// from this component.
 	pub fn extra_proof_size_bytes(&self) -> u64 {
 		let max_encoded_len = StoredInboundLaneData::<T, I>::max_encoded_len();
-		let relayers_count = self.data().relayers.len();
-		let actual_encoded_len =
-			InboundLaneData::<T::InboundRelayer>::encoded_size_hint(relayers_count)
-				.unwrap_or(usize::MAX);
+		let actual_encoded_len = self.data().encoded_size();
```
svyatonik, line 817: I'd still use `encoded_size_hint`. `encoded_size` does real encoding, which takes time and do allocations. And we're in runtime here => performance really matters

svyatonik (approved):
LGTM, but please double check the `proof_size_refund_from_receive_messages_proof_works` test - imo the strict less should remain there, or the test needs to be fixed/removed. Otherwise it doesn't actually test anything

modules/messages/src/lib.rs
serban300: Sorry, you're right. It was related to the fact that we didn't take into account the number of messages here. Since before it didn't matter. Fixed.

modules/messages/src/lib.rs
serban300: Done

PR #2036, opened 2023-04-12
Impl review suggestions from #2021

svyatonik:
https://github.com/paritytech/parity-bridges-common/pull/2021#discussion_r1162728894
https://github.com/paritytech/parity-bridges-common/pull/2021#discussion_r1162729997

### DIFF
--- a/bin/runtime-common/src/messages_call_ext.rs
+++ b/bin/runtime-common/src/messages_call_ext.rs
@@ -310,8 +310,6 @@ fn unrewarded_relayers_occupation<T: Config<I>, I: 'static>(
 		free_relayer_slots: T::MaxUnrewardedRelayerEntriesAtInboundLane::get()
 			.saturating_sub(inbound_lane_data.relayers.len() as MessageNonce),
 		free_message_slots: {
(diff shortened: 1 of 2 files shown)


serban300 (approved):

bin/runtime-common/src/messages_call_ext.rs
```diff
@@ -57,45 +54,53 @@ pub struct UnrewardedRelayerOccupation {
...
-
-			return true
+				self.unrewarded_relayers.free_message_slots == 0;
+			
+			return !empty_transactions_allowed;
 		}
 
 		// otherwise we require bundled messages to continue stored range
-		*self.bundled_range.start() != self.best_stored_nonce + 1
+		*self.base.bundled_range.start() != self.base.best_stored_nonce + 1
```
serban300, line 88: Nit: Maybe we could extract `*self.base.bundled_range.start() != self.base.best_stored_nonce + 1` in another method let's say `BaseMessagesProofInfo::is_obsolete()` since it's used both here and for `ReceiveMessagesDeliveryProofInfo`. In order to avoid duplication.

serban300 (approved):

PR #2037, opened 2023-04-12
Define `RangeInclusiveExt`

### LINKED ISSUE #2032: Introduce function/trait method to compute number of messages in the range
https://github.com/paritytech/parity-bridges-common/pull/2031#pullrequestreview-1378508881

`ExactSizeIterator` is not implemented for `RangeInclusive<u64>`, so we can't use `.len()`. `.count()` is probably ineffective (but actually maybe rust compiler is smart enough to optimize that?). So we're using construct like this:
```rust
					received_range
						.end()
						.checked_sub(*received_range.start())
						.and_then(|x| x.checked_add(1))
						.unwrap_or(<something-else>),
```

Let's extract all except the `unwrap_or()` to a separate function/trait method and use it across the code.

serban300:
Fixes #2032

### DIFF
--- a/primitives/runtime/src/lib.rs
+++ b/primitives/runtime/src/lib.rs
@@ -523,7 +523,9 @@ impl<T> Debug for StrippableError<T> {
 	}
 }
 
+/// A trait defining helper methods for `RangeInclusive` (start..=end)
 pub trait RangeInclusiveExt<Idx> {
+	/// Computes the length of the `RangeInclusive`, checking for underflow and overflow.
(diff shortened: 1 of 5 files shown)


primitives/runtime/src/lib.rs
```diff
@@ -523,6 +523,21 @@ impl<T> Debug for StrippableError<T> {
 	}
 }
 
+pub trait RangeInclusiveExt<Idx> {
```
svyatonik, line 526: Please add some docs for public trait and method

svyatonik (approved):

PR #2048, opened 2023-04-17
Revert dispatch-results

svyatonik:
@serban300 we've discussed that we don't need it anymore, right? I'm looking at PRs for next audit portion and #2027 is marked as "AuditRequired" => if we don't need it, then let's revert?


acatangiu (approved):

bkontur (approved):

serban300 (approved):
Thanks! And sorry, it was on my list, but didn't get to it yet.

PR #2066, opened 2023-04-24
[relay-substrate-client] Bump jsonrpsee

serban300:
Part of #2062 

Buming jsonrpsee only for `relay-substrate-client` for the moment. 

In order to also bump it for `millau` and `rialto-parachain` we need substrate to bump it first.

### DIFF
--- a/relays/client-substrate/src/error.rs
+++ b/relays/client-substrate/src/error.rs
@@ -138,13 +138,11 @@ impl Error {
 impl MaybeConnectionError for Error {
 	fn is_connection_error(&self) -> bool {
 		match *self {
-			Error::RpcError(RpcError::Transport(_))
-				// right now if connection to the ws server is dropped (after it is already established),
-				// we're getting this error
-				| Error::RpcError(RpcError::Internal(_))
-				| Error::RpcError(RpcError::RestartNeeded(_))
-				| Error::ClientNotSynced(_) => true,
-			Error::FailedToReadBestFinalizedHeaderHash { ref error, .. } => error.is_connection_error(),
+			Error::RpcError(RpcError::Transport(_)) |
+			Error::RpcError(RpcError::RestartNeeded(_)) |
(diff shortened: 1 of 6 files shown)


relays/client-substrate/src/error.rs
```diff
@@ -138,13 +138,11 @@ impl Error {
 impl MaybeConnectionError for Error {
 	fn is_connection_error(&self) -> bool {
 		match *self {
-			Error::RpcError(RpcError::Transport(_))
-				// right now if connection to the ws server is dropped (after it is already established),
-				// we're getting this error
-				| Error::RpcError(RpcError::Internal(_))
```
svyatonik, line 144: There's no internal anymore, or has it been renamed?

svyatonik (approved):
LGTM, Two things:
- we may want to ask dependabot to ignore this dependency, because we want to use the same version as the referenced Substrate does. wdyt?;
- please check if this `Internal` variant has gone, or has been renamed - we need to be careful with this `is_connection_error`

relays/client-substrate/src/error.rs
serban300: It was removed: https://github.com/paritytech/jsonrpsee/commit/1a2a19906ae00b9e19d67b14595610e7fb98091d#diff-363a23d98a6c2addecef2665cf68deec178a1dc19d7b625439d4cc50f7f4af50

serban300:
> we may want to ask dependabot to ignore this dependency, because we want to use the same version as the referenced Substrate does. wdyt?;

For Millau and RialtoParachain I think it would be good to ignore this dependency. Usually we can't update it there if substrate wasn't updated. For the `relay-substrate-client` I don't know. I'm afraid we might forget to update it when updating substrate. 

I can also mark this PR as draft and come back to it after substrate is updated in order to update everything at once.

What do you think would be best ?

PR #2072, opened 2023-04-25
Bump `subxt` and `subxt-codegen` to 0.28.0

serban300:
Closes #2045

The latest version of `subxt-codegen` exposes the possibility to:
- not add comments to the generated code
- only generate the runtime, without the subxt API
- not use the default derives

So we take advantage of this


svyatonik (approved):
-8k lines looks good :)

PR #2075, opened 2023-04-26
Fixed off-by-one when confirming rewards in messages pallet

svyatonik:
For example when messages `235..=363` are delivered by single relayer and only message `235` reward is confirmed, then the `235` was not removed from delivered messages and total number unrewarded messages was above the limit (e.g. `129` vs `128`).

### DIFF
--- a/modules/messages/src/inbound_lane.rs
+++ b/modules/messages/src/inbound_lane.rs
@@ -548,7 +548,7 @@ mod tests {
 	}
 
 	#[test]
-	fn no_off_by_one_when_first_message_of_single_relayer_is_confirmed() {
+	fn first_message_is_confirmed_correctly() {
 		run_test(|| {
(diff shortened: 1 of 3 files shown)


modules/messages/src/inbound_lane.rs
```diff
@@ -545,4 +546,29 @@ mod tests {
 			);
 		});
 	}
+
+	#[test]
+	fn no_off_by_one_when_first_message_of_single_relayer_is_confirmed() {
```
serban300, line 551: Nit: I would rename this to something like `first_message_of_single_relayer_is_confirmed_correctly`

serban300 (approved):

modules/messages/src/inbound_lane.rs
svyatonik, line 551:
```suggestion
	fn first_message_is_confirmed_correctly() {
```

PR #2080, opened 2023-04-26
Removed relayer_account: &AccountId from MessageDispatch 

bkontur:
because we use it as :

```
relayer_at_this_chain: &AccountId,
...
...

let dispatch_result = Dispatch::dispatch(
			relayer_at_this_chain,
```

### DIFF
--- a/bin/runtime-common/src/messages_xcm_extension.rs
+++ b/bin/runtime-common/src/messages_xcm_extension.rs
@@ -26,7 +26,7 @@ use bp_messages::{
 	target_chain::{DispatchMessage, MessageDispatch},
 	LaneId,
 };
-use bp_runtime::{messages::MessageDispatchResult, AccountIdOf, Chain};
+use bp_runtime::messages::MessageDispatchResult;
 use codec::{Decode, Encode};
 use frame_support::{dispatch::Weight, CloneNoBound, EqNoBound, PartialEqNoBound};
 use pallet_bridge_messages::WeightInfoExt as MessagesPalletWeights;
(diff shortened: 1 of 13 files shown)


bin/runtime-common/src/messages_xcm_extension.rs
```diff
@@ -46,23 +46,13 @@ pub enum XcmBlobMessageDispatchResult {
...
 /// [`XcmBlobMessageDispatch`] is responsible for dispatching received messages
-pub struct XcmBlobMessageDispatch<SourceBridgeHubChain, TargetBridgeHubChain, DispatchBlob, Weights>
-{
-	_marker: sp_std::marker::PhantomData<(
-		SourceBridgeHubChain,
-		TargetBridgeHubChain,
-		DispatchBlob,
-		Weights,
-	)>,
+pub struct XcmBlobMessageDispatch<RelayerAccountChain, DispatchBlob, Weights> {
```
svyatonik, line 49: Relayer has accounts on both chains, so imo `SourceChain` or `BridgedChain` would be better. We also may remove this argument (`_relayer_account`) at all, if it isn't required for XCM dispatch.

svyatonik (approved):
LGTM

bin/runtime-common/src/messages_xcm_extension.rs
svyatonik: Sorry - I mean `TargetChain` or `ThisChain`, not the `SourceChain`/`BridgedChain` :)

bkontur: well, we dont use it here
and [here](https://github.com/paritytech/parity-bridges-common/blob/master/primitives/messages/src/target_chain.rs#L107-L110) is writtern: 
```
        /// If your configuration allows paying dispatch fee at the target chain, then
	/// it must be paid inside this method to the `relayer_account`.
	fn dispatch(
		relayer_account: &AccountId,
```

but we call this in extrinsic, where relayer_account pays for that and after dispatch we do rewards
`T::DeliveryPayments::pay_reward(`

so I am not sure, can we remove `relayer_account: &AccountId,` from dispatch trait?

svyatonik: I think so - our rewards mechanism has changed significantly and now we're going to support XCM messaging only. So let's remove imo (as well as this obsolete comment). Feel free to merge as is - we may remove later

svyatonik: > what is target chain chain [here](https://github.com/paritytech/parity-bridges-common/blob/master/primitives/messages/src/target_chain.rs#L107-L110) ? is it BridgeHub who does routing (ump, hrmp)?

Target chain is the chain where `dispatch` method is called. In current architecture it is the target bridge hub (who's receiving messages).

> BridgeBlobDispatcher does not count with paying anything let _ = send_xcm::<Router>

Yes, that's why I'm proposing to remove this argument and comment at all

acatangiu: I agree, it'd be better to be specific which chain is used - relayer has account on both chains (both ends of the bridge) so `RelayerAccountChain` is ambiguous.

PR #2089, opened 2023-04-28
Add weight of refund extension post_dispatch to the weights of messages pallet

### LINKED ISSUE #2024: Add weight of our signed extensions to the messages pallet calls
So bridge hub will have our signed extensions that reject obsolete transactions and pay reward/slash relayers. The obsolete-check reads the same storage entries as the call itself does, so we assume it brings no additional weight (it does, but not that much). However, the `post_dispatch` may add/update entries of the `pallet-bridge-relayers` rewards map. So we should add this weight to our call weights (delivery and confirmation). Maybe something as simple as `DbWeight.reads_writes(1, 1).set_proof_size(<proof-size-that-we-may-get-from-relayers-pallet-benchmarks>)`. But maybe some benchmark is ok too.

svyatonik:
closes #2024 

So the current issue is that: we do some non-trivial operations in our signed extension (like rewarding and slashing). But the weight of such operations is just ignored. 

The solution: our signed extension doesn't look like regular signed extension, which performs same operations for all extrinsics. So we can't add its weight to the base extrinsic weight. It looks like the part of delivery/confirmation transactions. So let's add its weight to those calls weights.

Operations that the extension performs in its `pre_dispatch` method are either trivial, or are touching the same storage entries that the call itself does (lane state and relayer registration). And after reads are performed, they're read from the cache (afair). So we may say that the weight of `pre_dispatch` is zero.

The `post_dispatch` either slashes the relayer, or registers its reward. So  for delivery transaction we add extra weight: `max(slash-relayer-weight, reward-relayer-weight)`. Then, in post-dispatch we refund `slash-relayer-weight.saturating_sub(reward-relayer-weight)` if we are rewarding. For confirmation transaction, everything is simpler - we always add the `reward-relayer-weight` and never do any refunds.

When we'll be adding this to the bridge hub runtimes, we'll need to "connect" `receive_messages_proof_overhead_from_runtime` and `receive_messages_delivery_proof_overhead_from_runtime` methods of two pallets (`pallet_bridge_messages` and `pallet_bridge_relayers`).


acatangiu:
bot rebase

bin/runtime-common/src/integrity.rs
```diff
@@ -316,6 +328,20 @@ pub fn check_message_lane_weights<C: Chain, T: frame_system::Config>(
 		this_chain_max_unrewarded_relayers,
 		this_chain_max_unconfirmed_messages,
 	);
+
+	// check that extra weights of delivery/confirmation transactions include the weight
+	// of `RefundBridgedParachainMessages` operations. This signed assumes the worst case
```
acatangiu, line 333: I think this was supposed to be
```suggestion
	// of `RefundBridgedParachainMessages` operations. This signed extension assumes the worst case
```

bin/runtime-common/src/integrity.rs
```diff
@@ -316,6 +328,20 @@ pub fn check_message_lane_weights<C: Chain, T: frame_system::Config>(
...
+	// (i.e. slashing if delivery transaction was invalid) and refunds some weight if
+	// assumption was wrong (i.e. if we did refund instead of slashing). If pallet weights
+	// do not account weights of refund extension, the extension will do refund when it
+	// doesn't need to.
```
acatangiu, line 337: maybe?

```suggestion
	// assumption was wrong (i.e. if we did refund instead of slashing). This check
	// ensures the extension will not refund weight when it doesn't need to (i.e. if pallet
	// weights do not account weights of refund extension).
```

modules/relayers/src/weights_ext.rs
```diff
@@ -0,0 +1,49 @@
...
+	/// confirmation call, if `RefundBridgedParachainMessages` signed extension is deployed at
+	/// runtime level.
+	fn receive_messages_delivery_proof_overhead_from_runtime() -> Weight {
+		Self::register_relayer_reward()
+	}
+
+	/// Returns weight that we need to deduct from the message delivery call weight that has
+	/// completed successfully.
+	///
+	/// Ususally, the weight of `slash_and_deregister` is larger than the weight of the
```
acatangiu, line 41:
```suggestion
	/// Usually, the weight of `slash_and_deregister` is larger than the weight of the
```

acatangiu (approved):
Very nice!

PR #2103, opened 2023-05-03
Fix sized messages (Follow-up on #2064)

serban300:


relays/bin-substrate/Cargo.toml
```diff
@@ -49,6 +49,9 @@ relay-utils = { path = "../utils" }
 relay-westend-client = { path = "../client-westend" }
 relay-wococo-client = { path = "../client-wococo" }
 rialto-runtime = { path = "../../bin/rialto/runtime" }
+# we are not using this runtime to craft callsour transactions, but we still need it
```
acatangiu, line 52:
```suggestion
# we are not using this runtime to craft calls or transactions, but we still need it
```

svyatonik (approved):
Thank you!

acatangiu (approved):

PR #2106, opened 2023-05-03
Use generated runtimes for BHR/BHW

### LINKED ISSUE #2102: Generate Rococo/WococoBridgeHub runtimes using runtime-codegen
Generate Rococo/WococoBridgeHub runtimes using runtime-codegen

serban300:
Fixes #2102


acatangiu (approved):
Changes look good (didn't look at the actual generated code). Does the relayer compiled from this still work? :see_no_evil:

svyatonik (approved):
Thank you, looks cool! Two points:
1) since new (generated) files are `1300` LOC,, we can deduplicate them. Because RBH and WBH are sharing runtimes. OTOH they may diverge in the future, so feel free to ignore
2) we may add some script and put it into the `./scripts` folder for easier update in the future, not only the readme :) The scripts folder is where I'll go first when I need to do something like that :) But again - feel free to ignore, because the second place where I'd go to is the autogenerated file header, where it says what I need to type :)

svyatonik:
Thank you :)

PR #2107, opened 2023-05-03
Add `MessagesPalletInstance` for integrity tests

bkontur:
we have multiple instance in Cumulus and without this it wont work


svyatonik (approved):
Thanks. Sorry for missing that :/

acatangiu (approved):

PR #2114, opened 2023-05-05
Propagate message verification errors

### LINKED ISSUE #1881: Unify runtime error types
may be fixed separately or as a part of #1666

We are using pallet-level errors, but there are errors that are dropped somewhere inside the "glue" runtime code and they are then logged (logs are unavailable for release runtimes) and are then merged into some "big" pallet-level error. E.g. when messages pallet is verifying messages proof, there's a lot of bad stuff that can happen there - e.g. invalid messages range, unknown header, invalid storage root, duplicate items in the storage proof, unused items in the storage proof, ... But in the end the transaction will fail with the `InvalidMessagesProof` error, which isn't super-informative.

Let's make either bridges-level error enum or pallet-level enums of what can happen during the call and use it (along with `frame_support::PalletError` - cheers to @serban300 for finding this) instead of `&' static str` and other custom errors.

serban300:
Related to #1881


modules/parachains/src/lib.rs
```diff
@@ -91,6 +91,8 @@ pub mod pallet {
 		BoundedStorageValue<<T as Config<I>>::MaxParaHeadDataSize, ParaStoredHeaderData>;
 	/// Weight info of the given parachains pallet.
 	pub type WeightInfoOf<T, I> = <T as Config<I>>::WeightInfo;
+	type GrandpaPallet<T, I> =
```
bkontur, line 94: hmm, this is little bit tricky, it assumes that the same Instance1...N is used for pallet_bridge_parachains and its relevant pallet_bridge_grandpa,

this could be problem for example:
`pallet_bridge_grandpa::<Instance1>`
`pallet_bridge_parachains::<Instance1>` -> needs `pallet_bridge_grandpa::<Instance1>`
`pallet_bridge_parachains::<Instance2>` -> needs also `pallet_bridge_grandpa::<Instance1>`

so mayvbe we should better introduce explicit `GrandpaPalletInstance` -> `type GrandpaPallet<T, GI> =`

modules/parachains/src/lib.rs
```diff
@@ -337,112 +333,113 @@ pub mod pallet {
...
-						// if relayer has specified invalid parachain head hash, ignore the head
-						// (this isn't strictly necessary, but better safe than sorry)
-						let actual_parachain_head_hash = parachain_head.hash();
-						if parachain_head_hash != actual_parachain_head_hash {
+			)
+			.map_err(Error::<T, I>::HeaderChainStorageProof)?;
+
+			for (parachain, parachain_head_hash) in parachains {
+				let parachain_head =
+					match Pallet::<T, I>::read_parachain_head(&mut storage, parachain) {
```
bkontur, line 344: just blind guess, could we use here `Self` instead of `Pallet::<T, I>` ?

bkontur (approved):
lgtm, but better wait for another eyes :D

modules/parachains/src/lib.rs
```diff
@@ -337,112 +333,113 @@ pub mod pallet {
...
+							Self::deposit_event(Event::MissingParachainHead { parachain });
+							continue
+						},
+					};
+
+				// if relayer has specified invalid parachain head hash, ignore the head
+				// (this isn't strictly necessary, but better safe than sorry)
+				let actual_parachain_head_hash = parachain_head.hash();
+				if parachain_head_hash != actual_parachain_head_hash {
+					log::trace!(
```
bkontur, line 376: looks like formatting is no good for this trace

modules/parachains/src/lib.rs
```diff
@@ -337,112 +333,113 @@ pub mod pallet {
 				parachains.len() as _,
 			);
 
-			pallet_bridge_grandpa::Pallet::<T, T::BridgesGrandpaPalletInstance>::parse_finalized_storage_proof(
+			let mut storage = GrandpaPallet::<T, I>::storage_proof_checker(
```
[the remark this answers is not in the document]
serban300: The problem is that with `pallet_bridge_grandpa::Pallet...`, `cargo fmt` doesn't work because the line is too big no matter what. That's why I added the shortcut.

modules/parachains/src/lib.rs
serban300: You're right. I didn't notice it. Fixed.

primitives/messages/src/target_chain.rs
```diff
@@ -171,14 +170,13 @@ const ALL_INBOUND_MESSAGES_REJECTED: &str =
...
 {
-	type Error = &'static str;
 	type MessagesProof = MessagesProof;
 
 	fn verify_messages_proof(
 		_proof: Self::MessagesProof,
 		_messages_count: u32,
-	) -> Result<ProvedMessages<Message>, Self::Error> {
-		Err(ALL_INBOUND_MESSAGES_REJECTED)
+	) -> Result<ProvedMessages<Message>, VerificationError> {
```
boundless-forest, line 178: `/// Error message that is used in ForbidOutboundMessages implementation.`

The comment seems to be staled, should be `ForbidInboundMessages`

serban300: You're right. Done.

svyatonik (approved):

PR #2123, opened 2023-05-09
Remove some `expect()` statements

serban300:
Related to #2038

Remove some `expect()` statements in order to avoid unpredictable failures.

### DIFF
--- a/modules/messages/src/inbound_lane.rs
+++ b/modules/messages/src/inbound_lane.rs
@@ -40,7 +40,7 @@ pub trait InboundLaneStorage {
 	/// Return maximal number of unconfirmed messages in inbound lane.
 	fn max_unconfirmed_messages(&self) -> MessageNonce;
 	/// Get lane data from the storage.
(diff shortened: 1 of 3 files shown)


modules/messages/src/benchmarking.rs
```diff
@@ -443,11 +443,12 @@ benchmarks_instance_pallet! {
 
 fn send_regular_message<T: Config<I>, I: 'static>() {
 	let mut outbound_lane = outbound_lane::<T, I>(T::bench_lane_id());
-	outbound_lane.send_message(vec![]);
+	outbound_lane.send_message(vec![]).expect("We craft valid messages");
 }
 
 fn receive_messages<T: Config<I>, I: 'static>(nonce: MessageNonce) {
-	let mut inbound_lane_storage = inbound_lane_storage::<T, I>(T::bench_lane_id());
```
bkontur, line 450: I think `fn inbound_lane_storage` was used just here, so maybe we could remove it now?

PR #2124, opened 2023-05-09
Cosmetics

serban300:


bkontur (approved):

primitives/messages/src/lib.rs
```diff
@@ -318,16 +329,25 @@ pub struct UnrewardedRelayersState {
 	pub last_delivered_nonce: MessageNonce,
 }
 
+impl UnrewardedRelayersState {
+	// Verify that the relayers state corresponds with the `InboundLaneData`.
+	pub fn is_valid<RelayerId>(&self, lane_data: &InboundLaneData<RelayerId>) -> bool {
```
svyatonik, line 334: I suggest either to check `messages_in_oldest_entry` here too, or o rename this method to something less ambitious :) `is_valid` imo means that the whole struct here must be valid and we are not checking this field here

svyatonik (approved):
LGTM

PR #2134, opened 2023-05-16
Bump polkadot/substrate

bkontur:


svyatonik (approved):

primitives/header-chain/src/justification.rs
```diff
@@ -49,6 +49,25 @@ pub struct GrandpaJustification<Header: HeaderT> {
 	pub votes_ancestries: Vec<Header>,
 }
 
+// TODO: remove and use `RuntimeDebug` (https://github.com/paritytech/parity-bridges-common/issues/2136)
```
bkontur, line 52: @svyatonik @serban300 
I have no idea why this does not work in Cumulus for bridge-hub-rococo-runtime, 
probably some Cargo.lock/Cargo.toml + features propagation problem or I missed something.
Tried severel things but nothing worked.

`cargo check --all --features try-runtime` works for bridges repo and all rialto/millau runtimes, but does not work for Cumulus when `RuntimeDebug` is used here, strange.

PR #2138, opened 2023-05-17
Fix nodes startup

svyatonik:
[#13384](https://github.com/paritytech/substrate/pull/13384)

PR #2140, opened 2023-05-17
Ws-port argument has been repalced with rpc-port

svyatonik:
Follow up for #2134 and #2138

This PR:
- [x] RialtoParachain collator nodes are not starting because `ws-port` argument has been removed;
- [x] RialtoParachain <> Millau bridge is not working
  - [x] RialtoParachain collators have zero parachain peers and zero relay chain peers
  - [x] "Could not send block production result to proposer!" in collator logs

### DIFF
--- a/bin/rialto-parachain/node/src/service.rs
+++ b/bin/rialto-parachain/node/src/service.rs
@@ -502,8 +502,8 @@ pub async fn start_node(
 					keystore,
 					force_authoring,
 					slot_duration,
-					// We got around 500ms for proposing
-					block_proposal_slot_portion: SlotProportion::new(1f32 / 24f32),
+					// We got around 1000ms for proposing
+					block_proposal_slot_portion: SlotProportion::new(1f32 / 12f32),
 					telemetry,
 					max_block_proposal_slot_portion: None,
 				},
--- a/deployments/networks/millau.yml
+++ b/deployments/networks/millau.yml
@@ -20,7 +20,7 @@ services:
       - --unsafe-rpc-external
       - --prometheus-external
     environment:
-      RUST_LOG: runtime=trace,rpc=debug,txpool=trace,runtime::bridge=trace,sc_basic_authorship=trace,beefy=trace,xcm=trace,trace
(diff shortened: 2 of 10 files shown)


svyatonik:
> uff, sorry for that, I knew that I forgot something, to check dockers yesterday :(

It's fine :) Rialto<>Millau exist to be broken and detect such changes

bkontur (approved):

deployments/networks/millau.yml
```diff
@@ -20,7 +20,7 @@ services:
       - --unsafe-rpc-external
       - --prometheus-external
     environment:
-      RUST_LOG: runtime=trace,rpc=debug,txpool=trace,runtime::bridge=trace,sc_basic_authorship=trace,beefy=trace,xcm=trace
+      RUST_LOG: runtime=trace,rpc=debug,txpool=trace,runtime::bridge=trace,sc_basic_authorship=trace,beefy=trace,xcm=trace,trace
```
svyatonik, line 23: TODO: do not forget to revert `RUST_LOG=trace` for network nodes

bin/rialto-parachain/node/src/service.rs
```diff
@@ -502,8 +502,8 @@ pub async fn start_node(
 					keystore,
 					force_authoring,
 					slot_duration,
-					// We got around 500ms for proposing
-					block_proposal_slot_portion: SlotProportion::new(1f32 / 24f32),
+					// We got around 1000ms for proposing
```
svyatonik, line 505: This has caused problems for me - parachain was unable to produce any blocks within `500ms`, so I've increased it to `1s`.

deployments/networks/rialto-parachain.yml
```diff
@@ -9,68 +9,74 @@ services:
     entrypoint: >
       /home/user/rialto-parachain-collator
       --alice
+      --node-key=80cf382988364291a7968ae7825c01f68c50d679796a8983237d07fe0ccf363b
```
svyatonik, line 12: There were some recent changes in networking && I guess our parachain nodes can't connect to each other because of that => I've added stuff similar to what we have in Rialto and Millau

serban300 (approved):

PR #2154, opened 2023-05-24
local.Dockerfile for building images from locally built binaries

### LINKED ISSUE #2141: Build docker images by copying prebuilt binary
Our docker builds are super slow (git clone, cargo deps, release build from the scratch). Let's add dockerfile that will just copy prebuilt image from the target directory. This:
- will make builds super fast, because most probably you've already cloned and built all deps;
- will make it easier to build custom images e.g. with overrriden dependencies, sometimes required to debug something.

Like this (still experimenting):
```Dockerfile
# Builds images used by the bridge.
#
# In particular, it can be used to build Substrate nodes and bridge relayers. The binary that gets
# built can be specified with the `PROJECT` build-arg. For example, to build the `substrate-relay`
# you would do the following:
#
# `docker build . -t local/substrate-relay --build-arg=PROJECT=substrate-relay`
#
# See the `deployments/README.md` for all the available `PROJECT` values.

FROM docker.io/library/ubuntu:20.04 as runtime

USER root
WORKDIR /home/root

# show backtraces
ENV RUST_BACKTRACE 1
ENV DEBIAN_FRONTEND=noninteractive

RUN set -eux; \
	apt-get update && \
	apt-get install -y --no-install-recommends \
        curl ca-certificates libssl-dev && \
    update-ca-certificates && \
	groupadd -g 1000 user && \
	useradd -u 1000 -g user -s /bin/sh -m user && \
	# apt clean up
	apt-get autoremove -y && \
	apt-get clean && \
	rm -rf /var/lib/apt/lists/*

# switch to non-root user
USER user

WORKDIR /home/user

ARG PROFILE=release
ARG PROJECT=substrate-relay

COPY --chown=user:user ./target/${PROFILE}/${PROJECT} ./

# check if executable works in this container
RUN ./${PROJECT} --version

ENV PROJECT=$PROJECT
ENTRYPOINT ["/bin/sh"]
```

And:
```
docker build . --build-arg=PROJECT=rialto-parachain-collator -f local.Dockerfile -t local/rialto-parachain-collator
```

.dockerignore (just an example):
```
*
!target/release/millau-bridge-node
!target/release/rialto-bridge-node
!target/release/rialto-parachain-collator
!target/release/substrate-relay
```

svyatonik:
closes #2141

### DIFF
--- a/local.Dockerfile
+++ b/local.Dockerfile
@@ -25,8 +25,8 @@ RUN set -eux; \
 	apt-get install -y --no-install-recommends \
         curl ca-certificates libssl-dev && \
     update-ca-certificates && \
-	groupadd -g 1000 user && \
-	useradd -u 1000 -g user -s /bin/sh -m user && \
+	groupadd -g 1001 user && \
+	useradd -u 1001 -g user -s /bin/sh -m user && \
 	# apt clean up
 	apt-get autoremove -y && \
 	apt-get clean && \
--- a/local.Dockerfile
+++ b/local.Dockerfile
@@ -10,6 +10,7 @@
 #
 # You may use `scripts/build-containers.sh` to build all binaries and images at once.
 
+# This image needs to be binary compatible with the host machine (where images are built).
 FROM docker.io/library/ubuntu:20.04 as runtime
 
 USER root
@@ -39,7 +40,6 @@ WORKDIR /home/user
 ARG PROFILE=release
 ARG PROJECT=substrate-relay
 
-# assume that the host machine is binary compatible with Ubuntu
 COPY --chown=user:user ./target/${PROFILE}/${PROJECT} ./
 
 # check if executable works in this container
--- a/local.Dockerfile
+++ b/local.Dockerfile
@@ -1,0 +1,47 @@
+# Builds images used by the bridge using locally built binaries.
+#
+# In particular, it can be used to build Substrate nodes and bridge relayers. The binary that gets
+# built can be specified with the `PROJECT` build-arg. For example, to build the `substrate-relay`
+# you would do the following:
+#
+# `docker build ./target -f local.Dockerfile -t local/substrate-relay --build-arg=PROJECT=substrate-relay`
+#
+# See the `deployments/README.md` for all the available `PROJECT` values.
+
+FROM docker.io/library/ubuntu:20.04 as runtime
+
+USER root
+WORKDIR /home/root
+
+# show backtraces
+ENV RUST_BACKTRACE 1
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN set -eux; \
+	apt-get update && \
+	apt-get install -y --no-install-recommends \
+        curl ca-certificates libssl-dev && \
+    update-ca-certificates && \
+	groupadd -g 1000 user && \
+	useradd -u 1000 -g user -s /bin/sh -m user && \
+	# apt clean up
+	apt-get autoremove -y && \
+	apt-get clean && \
+	rm -rf /var/lib/apt/lists/*
+
+# switch to non-root user
+USER user
+
+WORKDIR /home/user
+
+ARG PROFILE=release
+ARG PROJECT=substrate-relay
+
+# assume that the host machine is binary compatible with Ubuntu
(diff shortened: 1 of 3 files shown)


local.Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+#
+# In particular, it can be used to build Substrate nodes and bridge relayers. The binary that gets
+# built can be specified with the `PROJECT` build-arg. For example, to build the `substrate-relay`
+# you would do the following:
+#
+# `docker build ./target -f local.Dockerfile -t local/substrate-relay --build-arg=PROJECT=substrate-relay`
+#
+# See the `deployments/README.md` for all the available `PROJECT` values.
+
+FROM docker.io/library/ubuntu:20.04 as runtime
```
serban300, line 13: Nit: This needs to be the same as the local OS. Maybe it would be worth having a comment about this.

bkontur (approved):

local.Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+# show backtraces
+ENV RUST_BACKTRACE 1
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN set -eux; \
+	apt-get update && \
+	apt-get install -y --no-install-recommends \
+        curl ca-certificates libssl-dev && \
+    update-ca-certificates && \
+	groupadd -g 1000 user && \
```
serban300, line 27: I'm using ubuntu 23.04 and gid 1000 doesn't work for me. Could we use another value here ? 1001 or 2000 for example ?

scripts/build-containers.sh
```diff
@@ -1,7 +1,22 @@
...
+        time cargo build -p millau-bridge-node --release
+        time cargo build -p rialto-parachain-collator --release
+    fi
+
+    # following (using DOCKER_BUILDKIT) requires docker 19.03 or above
+    DOCKER_BUILDKIT=1 time docker build . -f local.Dockerfile -t local/substrate-relay --build-arg=PROJECT=substrate-relay
+    DOCKER_BUILDKIT=1 time docker build . -f local.Dockerfile -t local/rialto-bridge-node --build-arg=PROJECT=rialto-bridge-node
+    DOCKER_BUILDKIT=1 time docker build . -f local.Dockerfile -t local/millau-bridge-node --build-arg=PROJECT=millau-bridge-node
+    DOCKER_BUILDKIT=1 time docker build . -f local.Dockerfile -t local/rialto-parachain-collator --build-arg=PROJECT=rialto-parachain-collator
+fi
```
serban300, line 22: Nit: Could we also clean the old images at the end ?

serban300 (approved):

local.Dockerfile
svyatonik: Moved compatibility comment here. Out of curiosity - `substrate-relay` that is built on 23.04 doesn't work on 20.04?

local.Dockerfile
svyatonik: Changed to 1001

scripts/build-containers.sh
svyatonik: I'm not sure what old images do you mean?

local.Dockerfile
serban300: Doesn't seem to work. At least I'm getting the following error:

```
./substrate-relay: error while loading shared libraries: libssl.so.3: cannot open shared object file: No such file or directory
The command '/bin/sh -c ./${PROJECT} --version' returned a non-zero code: 127
```

scripts/build-containers.sh
serban300: For example when we generate a new `local/rialto-bridge-node:latest`, AFAIK the previous `local/rialto-bridge-node:latest` will not be removed automatically. Or at least there will be layers of its storage that are not needed anymore, but they are kept. It would be nice to remove them. Not sure if possible. I usually run `docker rmi -f $(docker images -f dangling=true -q)`, but this deletes all dangling images.

svyatonik: OK, thanks for explanation. But imo let's not do this in script - removal looks like a manual action to me. And `docker system prune` (or something like that) isn't super complex to call sometimes.

serban300: Ok, sure

Issue #665, opened 2021-01-20
Storage Proof Verification Profiling/ Use compact proof


tomusdrw:
Currently the `StorageProofChecker` is simply just creating a (partially complete) trie of storage proof elements and the reads are preformed using `read_trie_value` function.
The proofs often contains multiple storage entries - each message has it's own storage entry based on it's nonce.

We should check if it isn't better to perform proof verification and reads separately, i.e.:
1. We first verify the proof using `sp_trie::verify_trie_proof` function
2. We then query the values from a simple in-memory kvdb hashmap or get an iterator.

This will also ensure that all entries in the proof are actually correct (not only the ones we read).

EmmanuellNorbertTulbure:
Dependency on https://github.com/paritytech/parity-bridges-common/issues/1950

serban300:
Yes, the proof generated by using `sp_trie::generate_trie_proof` doesn't contain the values. It only contains the keys. For example I generated a proof for a trie db with only one entry (`key0=value0`), and it looks like this: `[72, 107, 101, 121, 48, 0]`. `[107, 101, 121, 48] = key0` in ascii. Not sure what 72 and 0 are for, but it definitely doesn't contain the value.

And `sp_trie::verify_trie_proof` can't be called on a proof generated using `prove_read`. It returns `ExtraneousValue` error.

So, if I understand correctly, with this approach we would need:
- a hashdb containing all the key-value pairs
- the proof generated using `sp_trie::generate_trie_proof`
- getting all the key-value pairs from the hash db and using them to call `sp_trie::verify_trie_proof`

Still researching

svyatonik:
@serban300 I think Tomek wanted to compare performance of current approach, where the storage proof includes all intermediate trie nodes and leaf values VS approach where the storage proof is just intermediate nodes and we deliver messages separately. So e.g. now the `FromBridgedChainMessagesProof` is the struct with `5` fields, including `storage_proof` and `nonces_start..=nonces_end`. And we are reading values from the proof. An alternative option could be the structure with one additional field: `messages`, which will have all encoded messages. Then we don't need leaf nodes in the `storage_proof` field. So the goal of this issue is to compare performance of two options and check if performance of the second option is larger than the performance of the first one.

serban300:
Compared the performance with the 2 approaches for 10k messages of 10KB each. Here are the results:

```
small/verify_trie_proof time:   [79.523 ms 79.564 ms 79.606 ms]
[...]

Issue #2147, opened 2023-05-22
Investigate error verifying millau compact proof


serban300:
When generating a compact proof for 1 Millau message (1 message generated on Millau in order to be sent to Rialto) using `generate_trie_proof(...)` and immediately checking it using `verify_trie_proof(...)` I'm getting a `RootMismatch` error.

The same scenario works without any problems when generating Rialto -> Millau messages. 

I'm guessing it's related to the custom hasher used by Millau.

serban300:
Verifying the Millau compact proof works when using the `LayoutV0` `TrieLayout`: `verify_trie_proof::<LayoutV0<H>, _, _, _>(...)`.

I guess it's something related to the [`MAX_INLINE_VALUE`](https://github.com/paritytech/trie/blob/2af70c28bcfe1a99a7a790233eacbd87a25a2f77/trie-db/src/lib.rs#L491)

Researching

serban300:
Fixed by #2155

PR #2155, opened 2023-05-24
Use compact proofs for messages delivery

serban300:
Fixes: #665, #2147

Performance improvement: Almost 50% as measured in #2147 
Small size improvement: for 4096 entries of 2048 bytes each and keys of the form `format!("key{}", NUM_MSGS - i)`
- read proof: 8674561 bytes
- compact proof: 8430549 bytes


primitives/runtime/src/vec_db.rs
```diff
@@ -0,0 +1,274 @@
...
+		match self.get_and_decode_mandatory(key) {
+			Ok(val) => Ok(Some(val)),
+			Err(VecDbError::UnavailableKey | VecDbError::EmptyVal) => Ok(None),
+			Err(e) => Err(e),
+		}
+	}
+
+	/// Checks if each key was read.
+	pub fn ensure_no_unused_keys(&self) -> Result<(), VecDbError> {
+		for (key, _) in &self.db {
```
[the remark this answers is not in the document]
serban300: I think that `verify_trie_proof()` does that internally. I mean if we call `generate_trie_proof()` for `{ k1 => v1, k2 => v2 }` and we only `verify_trie_proof()` for `{ k1 => v1 }` I remember that I tried that and it was thorowing an `ExtraneousNode` error.  but I'll double check and come back with an exact answear. And also I'll add a unit test.

primitives/runtime/src/vec_db.rs
```diff
@@ -0,0 +1,274 @@
+use frame_support::{PalletError, StateVersion};
```
svyatonik, line 1: Please add file header

primitives/runtime/src/vec_db.rs
```diff
@@ -0,0 +1,274 @@
...
+
+impl UntrustedVecDb {
+	/// Creates a new instance of `UntrustedVecDb`.
+	pub fn new<H: Hasher>(
+		read_proof: StorageProof,
+		root: TrieHash<LayoutV1<H>>,
+		mut keys: Vec<impl AsRef<[u8]> + Ord>,
+	) -> Result<Self, VecDbError> {
+		let mem_db = read_proof.into_memory_db();
+		let trie_db = TrieDBBuilder::<LayoutV1<H>>::new(&mem_db, &root).build();
```
svyatonik, line 49: Sorry, I still had no time to explore this layout stuff in details, so probably an absolute dumb q - is it fine that we're always using `LayoutV1` here, but switching between `LayoutV0` and `LayoutV1` when verifying proofs?

bin/runtime-common/src/messages_generation.rs
```diff
@@ -95,13 +98,32 @@ where
 				.expect("TrieMut::insert should not fail in benchmarks");
 			storage_keys.push(storage_key);
 		}
+
+		if add_duplicate_key {
```
svyatonik, line 102: Probably worth add test for [that](https://github.com/paritytech/parity-bridges-common/pull/2155#discussion_r1205446525) here too

svyatonik (approved):
Looks great, Thank you! Would be ideal to see the impact of this optimization on the whole transaction weight (because proof verification is a (significant) part of the delivery transaction). Also - as you've suggested in the related issue, let's do the same optimization for the confirmation transaction.

serban300:
> Would be ideal to see the impact of this optimization on the whole transaction weight (because proof verification is a (significant) part of the delivery transaction).

Thanks for the suggestions ! Would be actually very good to see this. What would be the best way to do this ? Should I run the benchmarks ?

svyatonik:
> Thanks for the suggestions ! Would be actually very good to see this. What would be the best way to do this ? Should I run the benchmarks ?

I think anyone could do that, even post-merge. Just run the same benchmarks (messages pallet) on new version and on the parent commit. So it isn't a blocker for merge - just something that would be interesting for the future us (how much did we gain with this optimization) :)

bin/runtime-common/src/messages_generation.rs
svyatonik: I mean test for extra trie node. But NVM - having test in the `vec_db.rs` is enough

serban300:
> > Thanks for the suggestions ! Would be actually very good to see this. What would be the best way to do this ? Should I run the benchmarks ?
> 
> I think anyone could do that, even post-merge. Just run the same benchmarks (messages pallet) on new version and on the parent commit. So it isn't a blocker for merge - just something that would be interesting for the future us (how much did we gain with this optimization) :)

Better to run the benchmarks before merging. Trying to do that now, but I'm getting some `InvalidMessagesProof` error. Will push new commits after I sort this out.

serban300:
I ran the benchmarks. I don't see big differences, thay are probably even withing the error margin. But from what I understand we only measure scenarios in which we send 1 or 2 messages and apart from that the map is clean. The differences should be noticeable with more messages and bigger maps.

@svyatonik wdyt ?

serban300:
Did another test to get more relevant performance results. I modified `receive_single_message_proof` to actually send 1001 messages instead of just 1 and also removed the error logs that can affect the performance:

```
	receive_single_message_proof {
		let relayer_id_on_source = T::bridged_relayer_id();
		let relayer_id_on_target = account("relayer", 0, SEED);
		T::endow_account(&relayer_id_on_target);

		// mark messages 1..=20 as delivered
		receive_messages::<T, I>(20);

		let (proof, dispatch_weight) = T::prepare_message_proof(MessageProofParams {
			lane: T::bench_lane_id(),
			message_nonces: 21..=1021,
			outbound_lane_data: None,
			is_successful_dispatch_expected: false,
			size: StorageProofSize::Minimal(EXPECTED_DEFAULT_MESSAGE_LENGTH),
		});
	}: receive_messages_proof(RawOrigin::Signed(relayer_id_on_target), relayer_id_on_source, proof, 1001, dispatch_weight)
	verify {
		assert_eq!(
			crate::InboundLanes::<T, I>::get(&T::bench_lane_id()).last_delivered_nonce(),
			1021,
		);
	}
```

The results are:

With the old code:

```
	fn receive_single_message_proof() -> Weight {
		// Proof Size summary in bytes:
		//  Measured:  `490`
		//  Estimated: `52645`
		// Minimum execution time: 10_002_807 nanoseconds.
		Weight::from_parts(10_105_449_000, 52645)
			.saturating_add(T::DbWeight::get().reads(3_u64))
			.saturating_add(T::DbWeight::get().writes(1_u64))
	}
``` 

With the new code:
```
	fn receive_single_message_proof() -> Weight {
		// Proof Size summary in bytes:
		//  Measured:  `490`
		//  Estimated: `52645`
		// Minimum execution time: 5_307_999 nanoseconds.
		Weight::from_parts(5_356_684_000, 52645)
			.saturating_add(T::DbWeight::get().reads(3_u64))
			.saturating_add(T::DbWeight::get().writes(1_u64))
	}
```

So the improvement is indeed significant (about 50%). But it's hard to notice when sending only one message.

acatangiu:
> So the improvement is indeed significant (about 50%). But it's hard to notice when receiving only one message.

1001 messages is also I think unrealistic for a single proof, but the results are very encouraging :100: 

If cost is mostly the same for 1 message, but only goes down (in comparison with old version) as more messages are included, then this definitely a clean win! :tada:

svyatonik:
So the good thing is that Serban has improved performance (thank you). The bad thing is that we don't see this improvement - call weight will be close to the previous value. So IMO we need a follow up PR with benchmarks update - we need some better estimation than we have now (roughly speaking: call weight = `NumberOfMessages * (WeightOfTwoMessages - WeightOfOneMessage)`). Probably instead of `1` and `2` messages use the `1..=MAX` interval - needs to be investigated.

That said, imo this PR is good to go.

serban300:
> So IMO we need a follow up PR with benchmarks update - we need some better estimation than we have now (roughly speaking: call weight = `NumberOfMessages * (WeightOfTwoMessages - WeightOfOneMessage)`). Probably instead of `1` and `2` messages use the `1..=MAX` interval - needs to be investigated.

Agree that it would be good to see if we can have some better estimation for the benchmarks. I will create an issue for this and do some research.

acatangiu (approved):
Nice!

primitives/runtime/src/vec_db.rs
```diff
@@ -0,0 +1,332 @@
...
+/// in the chain storage.
+#[derive(Clone, Default, Decode, Encode, Eq, PartialEq, RuntimeDebug, TypeInfo)]
+pub struct UntrustedVecDb {
+	proof: RawStorageProof,
+	db: Vec<(RawStorageKey, Option<DBValue>)>,
+}
+
+impl UntrustedVecDb {
+	/// Creates a new instance of `UntrustedVecDb`.
+	pub fn new<H: Hasher>(
```
bkontur, line 68: small nit, `new -> try_new`

primitives/runtime/src/vec_db.rs
```diff
@@ -0,0 +1,332 @@
...
+		// It's ok to use `LayoutV1` in this function, no matter the actual underlying layout,
+		// because we only perform read operations. When reading `LayoutV0` and `LayoutV1` lead to
+		// the same result.
+		let mem_db = read_proof.into_memory_db();
+		let trie_db = TrieDBBuilder::<LayoutV1<H>>::new(&mem_db, &root).build();
+
+		let trie_proof = generate_trie_proof::<LayoutV1<H>, _, _, _>(&mem_db, root, &keys)
+			.map_err(|_| VecDbError::UnableToGenerateTrieProof)?;
+
+		let mut entries = vec![];
```
bkontur, line 82: Couldnt we initialize `Vec::with_capacity(keys.len())` or something like that?

bkontur (approved):

primitives/runtime/src/vec_db.rs
serban300: Yes. Missed this. Thanks ! Done.

PR #2166, opened 2023-05-30
Get best and finalized head updates from subscription

### LINKED ISSUE #2074: Improve the way relayer get state updates
Right now messages relay that works with parachains polls target node every `12s` (parachain block time) to read state updates. If it awakes at `T` and actual parachain block time is e.g. `T + 1s`, it means that the relayer will get the updated state in another `11s`. During that period, it may decide to submit some transaction that contain obsolete nonces. And this transaction may be invalidated when e.g. next block is built. So it'll be ~40s for the whole scheme. We may decrease it down to `12-24s` just just by switching to 

Possible solutions:
- use new heads subscription instead of manual polling. This is a good solution, but let's have subscription sharing + caching (#1820) first;
- dumb, but easy-to-implement solution. We may just decrease the polling interval. We also may try to guess next block time. I.e. sleep `1s` first, then when we see block, schedule next awake in `11s`. If we don't see block, awake in `1s` and so on.

svyatonik:
related to paritytech/parity-bridges-common#2428 

~Want to leave it running for a while => draft~ Worked for 1:40 without any issues, so if approved, we may merge ~(not earlier than tomorrow though)~

### DIFF
--- a/relays/client-substrate/src/client/caching.rs
+++ b/relays/client-substrate/src/client/caching.rs
@@ -24,19 +24,23 @@ use crate::{
 	HashOf, HeaderIdOf, HeaderOf, NonceOf, SignedBlockOf, SimpleRuntimeVersion, Subscription,
 	TransactionTracker, UnsignedTransaction, ANCIENT_BLOCK_THRESHOLD,
 };
-use std::future::Future;
+use std::{future::Future, task::Poll};
 
-use async_std::sync::{Arc, Mutex, RwLock};
+use async_std::{
+	sync::{Arc, Mutex, RwLock},
+	task::JoinHandle,
+};
 use async_trait::async_trait;
 use bp_runtime::UnverifiedStorageProof;
 use codec::Encode;
 use frame_support::weights::Weight;
+use futures::{FutureExt, StreamExt};
 use quick_cache::unsync::Cache;
 use sp_core::{
 	storage::{StorageData, StorageKey},
 	Bytes, Pair,
 };
-use sp_runtime::transaction_validity::TransactionValidity;
+use sp_runtime::{traits::Header as _, transaction_validity::TransactionValidity};
 use sp_version::RuntimeVersion;
 
 /// `quick_cache::unsync::Cache` wrapped in async-aware synchronization primitives.
@@ -56,6 +60,9 @@ pub struct CachingClient<C: Chain, B: Client<C>> {
 struct ClientData<C: Chain> {
 	grandpa_justifications: Arc<Mutex<Option<SubscriptionBroadcaster<Bytes>>>>,
 	beefy_justifications: Arc<Mutex<Option<SubscriptionBroadcaster<Bytes>>>>,
+	background_task_handle: Arc<Mutex<JoinHandle<Result<()>>>>,
+	best_header: Arc<RwLock<Option<HeaderOf<C>>>>,
+	best_finalized_header: Arc<RwLock<Option<HeaderOf<C>>>>,
 	// `quick_cache::sync::Cache` has the `get_or_insert_async` method, which fits our needs,
 	// but it uses synchronization primitives that are not aware of async execution. They
 	// can block the executor threads and cause deadlocks => let's use primitives from
@@ -69,19 +76,32 @@ struct ClientData<C: Chain> {
 
 impl<C: Chain, B: Client<C>> CachingClient<C, B> {
 	/// Creates new `CachingClient` on top of given `backend`.
-	pub fn new(backend: B) -> Self {
+	pub async fn new(backend: B) -> Self {
 		// most of relayer operations will never touch more than `ANCIENT_BLOCK_THRESHOLD`
 		// headers, so we'll use this as a cache capacity for all chain-related caches
 		let chain_state_capacity = ANCIENT_BLOCK_THRESHOLD as usize;
+		let best_header = Arc::new(RwLock::new(None));
+		let best_finalized_header = Arc::new(RwLock::new(None));
+		let header_by_hash_cache = Arc::new(RwLock::new(Cache::new(chain_state_capacity)));
+		let background_task_handle = Self::start_background_task(
+			backend.clone(),
+			best_header.clone(),
+			best_finalized_header.clone(),
+			header_by_hash_cache.clone(),
+		)
+		.await;
 		CachingClient {
 			backend,
 			data: Arc::new(ClientData {
 				grandpa_justifications: Arc::new(Mutex::new(None)),
 				beefy_justifications: Arc::new(Mutex::new(None)),
+				background_task_handle: Arc::new(Mutex::new(background_task_handle)),
+				best_header,
+				best_finalized_header,
 				header_hash_by_number_cache: Arc::new(RwLock::new(Cache::new(
 					chain_state_capacity,
 				))),
-				header_by_hash_cache: Arc::new(RwLock::new(Cache::new(chain_state_capacity))),
+				header_by_hash_cache,
 				block_by_hash_cache: Arc::new(RwLock::new(Cache::new(chain_state_capacity))),
 				raw_storage_value_cache: Arc::new(RwLock::new(Cache::new(1_024))),
 				state_call_cache: Arc::new(RwLock::new(Cache::new(1_024))),
@@ -113,6 +133,7 @@ impl<C: Chain, B: Client<C>> CachingClient<C, B> {
 		Ok(value)
 	}
 
+	/// Subscribe to finality justifications, trying to reuse existing subscription.
 	async fn subscribe_finality_justifications<'a>(
 		&'a self,
 		maybe_broadcaster: &Mutex<Option<SubscriptionBroadcaster<Bytes>>>,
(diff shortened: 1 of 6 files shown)


relays/client-substrate/src/client/caching.rs
```diff
@@ -175,41 +183,71 @@ impl<C: Chain, B: Client<C>> Client<C> for CachingClient<C, B> {
...
+			.insert(best_finalized_header_hash, best_finalized_header);
+		Ok(best_finalized_header_hash)
 	}
 
 	async fn best_header(&self) -> Result<HeaderOf<C>> {
-		// TODO: if after https://github.com/paritytech/parity-bridges-common/issues/2074 we'll
-		// be using subscriptions to get best blocks, we may use single-value-cache here, but for
-		// now let's just call the backend
-		self.backend.best_header().await
+		let (result_sender, result_receiver) = bounded(1);
```
serban300, line 205: It seems a bit complicated to send a command to the worker for anything that we need. Could we hold the data that is updated by the background worker in shared memory (a separate structure) ? For example something like:

```
struct BackgroundWorkerData<C: Chain> {
	best_header: RwLock<Option<C::Header>>,
    ...
}

impl<C: Chain> BackgroundWorkerData<C> {
	async fn best_header<B: Client<C>>(&self, backend: &B) -> Result<C::Header> {
		Ok(self.best_header.read().await.clone().unwrap_or(backend.best_header().await?))
	}

    ...
}
```

The backround worker would write to it, and the client would read from it, making it easier to access the values. For example:
```
	async fn best_header(&self) -> Result<HeaderOf<C>> {
		self.worker_data.best_header(&self.backend).await
	}
```

svyatonik: I agree - `RwLock`s looks better here :) I had a sync-primitives allergy after investigating this recent relay-halt-issue, hence the channels. Fixed

relays/client-substrate/src/client/caching.rs
```diff
@@ -175,41 +212,45 @@ impl<C: Chain, B: Client<C>> Client<C> for CachingClient<C, B> {
...
+		self.backend.subscribe_finalized_headers().await
+	}
+
+	async fn subscribe_grandpa_finality_justifications(&self) -> Result<Subscription<Bytes>> {
+		let (result_sender, result_receiver) = bounded(1);
+		self.to_background_sender
+			.read()
+			.await
+			.try_send(ToBackground::SubscribeGrandpaJustifications(result_sender))?;
+		result_receiver.recv().await.map_err(Into::into)
```
serban300, line 244: Do we need to do this on the background worker ? Can we just do `self.backend.subscribe_grandpa_finality_justifications().await` here and also cache it ?

relays/client-substrate/src/client/caching.rs
svyatonik: @serban300 I was considering another your suggestion (below) and realized that there was another reason to do that using channels. For example if background task has exited (because it has encountered some error), and if we are using `Arc<RwLock<>>`, then there's no way for us to understand that. And if we keep using channels, we could understand that and report an error to the caller. So I'm thinking of reverting my last commit (https://github.com/paritytech/parity-bridges-common/pull/2166/commits/ff6af52cbf53b51b86284c849c47cd05a8b0ef46).

We can also start background task for processing the subscriptions and change type of `best_header` from `Arc<RwLock<Option<_>>>` and `Arc<RwLock<Option<Result<_, _>>>>`, but imo channels are more reliable. WDYT?

serban300: We could also have a `background_worker_is_running` flag maybe ? And if it's false, we don't return from the cache. For best_header and best_finalized_header.

serban300: Or just clear the caches if the background worker task exits ?

svyatonik: My point is that relying on the fact that background task will set something **before** it stops doesn't feels like a super-reliable solution. I could just change the types from `Arc<RwLock<Option<_>>>` to `Arc<RwLock<Option<Result<_, _>>>>` and set it to `Some(Err())` before it exits. Do you think it is better than reverting to channels?

serban300: I think there should be a reliable way to perform some logic after the background task closes. For example `async_std::task::spawn()` returns a `JoinHandle`. Maybe we can poll that. If there is any result, it should mean that the background task exited ? I will do some experiments to see if that's true (to see for example if this works correctly if the task is killed) and if it's usable. Will get back with more details.

But I wouldn't use `Arc<RwLock<Option<Result<_, _>>>>` . It adds more complexity. I would just clear the caches.

serban300: Looked a bit into it. The only way I could find to break the async task and prevent it from running the logic after `do_background_worker()` was to have a panic inside `do_background_worker`. To handle this we could have a guard struct that is moved to the background worker task when it is spawned and that implements `Drop`. and on drop it performs the cache cleanup logic. The problem is that drop is not async, so we would have to use standard RwLocks or Mutexes if possible. 

Another option would be to have just a Ping command and send it through a channel before querying the cache. If the ping fails, we don't use the cache.

Anyway, maybe more research would be needed and it's not worth blocking the task. For me it's ok to revert the last commit as you suggested, and merge it like this for the moment. And after that I can do some more research about the guard solution or the ping or others.

svyatonik: Let's sleep over it - I don't think there's anything better than just call `receiver.recv()?`. Everything else is more complicated, so probably channels are the way to go. But let's reconsider on next week. Anyway - thanks for review and for suggestions :)

svyatonik:
@serban300 I've force pushed, because a lot has changed since that PR was opened. I've changed code to use `JoinHandle` + `RwLock`s. PTAL when you have a minute (not urgent)

serban300 (commented):
The code looks good. Only left 2 comments. 

Also one higher level question related to the issue:
> Right now messages relay that works with parachains polls target node every 12s (parachain block time) to read state updates. If it awakes at T and actual parachain block time is e.g. T + 1s, it means that the relayer will get the updated state in another 11s. 

I'm not very familiar with the messages relay code, but isn't the same happening after the current change ? If the relay awakes at T and actual parachain block time is e.g. T + 1s, won't it get the old state even with this change ? Because the subscription will provide a new state also at T + 1s ?

relays/client-substrate/src/client/caching.rs
```diff
@@ -132,6 +153,98 @@ impl<C: Chain, B: Client<C>> CachingClient<C, B> {
...
+	}
+
+	/// Try to get header, read elsewhere by background task through subscription.
+	async fn read_header_from_background<'a>(
+		&'a self,
+		header: &Arc<RwLock<Option<HeaderOf<C>>>>,
+		read_header_from_backend: impl Future<Output = Result<HeaderOf<C>>> + 'a,
+	) -> Result<HeaderOf<C>> {
+		// ensure that the background task is active
+		self.ensure_background_task_active().await?;
```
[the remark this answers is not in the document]
svyatonik: If we'll simply return `read_header_from_backend()` result, our background task will never be restarted and this PR won't work (because we'll keep calling backend client methods instead of using subscriptions)  :) Instead, we return connections error, forcing relay loop to call the `reconnect`, which in turn restarts the task.

relays/client-substrate/src/client/caching.rs
```diff
@@ -161,6 +274,16 @@ impl<C: Chain, B: Client<C>> Client<C> for CachingClient<C, B> {
 		// since we have new underlying client, we need to restart subscriptions too
 		*self.data.grandpa_justifications.lock().await = None;
 		*self.data.beefy_justifications.lock().await = None;
+		// also restart background task too
+		*self.data.best_header.write().await = None;
+		*self.data.best_finalized_header.write().await = None;
```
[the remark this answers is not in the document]
svyatonik: Normally not, but e.g. in case of chain restart it'll clear the old (invalidated) data. Without that, the relay will keep getting errors (because new headers will be from the new chain and old headers from the old chain version). Not that it is the super-important thing, but I'd prefer to be careful here :)

svyatonik:
> I'm not very familiar with the messages relay code, but isn't the same happening after the current change ? If the relay awakes at T and actual parachain block time is e.g. T + 1s, won't it get the old state even with this change ? Because the subscription will provide a new state also at T + 1s ?

That is true, thanks for spotting that! I think to really close the #2428, we shall also replace polling with subscriptions in the relay code.

relays/client-substrate/src/client/caching.rs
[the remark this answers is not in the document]
serban300: Oh, I didn't realize this. Yes, you're right.

svyatonik:
> > I'm not very familiar with the messages relay code, but isn't the same happening after the current change ? If the relay awakes at T and actual parachain block time is e.g. T + 1s, won't it get the old state even with this change ? Because the subscription will provide a new state also at T + 1s ?
> 
> That is true, thanks for spotting that! I think to really close the #2428, we shall also replace polling with subscriptions in the relay code.

Let's do it in a separate PRs though (one per relay loop) - this one is already good enough imo to be deployed :)

serban300:
> > > I'm not very familiar with the messages relay code, but isn't the same happening after the current change ? If the relay awakes at T and actual parachain block time is e.g. T + 1s, won't it get the old state even with this change ? Because the subscription will provide a new state also at T + 1s ?
> > 
> > 
> > That is true, thanks for spotting that! I think to really close the #2428, we shall also replace polling with subscriptions in the relay code.
> 
> Let's do it in a separate PRs though (one per relay loop) - this one is already good enough imo to be deployed :)

Ok, sounds good !

serban300 (approved):

Issue #2165, opened 2023-05-30
Improve `receive_messages_proof` benchmarks accuracy


serban300:
Right now we estimate the weight of `receive_messages_proof` roughly with the following formula: call weight = NumberOfMessages * (WeightOfTwoMessages - WeightOfOneMessage)

This formula seems to lose accuracy as the actual number of received messages increases. More details [here](https://github.com/paritytech/parity-bridges-common/pull/2155#issuecomment-1566697426).

We should investigate if we can improve the per message accuracy

serban300:
I've been doing some experiments.

If we introduce another benchmark (let's say `receive_n_messages_proof`) similar to `receive_single_message_proof` but which runs on the range 1..MaxUnconfirmedMessagesAtInboundLane we get much more consistent results and the improvement brought by compact proofs is visible. 

But if we do this, `receive_single_message_proof_with_outbound_lane_state`, `receive_single_message_proof_1_kb`, `receive_single_message_proof_16_kb`, `receive_single_message_proof_with_dispatch` are not really compatible anymore. In order to make them compatible we would have to run them on 1..MaxUnconfirmedMessagesAtInboundLane as well, but that would take a lot of time.

Still thinking of a solution that would output consistent results in a reasonable time.

acatangiu:
What does "a lot of time" mean? Benchmarking doesn't happen very often and I think everybody is fine with compromising benchmarking speed over weight accuracy.

If it doesn't take ridiculously long, and we don't have a better alternative, I think it's ok with what you have now.

@svyatonik wdyt?

serban300:
> What does "a lot of time" mean? Benchmarking doesn't happen very often and I think everybody is fine with compromising benchmarking speed over weight accuracy.

On my machine 2-3 minutes for `receive_n_messages_proof`, 2-3 minutes for `receive_n_messages_proof_with_outbound_lane_state` and probably a bit more for `receive_n_message_proof_with_dispatch`.

For `receive_single_message_proof_1_kb` and `receive_single_message_proof_16_kb` I think we can keep them as they are.

svyatonik:
We have had benchmarks which have been running for ~40 minutes and (imo) it is too much even for those rare cases. If they're running for couple of minutes, then it should be fine.

---

Re the weight formula change - it is a very delicate topic. I'll be super glad if we can make it easier, but imo the proof has many dimensions which affect the final weight, so it isn't trivial.

[...]

PR #2175, opened 2023-06-02
Messages pallet benchmarks - cosmetics

serban300:
Somewhat related to #2165, but only cosmetics. No accuracy improvements yet.

- Migrated the messages pallet benchmarks to benchmarks v2
- Some deduplication

Also replaced `receive_single_message_proof_1_kb` and `receive_single_message_proof_16_kb` with `receive_single_message_n_kb_proof` . Seems to lead to similar results. But not sure how it's best. I can change it back if it was better.


svyatonik (approved):
Looks good!

acatangiu (commented):
👍

PR #2176, opened 2023-06-02
Improve receive_messages_proof benchmarks accuracy

### LINKED ISSUE #2165: Improve `receive_messages_proof` benchmarks accuracy

serban300:
Fixes #2165 

For the record, with the previous code without compact proofs, the new benchmark outputs:

```
	fn receive_n_messages_proof(n: u32, ) -> Weight {
		// Proof Size summary in bytes:
		//  Measured:  `490`
		//  Estimated: `52645`
		// Minimum execution time: 36_973_000 picoseconds.
		Weight::from_parts(38_623_000, 0)
			.saturating_add(Weight::from_parts(0, 52645))
			// Standard Error: 11_632
			.saturating_add(Weight::from_parts(11_585_692, 0).saturating_mul(n.into()))
			.saturating_add(T::DbWeight::get().reads(3))
			.saturating_add(T::DbWeight::get().writes(1))
	}
```

vs

```
	fn receive_n_messages_proof(n: u32, ) -> Weight {
		// Proof Size summary in bytes:
		//  Measured:  `490`
		//  Estimated: `52645`
		// Minimum execution time: 34_644_000 picoseconds.
		Weight::from_parts(17_835_495, 0)
			.saturating_add(Weight::from_parts(0, 52645))
			// Standard Error: 4_354
			.saturating_add(Weight::from_parts(7_517_907, 0).saturating_mul(n.into()))
			.saturating_add(T::DbWeight::get().reads(3))
			.saturating_add(T::DbWeight::get().writes(1))
	}
```

now. So the performance improvement is more noticeable.

### DIFF
--- a/modules/messages/src/weights_ext.rs
+++ b/modules/messages/src/weights_ext.rs
@@ -40,13 +40,6 @@ pub fn ensure_weights_are_correct<W: WeightInfoExt>() {
 	// benchmarked using `MaxEncodedLen` approach and there are no components that cause additional
 	// db reads
 
-	// verify `receive_messages_proof` weight components
-	assert_ne!(W::receive_messages_proof_overhead().ref_time(), 0);
-	assert_ne!(W::receive_messages_proof_overhead().proof_size(), 0);
-	// W::receive_messages_proof_messages_overhead(1).ref_time() may be zero because:
-	// the message processing code (`InboundLane::receive_message`) is minimal and may not be
-	// accounted by our benchmarks
-	assert_eq!(W::receive_messages_proof_messages_overhead(1).proof_size(), 0);
 	// W::receive_messages_proof_outbound_lane_state_overhead().ref_time() may be zero because:
 	// the outbound lane state processing code (`InboundLane::receive_state_update`) is minimal and
 	// may not be accounted by our benchmarks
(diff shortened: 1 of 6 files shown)


modules/messages/src/weights_ext.rs
```diff
@@ -353,20 +343,11 @@ pub trait WeightInfoExt: WeightInfo {
...
-		let weight_of_two_messages_and_two_tx_overheads =
-			Self::receive_single_message_proof().saturating_mul(2);
-		let weight_of_two_messages_and_single_tx_overhead = Self::receive_two_messages_proof();
-		weight_of_two_messages_and_two_tx_overheads
-			.saturating_sub(weight_of_two_messages_and_single_tx_overhead)
-	}
-
 	/// Returns weight that needs to be accounted when receiving given a number of messages with
 	/// message delivery transaction (`receive_messages_proof`).
 	fn receive_messages_proof_messages_overhead(messages: MessageNonce) -> Weight {
```
svyatonik, line 348: We don't need this method at all, if I'm not mistaken - you're using `receive_n_messages_proof(N)` instead, right?

svyatonik (approved):

modules/messages/src/weights_ext.rs
serban300: You're right. Removed it.

PR #2181, opened 2023-06-05
Move Chain::ID from relay-level Chain to primitives-level Chain

svyatonik:
related to #1666 

Since we're adding `type BridgedChain: Chain` to the messages pallet configuration, we may use `BridgedChain::ID` instead of the dedicated pallet constant (`type BridgedChainId: Get<ChainId>`).

This PR moves this field from the relay-chain level `Chain` definition to the primitives level `Chain` definition. Logic is not changed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9826,7 +9826,6 @@ dependencies = [
  "bp-header-chain",
  "bp-parachains",
  "bp-polkadot",
- "bp-runtime",
  "bridge-runtime-common",
  "parity-scale-codec",
  "relay-substrate-client",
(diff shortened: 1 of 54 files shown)


bkontur:
what about also moving out this [constants](https://github.com/paritytech/parity-bridges-common/blob/master/primitives/runtime/src/lib.rs#L63-L100) from `primitives/runtime`?
I guess, this is not the best place for them, `primitives/runtime` should be generic.

and if we need some guarantees for uniqueness per runtime, maybe we could check it in `integrity_test` e.g. `ensure_unique_chain_id`

svyatonik:
Removed IDs from `bp-runtime` - had to add some hack to `CustomNetworkId` implementation - will fix it in https://github.com/paritytech/parity-bridges-common/issues/2068

acatangiu (approved):

PR #2187, opened 2023-06-06
Use compact proofs for messages delivery confirmation

### LINKED ISSUE #2179: Use compact proofs for message delivery confirmation
- Adjust the benchmarks if necessary (See #2165 for example)
- Use compact proofs for message delivery confirmation similarly to how we migrated messages proofs (#2155)

serban300:
Closes #2179

No need to adjust the benchmarks since in the case of messages delivery confirmation we only attach the storage proof for a single key (the inbound lane).

### DIFF
--- a/bin/runtime-common/src/messages.rs
+++ b/bin/runtime-common/src/messages.rs
@@ -162,7 +162,7 @@ pub mod source {
 		let FromBridgedChainMessagesDeliveryProof { bridged_header_hash, storage_proof, lane } =
 			proof;
 		let mut storage =
-			B::BridgedHeaderChain::storage_proof_checker(bridged_header_hash, storage_proof)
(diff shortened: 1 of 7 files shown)


svyatonik (approved):
Looks good, thank you! My understanding here is that we're doing it to be consistent with other proofs, because there's no any performance gain. If so, can you, please, consider migrating parachain storage proof to compact proof as well?

serban300:
> Looks good, thank you! My understanding here is that we're doing it to be consistent with other proofs, because there's no any performance gain. 

Yes, yes, it's for consistency. 

> If so, can you, please, consider migrating parachain storage proof to compact proof as well?

Yes, this is next on my list.

acatangiu (approved):

PR #2189, opened 2023-06-06
Move messages delivery proof verification to pallet

svyatonik:
related to #1666

No logic changes (except some tests logic), so no audit required.

This PR moves delivery proof verfication code from `bridge-runtime-common` to messages pallet and also removes `TargetHeaderChain` trait. One change that may attract attention is that the `TargetHeaderChain` was not only verifying proof, but also was checking outbound message size. Now it is gone. But the message size is still checked when (before) the message is inserted in the outbound messages map - that is done by the `StoredMessagePayload`, which is a `BoundedVec`.


modules/messages/README.md
```diff
@@ -121,16 +121,18 @@ The pallet provides no means to get the result of message dispatch at the target
...
-size of the chain and so on. And when the relayer sends us a confirmation transaction, this
-implementation must be able to parse and verify the proof of messages delivery. Normally, you would
-reuse the same (configurable) type on all chains that are sending messages to the same bridged
-chain.
+### How to plug-in Messages Module to Send and Receive Messages from the Bridged Chain?
+
+The `pallet_bridge_messages::Config` trait has 2 main associated types that are used to work with
+inbound messages. The `pallet_bridge_messages::BridgedChain` defines basic primitives of the bridged
+chain. The `pallet_bridge_messages::BridgedHeaderChain` defines the way we access the bridged chain
+headers in our runtime. You may use `pallet_bridge_grandpa` if you're bridging with chain thatuses
```
acatangiu, line 130:
```suggestion
headers in our runtime. You may use `pallet_bridge_grandpa` if you're bridging with chain that uses
```

modules/messages/src/tests/pallet_tests.rs
```diff
@@ -478,7 +465,7 @@ fn receive_messages_delivery_proof_rewards_relayers() {
 		assert!(TestDeliveryConfirmationPayments::is_reward_paid(TEST_RELAYER_B, 1));
 	});
 }
-
+/* TODO
```
acatangiu, line 468: will this be done later as part of #1666?

relays/bin-substrate/src/cli/encode_message.rs
```diff
@@ -134,9 +134,8 @@ pub(crate) fn compute_maximal_message_size(
 	let maximal_source_extrinsic_size =
 		maximal_source_extrinsic_size - service_tx_bytes_on_source_chain;
 	let maximal_message_size =
-		bridge_runtime_common::messages::target::maximal_incoming_message_size(
-			maximal_target_extrinsic_size,
-		);
+		bp_messages::maximal_incoming_message_size(maximal_target_extrinsic_size);
+	println!("{} {}", maximal_message_size, maximal_source_extrinsic_size);
```
acatangiu, line 138: leftover? if not, at least add some details to the printed values?

acatangiu (approved):

Issue #1950, opened 2023-03-08
Use host functions for proof verification once available


acatangiu:
Depending on discussion here https://github.com/paritytech/srlabs_findings/issues/280

serban300:
@acatangiu is this issue still actual ? I see that the mentioned issue was resolved. And we seem to have the dependencies mentioned there.

acatangiu:
I honestly don't remember the full context here - we all discussed this in the team retreat and I quickly opened this issue but obviously didn't provide enough context for our future selves :see_no_evil: 

@svyatonik do you remember the details here?

svyatonik:
Let's check if the updated code is actually used by us when we check storage proofs (may also be related to this optimization stuff you're working now on, @serban300 ). If we use it, let's close

serban300:
Ok, will check after doing the optimization

serban300:
Checked. The updated code only touches the Memory and Trie DBs. We also use the recorder which was not updated. 

But if we switch all proofs to the new `VecDb`, I think we should be ok. I should check if we use the Recorder in other places, but I suppose not. For the moment, the currently opened PR, only switches the messages delivery proofs: #2155 . If this works ok and the feedback is positive, I'll switch the other proofs to use the `VecDb`.

serban300:
Fixed by #2194

PR #2194, opened 2023-06-07
Use compact proofs for parachains

### LINKED ISSUE #2188: Use compact proofs for parachains
Use compact proofs for parachains similarly to how we migrated messages proofs (https://github.com/paritytech/parity-bridges-common/pull/2155)

serban300:
Fixes #2188


svyatonik (approved):
Great, thank you! I assume it also solves the #1950, right?

One possible improvement (from my PoV) is maybe renaming `VecDbError` to `StorageProofError` (and maybe even `UntrustedVecDb` -> `UnverifierStorageProof` and so on). We now use one "kind" of storage proofs for everything and imo `StorageProof` gives more context that `VecDb` :) But that's my personal opinion - feel free to keep current namings

serban300:
> current

Makes sense, but I would like to spend some more time thinking about naming. I am also thinking of renaming `ProofSize` to something else, but not sure to what, and maybe delete the `storage_proof.rs` file alltogether. I will think some more about it and do it in a future PR.

PR #2198, opened 2023-06-08
Storage proofs related renamings

serban300:
Some renamings since we migrated all the storage proofs to the compact ones

UntrustedVecDb -> UnverifiedStorageProof
TrustedVecDb -> VerifiedStorageProof


bin/millau/runtime/src/lib.rs
```diff
@@ -1083,7 +1083,7 @@ impl_runtime_apis! {
 				fn prepare_parachain_heads_proof(
 					parachains: &[bp_polkadot_core::parachains::ParaId],
 					parachain_head_size: u32,
-					proof_size: bp_runtime::StorageProofSize,
```
acatangiu, line 1086: This rename I am leaning against. `StorageSize` can be misleading as opposed to `StorageProofSize` which describes it completely.

acatangiu (approved):
Don't agree with `StorageProofSize` -> `StorageSize`. Other than that, everything looks good!

Issue #2137, opened 2023-05-16
`SharedSubscriptionFactory` and `Subscription` are tightly coupled


serban300:
Related to this thread: https://github.com/paritytech/parity-bridges-common/pull/2129#discussion_r1195037730

Investigate if there is a better alternative

PR #2206, opened 2023-06-13
Avoid using extra background task for polling subscription when possible

serban300:
I hope I'm not missing something related to async contexts, but it seems to me that sometimes we can avoid creating extra background tasks for polling subscriptions. For example for the `TransactionTracker`.

Noticed this while researching #2137

### DIFF
--- a/relays/client-substrate/src/client/mod.rs
+++ b/relays/client-substrate/src/client/mod.rs
@@ -33,7 +33,7 @@ mod rpc_api;
 mod subscription;
 
 pub use client::Client;
-pub use subscription::{SharedSubscriptionFactory, Subscription};
+pub use subscription::{SharedSubscriptionFactory, Subscription, UnderlyingSubscription};
 
 /// Type of RPC client with caching support.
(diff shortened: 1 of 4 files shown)


serban300:
> Changes looks good to me (maybe some docs for `struct Unwrap` and its pub methods would be good)! 

You're right. Will add some docs.

> But what extra task do you mean - I've been searching for `spawn` calls removal, but haven't found anything

I mean here:
```
			Ok(TransactionTracker::new(
				self_clone,
				stall_timeout,
				tx_hash,
				Subscription::new(
					C::NAME.into(),
					"transaction events".into(),
					Box::new(subscription.map_err(Into::into)),
				)
				.await?,
			))
```

`Subscription::new()` was calling `SharedSubscriptionFactory::new()` which calls `async_std::task::spawn(background_worker())`

And now it's changed to:
```
			Ok(TransactionTracker::new(
				self_clone,
				stall_timeout,
				tx_hash,
				Box::new(Unwrap::new(C::NAME.into(), "transaction events".into(), subscription)),
			))
```

PR #2211, opened 2023-06-15
Prune messages from confirmation tx body, not from the on_idle

### LINKED ISSUE #2210: Prune confirmed messages in the confirmation transaction
Related to #1760

Right now we are pruning delivered messages from `on_idle` callback. Why? Because in previous bridge version we didn't wanted relayer to pay additional for the pruning - because this can't be predicted when message is sent. So this has been done in `fn send_message` and now it is migrated to the `on_idle`.

This concern ^^^ is no longer valid, because we refund relayer the full transaction cost. And current implementation is an obstacle to dynamic lanes implementation - we need some way (better than round robin) to prune messages from (potentially) unlimited number of lanes.

svyatonik:
closes #2210

See #2210 for rationale. I've also updated weights to see the effect don't look at the absolute numbers, only at db reads + writes.

I've left `audit-maybe` label - it'll probably need to be audited as a part of dynamic-lanes audit.


modules/messages/src/outbound_lane.rs
```diff
@@ -150,41 +145,17 @@ impl<S: OutboundLaneStorage> OutboundLane<S> {
 
 		ensure_unrewarded_relayers_are_correct(confirmed_messages.end, relayers)?;
 
+		// prune all confirmed messages
```
svyatonik, line 148: Two things to note here:
1) this removal is accounted by our benchmarks (see changes in weights db writes);
2) we do weight refund in delivery confirmation transaction, but it doesn't need to be changed because it does refund based on `total_messages` from the call arguments and `confirmed_messages` range, returned by this function. So all should be fine here

serban300 (approved):

acatangiu (approved):
Looks good!

svyatonik:
Forgot to add this warning: when we'll be upgrading existing pallets to this version, we shall not forget to prune old messages and update lane state accordingly

PR #2214, opened 2023-06-16
Add state to message lanes

svyatonik:
related to #1760 and #2213

This PR introduces important change to runtime - now if the lane does not exists in the storage (`InboundLanes` and `OutboundLanes`), we don't create it automatically. Instead we are (will for the 2nd option) provide two options:
1) specify set of lanes at genesis block;
2) manage lanes externally - like by the pallet from #2213.

~The only thing that is left here is compiling and running everything locally to be sure that I haven't forgot anything. But it isn't a stopped for review~ All is good


modules/messages/src/lib.rs
```diff
@@ -710,39 +716,43 @@ impl<T: Config<I>, I: 'static> InboundLaneStorage for RuntimeInboundLaneStorage<
...
-		match self.cached_data {
-			Some(ref data) => data.clone(),
-			None => {
-				let data: InboundLaneData<AccountIdOf<BridgedChainOf<T, I>>> =
-					InboundLanes::<T, I>::get(self.lane_id).into();
-				self.cached_data = Some(data.clone());
-				data
-			},
-		}
+		self.cached_data.clone()
```
serban300, line 719: Nit: I would rename the function to just `data()`, considering that we're not initializing it here anymore.

serban300 (approved):

modules/messages/src/lib.rs
serban300: Also I don't think it needs to receive `&mut self` anymore. `&self` should be enough.

PR #2221, opened 2023-06-20
Change LaneId underlying type from [u8; 4] to H256

svyatonik:
related to #2213, #1760 and #1666

This PR changes the lane identifier from `LaneId([u8; 4])` to `LaneId(H256)`. This gives us ability to generate unique lane identifiers on both sides of the bridge given that two bridge endpoints have their own unique identifiers. In bridge testnets it is the `ChainId` (which will also be deprecated soon in #1666) and in XCM world it is the location within `GlobalConsensus`. So now e.g. lane for serving bridge between two asset hubs (`Kusama.AssetHub` <> `Polkadot.AssetHub`) doesn't need to be hardcoded anywhere - it's just a `blake2_256(order(Kusama.AssetHub.XcmLocation, Polkadot.AssetHub.Location).encode())`

I've started implementing map of `(location1, location2) => LaneId` in #2213, but decided to do this refactor first - it'll make our life significantly easier. For example we won't get any accidental collisions due to lack of "bridge-open" synchronization protocol.

**IMPORTANT**: there are two breaking changes in this PR, which would require storage migrations:
1) all storage values that contains `LaneId` either in values and/or keys. This include: `pallet_bridge_messages::InboundLanes`, `pallet_bridge_messages::OutboundLanes`, `pallet_bridge_messages::OutboundLanes`, `pallet_bridge_relayers::RelayerRewards`;
2) the encoding of `bp_relayers::RewardsAccountParams` has changed, but it is already included in the migration from (1);
3) the (1) and (2) implies the change in sovereign accounts that are refunding relayers and pay rewards. So their balance shall be transferred to new accounts. 

This is a breaking change and would require storage migration on all chains with previous version of pallets.

Obviously, breaking changes also makes new relayer incompatible with old runtimes and vice versa.

Please say now if you have something against this approach or know a better approach. Meanwhile I'll work on remaining TODOs here:
- [x] separate PR to box proof argument in delivery call of messages pallet - after this PR it is too large;
- [x] cleanup;
- [x] fix testnets + manual tests.


primitives/relayers/src/lib.rs
```diff
@@ -185,21 +185,21 @@ mod tests {
...
 
 		assert_eq!(
 			PayRewardFromAccount::<(), H256>::rewards_account(RewardsAccountParams::new(
-				LaneId([0, 0, 0, 0]),
+				LaneId::new(1, 2),
 				*b"test",
 				RewardsAccountOwner::BridgedChain
 			)),
-			hex_literal::hex!("62726170000000007465737401726577617264732d6163636f756e7400000000")
+			hex_literal::hex!("627261706087240931c736c98ccb5748ffe8b0a946fb1bb382db195ad826ec3e")
```
svyatonik, line 202: This change is obviously wrong - accounts need to be different. I'm gonna fix it + add note to description

primitives/messages/src/lib.rs
```diff
@@ -154,21 +155,62 @@ impl OperatingMode for MessagesOperatingMode {
...
+/// ```
+#[derive(Clone, Copy, Decode, Encode, Eq, Ord, PartialOrd, PartialEq, TypeInfo, MaxEncodedLen)]
 #[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
-pub struct LaneId(pub [u8; 4]);
+pub struct LaneId(H256);
+
+impl LaneId {
+	/// Create lane identifier from two locations.
+	pub fn new<T: Ord + Encode>(endpoint1: T, endpoint2: T) -> Self {
+		LaneId(
```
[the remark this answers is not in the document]
svyatonik: Fixed

primitives/relayers/src/lib.rs
svyatonik: Fixed

serban300 (approved):
Looks good !

PR #2224, opened 2023-06-21
Add test method for growing storage proof size

serban300:
Related to #2212 

- Add `grow_storage_proof` test method (only adding it for the moment. Still thinking what would be the best way to make use of it)
- Guard storage proof related test methods behind `test-helpers` feature


svyatonik (approved):
Looks good!

PR #2226, opened 2023-06-22
Renamings

serban300:
StorageProofSize -> StorageSize
receive_single_message_n_bytes_proof -> receive_single_n_bytes_message_proof
receive_single_message_n_bytes_proof_with_dispatch -> receive_single_n_bytes_message_proof_with_dispatch

Related to #2212


svyatonik (approved):
Since you're going to fill up trie somewhere else, maybe it worth removing `enum StorageSize` at all? Like we'll only care about the message (aka leaf) size in currently existing code, right? And your trie-fill-code will be always running after the minimal trie is built?

svyatonik:
> Either we'll always fill the minimal trie with a hardcoded number of common-prefixes. Or we should have a benchmark for the proof size. Not sure what would be best. But anyway, I think we need to take into account the size of the storage (leafs) as well. Because it impacts the weight. So I think we should keep the StorageSize enum.

True. I mean - we leave `StorageSize`, but it'll be just `u32` - i.e. leaf/message size :)

svyatonik:
(I mean it's just my opinion - feel free to leave as is)

serban300:
> > Either we'll always fill the minimal trie with a hardcoded number of common-prefixes. Or we should have a benchmark for the proof size. Not sure what would be best. But anyway, I think we need to take into account the size of the storage (leafs) as well. Because it impacts the weight. So I think we should keep the StorageSize enum.
> 
> True. I mean - we leave `StorageSize`, but it'll be just `u32` - i.e. leaf/message size :)

Oh, you might be right. Looking into it.

svyatonik (approved):
:+1:

acatangiu (approved):

PR #2246, opened 2023-07-07
Backport to polkadot-staging: Deduplicate Grandpa consensus log reading logic (#2245)

serban300:


svyatonik (approved):
IMO better not to touch `polkadot-staging` without significant reasons - like bug fixes or absolutely required refactorings. But this change is small enough, so let's merge.

serban300:
> IMO better not to touch `polkadot-staging` without significant reasons - like bug fixes or absolutely required refactorings. But this change is small enough, so let's merge.

Agree. Just merging it to make sure we won't have conflicts when fixing #39.

PR #2261, opened 2023-07-13
Implement additional require primitives for dynamic fees directly for pallet-xcm-bridge-hub

svyatonik:
related to paritytech/parity-bridges-common#2434
alternative to paritytech/parity-bridges-common#2233 

The idea is to:
- [x] add backoff to inbound bridge queue (aka outbound XCMP queue between target BH and target chain) - if it has too many enqueued messages, we assume that the target chain is unable to process messages in a timely fashion and stop accepting message delivery transactions: https://github.com/paritytech/parity-bridges-common/commit/e19c82a723bb9d80e7bc54bb82c9ecfa746fe59e
- [ ] add backoff mechanism to outbound bridge queue - if it has too many enqueued messages, we assume that the inbound bridge queue (at the bridged, target BH) contains too many messages (or if there's no relayer running) and stop accepting new messages to this queue by pausing inbound "end" of XCMP channel between sending chain and its sibling BH. It requires some parts of paritytech/parity-bridges-common#2213 and paritytech/parity-bridges-common#2406
- [x] XCMP queue has its own backoff mechanism and because of that, messages will eventually  start piling up at the sending chain outbound XCMP queue "end". So it can estimate the number of messages in the whole bridge pipeline and start applying exponential fee or something like that

This PR requires corresponding changes to the relayer. It'll be done in a follow up/parallel PR.

===

Update: this ^^^ information is a bit obsolete, but overall it stays the same.Main things to notice in this PR:
- I've introduced a separate `BridgeId` struct, so there's now both `BridgeId` and `LaneId`. Both are `H256` and are matching **now**. The reason is that we shall use versioned XCM structs in `BridgeId` AND version may change at some point. Don't know what we'll do when it happens, but we must be prepared for that;
- `messages_xcm_extension.rs` file is removed and required implementations are moved to the `pallet-xcm-bridge-hub` (`HaulBlobExporter`, `MessageDispatch` and so on);
- there's now `LocalXcmChannelManager`, which is able to report channel status + suspend/resume the channel. I've removed `bp-xcm-bridge-hub-router` crate because of that.


bin/runtime-common/src/messages_xcm_extension.rs
```diff
@@ -163,3 +210,121 @@ impl<H: XcmBlobHauler> HaulBlob for XcmBlobHaulerAdapter<H> {
 			})
 	}
 }
+
+// TODO: use following structures in the `pallet-message-queue` configuration
+
+// TODO: it must be a part of `pallet-xcm-bridge-hub`
+pub struct LocalInboundXcmChannelSuspender<Origin, Inner>(PhantomData<(Origin, Inner)>);
```
svyatonik, line 217: Open question: in https://github.com/paritytech/cumulus/pull/2157, the `QueueChangeHandler` of `pallet_message_queue` is not set (???) to the `cumulus-pallet-xcmp-queue`. Given that, our backpressure won't actually work - we are pausing dispatch here, but the transport channel will stay opened.

svyatonik: This: https://github.com/paritytech/cumulus/pull/2157/files#diff-3d3febd8cfcd42ba5bb949d63fdafb9a71da3db1feb35a45601404b64fcd70c4R575 and e.g. this https://github.com/paritytech/cumulus/pull/2157/files#diff-447906ba563849a582953cc1753f58164c23d4902539e37c5dde387113a3aa5bR411

svyatonik:
renamed, because most of the code required for dynamic fees is now in a separate PR: https://github.com/paritytech/parity-bridges-common/pull/2312

modules/xcm-bridge-hub/src/backpressure.rs
```diff
@@ -0,0 +1,131 @@
+// Copyright 2019-2021 Parity Technologies (UK) Ltd.
```
svyatonik, line 1: This file should be the copy of the similar primitives (`LocalXcmQueueMessageProcessor` and `LocalXcmQueueSuspender`) from #2294. The only difference is how we "compute" whether the inbound XCM queue should be suspended or not. In v1 PR it looks at the number of queued messages at the outbound lane. Here it should take into account all outbound lanes, opened by the same origin (a bit WIP)

serban300 (approved):

modules/xcm-bridge-hub/src/exporter.rs
```diff
@@ -123,6 +141,149 @@ where
...
+					);
+
+					return
+				},
+			}
+		}
+
+		// and forget that we have previously suspended the bridge
+		SuspendedBridges::<T, I>::mutate(|suspended_bridges| {
+			suspended_bridges.retain(|b| *b == bridge_id);
```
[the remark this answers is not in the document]
svyatonik: :facepalm: Thanks! It should've been spotted by unit tests, but they're not here yet :/

primitives/xcm-bridge-hub/src/lib.rs
```diff
@@ -26,58 +26,112 @@ use frame_support::{
...
+			fn encode(&self) -> sp_std::vec::Vec<u8> {
+				self.0.clone()
+			}
+		}
+
+		Self(LaneId::new(
+			EncodedVersionedInteriorMultiLocation(universal_location1.encode()),
+			EncodedVersionedInteriorMultiLocation(universal_location2.encode()),
+		))
+	}
```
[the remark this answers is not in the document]
svyatonik: filed #2500

PR #2262, opened 2023-07-13
GRANDPA: optimize votes_ancestries when needed

### LINKED ISSUE #2258: Issue when optimizing GRANDPA justifications
When optimizing a GRANDPA justification [we stop checking precommits after the weight of the precommit votes added so far >= required threshold](https://github.com/paritytech/parity-bridges-common/blob/ee49df62367e6087198d83c8de30a648a1dbdaf5/primitives/header-chain/src/justification.rs#L261). If however, there is a later valid precommit that adds a new fork to the ancestry, we might incorrectly fail [here](https://github.com/paritytech/parity-bridges-common/blob/ee49df62367e6087198d83c8de30a648a1dbdaf5/primitives/header-chain/src/justification.rs#L324) since that header will be unvisited.

Maybe we should just optimize the `votes_ancestries` as well.

serban300:
Fixes #2258 
Also needed for #39


primitives/header-chain/src/justification.rs
```diff
@@ -196,45 +214,82 @@ impl VerificationCallbacks for StrictVerificationCallbacks {
...
+impl<Header: HeaderT> OptimizationCallbacks<Header> {
+	fn optimize(self, justification: &mut GrandpaJustification<Header>) {
+		for invalid_precommit_idx in self.extra_precommits.into_iter().rev() {
 			justification.commit.precommits.remove(invalid_precommit_idx);
 		}
-		justification
+		if !self.redundant_votes_ancestries.is_empty() {
+			justification
+				.votes_ancestries
+				.retain(|header| self.redundant_votes_ancestries.get(&header.hash()).is_none())
```
svyatonik, line 251:
```suggestion
				.retain(|header| self.redundant_votes_ancestries.contains(&header.hash()))
```

primitives/header-chain/src/justification.rs
```diff
@@ -315,76 +356,99 @@ where
 			authorities_set_id,
 			&mut signature_buffer,
 		) {
-			return Err(Error::InvalidAuthoritySignature)
+			callbacks.on_invalid_authority_signature(precommit_idx)?;
```
svyatonik, line 359: If we are treating such justifications as normal now, then maybe let's a test for that?

primitives/header-chain/src/justification.rs
```diff
@@ -315,76 +356,99 @@ where
...
-		precommit_target: &Header::Hash,
-	) -> Result<Self, Error> {
-		let mut current_hash = *precommit_target;
+	/// Returns a route if the precommit target block is a descendant of the `base` block.
+	pub fn ancestry(
+		&self,
+		precommit_target_hash: &Header::Hash,
+		precommit_target_number: &Header::Number,
+	) -> Option<Vec<Header::Hash>> {
+		if precommit_target_number < &self.base.number() {
```
svyatonik, line 415: Just wonder - is that an optimization or fix of some issue? If that's a fix, can we please have a test for that as well?

primitives/header-chain/src/justification.rs
```diff
@@ -315,76 +356,99 @@ where
...
+				None => return None,
 			};
 		}
-		Ok(self)
+
+		Some(route)
+	}
+
+	fn mark_route_as_visited(&mut self, mut route: Vec<Header::Hash>) {
+		for hash in route.drain(..) {
```
svyatonik, line 446:
```suggestion
		for hash in route {
```
?

svyatonik (approved):
Looks good, but would be ideal if we'll have some tests for new fixes! Thank you!

primitives/header-chain/src/justification.rs
serban300: Done.

primitives/header-chain/src/justification.rs
```diff
@@ -315,76 +356,99 @@ where
...
 
-	// check that the cumulative weight of validators voted for the justification target (or one
-	// of its descendents) is larger than required threshold.
-	if cumulative_weight >= threshold {
-		Ok(())
-	} else {
-		Err(Error::TooLowCumulativeWeight)
+	// check that there are no extra headers in the justification
+	if !chain.is_fully_visited() {
+		callbacks.on_redundant_votes_ancestries(chain.unvisited)?;
```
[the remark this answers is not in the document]
serban300: Done.

primitives/header-chain/src/justification.rs
serban300: It's just cosmetics. Moved the code from [here](https://github.com/paritytech/parity-bridges-common/blob/a283fb89ce31259c65e619f905eedb353f9f109b/primitives/header-chain/src/justification.rs#L288). Seemed to fit in the `AncestryChain` logic.

Added a test.

svyatonik (approved):
Thank you!

PR #2268, opened 2023-07-14
Misbehavior in the xcm-bridge-hub pallet

svyatonik:
related to https://github.com/paritytech/parity-bridges-common/issues/1760, https://github.com/paritytech/parity-bridges-common/pull/2233, https://github.com/paritytech/parity-bridges-common/pull/2213, https://github.com/paritytech/parity-bridges-common/pull/2261

The future of this PR depends on the way we'll implement dynamic fees. If they will be implemented using state reporting, we'll need this PR as well. If they will be implementing using backpressure, then we need to revisit this code again and decide whether we need other kinds misbehavior(s) at all.


svyatonik:
https://github.com/paritytech/parity-bridges-common/pull/2294#issuecomment-1659668483

PR #2276, opened 2023-07-17
Some code grooming

svyatonik:
most of #23

I've added:
- `warn(missing_docs)` for all crates that are "exported" to the Cumulus repo;
- removed workarounds for some old nightly clippy version that are now obsolete;
- removed obsolete deny rules.


bkontur (approved):
lgmt,
And do we need/want backport this also to polkadot-staging?

svyatonik:
> lgmt, And do we need/want backport this also to polkadot-staging?

I'd avoid that - imo let's not push commits that are not strongly necessary there. We can update that later, during next "big merge" step.

PR #2277, opened 2023-07-17
Update dependecies

svyatonik:
another deps update needed for #2261
the PR is not yet ready - just showing some progress on that

Unfortunately, I had to revert most of changes to bridges subtree from https://github.com/paritytech/cumulus/pull/2790. I've missed the fact that the `Block` is now used in bridge primitives, which is not good - we don't have an access to extrinsic type there, so we don't (and won't need) to know the block type. So I'm reverting changes to our `Chain` trait. This also means that I'll backport those (my) changes to the `polkadot-staging` after merging this PR to master.

We also need to think of adding our crates to workspace level `Cargo.toml` in Cumulus. Some stuff here doesn't even compiles (my guess - haven't checked it in Cumulus subtree).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -268,6 +268,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "aquamarine"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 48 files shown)


bkontur (approved):
I went through, looks ok to me

serban300 (approved):

acatangiu (approved):

PR #2281, opened 2023-07-19
Port "Update dependecies (#2277)" to polkadot-staging

svyatonik:
I also had to port previous update-refs commits to master (actually I've made `cargo update` + copypasted fixes instead of that), because otherwise we won't be able to update Cumulus subtree properly. I think we need to make some changes to dependabot and `update-refs` PRs to make our life easier (left comments here: #2280).

Most notable fixes outside of #2277:
- impl `Deserialize` and `Serialize` in `no_std`;
- use `polkadot_primitives::v5` instead of `polkadot_primitives::v4` (node and runtime specific, so not important for Cumulus, but important for testing relayer).

I've also built + left Rialto<>Millau bridge running locally for a while - looks like relayer wotks  fine.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,6 +22,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "addr2line"
+version = "0.20.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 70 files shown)


bkontur (approved):
lgtm

PR #2285, opened 2023-07-20
Implement ExportXcm and MessageDispatch for pallet-xcm-bridge-hub

svyatonik:
This PR relates to:
- #1760, because for dynamic lanes we'll need an analog of `XcmBlobHaulerAdapter` that supports dynamic configuration;
- #2261, because in next PRs I'm going to add some code (that uses methods and storage of `pallet-xcm-bridge-hub`) to both `ExportXcm` and `MessageDispatch` to support backpressure on bridge/XCM queues.

So this PR adds implementation of `ExportXcm` and `MessageDispatch` to the recent `pallet-xcm-bridge-hub`. In the future we'll migrate Rialto<>Millau and then other bridges to this pallet, so `bridge_runtime_common::XcmBlobHaulerAdapter` and `bridge_runtime_common::XcmBlobMessageDispatch` will dissolve. Right now it's just a code that isn't used anywhere, bu as I mentioned, I need it to progress on #2261.

If we'll be implementing dynamic fees in V1 (without dynamic lanes), we'll probably keep using existing `bridge_runtime_common::XcmBlobHaulerAdapter` and `bridge_runtime_common::XcmBlobMessageDispatch`. Then we'll need a totally separate PR to the `polkadot-staging` branch.

### DIFF
--- a/modules/xcm-bridge-hub/src/exporter.rs
+++ b/modules/xcm-bridge-hub/src/exporter.rs
@@ -127,7 +127,7 @@ where
 ///
 /// We are using `HaulBlobExporter`, which requires `HaulBlob` implementation. It assumes that
 /// there's a single channel between two bridge hubs - `HaulBlob` only accepts the blob and nothing
-/// else. But bridge messages pallet maye have a dedicated channel (lane) for every pair of bridged
+/// else. But bridge messages pallet may have a dedicated channel (lane) for every pair of bridged
 /// chains. So we are using our own `ExportXcm` implementation, but to utilize `HaulBlobExporter` we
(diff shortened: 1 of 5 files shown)


modules/xcm-bridge-hub/src/exporter.rs
```diff
@@ -0,0 +1,184 @@
+// Copyright 2019-2021 Parity Technologies (UK) Ltd.
```
svyatonik, line 1: Here, I had to change the original `XcmBlobHaulerAdapter` a bit, because it doesn't work well with dynamic lanes. So it needs more careful review

acatangiu (approved):

PR #2294, opened 2023-07-24
Dynamic fees for bridges-v1

svyatonik:
https://gist.github.com/svyatonik/6d4d9f58b800ac951dcac6c1aa827c67#file-bridgequeues-png

This PR introduces all logic and components required for dynamic fees in v1, that we may build in this repo:
- message delivery transaction will be rejected if message dispatcher is "inactive". Message dispatcher must become "inactive" when the XCM channel (XCMP) between target bridge hub and target asset hub is suspended. This suspension occurs e.g. when the target asset hub is unable to process inbound XCM messages in a timely fashion. Review hints: `XcmBlobMessageDispatch` and messages pallet;
- added artificial limit to the bridge outbound queue. When it is full, we stop accepting new bridge messages. This can occur e.g. if relayers are not running or if dispatcher at the target bridge hub is "inactive" and relayers can't deliver any messages. Review hints: `XcmBlobHaulerAdapter`, `LocalXcmQueueManager`, `LocalXcmQueueMessageProcessor` and `LocalXcmQueueSuspender`;
- a new pallet that must be deployed at asset hubs (at sending chains) - `pallet-xcm-bridge-hub-router`. Now it exists only to support dynamic fees. In the future, we'll (probably) add a support for dynamic bridges to it. Dynamic fee is based on https://github.com/paritytech/polkadot/issues/6843. We start increasing fee factor when the channel between sending asset hub and its sibling bridge hub is suspended and/or has many queued messages.

TODOs in this PR:
- [x] tests;
- [x] review remaining TODOs in the code;
- [x] weights for the new pallet;
- [x] add support of delivery transaction logic changes to our signed extensions;
- [ ] ~add support of delivery transaction logic changes to our relay~ I'd like to do it in a separate PR. It isn't actually necessary, because signed extensions will do the job, but desirable to make relayer "polite".

But before investing a time in that ^^^, I'd like to check if all my assumptions are correct and test it using local chains. For that, we need:
- [x] add new branch to Cumulus, with this code and https://github.com/paritytech/cumulus/pull/2157 (started work on that: https://github.com/paritytech/cumulus/compare/sv-bridges-and-new-message-queue: this branch contains updated bridges subtree (this PR) + `oty-message-queue` + `bko-transfer-asset-via-bridge-pallet-xcm` + `bridge-hub-kusama-polkadot`);
- [x] implement `IsChannelActive` for outbound part of XCM queue between target BH and target AH;
- [x] implement `LocalXcmQueue` for outbound part of XCM queue between source AH and source BH;
- [x] add some scripts to send many messages and check how the whole scheme reacts to that. E.g.: ensure that every queue has a backpressure mechanism and it works as expected, ensure that all bridge queues works as a whole (they're able to trigger backpressure mechanism at the first queue eventually). See https://github.com/paritytech/parity-bridges-common/pull/2294#issuecomment-1653115033


modules/xcm-bridge-hub-router/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+	pub trait Config<I: 'static = ()>: frame_system::Config {
+		/// Universal location of this runtime.
+		type UniversalLocation: Get<InteriorMultiLocation>;
+		/// Relative location of the sibling bridge hub.
+		type SiblingBridgeHubLocation: Get<MultiLocation>;
+		/// The bridged network that this config is for.
+		type BridgedNetworkId: Get<NetworkId>;
+
+		/// Actual message sender (XCMP/DMP) to the sibling bridge hub location.
+		type ToBridgeHubSender: SendXcm + LocalXcmQueue;
```
svyatonik, line 59: Need to split it. Something like that:
```rust
type ToBridgeHubSender: SendXcm;
type LocalXcmQueue: IsChannelActive;
```

And also use similar `IsChannelActive` in the message dispatch implementation. Probably we need to add 2 methods to cumulus XCMP pallet that will return `true` if inbound/outbound channel is active and provide some easy adapter for implementing `IsChannelActive` using functions that return `bool`.

modules/xcm-bridge-hub-router/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+				*f = f.saturating_mul(total_factor);
+				*f
+			});
+		}
+	}
+}
+
+/// We'll be using `SovereignPaidRemoteExporter` to send remote messages over the sibling/child
+/// bridge hub.
+type ViaBridgeHubExporter<T, I> = SovereignPaidRemoteExporter<
```
svyatonik, line 123: Two things here:
1) we need to check that the bridge hub doesn't allow unpaid execution after this is changed;
2) in v2 - do we need to check at the bridge hub that the fee is enough? And how to do that?

modules/xcm-bridge-hub-router/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+
+		// TODO: we can reject large messages here (now they're just dropped at the bridge hub)!!!
+
+		// just use exporter to validate destination and insert instructions to pay message fee
+		// at the sibling/child bridge hub
+		//
+		// the cost will include both cost of: (1) to-sibling bridg hub delivery (returned by
+		// the `Config::ToBridgeHubSender`) and (2) to-bridged bridge hub delivery (returned by
+		// `Self::exporter_for`)
+		ViaBridgeHubExporter::<T, I>::validate(dest, xcm)
```
svyatonik, line 177: Another dependency we'll need for that is https://github.com/paritytech/cumulus/pull/2342. So idea is:
1) bridge backpressure mechanism involves many chains, so it isn't instant - it may take some to suspend the queue between source AH and source BH. And we can't allow senders to send messages over the bridge for a constant cost while this mechanism activates;
2) so we need to charge dynamic fees for sending messages from source AH and source BH even when we don't have any info about other bridge queues. That's where the mentioned PR should help us - the total transport cost that the sender pays is the `BridgeFee + HrmpFee`. So `BridgeFee` will grow when we have a report about overloaded bridge queues and `HrmpFee` shall be maintained independently

modules/xcm-bridge-hub-router/src/lib.rs
svyatonik: First part done

modules/xcm-bridge-hub-router/src/lib.rs
svyatonik: Actually both points may be done for v2 - in v1 we only deal with system parachains. But for v2 we need to check that the execution is paid.

As for the 2nd - IIUC we don't need to check anything apart from the fact that the execution is paid. If the sending chain is not able to limit the messages flow, it'll end up with messages piling up at its storage. And checking fee value here (at the source bridge chain) doesn't matter much.

svyatonik:
Marked onice because of dependencies

serban300 (approved):
Looks good !

modules/xcm-bridge-hub-router/src/lib.rs
```diff
@@ -0,0 +1,395 @@
...
+const MESSAGE_SIZE_FEE_BASE: FixedU128 = FixedU128::from_rational(1, 1000); // 0.001
+
+/// Maximal size of the XCM message that may be sent over bridge.
+///
+/// This should be less than the maximal size, allowed by the messages pallet, because
+/// the message itself is wrapped in other structs and is double encoded.
+pub const HARD_MESSAGE_SIZE_LIMIT: u32 = 32 * 1024;
+
+/// The target that will be used when publishing logs related to this pallet.
+pub const LOG_TARGET: &str = "runtime::bridge-hub-router";
```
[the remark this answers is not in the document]
svyatonik: True, but all other (bridge) pallet log targets are starting with `runtime::bridge`, so you can get all related logs using `RUST_LOG=runtime::bridge=trace`. OTOH xcm logs are using `xcm::` prefix like `xcm::currency_adapter`, `xcm::barriers` and so on. So I wonder if maybe we should make it `xcm::bridge-hub-router` or something like that? Or `runtime::bridge` is fine? WDYT?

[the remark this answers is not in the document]
serban300: True, I see. Then probably something that starts with `bridge` would be better. But then the fact that the log target differs from the pallet name would be confusing to me. If the pallet name is `pallet-xcm-bridge-hub-router`, I would expect the log target to be `xcm-bridge-hub-router`. I honestly don't know what would be better, but I don't have a strong preference here and it's not a blocker.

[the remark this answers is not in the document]
svyatonik: On a second though, maybe `xcm::` prefix is better for this pallet - it will be deployed at asset hubs (and other sending chains - other bridge pallets are deployed at bridge hubs.

svyatonik:
Since we've selected this option (backpressure) for dynamic fees, I'm going to close #2233 and #2268 for now. Will reopen if required:
- #2268 - we don't need it at all if we'll keep using backpressure;
- #2233 - most of its code is already in this PR and I'm going to open another PR to `master` with this code (for v2).

primitives/xcm-bridge-hub-router/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "bp-xcm-bridge-hub-router"
```
bkontur, line 2: @svyatonik 
just nit, what about renaming just to the `bp-xcm-router` + directory to `xcm-router`,
there is only `LocalXcmChannel` which seems pretty generic,
I dont know if more code is coming here, but I dont see here anything specific to `bridge-hub`

svyatonik: Don't know, my thinking was: we have our bridge pallets and pallet primitives (public types and traits, used in pallet configuration) are in the dedicated primitives crate/folder. Hence this name.

I could rename it to something that has "bridge" inside :) WDYT about `LocalBridgeHubXcmChannel`?

bkontur: ok, ok, let me walk through the whole PR first,
and then I will come back here :)

modules/messages/src/lib.rs
```diff
@@ -708,6 +720,9 @@ fn send_message<T: Config<I>, I: 'static>(
 		.send_message(encoded_payload)
 		.map_err(Error::<T, I>::MessageRejectedByPallet)?;
 
+	// return number of messages in the queue to let sender know about its state
+	let enqueued_messages = lane.data().queued_messages().checked_len().unwrap_or(0);
```
bkontur, line 724: @svyatonik 
I see two usages of `.queued_messages().checked_len().unwrap_or(0);`,
would it be worth to wrap it to `OutboundLaneData` e.g.
```
	pub fn queued_messages_len(&self) -> MessageNonce {
		self.queued_messages().checked_len().unwrap_or(0)
	}
``

**Comment by bkontur** [comment]
@svyatonik 
Slava, maybe I am wrong, but I think we are missing check/validation for source/target vs LaneId.

at least, I am sure that in actual Cumulus PRs we dont do that:
https://github.com/paritytech/cumulus/pull/2528
https://github.com/paritytech/cumulus/pull/2762

so if we open hrmp between whatever parachain to BridgeHub, all messages will go through the same hard-coded one lane, dedicated to AssetHubKusama/AssetHubPolkadot.

I think that we should be able to add it on MessageExporter level:
```
        fn validate(
		network: NetworkId,
		channel: u32,
		universal_source: &mut Option<InteriorMultiLocation>,
		destination: &mut Option<InteriorMultiLocation>,
		message: &mut Option<Xcm<()>>,
	) -> SendResult<Self::Ticket>;
```

Do you have this kind of validation for dynamic lanes?
I could add it for BridgeHubKusama/Polkadot somehow hard-coded.

wdyt?

File: primitives/xcm-bridge-hub-router/src/lib.rs [hunk 7]
```diff
@@ -0,0 +1,31 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Primitives of the `xcm-bridge-hub-router` pallet.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+/// Local XCM channel that may report whether it is congested or not.
+pub trait LocalXcmChannel {
```
bkontur, line 22: @svyatonik 
what about rename `LocalXcmChannel` -> `XcmChannelStatusProvider`?

suggestion is part of https://github.com/paritytech/parity-bridges-common/pull/2314

modules/xcm-bridge-hub-router/src/lib.rs
```diff
@@ -0,0 +1,400 @@
...
+		// `Self::exporter_for`)
+		ViaBridgeHubExporter::<T, I>::validate(dest, xcm)
+			.map(|(ticket, cost)| ((message_size, ticket), cost))
+	}
+
+	fn deliver(ticket: Self::Ticket) -> Result<XcmHash, SendError> {
+		// use router to enqueue message to the sibling/child bridge hub. This also should handle
+		// payment for passing through this queue.
+		let (message_size, ticket) = ticket;
+		let xcm_hash = T::ToBridgeHubSender::deliver(ticket)?;
```
bkontur, line 246: @svyatonik 
I dont know but I have a feeling that we should better use here for deliver the same thing what we used for validate,
I know that `ToBridgeHubSender` is Router of SovereignPaidRemoteExporter, but e.g. we could have a problem when somebody change/extend `SovereignPaidRemoteExporter::deliver` with some logic.
```suggestion
		let xcm_hash = ViaBridgeHubExporter::<T, I>::deliver(ticket)?;
```

suggestion is part of https://github.com/paritytech/parity-bridges-common/pull/2314

primitives/xcm-bridge-hub-router/Cargo.toml
bkontur: ok, from what I see now:

I think that everywhere we use word BridgeHub, we are referencing "system parachain BridgeHub".

`modules/xcm-bridge-hub-router`:
- I thought that that `xcm-bridge-hub-router` referenced to the very specific "system parachain BridgeHub"
- but I dont see any speciality about `BridgeHub` word or something which referenced "system parachain BridgeHub", bridge is just MultiLocation in this context which we call with `SendXcm`
- so I think, maybe name `modules/xcm-bridge-router` or `modules/bridge-xcm-router` (Router to XcmBridge or XcmRouter to Bridge?) is maybe enough

`primitives/xcm-bridge-hub-router`
- my thinking was about `primitives/xcm-router` or `primitives/xcm-bridge-router` or `primitives/bridge-xcm-router` as very generic primitives (that anybody can use for their own implementation of "xcm bridge router"), not related to any specific pallet

e.g. if we check `bin/runtime-common/src/messages_xcm_extension.rs` - which looks like generic/common xcm stuff but we have [here import](https://github.com/paritytech/parity-bridges-common/pull/2294/files#diff-f94bdda84771d04e176e0b8a789e3416426ad6b76d798f04b375a7f8600e660aR30) `use bp_xcm_bridge_hub_router::LocalXcmChannel;`  which looks pretty specific to "system parachain BridgeHub"

svyatonik: Not exactly. When bridge hub is mentioned, I'm mostly talking about the chain with [`pallet-xcm-bridge-hub`](https://github.com/paritytech/parity-bridges-common/tree/master/modules/xcm-bridge-hub) deployed - it supports multiple lanes, so it is technically a hub. And initially those two pallets were closely tied together, because they were need to talk to each other to ask and report bridge queues/lanes status. And maybe they'll be tied together in the future, because of https://github.com/paritytech/parity-bridges-common/pull/2313#discussion_r1281424022. However, I'm not sure yet - it needs more thinking.

There are some time anomalies with dynamic fees and dynamic lanes - the `pallet-xcm-bridge-hub` was born first, but `pallet-xcm-bridge-hub-router` will (likely) be deployed first. That said, I don't have anything against renaming those pallets to `pallet-xcm-bridge` and `pallet-xcm-bridge-router`, as you suggested. Although imo `hub` suffix gives more context here. Please comment to enact this change :)

modules/messages/src/lib.rs
svyatonik: True. Let's add another method to `RangeInclusiveExt` - i.e. `saturated_len()`. I think it is used everywhere, except one single place where we actually need the `checked_` version. I'll do that a bit later

svyatonik: done

primitives/xcm-bridge-hub-router/Cargo.toml
bkontur: ok, dont need to change anything, 
"bridge hub" as "multi lane" is clear now,
but just please update Docs with this distinction to avoid confusion with "system parachain bridge hub"

modules/messages/src/lib.rs
```diff
@@ -281,6 +283,9 @@ pub mod pallet {
 				Error::<T, I>::TooManyMessagesInTheProof
 			);
 
+			// if message dispatcher is currently inactive, we won't accept any messages
+			ensure!(T::MessageDispatch::is_active(), Error::<T, I>::MessageDispatchInactive);
```
bkontur, line 287: @svyatonik 
this is something I am not sure about,
on BridgeHubKusama we use this pallet for MessageDisptach of transfered message from BridgeHubPolkadot,
but now here we dont care if we disptaching to AssetHubKusama or to other parachain,

here https://github.com/paritytech/parity-bridges-common/pull/2318 and : https://github.com/paritytech/parity-bridges-common/issues/2315 you are working on backpreassure for "logical" sending channel

so I think this is opposite problem, we should also check `is_active/is_congested` for logical/psysical channel,
so I guess there should be some connection here between "lane" and "target destination"

I am going to try to PoC in https://github.com/paritytech/parity-bridges-common/pull/2319,
I have some thoughts like:
```
fn is_congested(dest: MultiLocation)
``

at least we dont need to write Adapter for every delivery parachain like here:
https://github.com/paritytech/cumulus/compare/bridge-hub-kusama-polkadot...sv-bridges-and-new-message-queue#diff-66b0425b5f194d88a1c1926558bccae96e56d6088b2123ed855b595a917997d3R492
but instead of `LocalXcmChannelWithSiblingAssetHub` we could have only one `XcmChannelStatusProviderForSiblingParachains`, which could parse `Parachain(para_id)` from MultiLocation and check `OutboundXcmpStatus` for para_id in generic way

Comment (svyatonik) [reply modules/messages/src/lib.rs:287]: @bkontur That's true - I was changing that in the relevant PR for v2 (https://github.com/paritytech/parity-bridges-common/pull/2261/files#diff-53b0bea1cfd5286348fbb357975027ca0832393dd64c9421cf4812286f7cc2c3R57-R61). For v1 let's assume that we will have just one lane - no need to complicate that.

Comment (bkontur) [reply modules/messages/src/lib.rs:287]: ahaaa, ok, cool, and lets hope we dont need to handle polkadot-collectives in v1 :)

at the end, we should define/update somewhere in docs very strict and clear terminology/definition for logical/physical channel vs lane+source, lane+destination and what is under backpreassure and dynamic fees - if it is logical or physical chanel or both

File: primitives/xcm-bridge-hub-router/src/lib.rs (hunk 7 above)
Comment (bkontur) [reply primitives/xcm-bridge-hub-router/src/lib.rs:22]: new suggestion: https://github.com/paritytech/parity-bridges-common/pull/2319 :)

**Review by bkontur — APPROVED** [review]
@svyatonik 
LGMT :)
I just left there two nits for docs update:
1. "bridge hub" as "multi lane"
2. logical/physical channel vs lane terminology...

**Comment by bkontur** [comment]
bot fmt

**Comment by svyatonik** [comment]
@bkontur do we have a bot now? :)

**Comment by alvicsam** [comment]
bot fmt

**Comment by alvicsam** [comment]
bot clean

File: bin/runtime-common/src/messages_xcm_extension.rs [hunk 10]
```diff
@@ -46,16 +53,25 @@ pub enum XcmBlobMessageDispatchResult {
...
+		BlobDispatcher: DispatchBlob,
+		Weights: MessagesPalletWeights,
+		Channel: XcmChannelStatusProvider,
+	> MessageDispatch for XcmBlobMessageDispatch<BlobDispatcher, Weights, Channel>
 {
 	type DispatchPayload = XcmAsPlainPayload;
 	type DispatchLevelResult = XcmBlobMessageDispatchResult;
 
+	fn is_active() -> bool {
+		!Channel::is_congested()
```
svyatonik, line 72: This code is executed at the target bridge hub. I suggest to use the following code for the `Channel` implementation:
```rust
pub struct LocalXcmChannelWithSiblingAssetHub;

impl bp_xcm_bridge_hub_router::XcmChannelStatusProvider for LocalXcmChannelWithSiblingAssetHub {
	fn is_congested() -> bool {
		// let's find the channel with the sibling AH
		let sibling_asset_hub_id: cumulus_primitives_core::ParaId = xcm_config::SiblingAssetHubParId::get().into();
		let outbound_channels = cumulus_pallet_xcmp_queue::OutboundXcmpStatus::<Runtime>::get();
		let channel_with_sibling_asset_hub = outbound_channels.iter()
			.filter(|c| c.recipient() == sibling_asset_hub_id)
			.next();

		// no channel => it is empty, so not congested
		let channel_with_sibling_asset_hub = match channel_with_sibling_asset_hub {
			Some(channel_with_sibling_asset_hub) => channel_with_sibling_asset_hub,
			None => return false,
		};

		// suspended channel => it is congested
		if channel_with_sibling_asset_hub.is_suspended() {
			return true;
		}

		// TODO: the following restriction is arguable, we may live without that, assuming that there
		// can't be more than some `N` messages queued at the bridge queue (at the source BH) AND before
		// accepting next (or next-after-next) delivery transaction, we'll receive the suspension signal
		// from the target AH and stop accepting delivery transactions

		// it takes some time for target AH to suspend inbound channel with the target BH and during that
		// we will keep accepting new message delivery transactions. Let's also reject new deliveries if
		// there are too many "pages" (concatenated XCM messages) in the target BH -> target AH queue.
		const MAX_QUEUED_PAGES_BEFORE_DEACTIVATION: u16 = 4;
		if channel_with_sibling_asset_hub.queued_pages() > MAX_QUEUED_PAGES_BEFORE_DEACTIVATION {
			return true;
		}

		true
	}
}
```

modules/xcm-bridge-hub-router/src/lib.rs
```diff
@@ -0,0 +1,557 @@
...
+	impl<T: Config<I>, I: 'static> Hooks<BlockNumberFor<T>> for Pallet<T, I> {
+		fn on_initialize(_n: BlockNumberFor<T>) -> Weight {
+			// TODO: make sure that `WithBridgeHubChannel::is_congested` returns true if either
+			// of XCM channels (outbound/inbound) is suspended. Because if outbound is suspended
+			// that is definitely congestion. If inbound is suspended, then we are not able to
+			// receive the "report_bridge_status" signal (that maybe sent by the bridge hub).
+
+			// if the channel with sibling/child bridge hub is suspended, we don't change
+			// anything
+			if T::WithBridgeHubChannel::is_congested() {
```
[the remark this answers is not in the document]
svyatonik: The base fee then need to include both bridge fee and HRMP fee. And then instead of the `BridgeFee * BridgeFeeFactor + HrmpFee * HrmpFeeFactor`, we'll temporary have the `(BridgeFee  + HrmpFee) * BridgeFeeFactor` until https://github.com/paritytech/cumulus/pull/2342 is implemented. This is fine, because the exponential factor is present in the formula.

acatangiu (approved):

bkontur:
bot fmt

modules/xcm-bridge-hub-router/src/lib.rs
```diff
@@ -0,0 +1,557 @@
...
+
+			log::info!(
+				target: LOG_TARGET,
+				"Received bridge status from {:?}: congested = {}",
+				bridge_id,
+				is_congested,
+			);
+
+			Bridge::<T, I>::mutate(|bridge| {
+				bridge.is_congested = is_congested;
```
bkontur, line 172: @svyatonik 
just asking, shouldnt we increase delivery fee factor here, when we receive `is_congested=true`?
because, next message to BridgeHub right after `report_bridge_status(is_congested=true)` will have normal fee,

I think this is the situation now:
```
send message 1 -> normal fee
send message 2 -> normal fee
receive `report_bridge_status(is_congested=true)`
send message 3 -> normal fee    // ??? normal or increased fee ???
send message 4 -> increased fee
```

svyatonik: We could, but the effect will be minimal. Like right now we are increasing fee by 1% - it doesn't matter much if **next** message will be 1% more expensive (1 EUR vs 1.01). The main thing is that sending a bunch of messages through congested bridge will be much more (exponentially) expensive. So this effect is for a "long distance".

bkontur: aha, ok, thank you, got it

bkontur:
bot rebase

Polkadot-Forum:

Issue #2296, opened 2023-07-24
Store justifications accepted by `submit_finality_proof()`


serban300:
Related to https://github.com/paritytech/parity-bridges-common/issues/39

We need to store the justifications accepted by `submit_finality_proof()` and maybe have a subscription that would emit notifications related to new accepted proofs

serban300:
Maybe we can use [`subscribeStorage`](https://polkadot.js.org/docs/substrate/rpc/#subscribestoragekeys-vecstoragekey-storagechangeset) for getting notifications when a new proof is accepted

serban300:
Resolved by the referenced PRs

PR #2298, opened 2023-07-25
GRANDPA module: store accepted justifications

serban300:
Related to #2296


modules/grandpa/src/lib.rs
```diff
@@ -319,6 +319,16 @@ pub mod pallet {
 	pub type BestFinalized<T: Config<I>, I: 'static = ()> =
 		StorageValue<_, BridgedBlockId<T, I>, OptionQuery>;
 
+	/// Justification of the best finalized header.
+	///
+	/// NOTE: This storage entry is only used by relayers to check for equivocations.
```
svyatonik, line 324: IMO events are exactly for that - for stuff that should never be read by runtime code && only accessible outside. I guess you're adding it here as a separate storage item because maybe it is easier to write the relayer code - you don't need to deal with pallet inidices? Or is there some other reason?

I suggest to move it to events. I could imagine for example a batch call with `All(submit_finality_proof(BadProof), submit_finality_proof(GoodProof))` and if we'll only be looking at the state at the end of block, we may miss the `BadProof`. The same with the scheduled calls or XCM `Transact` calls - they can be executed from `on_idle`/`on_initialize` and we may miss something. System events will have a record for every justification of a block && we can easily get all submitted justifications by iterating events.

serban300: > I could imagine for example a batch call with `All(submit_finality_proof(BadProof), submit_finality_proof(GoodProof))` and if we'll only be looking at the state at the end of block, we may miss the `BadProof`.

Oh, you're right. Nice catch !



> I guess you're adding it here as a separate storage item because maybe it is easier to write the relayer code - you don't need to deal with pallet inidices? Or is there some other reason?

High level, yes, I think it would be easier to write the relayer code. Because we could use [`subscribeStorage`](https://polkadot.js.org/docs/substrate/rpc/#subscribestoragekeys-vecstoragekey-storagechangeset) and get a notification each time `BestJustification` changes. With events from what I understand we can't have a subscription for a certain event. We just have to iterate over all the events each block and look for the ones that we need. I'll do some experiments. But that's what I understood from the documentation.

svyatonik: > High level, yes, I think it would be easier to write the relayer code. Because we could use [subscribeStorage](https://polkadot.js.org/docs/substrate/rpc/#subscribestoragekeys-vecstoragekey-storagechangeset) and get a notification each time BestJustification changes. With events from what I understand we can't have a subscription for a certain event. We just have to iterate over all the events each block and look for the ones that we need. I'll do some experiments. But that's what I understood from the documentation.

True, but I'm not sure how this RPC subscription works with multiple changes ar the single block. E.g. the same `All(submit_finality_proof(BadProof), submit_finality_proof(GoodProof))` - it'll emit the single event with `GoodProof` or two with `BadProof` and `GoodProof`. I think the latter, because it is implemented at the client level so it doesn't know what happens inside the block. I may be wrong here, though.

Also - I wouldn't rely on subscriptions - it should be combined with simple storage reads. Subscriptions have some caveats, that maybe are ok for regular usages, but for something that you're working on (i.e. component that can't tolerate any missing justifications). Like IIRC e.g. finality subscription may skip ranges of blocks if it finalizes many blocks at once. Also I'm not sure if there's a limit on number of active subscriptions on a node? Also - regular network outages, which can lead to missing blocks => justifications.

So imo your component may use subscriptions, but it should also track latest block where you have read justifications. And if something goes wrong (network or process crash), it should iterate all missed blocks on restart and check justifications from there. Maybe it'd be easier to avoid subscriptions at all and use regular polling, but it is up to you

serban300: Done. PTAL !

svyatonik: Sorry for the delay - I promise to review tomorrow

primitives/header-chain/src/lib.rs
```diff
@@ -225,6 +225,13 @@ pub trait ChainWithGrandpa: Chain {
 	/// the same name.
 	const WITH_CHAIN_GRANDPA_PALLET_NAME: &'static str;
 
+	/// Name of the runtime API method that is returning the GRANDPA justifications accepted
+	/// by the `submit_finality_proofs` extrinsic in the queried block.
+	///
+	/// Keep in mind that this method is normally provided by the other chain, which is
+	/// bridged with this chain.
+	const ACCEPTED_FINALITY_PROOFS_METHOD: &'static str;
```
svyatonik, line 233: Do we really need this at primitives level? We don't need it in the runtime, only from relay code - maybe worth moving it to `relays/*`?

primitives/runtime/src/chain.rs
```diff
@@ -319,6 +319,45 @@ macro_rules! decl_bridge_finality_runtime_apis {
...
 
+/// Convenience macro that declares bridge GRANDPA finality runtime apis and related constants
+/// for a chain.
+/// This includes:
+/// - chain-specific bridge runtime APIs:
+///     - `<ThisChain>GrandpaFinalityApi`
+/// - constants that are stringified names of runtime API methods:
+///     - `<THIS_CHAIN>_GRANDPA_JUSTIFICATIONS_METHOD`
+/// The name of the chain has to be specified in snake case (e.g. `rialto_parachain`).
+#[macro_export]
```
svyatonik, line 330: Maybe we could change the existing `decl_bridge_finality_runtime_apis` macro to generate this additional method in the same trait. E.g. for parachains the "call" will be the `decl_bridge_finality_runtime_apis!(bridge_hub_wococo)`, but for standalone/relay chains it'll be something like `decl_bridge_finality_runtime_apis!(kusama, grandpa)`? But feel free to leave as is

svyatonik (approved):
Looks good, but maybe would be better to move `bp_header_chain::ChainWithGrandpa::ACCEPTED_FINALITY_PROOFS_METHOD` to relay level (e.g. to add to removed `relay_substrate_client::ChainWithGrandpa`).

primitives/header-chain/src/lib.rs
serban300: Done.

primitives/runtime/src/chain.rs
serban300: I added a new API for it for 2 reasons:
1. I don't know if it make sense for parachains to implement it. 
2. There will be a separate implementation (so probably a separate API) for BEEFY

Personally I would keep it as it is for the moment. We can change it in the future if needed.

svyatonik: > 1. I don't know if it make sense for parachains to implement it.

That's exactly why I have suggested to add this 2nd arg to existing macro :) So it'll generate trait with single (old) method for parachains and two methods (old and new) for standalone/relay chains :) But ok - let's leave as is - it's fine too

serban300: Oh, got it. I misread it the first time. Makes sense. Just wondering if it wouldn't be confusing to have the same API with different methods depending on the context. I'm not sure how it would be best. I would still go with a separate API for the beginning. But will keep this in mind and maybe we'll get a better picture as the implementation progresses.

serban300: Actually looks good. Done.

svyatonik (approved):
Cool, thank you!

PR #2313, opened 2023-08-01
(Suggestion) Ability to externalize configuration for `ExporterFor`

bkontur:
(Replaced `BridgedNetworkId/SiblingBridgeHubLocation` with `Bridges: ExporterFor`)

actual `ExportFor` impl for Pallet matches only bridged newtork, which is not enough,
e.g. if we use `pallet_xcm::reserve_transfer_assets` on AssetHubKusama and we set destination as polkadot-collectives `MultiLocation { parents: 2, X2(GlobalConsensus(Polkadot), Parachain(1001))}`,
nothing stops to send this message over bridge to polkadot-collectives,
so for AssetHub we need to control exactly where we allow to send assets,

so for AssetHubs few days ago I added implementation for `ExportFor` which matches also remote destination:
https://github.com/paritytech/cumulus/pull/2762/commits/5103de8d3e9ee22df559b337a94c79f6428e2eb5

**TODO:**
- [x] how to handle `maybe_payment`? subsume with `(T::FeeAsset::get(), fee)`?
- [x] fix compiling other runtimes (millau, ...) :)

### DIFF
--- a/modules/xcm-bridge-hub-router/src/lib.rs
+++ b/modules/xcm-bridge-hub-router/src/lib.rs
@@ -73,10 +73,10 @@ pub mod pallet {
 
 		/// Universal location of this runtime.
 		type UniversalLocation: Get<InteriorMultiLocation>;
-		/// Relative location of the sibling bridge hub.
-		type SiblingBridgeHubLocation: Get<MultiLocation>;
-		/// The bridged network that this config is for.
-		type BridgedNetworkId: Get<NetworkId>;
+		/// Configuration for supported **bridged networks/locations** with **bridge location** and
+		/// **possible fee**. Allows to externalize better control over allowed **bridged
+		/// networks/locations**.
+		type Bridges: ExporterFor;
 
 		/// Actual message sender (`HRMP` or `DMP`) to the sibling bridge hub location.
 		type ToBridgeHubSender: SendXcm;
@@ -177,13 +177,13 @@ type ViaBridgeHubExporter<T, I> = SovereignPaidRemoteExporter<
 impl<T: Config<I>, I: 'static> ExporterFor for Pallet<T, I> {
 	fn exporter_for(
 		network: &NetworkId,
-		_remote_location: &InteriorMultiLocation,
+		remote_location: &InteriorMultiLocation,
 		message: &Xcm<()>,
 	) -> Option<(MultiLocation, Option<MultiAsset>)> {
-		// ensure that the message is sent to the expected bridged network
-		if *network != T::BridgedNetworkId::get() {
+		// ensure that the message is sent to the expected bridged network/location.
+		let Some((bridge_hub_location, maybe_payment)) = T::Bridges::exporter_for(network, remote_location, message) else {
 			return None
-		}
+		};
 
 		// compute fee amount. Keep in mind that this is only the bridge fee. The fee for sending
 		// message from this chain to child/sibling bridge hub is determined by the
@@ -194,6 +194,8 @@ impl<T: Config<I>, I: 'static> ExporterFor for Pallet<T, I> {
 		let fee_factor = Self::delivery_fee_factor();
 		let fee = fee_factor.saturating_mul_int(fee_sum);
 
+		// TODO: how to handle `maybe_payment`? subsume with `(T::FeeAsset::get(), fee)`?
+
 		log::info!(
 			target: LOG_TARGET,
 			"Going to send message ({} bytes) over bridge. Computed bridge fee {} using fee factor {}",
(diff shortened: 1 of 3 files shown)


bkontur:
another small positive side-effect is that for bridging to ethereum or other consensus, we dont need to add another xcm-bridge-hub-router instance, we just change `Bridges` cfg

modules/xcm-bridge-hub-router/src/lib.rs
```diff
@@ -194,6 +194,8 @@ impl<T: Config<I>, I: 'static> ExporterFor for Pallet<T, I> {
 		let fee_factor = Self::delivery_fee_factor();
 		let fee = fee_factor.saturating_mul_int(fee_sum);
 
+		// TODO: how to handle `maybe_payment`? subsume with `(T::FeeAsset::get(), fee)`?
```
svyatonik, line 197: I think we can subsume and use it instead of `Config;:BaseFee`.

modules/xcm-bridge-hub-router/src/lib.rs
```diff
@@ -177,13 +177,13 @@ type ViaBridgeHubExporter<T, I> = SovereignPaidRemoteExporter<
...
 	fn exporter_for(
 		network: &NetworkId,
-		_remote_location: &InteriorMultiLocation,
+		remote_location: &InteriorMultiLocation,
 		message: &Xcm<()>,
 	) -> Option<(MultiLocation, Option<MultiAsset>)> {
-		// ensure that the message is sent to the expected bridged network
-		if *network != T::BridgedNetworkId::get() {
+		// ensure that the message is sent to the expected bridged network/location.
+		let Some((bridge_hub_location, maybe_payment)) = T::Bridges::exporter_for(network, remote_location, message) else {
```
svyatonik, line 184: What concerns me here is exactly:  "another small positive side-effect is that for bridging to ethereum or other consensus, we dont need to add another xcm-bridge-hub-router instance, we just change Bridges cfg". IDK what will be the cost of with-Ethereum bridge, but my guess it'll be a bit more expensive than ours. So maybe they'll be using different `BaseFee` and `ByteFee` and we'll need to have a different instance here, or use something more sophisticated configuration trait than `ExporterFor`.

Also - I admit that our current dynamic fees architecture will work fine with a single remote consensus. But I have not been thinking about multiple remote consensus systems. If with-Polkadot bridge is congested (e.g. because relayer is not running), shall we also pause the with-Ethereum bridge? It might be a problem and I have no answer right now. Thanks for spotting that. Please share your thoughts here.

svyatonik (approved):
LGTM, thanks for spotting that! But please check out comments first. My suggestions:
1) to change how we handle fee, returned by `ExporterFor`;
2) and keep the`BridgedNetwotkId` check to avoid using the same pallet instance for different consensus systems.

modules/xcm-bridge-hub-router/src/lib.rs
svyatonik: > But I have not been thinking about multiple remote consensus systems

Maybe we should do some hybrid solution for v2. I.e. if one of bridges is congested, then instead of blocking the XCM channel with sending chain, it shall send XCM message to that chain with the congested bridge id. And then, if it keeps receiving messages to be sent over this bridge, suspend the XCM channel. Anyway - imo that's related to v2, right? We shall track this in some issue

svyatonik: I've filed #2315 to track that

bkontur:
bot rebase

modules/xcm-bridge-hub-router/src/lib.rs
bkontur: ok, thank you, you're right with your concerns, I totally got it now,
I returned back `BridgedNetwotkId`, but as optional, that user can decide, if he needs separate router for bridged network or not, with consequences of the same BaseFee/ByteFee and backpressure #2315,
or do you thing we should be better strict here with `type BridgedNetworkId: Get<NetworkId>;`?

bkontur: > If with-Polkadot bridge is congested (e.g. because relayer is not running), shall we also pause the with-Ethereum bridge? It might be a problem and I have no answer right now. Thanks for spotting that. Please share your thoughts here.

some thoughts [here](https://github.com/paritytech/parity-bridges-common/issues/2315#issuecomment-1662062169)

bkontur: > Let's keep as is. For v1 (hopefully) we'll deal with the single remote network only, right? For v2 - let's see if/how we'll implement the #2315


well, hard to say, for example, there is a snowfork ethereum bridge comming, not sure if it will be part of v1 or v2?
other use-case is bridging polkadot-collectives to kusama-somewhere (propagating governance origins for technical fellowship), but thats probably v2 (at least after asset transfer)

modules/xcm-bridge-hub-router/src/lib.rs
bkontur: @svyatonik 
I tried to fix maybe_payment, please, check https://github.com/paritytech/parity-bridges-common/pull/2313/commits/623c9c0e5feae38e948907a6b5f2f1c0354a3fb7

svyatonik (approved):
Looks good,m thank you for the help

PR #2314, opened 2023-08-01
(Suggestion) Join `ToBridgeHubSender/WithBridgeHubChannel` to `ToBridgeHubChannel` + nits

bkontur:


primitives/xcm-bridge-hub-router/src/lib.rs
```diff
@@ -18,14 +18,25 @@
...
 		false
 	}
 }
+
+/// XCM channel with dedicated `Sender` and ability to report whether it is congested or not.
+pub trait ToXcmChannel: XcmChannelStatusProvider {
+	type Sender: SendXcm;
+}
+
+impl<T: SendXcm + XcmChannelStatusProvider> ToXcmChannel for T {
```
svyatonik, line 40: 1) IIUC we won't be able to use this blanket implementation, because those queues (HRMP/DMP/UMP) live in different repos (Polkadot/Cumulus) and we are only plugged into Cumulus
2) there are (right now) differences in `is_congested` implementation for the same queue. E.g. at `Source.AH` it just checks if the channel is suspended. At the Target.BH it also checks if there are more than `N` pages in the outbound queue. This is an arguable decision - maybe we don't need to check the number of pages in the outbound queue, but that's how it is implemented now.

primitives/xcm-bridge-hub-router/src/lib.rs
```diff
@@ -18,14 +18,25 @@
...
-impl LocalXcmChannel for () {
+impl XcmChannelStatusProvider for () {
 	fn is_congested() -> bool {
 		false
 	}
 }
+
+/// XCM channel with dedicated `Sender` and ability to report whether it is congested or not.
+pub trait ToXcmChannel: XcmChannelStatusProvider {
+	type Sender: SendXcm;
```
svyatonik, line 37: If you're keen to keep it, please add documentation for this public item :)

primitives/xcm-bridge-hub-router/src/lib.rs
svyatonik: (that's not the blocker - you may merge as is, but imo blanket impl is not needed here)

svyatonik (approved):
Don't have anything against merging two config options in one.

bkontur:
ok, lets go with two config options, anyway we could come back later,
closing this in favour of https://github.com/paritytech/parity-bridges-common/pull/2319

Issue #2310, opened 2023-07-31
Make finality loop code more extensible


serban300:
Related to https://github.com/paritytech/parity-bridges-common/issues/39

I think that conceptually we could add the logic that checks for equivocations in the finality loop. But in order to do this, we need to make the loop more extensible.

PR #2317, opened 2023-08-02
Finality loop refactoring

serban300:
Related to #2310 

Some refactoring for the finality loop in order to make it more extensible. Main changes:

- Moved the loop logic in a struct in order to make it easier to track and access its state 
- Split the logic in more structs and files
- Moved the tests to the corresponding files

Only 2 functional change:
- avoid restarting the loop when the `submit_finality_proof` tx fails. Seems that we just have to set `self.best_submitted_number = None` in order to avoid stalling.
- avoid restarting the loop when encountering an error while re/subscribing to the source finality stream (`ensure_finality_proofs_stream`). If it's not a connection error

Doing this since in the future we need to be more careful about restarting and we should avoid losing the loop state. If we'll want to check for equivocations inside the loop.

Tested this locally overnight and there were no issues. 

Posting this PR since there are already a lot of changes. Probably some more cosmetics and small changed will follow, but in future PRs.


bkontur:
is it something that we also need to backport to polkadot-staging?

serban300:
> is it something that we also need to backport to polkadot-staging?

Yes. Because the equivocation logic will come on top of it.

svyatonik:
@serban300 Sorry, I've missed the #2310 - my fault. Could you, please, elaborate your (draft) design for this component? Below are my thoughts - please don't take it as a requirement. You may have your own design, which may be better. So just PTAL and if you'll say we should go with this PR, let me know - I'll review it :)

---

From my perspective, it may be worth to consider a separate component (crate) - imo the only common thing between equivocations detector (ED) and finality relay is that they are listening to the source node justifications stream. Then the ED should (imo):
1) save them to some persistent storage - relayers have no storage right now;
2) poll/subscribe target chain for new events at every block - it should read all emitted justifications and save them to some persistent storage. Last "processed" block must also be saved in this storage to survive possible restarts. Relayers do not do that;
3) when it sees new justification at the target chain, it should read all saved justifications from (1) and try to find equivocations. All saved justifications from previous rounds must be discarded;
4) it also should track authorities set changes at both chains. Finality relayer partially do that;
5) (maybe) it should also have some mechanism to guarantee that the equivocation-submit transactions has actually been mined. I.e. not to prune ^^^ before equivocation has been submitted.

IMO adding all this stuff into finality relay will make it too complex simply because there is a lot of differences between those components. Like right now I think the messages relay is too complex and I'd like to split it into two different relays some day (but after we'll have that relayers sync protocol). Also:
1) finality relayer, when works in a complex mode has the "mocked" source client, which returns fake source block. I think this may cause some issues with ED. Not sure, though (https://github.com/paritytech/parity-bridges-common/blob/master/relays/lib-substrate-relay/src/finality/source.rs#L222);
2) the ED should be able to submit transactions to the source chain, which is the opposite to current finality relay.
3) it might be easier to deploy this ED (separate crate) e.g. on either chain collators. So that they may only run ED without actually relaying. Also I've been thinking of including finality relay into target chain collators code (include = remove finality relay at all and replace it with inherent or something like that), so it also might be easier for us to have a separate ED component.

serban300:
@svyatonik I was just thinking that we already have the ephemeral finality proofs here, and that we could add the logic that polls for accepted proofs at the target and then look for equivocations.

But I think you're right. A different component would be simpler and then it would be easier to deploy on collators. So I can go with the separate component approach. But since this PR is ready, please also review it when you have time. It doesn't add any equivocation logic. It only contains finality loop cosmetics. 

Not sure about porting it to `polkadot-staging` then. I think it would be best to do it. But we could also avoid it.

svyatonik:
Ok, thank you! I'll review it soon. Re `polkadot-staging` - imo maybe don't merge unless we strictly need it.

svyatonik (approved):
Did a brief review and it looks good. I'd like to do another deeper review if you'll be porting it to the `polkadot-staging`.

serban300:
> Did a brief review and it looks good. I'd like to do another deeper review if you'll be porting it to the `polkadot-staging`.

Thanks ! I won't be porting it for the moment. I will leave it to run in the CI for a couple of days at least. And start working on the equivocation component. If we'll have structs that we can reuse between the finality loop and the equivocation loop, maybe we'll port it. But I think we can discuss this then. And it will be a new PR anyway.

PR #2318, opened 2023-08-03
Dynamic fees v1: report congestion status to sending chain

### LINKED ISSUE #2315: Think of more sophisticated algorithm than blocking XCM channel between sending chain and bridge hub
#2312 has backported the `pallet-xcm-bridge-hub` router 1:1 from v1 version (#2294). But @bkontur has raised a valid concern here:https://github.com/paritytech/parity-bridges-common/pull/2313#discussion_r1281424022. E.g. if the same chain is bridging with two different consensus systems, is it fine to block both bridges if at least one lane in one of bridges is blocked. With single remote consensus, it is maybe fine, because logically they're using the single channel across the pipeline - it is a bridge specific that messages are delivered over multiple lanes. But if we are bridging with other consensus system, then both bridges will stall until we resolve a single lane issue. It doesn't looks good to me.

OTOH for the sending chain, the channel for all outbound messages is the same - it is the channel to the sibling/child bridge hub. So maybe it is fine?

This needs some more thinking, so I'm filing the separate issue.

svyatonik:
closes #2315

TODO:
- [x] required changes to the `pallet-xcm-bridge-hub-router` at the sending chain: https://github.com/paritytech/parity-bridges-common/commit/ac799a88e3e50969e28b3b8e0472e503d4cc9b57
- [x] required changes at the bridge hub: https://github.com/paritytech/parity-bridges-common/pull/2318/commits/d78f10664a120ccd9c18aae7d475fc497d3d0e1a and https://github.com/paritytech/parity-bridges-common/pull/2318/commits/c3648469362c94d3f2c44e433d9815d0c18bb8bf;
- [x] integrate into Cumulus and do manual tests: https://github.com/paritytech/cumulus/tree/sv-bridges-and-new-message-queue;
- [x] unit tests;
- [x] cleanup.

What this PR does:
1) instead of physically suspending the XCM channel Source.AH -> Source.BH, we are sending the "congested" message from Source.BH to Source.AH when the lane has more than `OUTBOUND_LANE_CONGESTED_THRESHOLD` messages;
2) upon receiving this message, the Source.AH starts increasing fee for every bridge message;
3) once oubound lane has more than `OUTBOUND_LANE_UNCONGESTED_THRESHOLD` messages, it sends the "uncongested" message to the Source.AH;
4) upon receiving this message, the Source.AH starts relieving the fee factor.

Obviously there's a period when the "congested" message is sent to the Source.AH, but is not yet dispatched. So messages may be sent with a constant (previous) fee. This may be alleviated by e.g. sending a ping message when `OUTBOUND_LANE_CONGESTED_THRESHOLD` messages are sent and if pong is not received in `N` blocks, start increasing the fee. This is what has been done in the https://github.com/paritytech/parity-bridges-common/pull/2233 initially. IDK if we need to do that, because:
1) the HRMP cost (which is a part of fee) still has its own exponential fee, which would make messages more expensive;
2) I've added another condition to start increasing the exponential fee - if inbound HRMP queue from Source.BH to Source.AH is suspended (meaning that Source.AH is failing to process inbound messages in a timely manner), then we trigger our own exponential fee increase mechanism.

So to me it looks like the above is enough for us, because it'll be expensive to fill up both queues (Source.AH -> Source.BH and Source.BH -> Source.AH). Also it needs to be coordinated carefully, so that the Source.BH -> Source.AH won't trigger the channel suspension. But please share your thoughts

UPD: in other words, with current implementation, we accept the fact that if the Source.BH -> Source.AH XCM channel is NOT congested, bridge messages may be sent with constant bridge fee (although HRMP fee may apply its own exponential factor meanwhile) while "congested" message is being delivered. We may also change the `is_congested` impl a bit and look at the number of unprocessed Source.BH -> Source AH messages and if it is larger than some `N`, consider it congested. The `N` may be for example the average number of messages, processed at a single Source.AH block.


modules/xcm-bridge-hub-router/src/lib.rs
```diff
@@ -100,67 +103,111 @@ pub mod pallet {
 	#[pallet::hooks]
 	impl<T: Config<I>, I: 'static> Hooks<BlockNumberFor<T>> for Pallet<T, I> {
 		fn on_initialize(_n: BlockNumberFor<T>) -> Weight {
-			// if XCM queue is still congested, we don't change anything
-			if T::WithBridgeHubChannel::is_congested() {
-				return T::WeightInfo::on_initialize_when_congested()
-			}
+			Bridge::<T, I>::mutate(|bridge| {
+				// TODO: make sure that `WithBridgeHubChannel::is_congested` returns true if either
```
svyatonik, line 107: Relevant commit: https://github.com/paritytech/cumulus/commit/c0e3225cca573a7ee64180e3bedade6048a9c945. I'll remove this TODO before merging

bkontur:
@svyatonik 
just nit, in your Cumulus branch, I have never seen CumulusXcm used anywhere, I guess this module is really not needed and maybe will be removed,
and for matching XCM origin (e.g. from sibling parachain for example for ForeignCreatetors) we used something like this:
```
                let origin_location = EnsureXcm::<Everything>::try_origin(origin.clone())?;
		if origin_location.eq(BridgeHubKusama::get()) {
  ....
                }
```

svyatonik:
> CumulusXcm

@bkontur Which code exactly are you talking about? Could you, please, share a link?

bkontur:
> > CumulusXcm
> 
> @bkontur Which code exactly are you talking about? Could you, please, share a link?

https://github.com/paritytech/cumulus/compare/bridge-hub-kusama-polkadot...sv-bridges-and-new-message-queue#diff-9a51956f7781a349d207a9d5f54ea0061666a856fb108d1a1190bd5178eaac50R647-R657

hmm, I am sure that I saw there `CumulusXcm` somewhere :D :D , nevermind :D

bin/rialto/runtime/src/millau_messages.rs
```diff
@@ -125,10 +128,16 @@ pub type ToMillauBlobExporter = HaulBlobExporter<
...
-	type MessageSenderOrigin = RuntimeOrigin;
+	type Runtime = Runtime;
+	type MessagesInstance = WithMillauMessagesInstance;
 	type SenderAndLane = MullauSenderAndLane;
 
+	type ToSendingChainSender = crate::xcm_config::XcmRouter;
+	type CongestedMessage = DummyXcmMessage;
+	type UncongestedMessage = DummyXcmMessage;
+
+	type MessageSenderOrigin = RuntimeOrigin;
```
bkontur, line 139: @svyatonik 
sorry, unrelated, to this, 

but can we remove this `MessageSenderOrigin` we dont really use it for anything, 
I think now it just brings confusion to configuration, because it is hard-coded and not the real sender (e.g. sibling parachain)
```
pallet_xcm::Origin::from(MultiLocation::new(1, crate::xcm_config::UniversalLocation::get())).into()
...
pallet_xcm::Origin::from(MultiLocation::here()).into()
```

I think this `message_sender_origin` had its place when `send_message` was exposed as a extrinsic, but it is not now

svyatonik: Yep, I agree there's a lot of things that may confuse in the configuration. I've been working on trying to make configuration easier in #1666. E.g. this origin is already gone. But I wanted to avoid pushing unnecessary changes to `polkadot-staging`, so it was "scheduled" for v2.

But this particular type is well isolated and if you think it'd be better to remove it in v1, I could prepare a separate PR.

primitives/xcm-bridge-hub-router/src/lib.rs
```diff
@@ -29,3 +33,18 @@ impl XcmChannelStatusProvider for () {
 		false
 	}
 }
+
+/// Current status of the bridge.
+#[derive(Clone, Decode, Encode, Eq, PartialEq, TypeInfo, MaxEncodedLen, RuntimeDebug)]
+pub struct BridgeState {
```
bkontur, line 39: @svyatonik 
I think, we should rename this to `BridgeHubState` and use bridge-hub instead of bridge here,
it brings a little bit confusion what is "bridge" and "bridge hub" here,
because we have "xcm router to bridge hub",
and also I would rename in router:
```
pub type Bridge<T: Config<I>, I: 'static = ()> = StorageValue<_, BridgeState, ValueQuery>;
pub type BridgeHub<T: Config<I>, I: 'static = ()> = StorageValue<_, BridgeHubState, ValueQuery>;
```

bkontur: hmm, how do you change it later?
```
pub type BridgeHub<T: Config<I>, I: 'static = ()> = StorageMap<_, LaneId, BridgeHubState, ValueQuery>;
```

bkontur: maybe `BridgeHubLaneState`?

modules/messages/src/lib.rs
```diff
@@ -492,6 +495,12 @@ pub mod pallet {
 				lane_id,
 			);
 
+			// notify others about messages delivery
+			T::OnMessagesDelivered::on_messages_delivered(
+				lane_id,
+				lane.data().queued_messages().checked_len().unwrap_or(0),
```
bkontur, line 501: @svyatonik 
what about `saturating_len()`?

primitives/xcm-bridge-hub-router/src/lib.rs
svyatonik: Yes, it'll be the map in the future. So it isn't the bridge hub state - it is the state of single bridge that is served by single bridge hub. And at bridge hub it has a dedicated lane. And there'll be something like `type BridgeId = LaneId`. So sending chains (AH), where this router will be deployed, will only deal with "bridges" and "bridge hub". At bridge hub we'll deal with both bridges and their dedicated lanes.

Now we have the single lane, so I could rename it to `BridgeHubState` as you suggested. But in v2, it'll be the `BridgeState` anyway

modules/messages/src/lib.rs
svyatonik, line 501:
```suggestion
				lane.data().queued_messages().saturating_len(),
```

bin/rialto/runtime/src/millau_messages.rs
bkontur: yes, I would even delete it directly in `polkadot-staging` now

primitives/xcm-bridge-hub-router/src/lib.rs
bkontur: ok, no need to rename `BridgeState`,
"bridge" is too vague , I mean "bridge" word is everywhere here :D :D :D,
at least from the doc and terminology needs to be clear what is "bridge" vs "lane" in this context

bin/rialto/runtime/src/millau_messages.rs
svyatonik: okay - will do soon :)

serban300 (commented):
Looks good. Just left a few small comments

modules/xcm-bridge-hub-router/src/lib.rs
```diff
@@ -100,67 +103,111 @@ pub mod pallet {
...
-			// ok - we need to increase the fee factor, let's do that
-			let message_size_factor = FixedU128::from_u32(message_size.saturating_div(1024))
-				.saturating_mul(MESSAGE_SIZE_FEE_BASE);
-			let total_factor = EXPONENTIAL_FEE_BASE.saturating_add(message_size_factor);
-			DeliveryFeeFactor::<T, I>::mutate(|f| {
-				let previous_factor = *f;
-				*f = f.saturating_mul(total_factor);
 				log::info!(
 					target: LOG_TARGET,
 					"Bridge queue is congested. Increased fee factor from {} to {}",
```
bkontur, line 205: @svyatonik 
maybe worth to `deposit_event(DeliveryFeeFactorChanged(old_factor, new_factor))`,
maybe just move to v2?

modules/xcm-bridge-hub-router/src/lib.rs
```diff
@@ -100,67 +103,111 @@ pub mod pallet {
...
+				}
 
-					T::WeightInfo::on_initialize_when_non_congested()
-				} else {
-					// we have not actually updated the `DeliveryFeeFactor`, so we may deduct
-					// single db write from maximal weight
-					T::WeightInfo::on_initialize_when_non_congested()
-						.saturating_sub(T::DbWeight::get().writes(1))
+				// if fee factor is already minimal, we don't change anything
+				if bridge.delivery_fee_factor == FixedU128::one() {
```
bkontur, line 124: I see we use `FixedU128::one()` on several places and also for default vaule: `BridgeState { delivery_fee_factor: FixedU128::one()`
is it worth to introduce some constant like MINIMAL_DELIVERY_FEE_FACTOR and use that instead of `FixedU128::one()` everywhere?

modules/xcm-bridge-hub-router/src/lib.rs
```diff
@@ -100,67 +103,111 @@ pub mod pallet {
...
-				return
-			}
+			let _ = Bridge::<T, I>::try_mutate(|bridge| {
+				let is_channel_with_bridge_hub_congested = T::WithBridgeHubChannel::is_congested();
+				let is_bridge_congested = bridge.is_congested;
+
+				// if outbound queue is not congested AND bridge has not reported congestion, do
+				// nothing
+				if !is_channel_with_bridge_hub_congested && !is_bridge_congested {
+					return Err(())
```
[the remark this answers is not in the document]
svyatonik: done

modules/xcm-bridge-hub-router/src/benchmarking.rs
```diff
@@ -18,11 +18,19 @@
 
 #![cfg(feature = "runtime-benchmarks")]
 
-use crate::{DeliveryFeeFactor, InitialFactor};
+use crate::{Bridge, Call};
```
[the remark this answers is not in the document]
svyatonik: Let's do it in a separate PR - it's just a cosmetics in the end :)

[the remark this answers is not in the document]
svyatonik: (I'll file an issue)

modules/xcm-bridge-hub-router/src/benchmarking.rs
```diff
@@ -35,15 +43,50 @@ pub trait Config<I: 'static>: crate::Config<I> {
 
 benchmarks_instance_pallet! {
 	on_initialize_when_non_congested {
-		DeliveryFeeFactor::<T, I>::put(InitialFactor::get() + InitialFactor::get());
+		Bridge::<T, I>::put(BridgeState {
+			is_congested: false,
+			delivery_fee_factor: FixedU128::one() + FixedU128::one(),
```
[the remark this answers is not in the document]
svyatonik: irrelevant after https://github.com/paritytech/parity-bridges-common/pull/2318#pullrequestreview-1562690669

bin/runtime-common/src/messages_xcm_extension.rs
```diff
@@ -144,155 +141,203 @@ impl SenderAndLane {
...
+	type Runtime: MessagesConfig<Self::MessagesInstance>;
+	/// Instance of the messages pallet that is used to send messages.
+	type MessagesInstance: 'static;
 	/// Returns lane used by this hauler.
 	type SenderAndLane: Get<SenderAndLane>;
 
-	/// Runtime message sender origin, which is used by [`Self::MessageSender`].
+	/// Actual XCM message sender (`HRMP` or `UMP`) to the sending chain
+	/// location (`Self::SenderAndLane::get().location`).
+	type ToSendingChainSender: SendXcm;
```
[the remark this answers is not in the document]
svyatonik: done

modules/xcm-bridge-hub-router/src/lib.rs
svyatonik: This fee factor management is a copy of the similar mechanism from DMP pallet. The pallet has no events. But I admit it may be interesting to see it in events. IDK - v1 or v2, wdyt?

modules/xcm-bridge-hub-router/src/lib.rs
svyatonik: done

modules/xcm-bridge-hub-router/src/lib.rs
svyatonik: #2323

modules/xcm-bridge-hub-router/src/benchmarking.rs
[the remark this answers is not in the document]
svyatonik: #2323

modules/xcm-bridge-hub-router/src/lib.rs
bkontur: cool, v2 sure

bin/runtime-common/src/messages_xcm_extension.rs
```diff
@@ -144,155 +141,199 @@ impl SenderAndLane {
...
+	}
+
+	/// Returns true if we have sent "congested" signal to the `sending_chain_location`.
+	fn is_congested_signal_sent(lane: LaneId) -> bool {
+		OutboundLanesCongestedSignals::<H::Runtime, H::MessagesInstance>::get(lane)
+	}
+
+	/// Send congested signal to the `sending_chain_location`.
+	fn send_congested_signal(sender_and_lane: &SenderAndLane) -> Result<(), SendError> {
+		send_xcm::<H::ToSourceChainSender>(sender_and_lane.location, H::CongestedMessage::get())?;
```
bkontur, line 322: @svyatonik 
last thin, 
I am not sure, if we should better parametrize with `BridgeId/LaneId` now instead of `Get<Xcm<()>>`,
probably for v1 is ok like this

bin/runtime-common/src/messages_xcm_extension.rs
```diff
@@ -144,155 +141,199 @@ impl SenderAndLane {
...
+		log::info!(
+			target: crate::LOG_TARGET_BRIDGE_DISPATCH,
+			"Sending 'congested' XCM message to {:?} to avoid overloading lane {:?}: there are\
+			{} messages queued at the bridge queue",
+			sender_and_lane.location,
+			sender_and_lane.lane,
+			enqueued_messages,
+		);
+
+		if let Err(e) = Self::send_congested_signal(sender_and_lane) {
```
bkontur, line 270: maybe also to trigger some event here, I know that inner `send_xcm` creates one,
but maybe could be useful for later analyse if something happens

bkontur (approved):
lgmt

bin/runtime-common/src/messages_xcm_extension.rs
svyatonik: (or even better - we'll introduce a logical HRMP subchannels and remove this message sending at all)

PR #2336, opened 2023-08-07
Grandpa: Store the authority set changes

serban300:
Related to #2296

Together with each justification we have to also store the authority set that verified it. Otherwise, if multiple justifications that enact an authority set change are accepted in the same block via `submit_finality_proof()`, we can only check for equivocations the justifications before the first authority set change.

### DIFF
--- a/primitives/chain-millau/Cargo.toml
+++ b/primitives/chain-millau/Cargo.toml
@@ -8,6 +8,9 @@ license = "GPL-3.0-or-later WITH Classpath-exception-2.0"
 
 [dependencies]
 
+# TODO: Consume `fixed-hash` from crates.io when the following fix is published:
+# https://github.com/paritytech/parity-common/commit/d3a9327124a66e52ca1114bb8640c02c18c134b8
+# Expected in a version > 0.8.0
 fixed-hash = { git = "https://github.com/paritytech/parity-common", branch = "master", default-features = false }
 hash256-std-hasher = { version = "0.15.2", default-features = false }
 impl-codec = { version = "0.6", default-features = false }
--- a/primitives/chain-millau/Cargo.toml
+++ b/primitives/chain-millau/Cargo.toml
@@ -8,7 +8,7 @@ license = "GPL-3.0-or-later WITH Classpath-exception-2.0"
 
 [dependencies]
 
-fixed-hash = { version = "0.8.0", default-features = false }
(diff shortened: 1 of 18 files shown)


serban300:
> Together with each justification we have to also store the authority set that verified it. Otherwise, if multiple justifications that enact an authority set change are accepted in the same block via `submit_finality_proof()`, we can only check for equivocations the justifications before the first authority set change.

Actually we can check all the justifications if we take the authority set from the source. But if there is an attack that also changes the authority set, we won't be able to check the rest of the justifications sent in the current block via `submit_finality_proofs`. Not sure if we would need this. Thinking.

serban300:
Thought some more about this and I think it's best to also store the authority set changes. This shouldn't take much space (it should only happen for mandatory headers) and it would enable us to track all the authority sets that were accepted, which seems safer. Otherwise in case of an attack, if the source authority set and the authority set at target start to differ we might miss some of the authority sets at the target. Also the equivocation detection code would be a bit simpler and we could check all the accepted justifications.

Marked the PR as ready for review. @paritytech/bridges-core PTAL

primitives/chain-millau/Cargo.toml
```diff
@@ -8,7 +8,7 @@ license = "GPL-3.0-or-later WITH Classpath-exception-2.0"
 
 [dependencies]
 
-fixed-hash = { version = "0.8.0", default-features = false }
+fixed-hash = { git = "https://github.com/paritytech/parity-common", branch = "master", default-features = false }
```
acatangiu, line 11: why can't we keep consuming from crates.io?

acatangiu (approved):

primitives/chain-millau/Cargo.toml
serban300: There's a clippy error that we get and that was fixed by this: https://github.com/paritytech/parity-common/commit/d3a9327124a66e52ca1114bb8640c02c18c134b8

And the change wasn't published in the latest release of `fixed-hash`. Also fixed-hash seems to be published quite rarely.

Since the problem is in Millau which is only used for testing I would consume fixed-hash from github from the moment and switch back to crates.io after the fix is published. I can add a TODO related to this.

serban300: Added a TODO related to this

Issue #2279, opened 2023-07-18
Check offchain for equivocations in accepted GRANDPA finality proofs


serban300:
Related to #39

We should have a process (maybe in the relayers) for checking if the GRANDPA finality proofs that were submitted via `submit_finality_proof` contain equivocations. 

For each accepted finality proof on the chain we should query a node for that block's finality and look for equivocations.

serban300:
There is one problem

When we get the justifications via `grandpa_subscribeJustifications`, we get:

```
Chain Polkadot => Block 16452795 justification: GrandpaJustification { round: 17735, commit: Commit { target_hash: 0x5d0484e714bdc581b1b3c3c50c49a5f66baa374340ed478d39673b108b14d0d8, target_number: 16452795, precommits: [...] }, votes_ancestries: [] }
Chain Polkadot => Block 16452796 justification: GrandpaJustification { round: 17736, commit: Commit { target_hash: 0x85005485781f8e8ad13802e9d792745db34c0a13e5f7619ddafc3bf9c7da1b07, target_number: 16452796, precommits: [...] }, votes_ancestries: [] }
Chain Polkadot => Block 16452797 justification: GrandpaJustification { round: 17738, commit: Commit { target_hash: 0xcff2990a57cccab2907b4d2033f89da7f9af692c94557e5a9624b12ca033c36b, target_number: 16452797, precommits: [...] }, votes_ancestries: [] }
Chain Polkadot => Block 16452798 justification: GrandpaJustification { round: 17739, commit: Commit { target_hash: 0xb6d2e66a6c8b19644235527c937a758362612bfc98685dbf4c617c47c4efb8bf, target_number: 16452798, precommits: [...] }, votes_ancestries: [] }
Chain Polkadot => Block 16452799 justification: GrandpaJustification { round: 17741, commit: Commit { target_hash: 0x03e35f716cf02c4f631037e6f372e037aa320c5f4b10a5c390d95cbd735f116a, target_number: 16452799, precommits: [...] }, votes_ancestries: [] }
Chain Polkadot => Block 16452800 justification: GrandpaJustification { round: 17742, commit: Commit { target_hash: 0xa70a1ce928cc5f5f8a33c2c5d1c93e84c11eb5ed4b667737f9c8f3fd40fbadaf, target_number: 16452800, precommits: [...] }, votes_ancestries: [] }
[...]

PR #2338, opened 2023-08-09
Add basic equivocation detection pipeline schema

serban300:
Related to #2279

- add basic equivocation detection pipeline schema
- add `report_equivocation` call builder


relays/lib-substrate-relay/src/finality/mod.rs
```diff
@@ -51,14 +50,7 @@ pub(crate) const RECENT_FINALITY_PROOFS_LIMIT: usize = 4096;
 
 /// Substrate -> Substrate finality proofs synchronization pipeline.
```
bkontur, line 51: @serban300 
I see you splitted `SubstrateFinalitySyncPipeline` to `SubstrateFinalityPipeline` and `SubstrateFinalitySyncPipeline`
but both have the same docs now:
```
/// Substrate -> Substrate finality proofs synchronization pipeline.
```

there is a distinction of `Sync` word, so it should probably reflect the docs somehow,

relays/lib-substrate-relay/src/finality/mod.rs
```diff
@@ -51,14 +50,7 @@ pub(crate) const RECENT_FINALITY_PROOFS_LIMIT: usize = 4096;
...
-	/// Headers of this chain are submitted to the `TargetChain`.
-	type SourceChain: Chain;
-	/// Headers of the `SourceChain` are submitted to this chain.
-	type TargetChain: ChainWithTransactions;
-
-	/// Finality engine.
-	type FinalityEngine: Engine<Self::SourceChain>;
+pub trait SubstrateFinalitySyncPipeline: SubstrateFinalityPipeline {
 	/// How submit finality proof call is built?
 	type SubmitFinalityProofCallBuilder: SubmitFinalityProofCallBuilder<Self>;
```
bkontur, line 55: why `SubmitFinalityProofCallBuilder` was not moved to `SubstrateFialityPipeline`?
I could see `SubstrateFinalityPipeline` as a abstract "holder of data" / "configuration" and `SubstrateFinalitySyncPipeline` adding some  behavior (also adds `[async_trait]`).


if we move `SubmitFinalityProofCallBuilder` to `SubstrateFinalityPipeline`,
then `SubstrateFinalitySyncPipeline` could be more like `GuardedSubstrateFinalityPipeline`

bkontur: bellow I see:

```
pub type SubstrateFinalityProof<P> = <<P as SubstrateFinalityPipeline>::FinalityEngine as Engine<
	<P as SubstrateFinalityPipeline>::SourceChain,
>>::FinalityProof;
```

so the proof is related to the `SubstrateFinalityPipeline`, so `SubmitFinalityProofCallBuilder` should be placed maybe there

relays/lib-substrate-relay/src/equivocation/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+/// Different ways of building `report_equivocation` calls.
+pub trait ReportEquivocationCallBuilder<P: SubstrateEquivocationDetectionPipeline> {
+	/// Build a `report_equivocation` call to be executed on the source chain.
+	fn build_report_equivocation_call(
+		equivocation_proof: EquivocationProofOf<P>,
+		key_owner_proof: KeyOwnerProofOf<P>,
+	) -> CallOf<P::SourceChain>;
+}
+
+/// Building the `report_equivocation` call when having direct access to the target chain runtime.
```
acatangiu, line 53: Most likely this will not live on a relay chain node, so direct access will not be available.

Instead this will have to be done through a signed extrinsic.

acatangiu (approved):
In a vacuum the changes in this PR aren't incorrect, but I'm curious to see where this is going, aka how these primitives will be actually used.

relays/lib-substrate-relay/src/equivocation/mod.rs
serban300: You're right. We'll also have to add an indirect version of this later. But we'll need to either create mocks or generate the runtimes schema from metadata. So for the moment I added only this direct version to be able to use it on millau and rialto hopefully. And then we'll see..

Polkadot-Forum:

PR #2348, opened 2023-08-15
Add equivocation detector crate and implement clients

serban300:
Working towards the equivocations detection loop

Related to #2279


primitives/header-chain/src/justification/verification/equivocation.rs
```diff
@@ -177,3 +172,39 @@ impl<'a, Header: HeaderT> JustificationVerifier<Header> for EquivocationsCollect
...
+	> for GrandpaEquivocationsFinder<C>
+{
+	type Error = Error;
+
+	fn find_equivocations(
+		verification_context: &JustificationVerificationContext,
+		synced_proof: &GrandpaJustification<HeaderOf<C>>,
+		source_proofs: &[GrandpaJustification<HeaderOf<C>>],
+	) -> Result<Vec<EquivocationProof<HashOf<C>, BlockNumberOf<C>>>, Self::Error> {
+		let source_proof_idx = match source_proofs
```
svyatonik, line 193: Since we consider theoretical `1/3 + 1` malicious validators attack (and network split), I'd avoid relying on the fact that the `source_proofs` is ordered by round number here. For the safety, let's do something like:
```rust
for source_proof in source_proofs.iter().filter(|p| p.round == synced_proof.round) {
...
}
```

primitives/header-chain/src/lib.rs
```diff
@@ -190,13 +193,48 @@ impl<Number: Codec> ConsensusLogReader for GrandpaConsensusLogReader<Number> {
...
-	/// The authority set introduced by the header.
-	pub authority_set: Option<AuthoritySet>,
+pub struct HeaderFinalityInfo<FinalityProof, FinalityVerificationContext> {
+	/// The header finality proof.
+	pub finality_proof: FinalityProof,
+	/// The new validation context introduced by the header.
+	pub new_verification_context: Option<FinalityVerificationContext>,
+}
+
+/// Grandpa-related info associated to a header. This info can be saved to storage.
```
svyatonik, line 205: I think it can only be saved to events, not to the storage :) But I wonder if we can use the single `HeaderGrandpaInfo` type for both saving and verification? Is there some significant difference? Why can't we e.g. use `AuthoritySet` in verification and convert it to `VoterSet` when required?

relays/lib-substrate-relay/src/finality_base/engine.rs
```diff
@@ -319,4 +341,82 @@ impl<C: ChainWithGrandpa> Engine<C> for Grandpa<C> {
...
+	async fn generate_source_key_ownership_proof(
+		source_client: &impl Client<C>,
+		at: C::Hash,
+		equivocation: &Self::EquivocationProof,
+	) -> Result<Self::KeyOwnerProof, SubstrateError> {
+		let set_id = equivocation.set_id();
+		let offender = equivocation.offender();
+
+		let opaque_key_owner_proof = source_client
+			.generate_grandpa_key_ownership_proof(at, set_id, offender.clone())
```
svyatonik, line 402: So far our relay has been able to work with non-archive nodes. It has never accessed ancient blocks (ancient block by default is something near `best - 200` or so). Do we plan to call this method for ancient blocks now? If so, it'll change requirement for nodes and e.g. we won't be ably to run equivocation detection using public RPC nodes.

svyatonik: I wonder how the regular equivocation reports work? I mean - on regular nodes? Do they assume that it is called at archive node?

svyatonik (commented):
So far looks good. I've left some comments. Will do a 2nd review round after we'll converge on those comments :)

relays/lib-substrate-relay/src/finality_base/engine.rs
svyatonik: > I think the grandpa client only checks for equivocations in the current round or recent rounds. So not being able to access ancient blocks shouldn't be a problem.

But are you planning to store collected justifications somewhere and check them on restart (e.g. if relayer is crashed or there were e.g. connection-with-the-target-node issue)? Then you'll probably need to access the ancient block state - to read events + to get equivocation proofs.

I don't know about historical pallet, but I doubt it'll help with reading events from ancient blocks. So probably we need to think of running this agent on nodes that are storing states for at least `<unbonding-period>` of blocks?

serban300: I'm not sure, but not in the initial equivocations detector version anyway. I would add this in a future PR after we have a basic equivocations detector. So I didn't plan that far. 

I'm not completely sure about this, because if the relayer is offline, we're also losing ephemeral finality proofs from the source as well, and even if we get the proofs that were synced at the target, we might not be able to look for equivocations. It might help in some situations. But anyway if the relayer was offline for more then 200 target blocks, we lose quite a lot of source ephemeral proofs.

> I don't know about historical pallet, but I doubt it'll help with reading events from ancient blocks. So probably we need to think of running this agent on nodes that are storing states for at least <unbonding-period> of blocks?

Yes, the historical pallet can only help us with key ownership proofs at best. If we're lucky it can generate key ownership proofs for older validator sets.

I don't know if we need to get older target blocks. If we get older target blocks, we probably won't have the ephemeral source proofs corresponding to them anyway. I was thinking at least for the initial version to have the relayer check for equivocations only in the newly generated blocks at target. If there are multiple relayers running, this should be ok. Even if at times 1 relayer fails or restarts.

> But are you planning to store collected justifications somewhere and check them on restart (e.g. if relayer is crashed or there were e.g. connection-with-the-target-node issue)?

I'm not completely sure about saving the state either, for the same reason. If the relayer is offline, we're also losing ephemeral finality proofs from the source as well, and even if we get the proofs that were synced at the target, we might not be able to look for equivocations. It might help in some situations. But anyway if the relayer was offline for more then 200 target blocks, we lose quite a lot of source ephemeral proofs.

svyatonik: It is true that ephemeral justifications will be lost if we'll lose connection to the source node. But we can collect them and use later if this connection is live but we e.g. have lost connection with the target node.

Ok, I thought that you're planning to impl that persistent storage in the initial version. Without it my comment is not actual anymore :)

svyatonik (approved):
LGTM

Polkadot-Forum:

PR #2357, opened 2023-08-16
Backport: Finality loop refactoring

serban300:
Backporting #2317 to polkadot-staging

@svyatonik sorry, as discussed on #2317 I tried to avoid this, but in the end I think it's better if we backport it. I would like to reuse the `FinalityProofsStream` and `FinalityProofsBuf` for the equivocations detector. Also there are some shared traits from the finality-base. 

Could you PTAL when you have time ? I know that you wanted to do a second review iteration if we backport it.


svyatonik (approved):
Ok, LGTM!

PR #2370, opened 2023-08-23
Try dependabot for multiple branches

svyatonik:
related to https://github.com/paritytech/parity-bridges-common/issues/2399
ref: https://github.com/dependabot/dependabot-core/issues/2159 + https://github.com/dependabot/dependabot-core/issues/2511

Not sure if that will work, so probably someone from @paritytech/ci could check. What we need is to run dependabot for two branches - `master` and `polkadot-staging`  to avoid manually cherry-picking updates. I was thinking about adding two values to `target-branch`, but https://github.com/dependabot/dependabot-core/issues/2511. So duplication seems like the only option. However I'm not sure if it works (there are some comments that say it works and other saying it does not).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -60,3 +60,6 @@ updates:
     versions:
     - ">= 0"
   rebase-strategy: disabled
+  target-branch:
+    - "master"
+    - "polkadot-staging"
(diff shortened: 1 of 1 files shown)


svyatonik:
Ok, let's try (next try will be on monday)

PR #2371, opened 2023-08-23
Manually update deps in polkadot staging

svyatonik:
I've manually cherry-picked all updated for last two weeks, but almost every pick has caused conflicts. We definitely need some better way to do that (#2280)


serban300 (approved):

acatangiu:
> I've manually cherry-picked all updated for last two weeks, but almost every pick has caused conflicts. We definitely need some better way to do that (https://github.com/paritytech/parity-bridges-common/issues/2280)

We're moving to monorepo in a couple of weeks and these kind of problems should go away.

svyatonik:
> We're moving to monorepo in a couple of weeks and these kind of problems should go away.

Are we (this repo) is moving? I thought it will be the next step (following the initial monorepo). Also in monorepo we'll probably need the similar stuff too? We'll have our own `polkadot-staging` there

acatangiu:
It's up to us to see what exactly from this repo we move there, and when. Current `bridges/` dir from cumulus will be in the monorepo since day 1.

My preference is to completely move this repo to `polkadot-sdk`, but for this we first need to migrate our nightly testnets from Rialto<>Millau to Rococo<>Wococo so we can retire Rialto and Millau and not have to carry them over.
Otherwise, I think everything in this repo can live in `polkadot-sdk`.

> We'll have our own `polkadot-staging` there

I guess we'll at most have features branches (for new features that can't be developed directly in `master`) - and follow same process as all other components.
E.g. V1 stuff would go as PRs to master with "needs-audit" label, and will be merged once audit is done.
V2 stuff would be developed in `bridges-feature-name-here` branch.

What would be different I guess is we might want some "special" CI to allow overnight runs of some PRs (bridge changes).

WDYT?

PR #2501, opened 2023-08-25
Pre-support logical HRMP channels in pallet-xcm-bridge-hub-router

svyatonik:
part of #2391 that may be implemented in our repo

This PR should add:
- [ ] the per-bridge delivery fee factor - now all bridges using the same `pallet-xcm-bridge-hub-router` instance are sharing the same fee factor. Logical HRMP channels should solve that, so it'd be better to have different fee factors for different bridges;
- [ ] outbound messages queue to the `pallet-xcm-bridge-hub-router`, where messages are saved when the bridge is suspended;
- [ ] code to serve outbound messages queue from `on_idle` when the bridge is resumed.

Opening it early for visibility + maybe someone would propose better solution for problems that I've met (will leave comments).

### DIFF
--- a/modules/xcm-bridge-hub-router/src/lib.rs
+++ b/modules/xcm-bridge-hub-router/src/lib.rs
@@ -30,7 +30,9 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
-use bp_xcm_bridge_hub::LocalXcmChannelManager;
+use bp_xcm_bridge_hub_router::{
+	bridge_id_from_locations, Bridge, BridgeId, LocalXcmChannelManager,
+};
 use codec::Encode;
 use frame_support::traits::Get;
 use sp_runtime::{FixedPointNumber, FixedU128, Saturating};
@@ -79,6 +81,10 @@ pub mod pallet {
 		/// Benchmarks results from runtime we're plugged into.
 		type WeightInfo: WeightInfo;
 
+		/// Maximal number of suspended bridges.
+		#[pallet::constant]
+		type MaxSuspendedBridges: Get<u32>;
+
 		/// Universal location of this runtime.
 		type UniversalLocation: Get<InteriorMultiLocation>;
 		/// Relative location of the sibling bridge hub.
@@ -87,16 +93,14 @@ pub mod pallet {
 		/// Also used for filtering `Bridges` by `BridgedNetworkId`.
 		/// If not specified, allows all networks pass through.
 		type BridgedNetworkId: Get<NetworkId>;
-		/// Configuration for supported **bridged networks/locations** with **bridge location** and
-		/// **possible fee**. Allows to externalize better control over allowed **bridged
-		/// networks/locations**.
-		type Bridges: ExporterFor;
 
 		/// Actual message sender (`HRMP` or `DMP`) to the sibling bridge hub location.
 		type ToBridgeHubSender: SendXcm;
 		/// Local XCM channel manager.
 		type LocalXcmChannelManager: LocalXcmChannelManager;
 
+		/// Base fee that is paid for every byte of the outbound message.
+		type BaseFee: Get<u128>;
 		/// Additional fee that is paid for every byte of the outbound message.
 		type ByteFee: Get<u128>;
 		/// Asset that is used to paid bridge fee.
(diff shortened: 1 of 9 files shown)


modules/xcm-bridge-hub-router/src/lib.rs
```diff
@@ -109,29 +113,32 @@ pub mod pallet {
...
-			log::info!(
-				target: LOG_TARGET,
-				"Bridge channel is uncongested. Decreased fee factor from {} to {}",
-				previous_factor,
-				delivery_fee_factor,
-			);
-
-			DeliveryFeeFactor::<T, I>::put(delivery_fee_factor);
-			T::WeightInfo::on_initialize_when_non_congested()
+			/*
```
svyatonik, line 116: Now we have multiple delivery fee factors - one per every bridge. So the most straightforward solution to decrease this fee factor from `on_initialize` is to loop over all bridges in every block `on_initialize` and decrease it. However it seems too dumb. Especially, given that `is_congested` call will return the same value for all bridges.

So what I'm currently thinking of is to have two fee factors. The first one will be shared by all bridges and will be decreased from `on_initialize` only when the channel with bridge hub is uncongested - let's call it `CongestionFeeFactor`. The other one will be per-bridge. We'll be decreasing it when a message is sent over the bridge. It'll be the `BridgeFeeFactor`.

So the total fee would be the `(base_fee + msg_len * byte_fee) * (CongestionFeeFactor + BridgeFeeFactor - 1)`.

modules/xcm-bridge-hub-router/src/lib.rs
```diff
@@ -87,16 +93,14 @@ pub mod pallet {
 		/// Also used for filtering `Bridges` by `BridgedNetworkId`.
 		/// If not specified, allows all networks pass through.
 		type BridgedNetworkId: Get<NetworkId>;
-		/// Configuration for supported **bridged networks/locations** with **bridge location** and
-		/// **possible fee**. Allows to externalize better control over allowed **bridged
-		/// networks/locations**.
-		type Bridges: ExporterFor;
```
svyatonik, line 93: I've removed this configuration - now the pallet has the separate `Bridges` **storage map** for that. Also - `BaseFee` is returned because of that. It solves the #2380

svyatonik: In other words, "opening the bridge" would now mean: (1) inserting an entry to the storage map (2) calling the `open_bridge` method of `pallet-xcm-bridge-hub` at sibling/child bridge hub. Right now it needs to be done using governance, but we may add some helper functions for that later

modules/xcm-bridge-hub-router/src/lib.rs
```diff
@@ -325,24 +340,51 @@ impl<T: Config<I>, I: 'static> SendXcm for Pallet<T, I> {
...
-		let xcm_hash = ViaBridgeHubExporter::<T, I>::deliver(ticket)?;
+		let (bridge_id, message_size, ticket) = ticket;
+		let bridge = Bridges::<T, I>::get(bridge_id).ok_or_else(|| SendError::Transport("UnknownBridge"))?;
+		let xcm_hash = if bridge.is_suspended() {
+			let xcm_hash = Default::default(); // TODO: what about message hash? is it important? where it is used?
+			let message_index = LatestSuspendedMessageIndex::<T, I>::mutate(|index| {
+				*index += 1;
+				*index
+			});
+			SuspendedMessages::<T, I>::insert(message_index, ticket.encode().try_into().expect("TODO"));
```
svyatonik, line 380: If the bridge is suspended (because some of its channels are congested), we have 2 options here imo:
1) save messages here and send them later, when bridge operations are resumed - that's what I'm trying to implement here;
2) (the easiest one) simply return an error. Then the sender will get a corresponding error and his transaction will fail. So tokens are not lost (given that the "delivery" is called by actual sender - if it's just a call somewhere in the middle of the path, it'll be lost)

WDYT is better?

acatangiu: hmm.. the first option makes using this pallet easier - no need to verify all paths leading to this pallet and whether messages/tokens can be lost, and the implementation here doesn't add much complexity so I say we stay with 1.

We'll ultimately see in audit 🙈 😆

modules/xcm-bridge-hub-router/src/lib.rs
```diff
@@ -325,24 +445,57 @@ impl<T: Config<I>, I: 'static> SendXcm for Pallet<T, I> {
...
-		let xcm_hash = ViaBridgeHubExporter::<T, I>::deliver(ticket)?;
+		let (bridge_id, message_size, ticket) = ticket;
+		let mut bridge =
+			Bridges::<T, I>::get(bridge_id).ok_or_else(|| SendError::Transport("UnknownBridge"))?;
+		let xcm_hash = if bridge.is_suspended() {
+			// TODO: this is the most worrysome code here - I'm not sure that it is ok to encode and
+			// save the **ticket**. Maybe we need to support just some ticket portion and then
+			// reconstruct the ticket?
+			let encoded_ticket = ticket.encode();
+			// TODO: not sure if there's some consistency for computing that?
```
svyatonik, line 485: I've seen how it is computed in other senders (`blake2_256`), but haven't seen how it is actually used. Shall we care about this value at all?

PR #2514, opened 2023-08-31
Fix benchmark with new XCM::V3 `MAX_INSTRUCTIONS_TO_DECODE`

bkontur:


bin/runtime-common/src/messages_benchmarking.rs
```diff
@@ -291,3 +281,44 @@ where
...
+		const MAX_INSTRUCTIONS_TO_DECODE: u32 = 100;
+		let xcm_number_of_instructions =
+			sp_std::cmp::min(xcm_number_of_instructions, MAX_INSTRUCTIONS_TO_DECODE as usize);
+		log::trace!(
+			target: "runtime::bridge-benchmarks",
+			"generate_xcm_builder_bridge_message_sample using xcm_number_of_instructions: {}", xcm_number_of_instructions
+		);
+
+		let xcm = xcm::VersionedXcm::<()>::V3(
+			vec![Instruction::ClearOrigin; xcm_number_of_instructions].into(),
```
svyatonik, line 315: Sorry - I'm too late to the party, but to generate more precise message, we could use the `ExpectPallet` pallet instruction instead of vec of `ClearOrigin`. Something like:
```rust
ExpectPallet {
		index: 0,
		name: vec![42; expected_message_size - <already-occupied-stuff>],
        module_name: vec![0],
       crate_major: 0,
       min_crate_minor: 0,
		module_name: Vec<u8>,
		#[codec(compact)]
		crate_major: u32,
		#[codec(compact)]
		min_crate_minor: u32,
	}
```

bkontur: no, no, you're not late, thats where I ended yesterday - to change ClearOrigin for some other instruction :D,
so thank you for `ExpectPallet` hint :)

bkontur:
@svyatonik 
there is one mention of "Rialto":
```
/// Number of extra bytes (excluding size of storage value itself) of storage proof, built at
/// Rialto chain. This mostly depends on number of entries (and their density) in the storage trie.
/// Some reserve is reserved to account future chain growth.
pub const EXTRA_STORAGE_PROOF_SIZE: u32 = 1024;
```
what would be better docs here?
or shouldnt be proof_size better chain-specific?

bkontur:
@svyatonik 
and another question,  what is the rationale behind `Weight::MAX / 1000` [here](https://github.com/paritytech/parity-bridges-common/blob/polkadot-staging/bin/runtime-common/src/messages_benchmarking.rs#L161)?

bin/runtime-common/src/messages_benchmarking.rs
```diff
@@ -84,7 +79,7 @@ fn prepare_inbound_message(
 /// function.
 pub fn prepare_message_proof_from_grandpa_chain<R, FI, B>(
 	params: MessageProofParams,
-	message_destination: InteriorMultiLocation,
+	message_generator: impl Fn(usize) -> MessagePayload,
```
bkontur, line 82: @svyatonik 
what is better here `MessagePayload` or `Vec<u8>`?

svyatonik:
> @svyatonik
> there is one mention of "Rialto":

It is chain specific - every primitives has this constant defined. Here we need it because we implement `WeightInfoExt` for `()`, which technically may be used for any (non-production) chain. So we need to have some "default" value.

We'll probably need to remove this constant at all in the future - I do not remember where it is used now tbh. Initially it was used to compute message fee (because delivery tx cost depends on storage proof size). For now - you may leave `Rialto` ot replace it with "Generic" or something like that

bin/runtime-common/src/messages_benchmarking.rs
svyatonik: `MessagePayload` is better :) It is alias `Vec<u8>` though

svyatonik (approved):
LGTM!

bin/runtime-common/src/messages_benchmarking.rs
bkontur: yes, I know it is the same, but wanted to be sure, thank you :)

svyatonik:
> @svyatonik and another question, what is the rationale behind `Weight::MAX / 1000` [here](https://github.com/paritytech/parity-bridges-common/blob/polkadot-staging/bin/runtime-common/src/messages_benchmarking.rs#L161)?

It's just some random value that will be :
1) enough to dispatch any message (`Weight::MAX / 1000` is large enough :) )
2) enough to fit (and pay for) dispatch of several messages - we are currently dispatching at most 2 messages from messages pallet benchmarks, so it has reserve for another 998

PR #2515, opened 2023-08-31
Fix benchmark with new XCM::V3 MAX_INSTRUCTIONS_TO_DECODE

bkontur:
`receive_single_message_proof_with_dispatch` benchmark has started to [fail](https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3499492) in monorepo: 
```
[2023-08-30 15:17:41] 2023-08-30 15:17:41 [XcmBlobMessageDispatch] DispatchBlob::dispatch_blob failed, error: InvalidEncoding - message_nonce: 21
```

Backport of this PR: https://github.com/paritytech/parity-bridges-common/pull/2514 (already merged in monorepo)

### DIFF
--- a/bin/runtime-common/src/messages_benchmarking.rs
+++ b/bin/runtime-common/src/messages_benchmarking.rs
@@ -42,7 +42,7 @@ use xcm::v3::prelude::*;
 /// Prepare inbound bridge message according to given message proof parameters.
 fn prepare_inbound_message(
 	params: &MessageProofParams,
-	successful_dispatch_message_generator: impl Fn(usize) -> MessagePayload,
+	successful_dispatch_message_generator: &impl Fn(usize) -> MessagePayload,
 ) -> MessagePayload {
 	let expected_size = params.proof_params.db_size.unwrap_or(0) as usize;
 
@@ -90,7 +90,7 @@ where
 			params.message_nonces.clone(),
 			params.outbound_lane_data.clone(),
 			params.proof_params,
-			|_| prepare_inbound_message(&params, message_generator),
+			|_| prepare_inbound_message(&params, &message_generator),
 			encode_all_messages,
 			encode_lane_data,
 			false,
@@ -137,7 +137,7 @@ where
 			params.message_nonces.clone(),
 			params.outbound_lane_data.clone(),
 			params.proof_params,
-			|_| prepare_inbound_message(&params, message_generator),
+			|_| prepare_inbound_message(&params, &message_generator),
 			encode_all_messages,
 			encode_lane_data,
 			false,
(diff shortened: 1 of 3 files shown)


bin/runtime-common/src/messages_benchmarking.rs
```diff
@@ -58,6 +58,11 @@ fn prepare_inbound_message(
 
 	// we don't need to be super-precise with `expected_size` here
 	let xcm_size = expected_size.saturating_sub(location_encoded_size);
+
+	// XCM V3 introduced limit for max instructions
+	const MAX_INSTRUCTIONS_TO_DECODE: u32 = 100;
+	let xcm_size = sp_std::cmp::min(xcm_size, MAX_INSTRUCTIONS_TO_DECODE as usize);
```
serban300, line 64: Looks like this might not respect the `expected_size` anymore. Isn't this an issue ?

bkontur: hmm, I see now, I think you are right.
I saw `size: StorageProofSize::Minimal(i),` ->  `i` as `1` at first :D , but now I see that it is generated by:
```
let i in EXPECTED_DEFAULT_MESSAGE_LENGTH .. EXPECTED_DEFAULT_MESSAGE_LENGTH * 16;
```

We calculate [`xcm_size`](https://github.com/paritytech/parity-bridges-common/blob/master/bin/runtime-common/src/messages_benchmarking.rs#L59-L61) and use it as number of `ClearOrigin` instruction,
so at least this calculation should do something like this:

```
	// we don't need to be super-precise with `expected_size` here
	let xcm_size = expected_size.saturating_sub(location_encoded_size);

        let xcm_number_of_instructions = xcm_size / Instruction::<()>::ClearOrigin.encoded_size();

       let xcm = xcm::VersionedXcm::<()>::V3(vec![Instruction::ClearOrigin; xcm_number_of_instructions].into()); 
```

But we still need to limit `xcm_number_of_instructions` to the [`MAX_INSTRUCTIONS_TO_DECODE`](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/src/v3/mod.rs#L72-L80).

Ok, I am going to try a different fix and little refactor, because [`prepare_inbox_message`](https://github.com/paritytech/parity-bridges-common/blob/master/bin/runtime-common/src/messages_benchmarking.rs#L63-L67) counts with `BridgeMessage` like encoded data, so if anybody else is going to have its own implementation it wont work.

Maybe we dont need `is_successful_dispatch_expected` and could be replaced by some callback which generates "random data/instructions".

bkontur: ok, I guess no need to refactor heavily because anybody could implement their own `prepare_message_proof_from_parachain` alternative (probably it will require to change `pub(crate)` to `pub` for `insert_header_to_parachains_pallet` )

bkontur: @serban300 
please, lets switch review to polkadot-staging (less noise) here: https://github.com/paritytech/parity-bridges-common/pull/2514

svyatonik (approved):

PR #2525, opened 2023-09-04
`master` branch: Use polkadot-sdk dependencies

serban300:
Porting #2524 to `master` + some extra fixes

### DIFF
--- a/bin/rialto/node/Cargo.toml
+++ b/bin/rialto/node/Cargo.toml
@@ -1,3 +1,11 @@
+[[bin]]
+name = "rialto-bridge-node-execute-worker"
+path = "src/workers/execute-worker.rs"
+
+[[bin]]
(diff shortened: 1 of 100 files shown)


bin/rialto/node/Cargo.toml
```diff
@@ -1,3 +1,11 @@
+[[bin]]
```
svyatonik, line 1: It doesn't break our test deployments, right? I mean - do we nee those binaries for relay/parachain to work? Do we still able to build that good old `rialto-bridge-node`, which has many references in our code?

svyatonik (approved):
:+1:

bin/rialto/node/Cargo.toml
serban300: Yes, it works. Tested locally.

We need these binaries. Otherwise the parachain won't produce blocks,

PR #2530, opened 2023-09-05
Try to fix docker build at CI

svyatonik:
Looks like `rialto-bridge-node-execute-worker` and `rialto-bridge-node-prepare-worker` weren't included into latest [`rialto-bridge-node` docker image](https://hub.docker.com/layers/paritytech/rialto-bridge-node/latest/images/sha256-876ec00666c5c7b0fb305f2d75f59d9b4967a7eb932b9dd16b31c0da7874c133?context=explore). Let's try this fix

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -240,6 +240,10 @@ build:
     - mkdir -p ./artifacts
     - strip ./target/release/rialto-bridge-node
     - mv -v ./target/release/rialto-bridge-node ./artifacts/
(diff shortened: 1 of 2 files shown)


serban300 (approved):
Sorry, I missed this. Thank you for fixing it !

svyatonik:
> Sorry, I missed this. Thank you for fixing it !

That is fine! I'm also not sure that this fix is the latest - but let's try :)

PR #2542, opened 2023-09-08
Add polkadot bulletin chain primitives

svyatonik:
This is primitives for the [Polkadot Bulletin Chain](https://github.com/zdave-parity/polkadot-bulletin-chain). It is going to be bridged with Polkadot BH. Some bridge details for bridge guys:
- Polkadot Bulletin Chain tokens have no economic value, so we are going to have whitelisted set of relayers and check it in [signed extension](https://github.com/zdave-parity/polkadot-bulletin-chain/blob/main/runtime/src/lib.rs#L415);
- at Bridge Hub, we'll just enable fee refund for useful transactions. We don't want any rewards - will probably run internal (Parity?) relayer for that;
- there'll be a separate Polkadot system parachain that will send messages to Bulletin Chain over bridge hub. So regular `ExportMessage` support. This also needs to be unpaid (?) at Polkadot.BH;
- some messages will also be headed from Bulletin Chain to this new system parachain. We also need to make sure that everything is free here. We need to trust the Bulletin Chain validators then (otherwise it could then spam us with messages and halt the BH);
- I have to add [a separate branch](https://github.com/paritytech/parity-bridges-common/tree/polkadot-v.1.0.0-audited) for this bridge, because Bulletin Chain codebase is using the `v1.0.0` branches and we need audited code there.

@zdave-parity @NachoPal In bridges we need to define chain primitives, because other (bridged) runtime need to know some facts about the chain it bridges with. In our case, Polkadot.BH will need to know primitives of Bulletin Chain and vice versa. This is also required for our offchain actors (relayers) - they should be able to craft valid transactions and make proper RPC calls. After we'll update subtree with this commit (ofc I'll need to backport it to `polkadot-v.1.0.0-audited` branch first), I'm gonna use typedefs from this crate in Bulletin Chain runtime to avoid code duplication.

### DIFF
--- a/primitives/chain-polkadot-bulletin/src/lib.rs
+++ b/primitives/chain-polkadot-bulletin/src/lib.rs
@@ -1,0 +1,110 @@
+// Copyright (C) Parity Technologies (UK) Ltd.
+// This file is part of Parity Bridges Common.
+
+// Parity Bridges Common is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Parity Bridges Common is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Bridges Common.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Polkadot Bulletin Chain primitives.
+
+#![warn(missing_docs)]
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use bp_header_chain::ChainWithGrandpa;
+use bp_runtime::{
+	decl_bridge_finality_runtime_apis, decl_bridge_messages_runtime_apis, Chain, ChainId,
+};
+use frame_support::{
+	dispatch::DispatchClass,
+	parameter_types,
+	weights::{constants::WEIGHT_REF_TIME_PER_SECOND, Weight},
+};
+use frame_system::limits;
+use sp_runtime::{Perbill, StateVersion};
+
+// This chain reuses most of Polkadot primitives.
+pub use bp_polkadot_core::{
+	AccountId, Balance, BlockNumber, Hash, Hasher, Header, Nonce, Signature,
+	AVERAGE_HEADER_SIZE_IN_JUSTIFICATION, MAX_HEADER_SIZE,
+	REASONABLE_HEADERS_IN_JUSTIFICATON_ANCESTRY,
+};
+
+/// Maximal number of GRANDPA authorities at Polkadot Bulletin chain.
+pub const MAX_AUTHORITIES_COUNT: u32 = 100;
+
+/// Name of the With-Polkadot Bulletin chain GRANDPA pallet instance that is deployed at bridged
+/// chains.
+pub const WITH_POLKADOT_BULLETIN_GRANDPA_PALLET_NAME: &str = "BridgePolkadotBulletinGrandpa";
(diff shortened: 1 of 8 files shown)


primitives/chain-polkadot-bulletin/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+impl Chain for PolkadotBulletin {
+	const ID: ChainId = *b"pdbc";
+
+	type BlockNumber = BlockNumber;
+	type Hash = Hash;
+	type Hasher = Hasher;
+	type Header = Header;
+
+	type AccountId = AccountId;
+	type Balance = Balance;
```
zdave-parity, line 93: The Bulletin Chain does not have balances at all. It's not that the tokens are worthless, they really just don't exist. I don't know if the bridges stuff supports not having a balance type, but if it doesn't it's at least worth a comment here explaining why we have to provide one and why that isn't a problem (assuming it isn't).

primitives/chain-polkadot-bulletin/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+};
+use frame_support::{
+	dispatch::DispatchClass,
+	parameter_types,
+	weights::{constants::WEIGHT_REF_TIME_PER_SECOND, Weight},
+};
+use frame_system::limits;
+use sp_runtime::{Perbill, StateVersion};
+
+// This chain reuses most of Polkadot primitives.
```
zdave-parity, line 35: Not sure it's a good idea to tie these things to Polkadot? If they are changed in Polkadot they're not going to automatically change in the Bulletin Chain...

zdave-parity: Should just take the definitions from `runtime/src/lib.rs` in the Bulletin Chain repo?

primitives/chain-polkadot-bulletin/src/lib.rs
svyatonik: That's a fair point. I don't think we will be using this type anywhere for something critical. I'll add a comment in next commits

primitives/chain-polkadot-bulletin/src/lib.rs
svyatonik: > Not sure it's a good idea to tie these things to Polkadot? If they are changed in Polkadot they're not going to automatically change in the Bulletin Chain...

True, but if types will change in Polkadot, they won't automatically change in bridges `bp-polkadot-core` crate as well :) Because polkadot uses its own crates and our `bp-polkadot-core` is just a minimal knowledge of Polkadot-like chains that we need to implement bridges. We need to track the possible type changes and act accordingly.

In bridges there are two "base" chain types that we are able to handle currently - standalone (Polkadot-like) chains and parachains (Cumulus-based mostly). To me the Bulletin Chain looks more Polkadot-like, hence the usage of the `bp_polkadot_core` crate. I could just copy paste those types and constants here, but it'll just increase amount of code and won't change anything in practice :) But feel free to insist - it isn't a big deal

> Should just take the definitions from runtime/src/lib.rs in the Bulletin Chain repo?

There are two huge issues with that:
- we are in a different repo and if we'll start referring bulleting chain crates here, it'll cause circular references;
- this crate will be used by polkadot bridge hub runtime and we don't want to embed (at least build) another runtime when building Polkadot BH.

Actually, with your chain it is easier and I'd like to "use typedefs from this crate in Bulletin Chain runtime to avoid code duplication" as the PR description states later, if you'll accept such PR. I.e. instead of having `type BlockNumber = u32;` in bulletin chain runtime, I'd like to have it here and then simply reuse in runtime. WDYT?

zdave-parity: I meant cut the definitions out of the Bulletin Chain's `runtime/src/lib.rs` and paste them here. They're just typedefs of Substrate types AFAIK, shouldn't be any dependency issues.

I'm happy to have the Bulletin Chain then get the definitions from here to avoid duplication; we're going to be depending on this stuff anyway.

Issue #2547, opened 2023-09-08
Tracking issue for Polkadot Bulletin Chain <> Polkadot Bridge Hub bridge


svyatonik:
All development should happen in master and then backported to the [polkadot-v.1.0.0-audited](https://github.com/paritytech/parity-bridges-common/tree/polkadot-v.1.0.0-audited) branch (and `polkadot-staging` eventually).

Tasks that would allow us to test it locally:
- [x] add Bulletin Chain primitives (#2542);
- [x] add required pallets to Polkadot Bridge Hub (on top of https://github.com/paritytech/polkadot-sdk/pull/1352) (https://github.com/paritytech/polkadot-sdk/tree/sv-bulletin-chain-bridge):
   - [x] make sure that we only register transaction fee refunds and are not paying any rewards to relayers;
- [x] add required pallets to Bulletin Chain (https://github.com/svyatonik/polkadot-bulletin-chain/tree/add-bridge-pallets):
   - [x] add XCM support to Bulletin Chain (hopefully someone from XCM team could help with that);
   - [x] ~decide whether we want to use utility pallet there~;
   - [x] make sure that we only accept bridge transactions from whitelisted relayers;
   - [x] do not register any rewards at Bulletin Chain, because tokens have no economic value there;
- [x] add bridge support to `substrate-relay`;
- [x] prepare something similar to Branislav setup (zombienet + binaries + document) to launch Polkadot + Polkadot Bridge Hub + Bulletin Chain (and another system parachain later) locally.
- [x] send some test message directly from Polkadot BH to Bulletin Chain (unless other Polkadot parachain is ready).

Follow up tasks: make actual PRs to Polkadot.BH once it is deployed.

A list of live unresolved problems:
- [x] we need to know real dispatch weight of the message. Bulletin chain uses immediate dispatch, so we need to decode the message (and calls) and use XCM weigher for estimating dispatch weight for message delivery transaction (https://github.com/zdave-parity/polkadot-bulletin-chain/pull/33).
- [x] rename Kawabunga to People Chain (see https://github.com/paritytech/polkadot-sdk/pull/2281)
- [x] add Rococo Bridge Hub <> Rococo Bulletin bridge (Rococo Bulletin chain will be deployed and will be 1:1 copy of Polkadot Bulletin)

svyatonik:
Rococo <> RococoBulletin bridge may be tested locally (until centralized deployment) with:
[...]

PR #2552, opened 2023-09-11
Polkadot Bulletin Chain client

svyatonik:
related to #2547


primitives/chain-polkadot-bulletin/src/lib.rs
```diff
@@ -63,6 +72,94 @@ pub const MAX_UNREWARDED_RELAYERS_IN_CONFIRMATION_TX: MessageNonce = 1024;
...
+		CheckEra<Hash>,
+		CheckNonce<Nonce>,
+		CheckWeight,
+	),
+	ValidateSigned,
+)>;
+
+/// Signed extension, used by Polkadot Bulletin.
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone, TypeInfo)]
+pub struct SignedExtension(SignedExtensionSchema);
```
svyatonik, line 94: Had to add this custom struct, because this chain has no transaction payment signed extension.

primitives/chain-polkadot-bulletin/src/lib.rs
```diff
@@ -63,6 +72,94 @@ pub const MAX_UNREWARDED_RELAYERS_IN_CONFIRMATION_TX: MessageNonce = 1024;
...
+
+impl sp_runtime::traits::SignedExtension for SignedExtension {
```
serban300, line 96: Maybe we could have another trait something like `SignedExtensionSchemaProvider {SignedExtensionSchema = ...}` and implement `sp_runtime::traits::SignedExtension` automatically for the types that implement `SignedExtensionSchemaProvider`.

serban300 (commented):
Looks good ! Left just one comment.

primitives/chain-polkadot-bulletin/src/lib.rs
svyatonik: That's a good idea, however it won't allow us to have a blanket implementation (I guess that is what you mean by "implement sp_runtime::traits::SignedExtension automatically", right?) of `sp_runtime::traits::SignedExtension`, because `sp_runtime::traits::SignedExtension` lives in the external crate (Substrate). And if we'll create some adapter for that, we won't be able to add custom methods  which we need here (e.g. `from_params` and so on). I know - that looks ugly, but probably in the future we'll switch to something better (auto detecting well known signed extensions and explicitly asking about unknown on relayer start? something that we'll need to think of in #2491)

serban300: Oh yes, you're right. I forgot about that.

serban300 (approved):

PR #2555, opened 2023-09-12
Equivocation detection loop: Reorganize block checking logic as state machine

serban300:
This way we'll be able to save the state in case of a failure. Actually saving the state is not implemented yet. But this PR is a step towards better error handling.


svyatonik (approved):

Polkadot-Forum:

PR #2558, opened 2023-09-12
prepare refund extension infra to add refund extension for messages from standalone chain

svyatonik:
#2547 

In with-Bulletin chain bridge, we need extension to refund messages transactions, coming from remote standalone (GRANDPA) chain. Our current extension only supports refund from remote parachain. This PR doesn't add this option, but prepares infrastructure for adding that - I'll prepare next PR with new extension tomorrow.

This PR does not change any logic - it's just a cosmetics to be able to reuse the same code for another extension


serban300 (approved):
Looks good !

PR #2568, opened 2023-09-15
Add unit tests for the equivocation detection loop

serban300:
Related to https://github.com/paritytech/parity-bridges-common/issues/2377

### DIFF
--- a/relays/equivocation/src/reporter.rs
+++ b/relays/equivocation/src/reporter.rs
@@ -25,11 +25,11 @@ use std::{
 	task::{Context, Poll},
 };
 
-pub struct EquivocationsReporter<P: EquivocationDetectionPipeline, SC: SourceClient<P>> {
-	pending_reports: Vec<TrackedTransactionFuture<SC::TransactionTracker>>,
+pub struct EquivocationsReporter<'a, P: EquivocationDetectionPipeline, SC: SourceClient<P>> {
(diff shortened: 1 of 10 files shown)


relays/equivocation/src/reporter.rs
```diff
@@ -81,3 +81,49 @@ impl<P: EquivocationDetectionPipeline, SC: SourceClient<P>> EquivocationsReporte
...
+						TrackedTransactionStatus::Finalized(HeaderId(1, 1))
+					}),
+					Box::pin(async {
+						polled_reports.lock().unwrap().push(2);
+						finished_reports.lock().unwrap().push(2);
+						TrackedTransactionStatus::Finalized(HeaderId(2, 2))
+					}),
+					Box::pin(async {
+						polled_reports.lock().unwrap().push(3);
+						async_std::task::sleep(Duration::from_secs(100)).await;
```
svyatonik, line 112: Could we maybe use [`std::future::pending()`](https://doc.rust-lang.org/std/future/fn.pending.html) here to avoid potential spurious failures?

svyatonik (approved):
Looks good, thank you!

relays/equivocation/src/reporter.rs
serban300: Good idea ! Done.

PR #2570, opened 2023-09-18
Support message relay limits override

svyatonik:
To estimate maximal number of messages and their total cumulative dispatch weight in messages delivery transaction we currently [use the transaction payment runtime APIs](https://github.com/paritytech/parity-bridges-common/blob/master/relays/lib-substrate-relay/src/messages/mod.rs#L460-L528). However [on some chains](https://github.com/zdave-parity/polkadot-bulletin-chain) there's no such pallet and it doesn't make any sense to add it there just for that. So let's add an option to configure bridge limits from hardcoded constants in the code. It maybe worth to add CLI options for that, but our CLI interface is already too complicated, so let's start with hardcoded constants and see if we need CLI options for that in the future.

### DIFF
--- a/relays/bin-substrate/src/cli/init_bridge.rs
+++ b/relays/bin-substrate/src/cli/init_bridge.rs
@@ -244,9 +244,14 @@ impl BridgeInitializer for PolkadotToPolkadotBulletinCliBridge {
 	fn encode_init_bridge(
 		init_data: <Self::Engine as Engine<Self::Source>>::InitializationData,
 	) -> <Self::Target as Chain>::Call {
-		relay_polkadot_bulletin_client::RuntimeCall::BridgePolkadotGrandpa(
-			relay_polkadot_bulletin_client::BridgePolkadotGrandpaCall::initialize { init_data },
-		)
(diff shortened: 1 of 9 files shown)


relays/bin-substrate/src/cli/init_bridge.rs
```diff
@@ -244,9 +244,14 @@ impl BridgeInitializer for PolkadotToPolkadotBulletinCliBridge {
 	fn encode_init_bridge(
 		init_data: <Self::Engine as Engine<Self::Source>>::InitializationData,
 	) -> <Self::Target as Chain>::Call {
-		relay_polkadot_bulletin_client::RuntimeCall::BridgePolkadotGrandpa(
-			relay_polkadot_bulletin_client::BridgePolkadotGrandpaCall::initialize { init_data },
-		)
+		type RuntimeCall = relay_polkadot_bulletin_client::RuntimeCall;
```
svyatonik, line 247: And yet another change here: bulletin chain has the sudo pallet - let's use it to initialize bridge.

serban300 (approved):

Issue #2486, opened 2021-06-02
[V2 Bridges] Coordinated relayers protocol for cheaper relaying


svyatonik:
related: https://github.com/paritytech/parity-bridges-common/issues/316#issuecomment-679015022 , paritytech/parity-bridges-common#377, paritytech/parity-bridges-common#439, paritytech/parity-bridges-common#954, paritytech/parity-bridges-common#978

So now if there are multiple concurrent relayers (we're mostly talking about message relayers) running, one/some would lose their funds because of overlapping transactions. I'm opening this issue to track all related discussions/questions and not to lose any information about that.

Specifically - I'm now working on non-altuistic relayers and if we'll be considering sessions+elections as suggested [here](https://github.com/paritytech/parity-bridges-common/issues/316#issuecomment-679015022), then we shall not treat **not delivering** message `M` as misbehavior if delivering message is non-profitable (e.g. because of conversion rate change). This would require some non-profitability proof (exchange rate) or some other strategy. 

I'm also thinking of single-tx sessions - when relayers are announcing their intention (backed by their funds) to submit **single** delivery transaction. And either best offer wins (e.g. offer that promises to deliver most messages in the shortest number of blocks), or, if there are several equivalent intentions, only one is picked.

svyatonik:
Some thoughts: https://gist.github.com/svyatonik/398bfd306541b693fe5afe37dc68f908. I'll start implementing something like that. Will need to do some refactorings before

svyatonik:
> Some thoughts: https://gist.github.com/svyatonik/398bfd306541b693fe5afe37dc68f908. I'll start implementing something like that. Will need to do some refactorings before

One important thing that I've missed in proposed design is that we may need to implement both stages simultaneously before v2. Otherwise if there are many opened lanes, we may face the issue when batch transactions (`[submit_finality_proof, submit_parachain_heads, receive_messages_proof]`) fail because GRANDPA and/or parachain finality has been already updated by another relayer.

Ideally (imo), we'll move finality relay to collator. Other options is to allow 1 free bridged finality submission for every `N` blocks (`N` may even be `1` here) + do not use refund mechanism for batch transactions.

Issue #2586, opened 2023-09-28
Concurrent relayers protocol: finality submissions


svyatonik:
https://github.com/paritytech/parity-bridges-common/issues/2486#issuecomment-1738662055

Let's extract this part from #2486, because it could/need to be implemented in parallel with #2486.

So the problem is that if we'll introduce relayers coordination protocol (e.g. using priorities) BUT will keep our refund-batch-transactions approach (where we refund relayer, which has submitted valid finality proofs + messages transactions), we'll face the issue where bridge throughput is dropped to `1` lane per BH block. Because relayers may (and will - both honest and malicious) submit batch transactions AND of those transactions only one will be mined - all others will be dropped, because they contain obsolete finality proofs.

The solution that I've initially proposed in #2486 is to move finality relay functionality to collators - they'll submit finality proofs using BH inherents. This looks ideal at first glance, but it may take a while + I feel that there are some undiscovered caveats. So we need a backup plan that we may implement even before adding finality inherents (read: in v2).

So this issue is about designing this backup solution + implementing it. The most straightforward way imo is to remove refund support for batch transactions AND allow `1` free finality submission (be it GRANDPA, parachain or BEEFY) for every `N` BH blocks (where `N` may be one). So it is kinda inherent, but anyone could submit it. We need to think of resolving concurrency issue here too, though - because relayers will be submitting the same transactions every `N`th block, which is not good. Maybe we shall select one relayer from registered set and force him to submit finality proof? Need some more thinking

svyatonik:
> The most straightforward way imo is to remove refund support for batch transactions AND allow `1` free finality submission (be it GRANDPA, parachain or BEEFY) for every `N` BH blocks (where `N` may be one). So it is kinda inherent, but anyone could submit it. We need to think of resolving concurrency issue here too, though - because relayers will be submitting the same transactions every `N`th block, which is not good. Maybe we shall select one relayer from registered set and force him to submit finality proof? Need some more thinking

This is exactly what have been implemented in https://github.com/paritytech/parity-bridges-common/issues/2871 - we now have a standalone relayers with free submissions of every `Nth` header. So idea is that anyone may submit (relay/parachain) finality for free with some period and for messages we'll need relayers coordination protocol.

PR #2597, opened 2023-10-02
Additional priority boost for relayers that have explicitly registered at lane

svyatonik:
related #2486

All that this PR does (apart from some small refactoring) is adding the `compute_per_lane_priority_boost` function. It adds priority boost to message delivery transactions of relayer at the head of the lane relayers queue. This queue is always empty now, so no logic changes yet. But the `compute_per_lane_priority_boost` most likely won't change at all or will be changed slightly.

The reason I'm opening PR in this state is that I want some feedback on this priority-based approach. While I'm pretty sure it is the best possible option for us, it still has some drawbacks. The main one is that it isn't super scalable - i.e. if we'll have other bridges on the bridge hub (read: with-Ethereum, with-Bulletin, ...), their transactions priority boost will be zero, which could cause some issues. On the other hand, if bridge hub will be extensively used (e.g. many lanes), we will need to solve the same problem even if we won't have those priority boosts.

Detailed info may be found: https://github.com/paritytech/parity-bridges-common/issues/2486 and expecially https://gist.github.com/svyatonik/398bfd306541b693fe5afe37dc68f908

### DIFF
--- a/modules/relayers/src/extension/priority.rs
+++ b/modules/relayers/src/extension/priority.rs
@@ -22,23 +22,81 @@
 //! single message with nonce `N`, then the transaction with nonces `N..=N+100` will
 //! be rejected. This can lower bridge throughput down to one message per block.
 
-use bp_messages::MessageNonce;
+use crate::{Config as RelayersConfig, Pallet as RelayersPallet};
+
+use bp_messages::{LaneId, MessageNonce};
 use frame_support::traits::Get;
-use sp_runtime::transaction_validity::TransactionPriority;
+use frame_system::{pallet_prelude::BlockNumberFor, Pallet as SystemPallet};
+use sp_runtime::{
+	traits::{One, Zero},
+	transaction_validity::TransactionPriority,
+};
 
 // reexport everything from `integrity_tests` module
 pub use integrity_tests::*;
 
-/// Compute priority boost for message delivery transaction that delivers
-/// given number of messages.
-pub fn compute_priority_boost<PriorityBoostPerMessage>(
+/// Compute total priority boost for message delivery transaction.
+pub fn compute_priority_boost<R: RelayersConfig>(
+	lane_id: LaneId,
 	messages: MessageNonce,
+	relayer: &R::AccountId,
 ) -> TransactionPriority
 where
-	PriorityBoostPerMessage: Get<TransactionPriority>,
+	usize: TryFrom<BlockNumberFor<R>>,
 {
+	compute_per_message_priority_boost::<R::PriorityBoostPerMessage>(messages)
+		.saturating_add(compute_per_lane_priority_boost::<R>(lane_id, relayer))
+}
+
+/// Compute priority boost for message delivery transaction, that depends on number
+/// of bundled messages.
+fn compute_per_message_priority_boost<PriorityBoostPerMessage: Get<TransactionPriority>>(
+	messages: MessageNonce,
+) -> TransactionPriority {
 	// we don't want any boost for transaction with single message => minus one
 	PriorityBoostPerMessage::get().saturating_mul(messages.saturating_sub(1))
+}
+
+/// Compute priority boost for message delivery transaction, that depends on
+/// the set of lane relayers and current slot.
+fn compute_per_lane_priority_boost<R: RelayersConfig>(
+	lane_id: LaneId,
+	relayer: &R::AccountId,
+) -> TransactionPriority
+where
+	usize: TryFrom<BlockNumberFor<R>>,
+{
+	// if there are no relayers, explicitly registered at this lane, noone gets additional
+	// priority boost
+	let lane_relayers = RelayersPallet::<R>::lane_relayers(&lane_id);
+	let lane_relayers_len: BlockNumberFor<R> = (lane_relayers.len() as u32).into();
+	if lane_relayers_len.is_zero() {
+		return 0
+	}
+
+	// we can't deal with slots shorter than 1 block
+	let slot_length: BlockNumberFor<R> = R::SlotLength::get().into();
+	if slot_length < One::one() {
+		return 0
+	}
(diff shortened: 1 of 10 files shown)


serban300:
Is this a solution for https://github.com/paritytech/parity-bridges-common/issues/2486#issuecomment-1738662055 ? An alternative to moving the finality relaying to the collator ? If so could we do that instead ? Seems more scalable and also seems better for the equivocation detection part.

svyatonik:
> Is this a solution for [#2486 (comment)](https://github.com/paritytech/parity-bridges-common/issues/2486#issuecomment-1738662055) ? An alternative to moving the finality relaying to the collator ? If so could we do that instead ? Seems more scalable and also seems better for the equivocation detection part.

No, this is not about the comment you've pointed. This is first step towards implementing #2486, of which I've separated everything related to finality to #2586. So it is a way to have a protocol, where relayer has some guarantees that if it'll be running relayer node, he could get some reward. Right now we have a chaos - everyone could submit a message delivery transaction, but who will win - the lucky one whose transaction will get to validators first AND/OR have a larger tip. By playing with priorities, we give some guarantee to honest relayer that if he'll be submitting valid and "best possible" transactions, he'll win during his slot.

For finality there's a separate #2586 and for all this thing there's that https://gist.github.com/svyatonik/398bfd306541b693fe5afe37dc68f908

modules/relayers/src/extension/priority.rs
```diff
@@ -22,31 +22,89 @@
...
+/// Compute priority boost for message delivery transaction, that depends on number
+/// of bundled messages.
+fn compute_per_message_priority_boost<PriorityBoostPerMessage: Get<TransactionPriority>>(
+	messages: MessageNonce,
+) -> TransactionPriority {
 	// we don't want any boost for transaction with single message => minus one
 	PriorityBoostPerMessage::get().saturating_mul(messages.saturating_sub(1))
 }
 
+/// Compute priority boost for message delivery transaction, that depends on
```
svyatonik, line 60: Another important (and fragile) thing here is that if validator own collator(s), he may simply ignore priorities and include affilated relayer transactions into blocks that he authors. So honest relayer may not receive rewards even if he is submitting proper transactions in assigned slots. Normally we assume that the collators set is honest and decentralized (so it is solved elsewhere), but here we build the important subsystem (relayer rewards) on top of that. 

Options I see:
- ignore that fact. If we admit the possibility of centralized collators set, then we also admit the possibility of censoring particular transactions - e.g. such collator set may simply halt one of bridges (lanes) by ignoring delivery transactions on those lanes. So still we need to solve it elsewhere;
- have some kind of misbehavior reports. But proving that the honest relayer has submitted valid transaction, that was not "mined" is impossible. All other solutions (e.g. only accepting delivery transactions from relayer that has submitted finality proof for next `N` blocks) still allow affilated collators to censorship such transactions and get rewarded;
- have a stricter laws. I.e. instead of bumping priority for relayer at the head of the queue, we simply reject all other relayer transactions during his slot. This won't protect us from censorship anyway.

So I suggest us to go with the option (1) - I don't think that we can (and should) solve it here. So just wanted to share and warn that this ^^^ is possible.

modules/relayers/src/extension/priority.rs
```diff
@@ -22,31 +22,89 @@
...
+	lane_id: LaneId,
+	relayer: &R::AccountId,
+) -> TransactionPriority
+where
+	usize: TryFrom<BlockNumberFor<R>>,
+{
+	// if there are no relayers, explicitly registered at this lane, noone gets additional
+	// priority boost
+	let lane_relayers = RelayersPallet::<R>::lane_relayers(lane_id);
+	let lane_relayers_len: BlockNumberFor<R> = (lane_relayers.len() as u32).into();
```
serban300, line 72: Nit: I think we can avoid this conversion. We can simply do `let lane_relayers_len = lane_relayers.len();` here and then change the following lines to something like

```
	let slot = match usize::try_from(current_block_number / slot_length) {
		Ok(slot) => slot,
		Err(_) => return 0,
	};
	// and then get the relayer for that slot
	let slot_relayer = &lane_relayers[slot % lane_relayers_len];
```

serban300 (approved):
> By playing with priorities, we give some guarantee to honest relayer that if he'll be submitting valid and "best possible" transactions, he'll win during his slot.

Thanks for the explanation ! Makes sense now.

modules/relayers/src/extension/priority.rs
serban300: I am not very familiar with collators. How would this work exactly ? How many collators would the relayer need to own in order to bypass our priorities mechanism ? Is there any way for the validators to realize that these collators are assigning wrong priorities to the bridge-related transactions and avoid these collators ? I'm wondering if it wouldn't be more advantageous for a relayer to run a set of modified collators instead of registering as a relayer, because:
- registering as a relayer requires staking, which is a disadvantage
- running a set of collators could lead to more mined transactions maybe ? So more rewards

modules/relayers/src/extension/priority.rs
svyatonik: I'd rather not to convert from block (or `slot`) to `usize`, because block number could be represented with e.g. `u128` and then the conversion will fail. For `usize` OTOH we know that the `usize` is `u32` in WASM environment (even if will ever be changed to `u64`, we deal with the in-storage and in-memory vec here, which shall never be that large) and since block number is at least `u32`, conversions from `usize` to `block`/`slot` should be fine

serban300: Oh, right, `lane_relayers` shouldn't be very large. Sounds good then.

svyatonik:
I'm going to merge it, because soon I'll start breaking #2605 into smaller PRs. Thanks for review @serban300 !

PR #2600, opened 2023-10-04
Reject transactions if bridge pallets are halted

svyatonik:
Halting one of bridges pallets is an extraordinary action that could only be performed by the root (governance) on live chains. We may remove this functionality (along with pallet owners) when we drop Rialto<>Millau in favor of local R<>W/P<>K bridge. 

However, we still have the code in v1 and it is a good idea to take care of careless registered relayers, who keep running the relayer while pallets are halted. Now they can be slashed if they submit transaction to the halted pallet. After this PR, our extensions will reject such transactions and they won't be "mined" => relayers won't be slashed.

### DIFF
--- a/bin/runtime-common/src/messages_call_ext.rs
+++ b/bin/runtime-common/src/messages_call_ext.rs
@@ -18,6 +18,7 @@ use crate::messages::{
 	source::FromBridgedChainMessagesDeliveryProof, target::FromBridgedChainMessagesProof,
 };
 use bp_messages::{target_chain::MessageDispatch, InboundLaneData, LaneId, MessageNonce};
+use bp_runtime::OwnedBridgeModule;
 use frame_support::{
 	dispatch::CallableCallFor,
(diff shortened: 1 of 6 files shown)


acatangiu (approved):

bin/runtime-common/src/messages_call_ext.rs
```diff
@@ -291,7 +302,7 @@ impl<
 				return sp_runtime::transaction_validity::InvalidTransaction::Stale.into()
 			},
 			Some(CallInfo::ReceiveMessagesDeliveryProof(proof_info))
-				if proof_info.is_obsolete() =>
+				if is_pallet_halted || proof_info.is_obsolete() =>
```
serban300, line 305: Is this needed ? If `is_pallet_halted == true`  and `self.call_info() == Some(_)` won't the first branch of this match be executed (`Some(proof_info) if is_pallet_halted => {...}`) ?

PR #2605, opened 2023-10-06
Finishing "Additional priority boost for relayers that have explicitly registered at lane"

svyatonik:
related to #2486
follow up for #2597 (hence on top of it)

Opening it to give additional context when reviewing #2597. While there's still a lot of remaining work + a ton of TODOs, it show additional pallet calls + other changes, required for #2597. A brief overview:
- new `LaneRelayersSet` structure. It contains the current active set of lane relayers + next set + a block number, when the `next_set` may replace the `active_set`;
- relayers in the `LaneRelayersSet` are ordered by the reward that they are willing to receive for delivering a single message. Relayers, which are ready to "work" for smaller reward will push our relayers, wanting larger reward. In theory it should eventually affect the message fee that is paid at the sending chain (AH) if we'll implement reporting (or sending chains will receive this information in some other way). For now it should make the bridge cheaper for mantainers;
- relayer registration is considered active even if current stake is lower than the required (i.e. if `RequiredStake` was increased after relayer has registered itself);
- relayer registration is considered IF relayer has registered itself at least at one lane. So even initially his registration should have finished at block `1_000`, registration may be active (so he can't claim his funds back) even at block `2_000` if he's registered at some lane;
- added draft prototypes for `register_at_lane` and `deregister_at_lane` calls that the relayer will be able to call;
- added `enact_next_relayers_set_at_lane` call - it shall be called explicitly by one of relayers from the `next_set` for free (or for regular fee for other submitters) to activate the `next_set`. Why separate call? Because otherwise, we need to dance with hooks (`on_initialize`, `on_idle`, ...) and this is not well-scalable.

There's a chance that I'll split this PR into several smaller PRs in the future, if it'll grow to unreviewable state.

### DIFF
--- a/modules/relayers/src/lib.rs
+++ b/modules/relayers/src/lib.rs
@@ -406,16 +406,18 @@ pub mod pallet {
 					Error::<T>::TooEarlyToActivateNextRelayersSet,
 				);
 
-				let next_next_set_may_enact_at = current_block_number.saturating_add(4); // TODO
+				let new_next_set_may_enact_at = current_block_number.saturating_add(4u32.into()); // TODO
 				lane_relayers.activate_next_set(
-					next_next_set_may_enact_at,
-					|relayer| Self::is_registration_active(relayer.relayer())
+					new_next_set_may_enact_at,
+					|relayer| Self::is_registration_active(relayer)
 				);
 
 				*lane_relayers_ref = Some(lane_relayers);
 
 				Ok::<_, Error<T>>(())
 			})?;
+
+			Ok(())
 		}
 	}
 
@@ -668,6 +670,10 @@ pub mod pallet {
 		TooLargeRewardToOccupyAnEntry,
 		///
 		NotRegisteredAtLane,
+		///
+		NoRelayersAtLane,
+		///
+		TooEarlyToActivateNextRelayersSet,
 	}
 
 	/// Map of the relayer => accumulated reward.
--- a/modules/relayers/src/lib.rs
+++ b/modules/relayers/src/lib.rs
@@ -27,7 +27,7 @@
 //
 // Or we may allow to buy lane registration using relayer rewards only - i.e. has has delivered 100 messages
 // and reward is 100 DOTs. He could reserve his 100 DOTs to buy lane registration slots. Every slot costs 1 DOT.
-// So after 100 slots, the registration becomes inactive. And he must renew it. 
+// So after 100 slots, the registration becomes inactive. And he must renew it.
 
 // TODO: additionally we could add a reward market - i.e. now we have boosts:
 // `messages_count * per_message + per_lane`. We could add another boost if relayer wants to receive lower reward.
@@ -41,7 +41,7 @@
 
 use bp_messages::LaneId;
 use bp_relayers::{
-	PaymentProcedure, Registration, RelayerRewardsKeyProvider, RewardsAccountParams, StakeAndSlash,
+	LaneRelayersSet, PaymentProcedure, Registration, RelayerRewardsKeyProvider, RewardsAccountParams, StakeAndSlash,
 };
 use bp_runtime::StorageDoubleMapKeyProvider;
 use codec::{Decode, Encode, MaxEncodedLen};
@@ -49,7 +49,7 @@ use frame_support::{DefaultNoBound, fail};
 use frame_system::Pallet as SystemPallet;
 use sp_arithmetic::traits::{AtLeast32BitUnsigned, Zero};
 use scale_info::TypeInfo;
-use sp_runtime::{traits::CheckedSub, BoundedVec, RuntimeDebug, Saturating};
+use sp_runtime::{traits::{CheckedSub, One}, BoundedVec, RuntimeDebug, Saturating};
 use sp_std::{collections::vec_deque::VecDeque, marker::PhantomData};
 
 pub use pallet::*;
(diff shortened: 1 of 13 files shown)


modules/relayers/src/payment_adapter.rs
```diff
@@ -61,47 +59,106 @@ where
...
-		// `u32::MAX`
-		let relayer_reward = T::Reward::saturated_from(messages).saturating_mul(delivery_fee);
-
-		if relayer != *confirmation_relayer {
-			Pallet::<T>::register_relayer_reward(lane_id, &relayer, relayer_reward);
-		} else {
-			confirmation_relayer_reward =
-				confirmation_relayer_reward.saturating_add(relayer_reward);
-		}
+		// TODO: THIS IS WRONG - `LaneRelayers` is for relayers, which are delivering messages from BRIDGED to THIS
```
svyatonik, line 137: I've missed that - the extension that boost priorities works at target chain. And reward is currently paid (`register_relayers_rewards` is called) at the source chain. So we do not know the reward that the relayer works for here (at the source chain). We will probably need to change the unrewarded relayers set to also store accumulated reward OR something else. Need to think about it a bit

svyatonik: Ok, so it has three solutions imo:
1) to pay reward at the target chain (target BH). I'm not currently considering it, but it can be implemented later by slightly changing the solution (3);
2) to include proof of `LaneRelayers` into confirmation transaction. This has some obvious flaws - e.g. when the set changes, while there are unconfirmed messages, we may need to provide proof of old set and of the new set. Also we'll need to split delivered messages into two (or even more groups) - one, delivered by the old set and one, delivered by the new set. Also, it'll make confirmation transaction heavier, which is not good;
3) let's then just compute total accumulated relayer reward into `InboundLaneData::relayers` vector. So we'll "fix" the reward during delivery transaction and then will simply report it back to the source chain, where it'll actually be paid. We could later implement (1) or implenent it partially by registering lower reward (or zero if reward is paid fully at the target chain). To be super-careful, we may limit the maximal paid reward at the source chain by `DeliveredMessages * MaxRewardPerMessage`. The `MaxRewardPerMessage` shall be represented on both sides of the bridge, though.

That said, I'm gonna implement this in this PR, but then will extract it into a separate PR (after finishing implementation). This PR is already non-trivial.

svyatonik: Changed in https://github.com/paritytech/parity-bridges-common/pull/2605/commits/5df0872b9ab3c9a02276fa6987be3fe0c195be8d. Need more tests + self review

svyatonik: Two other relevant commits: https://github.com/paritytech/parity-bridges-common/pull/2605/commits/1a3efe764bf8f62c6912252bf16402d063849595 + https://github.com/paritytech/parity-bridges-common/pull/2605/commits/9c134da66aef091f2c8f5ed0081e3eb8e9466844 . I had to fix `RewardAtSource` type to be `u64`. We can't make it generic, because relayers pallet works with all chains, which may use different types for rewards (`MultiAssets`, `u128`, ...). At the same time, we need to cmp them in the pallet. So it is fixed to `u64`, but the other (chain) source may map it to its tokens using any kind of conversions.

modules/relayers/src/lib.rs
```diff
@@ -241,6 +269,196 @@ pub mod pallet {
...
+					};
+
+					// cannot add another lane registration if relayer has already max allowed
+					// lane registrations
+					ensure!(
+						registration.register_at_lane(lane),
+						Error::<T>::FailedToRegisterAtLane
+					);
+
+					// TODO: ideally we shall use the candle auction for relayers selection (similar
```
svyatonik, line 310: Those two TODOs is the biggest unresolved issue remaining in this PR. Candle auction requires additional runtime storage, so probably not the best solution here if we want the solution to be scalable. Thinking on how to alleviate that.

svyatonik: TLDR: the problem is that our `next_set` has limited size (right now it is the same size as the `active_set`) and unhonest relayers may fill this set by claiming to work for the smallest possible reward (say, zero). Then, right before the bidding ends, they could deregister and epoch will start with an empty set, which leads us to the chaos we have now. Also they can hold their place for relayers, that are working for the largest possible reward. So even during "auction" there are relayers that are ready to work for 1DOT/message, they may be pushed out by spam relayers, working for zero and then, right before next epoch starts, they'll deregister and will be replaced by relayers working for 10DOT/message. So - last minute changes is a problem.

I have three solutions:
1) if `deregister_at_lane` is called by relayer, which has not relayed at least one message since he was included in the `next_set`, he is slashed with his lane stake. The only exception is when he is pushed from the relayers set by relayer with smaller `expected_reward`. This way spam relayers, who may fill the whole relayers set with reward = zero won't be able to deregister before next epoch start without losing their stake;
2) increase the size of the `next_set`. It would make it harder (more expensive) for spam relayers to fill the whole next set;
3) an alternate implementation of candle auction, based on relay chain randomness. Something like - during `on_initialize` we check if `random() < (epoch_end - current_block) / epoch_length` and then stop auction. So it likely ends near the epoch end, but noone knows the actual block. Relayers will likely to make actual stakes since beginning to actually stay in the final `next_set`. Relayers will know the stop in advance (once the relay block is ready), but since we'll do our computations from `on_initialize`, they won't be able to alter the set using transactions.

I think we need to implement all three options, but for this PR I'd start with the option (2) as the easiest one. Next one in the queue (in a follow up PR) is the (1) and the last is (3).

svyatonik: Option (2) is implemented in https://github.com/paritytech/parity-bridges-common/pull/2605/commits/081523c762c6ff613fe454c81976892f56e07c78

svyatonik:
It is ready for review, but I'd like to extract couple of other changes into separate PRs to make it easier for review => still draft

svyatonik:
Marking this PR ready for review - it is still large, but all remaining changes are about managing those new `ActiveLaneRelayersSet` and `NextLaneRelayersSet` sets, so imo it makes sense to review it altogether.

svyatonik:
I completely forgot about this gist I've made some time ago: https://gist.github.com/svyatonik/398bfd306541b693fe5afe37dc68f908

So #2597 + this PR + TODOs described in #2618 is the implementation of "Stage 1" from that gist, in particular this chapter: https://gist.github.com/svyatonik/398bfd306541b693fe5afe37dc68f908#relayers-queue-and-priority-boosts. I could provide more details if required, but feel free to review @paritytech/bridges-core . Thank you!

PR #2614, opened 2023-10-11
Millau, Rialto: accept equivocation reports

serban300:
Related to https://github.com/paritytech/parity-bridges-common/issues/2610

Add logic for accepting equivocation reports in Millau and Rialto runtimes.

Working towards adding some integration tests for the equivocation detection loop. Not sure how to do it yet, but I guess we will need Millau and Rialto to accept equivocation reports.


svyatonik (approved):
I'm not against adding that to Rialto<>Millau, but you shall be mentally ready that your work will be dropped in the near future (once we'll migrate to local P/K or R/W bridge) :) But now we do not have it, so Rialto<>Millau is fine for testing.

Re how to do it - I think the easiest option is to add subcommand to `substrate-relay` that would generate justification for some dummy block and submit it to the target chain (with some large tip). Authorities for Rialto/Milllau and local R/W/P/K are well known (`//Alice`, `//Bob`, ...), so iiuc it shouldn't be a problem. We may run it once per day or so.

serban300:
> but you shall be mentally ready that your work will be dropped in the near future (once we'll migrate to local P/K or R/W bridge)

Yes, I'm expecting this. This will be just something temporary in order to enable us to test the equivocation detection loop until we migrate to P/K or R/W .

> Re how to do it - I think the easiest option is to add subcommand to substrate-relay that would generate justification for some dummy block and submit it to the target chain (with some large tip). Authorities for Rialto/Milllau and local R/W/P/K are well known (//Alice, //Bob, ...), so iiuc it shouldn't be a problem. We may run it once per day or so.

Thanks for the suggestion ! Will try it.

PR #2633, opened 2023-10-18
Backport from `polkadot-sdk` with actual master

bkontur:


bin/rialto/node/Cargo.toml
```diff
@@ -47,7 +47,7 @@ polkadot-node-core-pvf-execute-worker = { git = "https://github.com/paritytech/p
 polkadot-node-core-pvf-prepare-worker = { git = "https://github.com/paritytech/polkadot-sdk", branch = "master" }
 polkadot-primitives = { git = "https://github.com/paritytech/polkadot-sdk", branch = "master" }
 polkadot-runtime-parachains = { git = "https://github.com/paritytech/polkadot-sdk", branch = "master" }
-polkadot-service = { git = "https://github.com/paritytech/polkadot-sdk", branch = "master", default-features = false, features = [ "full-node", "polkadot-native" ] }
+polkadot-service = { git = "https://github.com/paritytech/polkadot-sdk", branch = "master", default-features = false, features = [ "full-node", "westend-native" ] }
```
bkontur, line 50: @svyatonik `polkadot-runtime` is removed from `polkadot-sdk`, changed it to `westend-runtime` and lets see if it works. Do we need to change anything else polkadot->westend related for rialto?

acatangiu (approved):

bin/rialto/node/Cargo.toml
svyatonik: I don't think we need any runtime for that - the reason it is here (iirc) is because `polkadot-service` required at least one of `*-runtime` features to compile. If it does compile without it, we could remove this feature at all

svyatonik (approved):

bin/rialto/node/Cargo.toml
bkontur: ok, missing runtime feaure is not a problem, we need to fix probably feature flags propagation:
```
~/parity/parity-bridges-common$ zepter lint propagate-feature --feature runtime-benchmarks --left-side-feature-missing=ignore --workspace --feature-enables-dep="runtime-benchmarks:frame-benchmarking" --locked
crate 'millau-runtime'
  feature 'runtime-benchmarks'
    must propagate to:
      pallet-bridge-grandpa
crate 'rialto-parachain-runtime'
  feature 'runtime-benchmarks'
    must propagate to:
      bridge-runtime-common
      pallet-bridge-grandpa
      pallet-bridge-messages
      pallet-bridge-relayers
crate 'rialto-runtime'
  feature 'runtime-benchmarks'
    must propagate to:
      pallet-bridge-grandpa
      pallet-bridge-relayers
Found 7 issues (run with --fix to fix).
```

continue tomorrow, if this will work, maybe zepter will work in our CI also

PR #2634, opened 2023-10-18
Grandpa justifications: Avoid duplicate vote ancestries

serban300:
Fixes https://github.com/paritytech/parity-bridges-common/issues/2397

### DIFF
--- a/primitives/header-chain/src/justification/verification/mod.rs
+++ b/primitives/header-chain/src/justification/verification/mod.rs
@@ -21,6 +21,7 @@ pub mod optimizer;
 pub mod strict;
 
 use crate::{justification::GrandpaJustification, AuthoritySet};
+use std::collections::btree_map::Entry::{Occupied, Vacant};
 
 use bp_runtime::HeaderId;
 use finality_grandpa::voter_set::VoterSet;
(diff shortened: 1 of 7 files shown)


primitives/header-chain/src/justification/verification/mod.rs
```diff
@@ -44,23 +45,40 @@ pub struct AncestryChain<Header: HeaderT> {
...
 		let mut parents = BTreeMap::new();
 		let mut unvisited = BTreeSet::new();
-		for ancestor in &justification.votes_ancestries {
+		let mut ignored_idxs = Vec::new();
+		for (idx, ancestor) in justification.votes_ancestries.iter().enumerate() {
 			let hash = ancestor.hash();
-			let parent_hash = *ancestor.parent_hash();
-			parents.insert(hash, parent_hash);
-			unvisited.insert(hash);
+			match parents.entry(hash) {
```
svyatonik, line 66: maybe
```rust
if parents.insert(hash).is_none() {
					entry.insert(*ancestor.parent_hash());
					unvisited.insert(hash);
} else {
ignored_idxs.push(idx);
}
```

svyatonik (approved):
Thank you!

primitives/header-chain/src/justification/verification/mod.rs
svyatonik: No, no need to change :) It's just 6 LOC vs 9, but current version is ok too :)

serban300: I mean the `match parents.entry(hash) ...` seemed a bit better to me, because `parents.insert()` does almost the same inside:
```
    pub fn insert(&mut self, key: K, value: V) -> Option<V>
    where
        K: Ord,
    {
        match self.entry(key) {
            Occupied(mut entry) => Some(entry.insert(value)),
            Vacant(entry) => {
                entry.insert(value);
                None
            }
        }
    }
```

And apart from this we would have to do another if-else. This way we do everything at once. But I agree it doesn't make a big difference.

PR #2642, opened 2023-10-20
Fix master build

svyatonik:
https://gitlab.parity.io/parity/mirrors/parity-bridges-common/-/jobs/4024579

Just a [partial copy](https://github.com/paritytech/parity-bridges-common/blob/polkadot-staging/primitives/xcm-bridge-hub-router/src/lib.rs) of required files from `polkadot-staging`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1136,6 +1136,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "bp-asset-hub-rococo"
+version = "0.1.0"
+dependencies = [
+ "bp-xcm-bridge-hub-router 0.1.0",
+ "frame-support",
+ "parity-scale-codec",
+ "scale-info",
+]
+
+[[package]]
+name = "bp-asset-hub-wococo"
+version = "0.1.0"
+dependencies = [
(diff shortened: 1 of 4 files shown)


bkontur (approved):
hmm, just wondering how could have [this](https://github.com/paritytech/parity-bridges-common/pull/2638) passed without this yesterday?

svyatonik:
I think because asset hub primitives are missing from our top-level `Cargo.toml`. Do you think we need to add it there, @bkontur ?

svyatonik:
Actually, let's do it :)

bkontur:
> hmm, just wondering how could have [this](https://github.com/paritytech/parity-bridges-common/pull/2638) passed without this yesterday?

I see, nighlty pipeline for chain primitives, which we dont reference here in bridges repo,
also we forgot to add them to the main Cargo.toml

bkontur:
yes, lets better add them (also to polkadot-staging)

svyatonik:
> yes, lets better add them (also to polkadot-staging)

:+1: done

PR #2648, opened 2023-10-27
Deposit event when relayer reward is registered

svyatonik:
Think it is a good idea - e.g. when debugging we can't see account in parachain logs.


acatangiu (approved):

bkontur:
@svyatonik can we add this to the `polkadot-staging` also please?

PR #2649, opened 2023-10-27
Add basic zombienet test to be used in the future

svyatonik:
This PR adds basic zombienet test that can only be used **manually** now. This basic test starts local Rococo, Wococo and BH+AH parachains (using existing [Branislav scripts](https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/scripts/bridges_rococo_wococo.sh)), starts relay + makes asset transfer in both directions. It checks that the bridge is initialized, then that the assets are exchanged and then that relayers are rewarded for messages delivery. Later we may add more test conditions, but for beginning it is enough. In other words, this script repeats all required steps from the [bridge hubs readme](https://github.com/paritytech/polkadot-sdk/tree/master/cumulus/parachains/runtimes/bridge-hubs).

A few details on how it is made:
- bridge tests need two relay chains to be started. Unfortunately, [it is not supported by the zombienet yet](https://github.com/paritytech/zombienet/pull/792), so we need our custom test starter (which is the ./zombienet/run-tests.sh);
- our external process (relay) needs connection to nodes from both relay chains (and their parachains). So IIUC we can't use anything other than the native zombienet provider (i.e. no `k8s`);
- we need to wait for both networks to finish their tests before shutting them down. Otherwise one chain may e.g. get message and then immediately shut down, while other still has not received this message. Hence this hack with `sync-exit.sh`;
- a couple of things I've learned about zombienet:
  - custom js script that is invoked for tests is required to return proper value if we want to test it using "return is at least 1". So we can't just return **current best finalized block of the chain** - we need to wait until it satisfies condition and then return. So every js script has internal loop;
  - testing returned value outside of this script makes no sense because of previous point - if our script returns, we already know that it has returned correct value. Also I had some problems using comparators - I guess because sometime string was compared to number and so on. So the only condition that the test is successful now is that it has completed within given time;
- because we do not have `substrate-relay` binary in the `polkadot-sdk`, it'd be hard to run tests within docker container there. So this is only first step towards having something on CI. I assume that we'll start with manual testing, then after migrating code to `polkadot-sdk`, we could replace local script run with local container run (that would depend on prebuilt binaries), then after we'll start building (and publishing) `substrate-relay` docker there - we could do the same autiomatically on CI. And someday maybe we'll have multichain support in zombienet - then we could remove our dedicated test runner and use `zombienet test` instead.


bkontur (approved):
wow, very good :)
maybe, some small README.md how to run it or install any prerequisites?

bkontur:
> > wow, very good :) maybe, some small README.md how to run it or install any prerequisites?
> 
> Thanks to you for those scripts :) I've added README.md - ptal

very nice, so for rococo<>westend we will just need to add two `zndsl` files :)
I would go with this to the `polkadot-staging` also and check `verify-pallets-build.sh`.
It would be very cool, if we would be able to run this from bridges subtree in `polkadot-sdk`.

svyatonik:
> > > wow, very good :) maybe, some small README.md how to run it or install any prerequisites?
> > 
> > 
> > Thanks to you for those scripts :) I've added README.md - ptal
> 
> very nice, so for rococo<>westend we will just need to add two `zndsl` files :) I would go with this to the `polkadot-staging` also and check `verify-pallets-build.sh`. It would be very cool, if we would be able to run this from bridges subtree in `polkadot-sdk`.

Yep, the idea was to move it to the `polkadot-sdk` later. I'll prepare a PR for `polkadot-staging` first

PR #2660, opened 2023-10-30
Add basic zombienet test to be used in the future (polkadot-staging)

svyatonik:
* add basic zombienet test to be used in the future

* removed unneeded variables

* add README.md for zombienet folder

PR #2663, opened 2023-11-01
Drop Rialto <> Millau bridges

svyatonik:
We already have some troubles with maintaining those bridges (https://github.com/paritytech/parity-bridges-common/pull/2647#pullrequestreview-1703849299).

Since we already have a live setup (Rococo <> Wococo), which will be superseded with [Rococo <> Westend](https://github.com/paritytech/parity-bridges-common/pull/2647) bridge soon AND at least we have [some basic zombienet tests](https://github.com/paritytech/parity-bridges-common/pull/2660), I think the time has come to drop Rialto and Millau (and test deployments VM).

Couple of things to mention:
- also removed several unused subcommands of `substrate-relay` binary;
- also removed several scripts that will be unneeded after we'll move to `polkadot-sdk`;
- also removed fuzzer that has never been used;
- left the only docker image that should be built from this repo - the one with `substrate-relay` binary. We'll need to have a similar build step in `polkadot-sdk`, but until that, we'll be building it from here.


bkontur (approved):
@svyatonik Slava, you should write some condolence speech for Millau and Rialto :(,
Maybe I asked that before, but who and why did choose those names?

Will do the same for `polkadot-staging?

svyatonik:
> @svyatonik Slava, you should write some condolence speech for Millau and Rialto :(, Maybe I asked that before, but who and why did choose those names?
> 
> Will do the same for `polkadot-staging?

Let's just move forward with that. I think @tomusdrw was behind that :) Just some bridges (https://en.wikipedia.org/wiki/Rialto_Bridge + https://en.wikipedia.org/wiki/Millau_Viaduct)

sergejparity (approved):

acatangiu (approved):
40K lines fewer to maintain ❤️

tomusdrw:
> > @svyatonik Slava, you should write some condolence speech for Millau and Rialto :(, Maybe I asked that before, but who and why did choose those names?
> > Will do the same for `polkadot-staging?
> 
> Let's just move forward with that. I think @tomusdrw was behind that :) Just some bridges (https://en.wikipedia.org/wiki/Rialto_Bridge + https://en.wikipedia.org/wiki/Millau_Viaduct)

Indeed :)

PR #2664, opened 2023-11-01
Adjusted constant because for measuring we used mistakenly rococo constants

bkontur:


svyatonik (approved):

primitives/chain-asset-hub-westend/src/lib.rs
```diff
@@ -45,7 +45,7 @@ frame_support::parameter_types! {
 
 	/// Base delivery fee to `BridgeHubWestend`.
 	/// (initially was calculated by test `BridgeHubWestend::can_calculate_weight_for_paid_export_message_with_reserve_transfer` + `33%`)
-	pub const BridgeHubWestendBaseFeeInWocs: u128 = 1624803349;
+	pub const BridgeHubWestendBaseFeeInWocs: u128 = 487441010000;
```
svyatonik, line 48: Maybe worth renaming `InWocs` too :) But that isn't required

bkontur: oh, yeah, thanks, renamed

Issue #2662, opened 2023-10-30
Start version guard for the equivocation detection loop


serban300:
Related to this comment: https://github.com/paritytech/parity-bridges-common/pull/2647#discussion_r1376248266

bkontur:
if we start this loop, do we need any accounts on relay chains for submitting extrinsics?
can we use the bridge-hub-accounts? probably we would need to drip those accounts on relay chain, right?
if so, then we need to update deployment: https://github.com/paritytech/devops/issues/2922

serban300:
I don't think we need extra accounts for this. I think we'll use the same ones that we use for sending finality proofs

svyatonik:
> I don't think we need extra accounts for this. I think we'll use the same ones that we use for sending finality proofs

But we need to submit equivocations to relay chains (Rococo and Westend), where we don't have any accounts and never submit any bridge transactions? So I think we need some accounts there.

Also I think we need a separate processes/containers for that? IIRC we are not starting ED detection in a "normal" relay process (`headers-and-messages`). That's because we've been planning to have a single `headers-and-messages` container, but multiple ED "guards". For Rococo <> Westend they may be started in a single container, though (imo). But for P<>K we'll probably need more containers. I maybe missed something - sorry @serban300 if I'm wrong here, just double checking

serban300:
> But we need to submit equivocations to relay chains (Rococo and Westend), where we don't have any accounts and never submit any bridge transactions? So I think we need some accounts there.

Yes, actually you're right. We need relay chain accounts for this.

> Also I think we need a separate processes/containers for that? IIRC we are not starting ED detection in a "normal" relay process (`headers-and-messages`).

Yes, we don't start the ED loop in the `headers-and-messages` relayer. We need separate processes.

serban300:
Fixed by https://github.com/paritytech/parity-bridges-common/pull/2693 and https://github.com/paritytech/parity-bridges-common/pull/2678

PR #2678, opened 2023-11-07
[polkadot-staging] Start version guards for the ED loop

serban300:
Fixes https://github.com/paritytech/parity-bridges-common/issues/2662


svyatonik (approved):
LGTM

PR #2679, opened 2023-11-08
Fixing gitspiegel trigger workflow

mutantcornholio:
The first attept to use a workflow to protect GitLab CI from untrusted contributors failed, because GitHub doesn't pass secrets to workflows for PRs that originate from forks. 
 
This uses a different approach: instead of triggerring gitspiegel API directly from the workflow, we're just spawning an empty workflow with a specific path, and gitspiegel listens for `workflow_run` event to start mirroring.  

The idea is the same: for the first-time contributors, running workflows would require manual aciton and that would block mirroring. But this time, we don't need any secrets to make it work.

### DIFF
--- a/.github/workflows/gitspiegel-trigger.yml
+++ b/.github/workflows/gitspiegel-trigger.yml
@@ -1,5 +1,10 @@
 name: gitspiegel sync
 
+# This workflow doesn't do anything, it's only use is to trigger "workflow_run"
+# webhook, that'll be consumed by gitspiegel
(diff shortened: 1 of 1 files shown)


alvicsam (approved):

bkontur:
@mutantcornholio we use `polkadot-staging` branch for syncing to `polkadot-sdk` repo, but I see that e.g. this [PR](https://github.com/paritytech/parity-bridges-common/pull/2692) does not run pipelines, could this giespiegel be responsible for that?

PR #2695, opened 2023-11-14
Gitspiegel polkadot staging

svyatonik:
@mutantcornholio Could you, please, confirm that those two commits will fix our problem. And the problem is that when we open PRs to `polkadot-staging` branch, no checks (apart from CLA) are running on that PR. For us it is important to run the same set of checks on PRs to both `master` and `polkadot-staging` branches.

Example PRs:
- to `polkadot-staging`: https://github.com/paritytech/parity-bridges-common/pull/2692 (no any checks)
- to `main`: https://github.com/paritytech/parity-bridges-common/pull/2605

### DIFF
--- a/.github/workflows/gitspiegel-trigger.yml
+++ b/.github/workflows/gitspiegel-trigger.yml
@@ -1,0 +1,20 @@
+name: gitspiegel sync
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - unlocked
+      - ready_for_review
+      - reopened
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


bkontur (approved):

sergejparity (approved):

svyatonik:
> @mutantcornholio Could you, please, confirm that those two commits will fix our problem. And the problem is that when we open PRs to `polkadot-staging` branch, no checks (apart from CLA) are running on that PR. For us it is important to run the same set of checks on PRs to both `master` and `polkadot-staging` branches.

I think checks from above answers that. So nvm, thank you :)

PR #2705, opened 2023-11-15
fix lane id and bridged chain id

svyatonik:


bkontur:
haha, this is last thing which I am testing, added one more commit

bkontur (approved):

bkontur:
what about?
`wait -n -p COPROC_PID` -> `wait -n -p $COPROC_PID`

PR #2711, opened 2023-11-16
Add withdraw reserve assets to zombienet tests

### LINKED ISSUE #2671: Add withdraw reserve script calls to zombienet tests
see [a4089c8de3d5d36a7f0c7c8e3f9dc0a8ec779eb7](https://github.com/paritytech/polkadot-sdk/commit/a4089c8de3d5d36a7f0c7c8e3f9dc0a8ec779eb7)

svyatonik:
closes #2671

### DIFF
--- a/zombienet/tests/0001-asset-transfer-works-westend-to-rococo.zndsl
+++ b/zombienet/tests/0001-asset-transfer-works-westend-to-rococo.zndsl
@@ -22,5 +22,12 @@ asset-hub-rococo-collator1: js-script ../helpers/wrapped-assets-balance.js with 
 bridge-hub-rococo-collator1: js-script ../helpers/relayer-rewards.js with "5FLSigC9HGRKVhB9FiEo4Y3koPsNmBmLJbpXg2mp1hXcS59Y,0x00000002,0x62687764,BridgedChain,0" within 300 seconds
 bridge-hub-rococo-collator1: js-script ../helpers/relayer-rewards.js with "5FLSigC9HGRKVhB9FiEo4Y3koPsNmBmLJbpXg2mp1hXcS59Y,0x00000002,0x62687764,ThisChain,0" within 300 seconds
 
+# step 7: send wROC back to Alice at Rococo AH
+asset-hub-rococo-collator1: run ../scripts/invoke-script.sh with "withdraw-reserve-assets-from-asset-hub-rococo-local" within 60 seconds
+
+# step 8: elsewhere Rococo has sent wWND to //Alice - let's wait for it
+# (we wait until //Alice account increases here - there are no other transactionc that may increase it)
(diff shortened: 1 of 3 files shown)


zombienet/tests/0001-asset-transfer-works-westend-to-rococo.zndsl
```diff
@@ -22,5 +22,12 @@ asset-hub-rococo-collator1: js-script ../helpers/wrapped-assets-balance.js with
 bridge-hub-rococo-collator1: js-script ../helpers/relayer-rewards.js with "5FLSigC9HGRKVhB9FiEo4Y3koPsNmBmLJbpXg2mp1hXcS59Y,0x00000002,0x62687764,BridgedChain,0" within 300 seconds
 bridge-hub-rococo-collator1: js-script ../helpers/relayer-rewards.js with "5FLSigC9HGRKVhB9FiEo4Y3koPsNmBmLJbpXg2mp1hXcS59Y,0x00000002,0x62687764,ThisChain,0" within 300 seconds
 
+# step 7: send wWND back to Alice at Westend AH
```
bkontur, line 25: @svyatonik what about adding here some check that user already has som wWND?

this file is 0001-asset-transfer-works-**westend-to-rococo**.zndsl but we start on AssetHubRococo :)

bkontur (approved):
haha, I have also TODO for this: `- fix test case and add withdraw to the Slava's tests`, but I see there was an issue :)

zombienet/tests/0001-asset-transfer-works-westend-to-rococo.zndsl
svyatonik: > @svyatonik what about adding here some check that user already has som wWND?

It is tested above:
```
# step 5: elsewhere Westend has sent WOC to //Alice - let's wait for it
asset-hub-rococo-collator1: js-script ../helpers/wrapped-assets-balance.js with "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY,0,Westend" within 600 seconds
```

> this file is 0001-asset-transfer-works-westend-to-rococo.zndsl but we start on AssetHubRococo :)

Yeah, I know - it causes confusion :) I could move to sibling file, but since transaction is made at Rococo AH, I thought it is better to keep it here. So all scripts/tests that are checking Rococo parachains are happening here and Westend are happening in sibling file. Open to suggestions though

svyatonik: I've updated comments here: https://github.com/paritytech/parity-bridges-common/pull/2711/commits/e31ae6b7e6f54f331f22625bc4eb491212823c0c

Yes, technically "step 4" belongs to other tests file, but let's keep all Rococo transaction submitter script calls in 0001-asset-transfer-works-westend-to-rococo.zndsl and Westend in 0001-asset-transfer-works-rococo-to-westend.zndsl. When Zombienet will support multiple relay chains in single network/test file, it should be clearer

PR #2724, opened 2023-11-27
Added Rococo BH <> Rococo Bulletin bridge

svyatonik:
ref #2547

The idea is to have some test environment before adding this stuff to Polkadot Bridge Hub. People chain (which will be bridged to Bulletin Chain) is being added to Rococo/Westend here: https://github.com/paritytech/polkadot-sdk/pull/2281. @joepetrowski said that we'll be able to deploy another "copy" of bulletin chain to bridge with Rococo, so let's add support for Rococo <> Rococo Bulletin chain bridge then.

TODOs left here:
- [ ] final decision on whether we need another two crates - `bp_rococo_bulletin_chain` and `relay-rococo-bulletin-client`. Since we'll be deploying the same runtime, let's probably try to reuse Polkadot Bulletin crates? But it causes confuction - i.e. in Rococo runtime I had to use `BridgePolkadotBulletinGrandpa` pallet name instead of `BridgeRococoBulletinGrandpa`, so I'm slowly tending to adding other two crates :/
- [ ] test everything locally (at least that we are able to send dummy messages from Rococo BH to Rococo Bulletin) before merging.


bkontur (approved):
I think, so far, so good

acatangiu (approved):

PR #361, opened 2020-09-20
feat(bridge): [ISSUE-333] Bridge: Integrate cargo-spellcheck

TomaszWaszczyk:
I test how to contribute to the repository (that is why `Draft pull request`), when the PR will be ready to merge I will mark it as vanilla plain PR :D

PR #2727, opened 2023-12-05
Improved `ExportXcm::validate` implementation for BridgeHubs - step 1

bkontur:
## Motivation

At first, this PR addresses https://github.com/paritytech/parity-bridges-common/issues/2452. Currently, on BridgeHubs V1, we don't check the sender to verify if it is allowed to use a dedicated lane. As a result, anyone opening an HRMP channel to the local BridgeHub will be able to send messages to the dedicated lane for AssetHubs.

Secondly, there is an issue raised by Jakob (`#361`) regarding the splitting/improvement of `ExportXcm::validate` for the messages pallet. As we discussed [here](https://matrix.to/#/!cZvXHjPZkBNsuesmYv:parity.io/$Bq7Sk-t3gMA07LiZ5gFINtMiVvSHwri2bh1X6-UQEvg?via=parity.io&via=matrix.parity.io) on Matrix - we need a wrapper/adapter for HaulBlobExporter, as Slava proposed `EnsureMessagesPalletActive`.

## Solution

I reviewed Slava's master implementation for dynamic bridges, and he has already addressed almost both issues with `pallet-xcm-bridge-hub`. Therefore, I believe creating any temporary `EnsureMessagesPalletActive` is not necessary.

Instead, I backported `pallet-xcm-bridge-hub` with just `ExportXcm` capabilities and adapted it with `XcmBlobHauler` elements to minimize the impact on Bridges V1. We will still proceed with the hard-coded lane for AssetHubs.

You can find the `polkadot-sdk` copy of this PR for the Rococo/Westend bridge here: https://github.com/paritytech/polkadot-sdk/pull/2602

This PR also brings `polkadot-staging` closer to the `master`, making backports and migrations to V2 easier.

## Next step

So, if this is the way to go, we should focus on adding the `validate_message` feature to both `pallet-xcm-bridge-hub` and `pallet-bridge-messages`, which will resolve Jacob's issue, as a follow up PR.

Part of: https://github.com/paritytech/parity-bridges-common/issues/2452


svyatonik (approved):
Overall - LGTM. But this PR actually doesn't solve the #2452? I mean - this xcm exporter works on the source BH and we also need to check that the lane id is correct at the target BH.

bkontur:
> Overall - LGTM. But this PR actually doesn't solve the #2452? I mean - this xcm exporter works on the source BH and we also need to check that the lane id is correct at the target BH.

ah, yes, you're right, I was focused on that sender validation/export stuff,
let me prepare antoher PR above this

PR #2728, opened 2023-12-05
separate constants for average and worst case relay headers

svyatonik:
The idea is to move bridge fees towards expected after https://github.com/paritytech/polkadot-sdk/pull/2139 . There I've used existing `AVERAGE_HEADER_SIZE_IN_JUSTIFICATION` constants, which are actually worst possible header size. Once this PR is merged, we'll keep using worst possible const for computing maximal refund, but for computing fees we'll be using average header size. Side effects:
- the user will pay smaller fee for over-bridge messages. Let this fee be `FEE`;
- SA of AH at BHs will keep paying to relayers at most `MaxFEE`, which is larger than the `FEE`;
- if message proofs will be built using large headers, the difference `MaxFEE - FEE` will be paid by the SA of AH.

In real world, however, we'll be running our own realyer (at least for v1), which is configured to deliver messages/confirmations asap and will select the header without looking at its size. Since we expect headers to have the `AVERAGE_HEADER_SIZE_IN_JUSTIFICATION`, not the `WORST_HEADER_SIZE_IN_JUSTIFICATION`. So the actual fee in average will be closer to the `FEE`, not the `MaxFEE`.


bkontur (approved):
looks like you need to add `WORST_HEADER_SIZE_IN_JUSTIFICATION` to the `polkadot-bulletin` imports
```
pub use bp_polkadot_core::{
	AccountAddress, AccountId, Balance, Block, BlockNumber, Hash, Hasher, Header, Nonce, Signature,
	SignedBlock, UncheckedExtrinsic, AVERAGE_HEADER_SIZE_IN_JUSTIFICATION,
	EXTRA_STORAGE_PROOF_SIZE, MAX_HEADER_SIZE, REASONABLE_HEADERS_IN_JUSTIFICATON_ANCESTRY,
};
```

Issue #2413, opened 2023-05-22
Tests for on-demand relays


svyatonik:
Follow up for paritytech/parity-bridges-common#2129 and paritytech/parity-bridges-common#2135. Let's start implementing new `TestingClient` and add missing tests for on-demand relays (parachain and headers).

EmmanuellNorbertTulbure:
@svyatonik to re-check and continue if needed

svyatonik:
superseded by #3002

PR #2729, opened 2023-12-06
Select header that will be fully refunded in on-demand batch finality relay

svyatonik:
This is a better alternative of #2728. The idea is to change relayer a bit and let it wait for the header that will be fully refunded (when speaking of its size) before submitting batch delivery/confirmation transaction. That allows us to have better bridge fee estimations:
- we believe that our relayer will be trying to select such header and justification, so that it'll be fully covered by our refund mechanism. Relayer is economically incentivized to do that;
- so we can use the regular/average header size in our fee estimations and be sure that we won't refund more than this fee.

As a backup plan for the breaking changes (when header size suddenly grows), the relayer will be "selecting" smaller header only a few times (4 atm). If after 4 attempts, the call size is still above maximal, it'll submit what it has. 

The runtime code **logic** changes are minimal:
- before we have been always using `MAX_HEADER_SIZE + max_justification_size(AVERAGE_HEADER_SIZE)`, where `MAX_HEADER_SIZE` was `90_000` and max justification size was `2 * 20_000 + some overhead for other fields`. So max refundable call size was about `90_000 + 40_000 = 130_000`;
- now we have two max call sizes. One for mandatory header submission: `120kb + 2 *1024 + overhead for other fields = 122kb + overhead for other fields`. The other is for regular headers - it is somewhere near `3Kb + overhead for other fields`. Since we assume that mandatory header submission is mandatory for our bridge (and also rare), we could use the latter estimation for our batch transactions.

The "core" of relayer logic changes is in the `relays/lib-substrate-relay/src/on_demand/headers.rs` file.

What is left in this PR:
- [x] test it: tested locally by using slightly modified `polkadot` binary (session time for Rococo and Westend increased from 1m to 5m) + modified `substrate-relay` (average Rococo header size is set to `100` bytes):
```
[BridgeHubRococo_to_BridgeHubWestend_MessageLane_00000002] 2023-12-06 10:26:23 +00 DEBUG bridge [Rococo-to-BridgeHubWestend-on-demand-headers] Requested to prove Rococo head 120. Selected to prove Rococo head HeaderId(121, 0x2ddbcfea419006daf5361c32dfd59b581c738b2b771898a77bdf3d4ad3b624d1). But it is too large: 767 vs 740. Going to select next header
...
[BridgeHubRococo_to_BridgeHubWestend_MessageLane_00000002] 2023-12-06 10:26:27 +00 DEBUG bridge [Rococo-to-BridgeHubWestend-on-demand-headers] Requested to prove Rococo head 120. Selected to prove Rococo head HeaderId(122, 0x841844afd16fe63d05c692ac60d7c796ead7817a9e39d0372282e23e3e0e2e21). But it is too large: 767 vs 740. Going to select next header
...
[BridgeHubRococo_to_BridgeHubWestend_MessageLane_00000002] 2023-12-06 10:26:35 +00 DEBUG bridge [Rococo-to-BridgeHubWestend-on-demand-headers] Requested to prove Rococo head 120. Selected to prove Rococo head HeaderId(123, 0xeae7a1a926584410ee36e36094c1fe3fb7c859f93d09b447a15b0a22410f9dec). But it is too large: 767 vs 740. Going to select next header
...
[BridgeHubRococo_to_BridgeHubWestend_MessageLane_00000002] 2023-12-06 10:26:39 +00 DEBUG bridge [Rococo-to-BridgeHubWestend-on-demand-headers] Requested to prove Rococo head 120. Selected to prove Rococo head HeaderId(124, 0x4bfd833825ff96dc5d39524e2da8587d46029bbf3c32c700ead9a6f5f20aece3) (after 4 iterations)
```
- [x] add unit tests where possible: unfortunately, we still miss ability to write tests for on-demand relays (#2413), so this PR has been tested locally on Rococo <> Westend and it has just one unit test :(;
- [x] file an issue for adding similar functionality to non-batch on-demand relays. We do not need it now, but better have some issue to not forget this fact in the future: #2730


relays/lib-substrate-relay/src/finality_base/engine.rs
```diff
@@ -110,6 +124,13 @@ pub trait Engine<C: Chain>: Send {
 		header: &C::Header,
 		proof: &mut Self::FinalityProof,
 	) -> Result<(), SubstrateError>;
+	/// Returns `true` if selected header and proof are exceeding maximal expected
```
serban300, line 127: Nit: I would add a new line after the previous method declaration.

serban300 (approved):
Looks good !

relays/lib-substrate-relay/src/finality_base/engine.rs
bkontur: Nit: also "Returns `true`` vs `MaxExpectedCallSizeCheck`  is not valid, right?

bkontur (approved):

svyatonik:
> Do you want to update subtree? I could do it also maybe with #2727

Main thing I want to do here is to deploy relay once this PR is merged. Constants are not that important - tests in fellowship repo are "manual" atm until we'll do another round of upgrade dependencies. But yes - if you'll be doing update in #2727, let's bring those changes too

PR #2732, opened 2023-12-08
Backport latest changes from polkadot staging

svyatonik:
Looks like new relayer works ok - still going to watch it for another couple of days.

@bkontur I'm not porting https://github.com/paritytech/parity-bridges-common/pull/2727, because you've said it is already a port from master :) But if you need something from there in master, please point - I'll port it too


svyatonik:
> yes, I have just small nits:
> 
> https://github.com/paritytech/parity-bridges-common/pull/2727/files#diff-1059dbc95faa4274ddbe6051c6b0c66a8e297dbbf799ce2867884dd904db2277R5-R6
> 
> https://github.com/paritytech/parity-bridges-common/pull/2727/files#diff-9aac78fe025db5c4ca23686baaf581f1c9cdcd279cf8bd673fdec734ce646312R5-R6
> 
> and I would move this bridgeId/laneId check: https://github.com/paritytech/parity-bridges-common/blob/master/modules/xcm-bridge-hub/src/exporter.rs#L88-L98 before `PalletAsHaulBlobExporter::<T, I>::validate(` and change `Unroutable` to `NotAplicable` -> if we dont have configuration for (source, dest), maybe we could have another instance for the same `BridgedNetwork` and give it a try, we dont need to be strict here and say "we dont allow only one instance for the same BridgedNetwork`

Ok, done. Thank you!

bkontur:
and this https://github.com/paritytech/parity-bridges-common/pull/2733 :)

bkontur (approved):

PR #2733, opened 2023-12-08
Grafana update stuff

bkontur:


svyatonik (approved):
Wanted to wait for a couple of days to ensure it is gone :) But ok - let's merge :)

PR #2761, opened 2023-12-27
Port safe commits from master to polkadot-staging

svyatonik:
opening early for visibility - could be reviewed commit-by-commit

Since we are now closer to moving to `polkadot-sdk` than to developing `bridges-v2` + some parts of v2 have been already merged to the `polkadot-staging`, maybe we should start picking some commits from master and merge them to `polkadot-staging`. I'll start with safe commits that are not touching any critical code pieces. The final goal is to have https://github.com/paritytech/parity-bridges-common/issues/2457 in `polkadot-staging` - it'll should make bridges configuration much easier. So in next updates of polkadot-sdk and fellowship runtimes we'll decrease LOCs dedicated to bridges configuration.

All `v2` stuff (dynamic lanes, relayers protocol) will stay in master.

Additionally, it'll make backporting easier - right now not having https://github.com/paritytech/parity-bridges-common/pull/2171 and https://github.com/paritytech/parity-bridges-common/pull/2181 in `polkadot-staging` is painful


svyatonik:
Actually, there's a lot of changes that I'm going to port (#2457), so maybe it is better to have multiple PRs. So let's start with that one

serban300 (approved):

Issue #2776, opened 2024-01-04
`pallet-xcm-bridge-hub` should do validations in `ExportXcm::validate`


bkontur:
## Summary

To handle the `ExportMessage` instruction, we utilize the new `pallet-xcm-bridge-hub`, which implements `ExportXcm`. This functionality comprises two parts:
- `validate`
  - [here](https://github.com/paritytech/parity-bridges-common/blob/polkadot-staging/modules/xcm-bridge-hub/src/exporter.rs#L59-L75) we validate **lane** and prepare/extend/decode XCM to blob
- `deliver`
  - [here](https://github.com/paritytech/parity-bridges-common/blob/polkadot-staging/modules/xcm-bridge-hub/src/exporter.rs#L80-L84) is the most important place, where we enqueue XCM blob to the **messages pallet** queue

The enqueuing process in `send_message` involves several validations [here](https://github.com/paritytech/parity-bridges-common/blob/polkadot-staging/modules/messages/src/lib.rs#L699-L729).

## Problem/fix

Move validations from `send_message` to the `ExportXcm::validate`. 

Also `lane.save_message` does [validation for message length `MessageTooLarge`](https://github.com/paritytech/parity-bridges-common/blob/c3e469ee210410b69089ad4554e38cc66046a954/modules/messages/src/lib.rs#L873-L884), so maybe this could be added also to the `validate` but after `PalletAsHaulBlobExporter::<T, I>::validate`.

## Note

I would maybe start with `polkadot-staging` and then backport to `master`, `ExportXcm::validate` should differ only with hard-coded `laneIds` vs dynamic `bridges`.

serban300:
Ported to polkadot-sdk via: https://github.com/paritytech/polkadot-sdk/pull/2996

Closing

Polkadot-Forum:
This issue has been mentioned on **Polkadot Forum**. There might be relevant details there:

https://forum.polkadot.network/t/polkadot-kusama-bridge/2971/8

PR #2795, opened 2024-01-17
[`polkadot-staging`] Separate `send_message()` validations

serban300:
Addresses https://github.com/paritytech/parity-bridges-common/issues/2776 for the `polkadot-staging` branch


modules/messages/src/lib.rs
```diff
@@ -534,6 +534,7 @@ pub mod pallet {
 	}
 
 	#[pallet::error]
+	#[derive(PartialEq)]
```
svyatonik, line 537: I could remember I've been fixing [this lint](https://rust-lang.github.io/rust-clippy/master/#/derive_partial_eq_without_eq) some time ago - we could also probably derive the `Eq` here?

primitives/messages/src/source_chain.rs
```diff
@@ -125,10 +125,22 @@ pub trait MessagesBridge<Payload> {
 	/// Error type.
 	type Error: Debug;
 
+	/// Intermediary structure returned by `validate_message()`.
+	///
+	/// It can than be passed to `send_message()` in order to actually send the message
+	/// on the bridge.
+	type SendMessageArgs;
```
svyatonik, line 132: Let's probably remove this associated type and use a fixed `struct ValidatedMessage` (or something like that?). There's a lot of work done in the `master` branch to remove extra generic args/associated types in favor of fixed types. Or is there really a need for that?

primitives/messages/src/source_chain.rs
```diff
@@ -137,9 +149,14 @@ pub struct NoopMessagesBridge;
 
 impl<Payload> MessagesBridge<Payload> for NoopMessagesBridge {
```
svyatonik, line 150: Not a part of this PR - sorry. But I think this `NoopMessagesBridge` isn't used anywhere anymore - so if it causes additional issues, just remove it.

modules/messages/src/mock.rs
```diff
@@ -402,8 +402,10 @@ pub fn message(nonce: MessageNonce, payload: TestPayload) -> Message {
 }
 
 /// Return valid outbound message data, constructed from given payload.
-pub fn outbound_message_data(payload: TestPayload) -> MessagePayload {
-	payload.encode()
+pub fn outbound_message_data<T: Config<I>, I: 'static>(
```
svyatonik, line 405: Nit: you could remove generic args and just return `StoredMessagePayload<TestRuntime, ()>`

svyatonik (approved):
LGTM!

bkontur (approved):
lgmt, I like that `send_message` does not return a `Result`.  I guess, it is compliant with the philosophy of `ExportXcm::validate/deliver`.

modules/messages/src/lib.rs
serban300: Done

modules/messages/src/mock.rs
serban300: Done

primitives/messages/src/source_chain.rs
serban300: I added this in order to be able to define the actual `SendMessageArgs` struct inside the messages module. This way it doesn't have to expose a constructor. It can only be constructed there. So that it can't be created manually without going through the `validate_message()` function. WDYT ?

serban300: @svyatonik sorry for the ping. WDYT ? Would you have any concerns about keeping this associated type ?

svyatonik: Oh, sorry - missed that comment. Yeah - that makes sense. I think both options are fine in that case - leave as is, or change

PR #2827, opened 2024-02-06
[Do Not Merge until RBH/WBH upgrade] Use new submit_finality_proof_ex call in Rococo <> Westend bridge

svyatonik:
I've also:
- removed manual `runtime_wrapper.rs` that we have been using to craft WBH transactions and now we are using actual WBH runtime, generated by Serban' tool;
- regenerated RBH and WBH runtimes from [this branch](https://github.com/paritytech/polkadot-sdk/tree/sv-update-bridges-subtree).


bkontur (approved):
lgtm!

Do you think we could remove this TODO: https://github.com/paritytech/parity-bridges-common/blob/polkadot-staging/scripts/regenerate_runtimes.sh#L13-L15 ?

svyatonik:
> lgtm!
> 
> Do you think we could remove this TODO: https://github.com/paritytech/parity-bridges-common/blob/polkadot-staging/scripts/regenerate_runtimes.sh#L13-L15 ?

No, it is still actual :)

PR #2859, opened 2024-03-04
Some remaining fixes (mostly insubstantial) for relay

svyatonik:
I'm preparing a documentation on how to make a releases here && found some things in relayer that I'd like to fix before making an initial release for PBH <> KBH. Better to review as separate commits - it should be easier then:
- https://github.com/paritytech/parity-bridges-common/commit/9d6e752644df44d90cde564650308140aa35f906: by default we use bundled chain spec and transaction runtime versions when submitting a transcations and/or starting a relay guard. However, if there's no bundled version in the code (as e.g. for all relay chains), then we **silently** use the auto version. Which means that version guard is not started => relay will keep running after upgrade happens, which is not something that have been planned. I've changed the code - relay now exits if bundled version is `None` and we need to use it;
- https://github.com/paritytech/parity-bridges-common/commit/7f351ee56cf1a99dc8852c796388439eb934948c: update chain spec versions to actual versions for all supported (and live) chains;
- https://github.com/paritytech/parity-bridges-common/commit/caffe94b57a773cb95bb32a6a6a7b80c9c98d160: instead of manually written runtime wrappers, let's use autogenerated runtime for KBH and PBH. This has the drawback - since current version of PBH misses Bulletin chain support, the relay won't work in an emulated local environment. But let's see when it actually lands the PBH - there's no even live bridge between Rococo and Bulletin chain yet.

@serban300 If you gonna be doing local zombienet tests for Kusama <> Polkadot bridge, can you, please, use relayer, updated here - just to double check that I haven't accidentally broke anything. Thank you! (P.S.: I'm unable to do local zn tests these days)


serban300 (approved):

serban300:
> @serban300 If you gonna be doing local zombienet tests for Kusama <> Polkadot bridge, can you, please, use relayer, updated here - just to double check that I haven't accidentally broke anything. Thank you! (P.S.: I'm unable to do local zn tests these days)

Sure, will run the tests with this version of the relay. Will get back with the results.

serban300:
> > @serban300 If you gonna be doing local zombienet tests for Kusama <> Polkadot bridge, can you, please, use relayer, updated here - just to double check that I haven't accidentally broke anything. Thank you! (P.S.: I'm unable to do local zn tests these days)
> 
> Sure, will run the tests with this version of the relay. Will get back with the results.

Done. The tests run successfully.

svyatonik:
> > > @serban300 If you gonna be doing local zombienet tests for Kusama <> Polkadot bridge, can you, please, use relayer, updated here - just to double check that I haven't accidentally broke anything. Thank you! (P.S.: I'm unable to do local zn tests these days)
> > 
> > 
> > Sure, will run the tests with this version of the relay. Will get back with the results.
> 
> Done. The tests run successfully.

Thank you!

acatangiu (approved):

Issue #2871, opened 2024-03-11
Reducing bridge fees: short and long term options


svyatonik:
Bridge fees will be quite high in initial deployment. Omitting parts, that are not relevant to the bridge itself (`transfer_assets` transaction cost, HRMP, potential increased fee factors, ...), the fee consists of:

---
For Polkadot -> Kusama message is 1.93 DOTs:
1) cost of XCM execution at Polkadot BH: 0.01 DOT (which has been decreased significantly from 0.48 DOT in https://github.com/paritytech/polkadot-sdk/pull/3157). [Link](https://github.com/polkadot-fellows/runtimes/blob/e42821da8d85f721d0dd1670dfb23f4dd91bd3e8/system-parachains/bridge-hubs/bridge-hub-polkadot/primitives/src/lib.rs#L92)
2) cost of message delivery transaction at Kusama BH: 0.06 KSM, converted to 0.3 DOT. [Link](https://github.com/polkadot-fellows/runtimes/blob/e42821da8d85f721d0dd1670dfb23f4dd91bd3e8/system-parachains/bridge-hubs/bridge-hub-kusama/primitives/src/lib.rs#L105C57-L105C71)
3) const of message confirmation transaction at Polkadot BH: 1.62 DOT. [Link](https://github.com/polkadot-fellows/runtimes/blob/e42821da8d85f721d0dd1670dfb23f4dd91bd3e8/system-parachains/bridge-hubs/bridge-hub-polkadot/primitives/src/lib.rs#L100)
---
For Kusama -> Polkadot message is 0.39 KSMs:
1) cost of XCM execution at Kusama BH: almost zero (0.0004 KSM). [Link](https://github.com/polkadot-fellows/runtimes/blob/e42821da8d85f721d0dd1670dfb23f4dd91bd3e8/system-parachains/bridge-hubs/bridge-hub-kusama/primitives/src/lib.rs)
2) cost of message delivery transaction at Polkadot BH: 1.7 DOT, converted to 0.34 KSM. [Link](https://github.com/polkadot-fellows/runtimes/blob/e42821da8d85f721d0dd1670dfb23f4dd91bd3e8/system-parachains/bridge-hubs/bridge-hub-polkadot/primitives/src/lib.rs#L96C62-L96C76)
[...]

PR #2884, opened 2024-03-18
Add option to only relay free headers to substrate-relay

svyatonik:
related to #2871

Adds support of #2873 to our standalone GRANDPA and finality relays. The main thing that is left here is to actually test it and for that I need to get updated Rococo<>Westend runtimes first

UPD: obviously I'll also need to change tx submitter to support that new `free_execution_expected` flag


svyatonik:
I've updated zn test "one" here: https://github.com/paritytech/polkadot-sdk/tree/sv-try-new-bridge-fees - will need to add some more conditions to test that we are not actually spend any funds on finality

svyatonik:
Superseded by https://github.com/paritytech/polkadot-sdk/pull/4157. BUT. Chain-specific changes from this PR must be merged afteer `polkadot-sdk` is upgraded. So I'm leaving this PR open

Issue #2538, opened 2023-09-05
Migrate parity-bridges-common repo to polkadot-sdk repo


acatangiu:
The point is to move (ideally everything) from https://github.com/paritytech/parity-bridges-common repo to https://github.com/paritytech/polkadot-sdk repo.

Right now we're already manually syncing/copying (parts of the) following dirs:
```
bin		modules		docs		primitives		scripts
```
as they're needed by several system parachains (`BridgeHub`s and `AssetHub`s).

### Prerequisites for full migration

Main "blocker" is the fact that our bridge test deployments are using custom chains [millau](https://github.com/paritytech/parity-bridges-common/tree/master/bin/millau), [rialto](https://github.com/paritytech/parity-bridges-common/tree/master/bin/rialto) and custom parachain [rialto-parachain](https://github.com/paritytech/parity-bridges-common/tree/master/bin/rialto-parachain).

We are testing the bridge by [(re)deploying](https://github.com/paritytech/parity-bridges-common/tree/master/deployments)  the chains, the bridge and the relayers every morning with latest changes and letting it run for ~24h.

We don't want to bring in `Rialto, Millau` to `polkado-sdk` so first thing to do is move our nightly test deployments to use `Rococo<>Wococo` (and their BridgeHub and AssetHub parachains).
A blocker here is that RococoAH is deployed from KusamaAH code which doesn't contain bridging functionality, so we need https://github.com/paritytech/polkadot-sdk/issues/1128 + https://github.com/paritytech/polkadot-sdk/pull/1215.

### List

- [x] https://github.com/paritytech/polkadot-sdk/pull/1215
- [x] https://github.com/paritytech/polkadot-sdk/issues/1128
- [x] migrate [test deployments](https://github.com/paritytech/parity-bridges-common/tree/master/deployments) to `R<>W`
- [x] https://github.com/paritytech/parity-bridges-common/issues/2472
- [x] retire Rialto and Millau
- [x] move over code from https://github.com/paritytech/parity-bridges-common to https://github.com/paritytech/polkadot-sdk/tree/master/bridges

bkontur:
I had this question on last bridges meeting:

When we move production runtimes for Kusama/Polkadot to the `polkadot-fellows` repository, where do we place/keep relayer-related configuration/code?

`polkadot-sdk` or old `parity-bridges-common` or `polkadot-fellows` or somewhere else?

acatangiu:
[...]

PR #2885, opened 2024-03-18
Move generic CLI logic to different crate

serban300:
Related to https://github.com/paritytech/parity-bridges-common/issues/2538

This PR does a high-level decoupling of the generic CLI logic from the specific relay flavors.

Further improvements and cosmetics are probably possible, but will do that in future PRs.


serban300:
> Does it makes sense to move it to [existing crate](https://github.com/paritytech/parity-bridges-common/tree/master/relays/lib-substrate-relay) instead?

@svyatonik Thanks ! Good idea ! Done. PTAL !

PR #2902, opened 2024-03-28
Some relayer improvments

svyatonik:
I'll propagate changes to `polkadot-sdk` once https://github.com/paritytech/polkadot-sdk/pull/3859 is merged.

This PR just adds some CLI args that we haven't supported so far. E.g. to connect to OnFinality Polkadot node, you need to use `wss://polkadot.api.onfinality.io:443/public-ws`. You had no chance to specify that `path` component before this PR. Now you can use either:
```
    --source-host polkadot.api.onfinality.io \
    --source-port 443 \
    --source-secure \
    --source-suffix "/public-ws" \
```
or
```
--source-uri wss://polkadot.api.onfinality.io:443/public-ws
```

We'll deprecate the former method later, when we'll start using `*-uri` in all our relayers.


relays/client-substrate/src/lib.rs
```diff
@@ -57,10 +57,15 @@ pub use bp_runtime::{
...
 pub struct ConnectionParams {
+	/// Websocket endpoint URL. Overrides all other URL components (`host`, `port`, `path` and
+	/// `secure`).
+	pub uri: Option<String>,
 	/// Websocket server host name.
 	pub host: String,
 	/// Websocket server TCP port.
 	pub port: u16,
+	/// Websocket endpoint path at server.
+	pub path: Option<String>,
```
serban300, line 68: Should we mark `host`, `port`, `path` and `secure` as deprecated in order to not forget about them ?

serban300 (commented):
> I'll propagate changes to polkadot-sdk once https://github.com/paritytech/polkadot-sdk/pull/3859 is merged.

I can cherry-pick this commit into https://github.com/paritytech/polkadot-sdk/pull/3859 after it's merged into parity bridges-common or I can redo https://github.com/paritytech/polkadot-sdk/pull/3859 from scratch. I see that there are some more comments to address there anyway, and I'll probably fix at least part of them inside parity-bridges-common

relays/client-substrate/src/lib.rs
serban300: Ok, makes sense. 

> I could add a suffix to associated CLI arguments: "Deprecated: use uri instead" as a compromise. WDYT?

I don't know if it's necessary. We can leave it as it is. At most we can add a TODO comment, just to make sure we don't forget about it.

serban300 (approved):

PR #2910, opened 2024-04-01
Address migration comments

serban300:
Addressing comments from https://github.com/paritytech/polkadot-sdk/pull/3859

More details in the commit titles

### DIFF
--- a/tools/runtime-codegen/Cargo.toml
+++ b/tools/runtime-codegen/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 description = "Tool for generating bridge runtime code from metadata"
 authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2021"
-license = "GPL-3.0-or-later WITH Classpath-exception-2.0"
+license = "GPL-3.0-only"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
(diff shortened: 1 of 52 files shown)


tools/runtime-codegen/Cargo.toml
```diff
@@ -4,15 +4,16 @@ version = "0.1.0"
 description = "Tool for generating bridge runtime code from metadata"
 authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2021"
-license = "GPL-3.0-or-later WITH Classpath-exception-2.0"
+license = "GPL-3.0-only"
```
svyatonik, line 7: Should we change to "GPL-3.0-or-later WITH Classpath-exception-2.0"?

svyatonik (approved):

tools/runtime-codegen/Cargo.toml
svyatonik: Sorry, I mean - leave the previous version : "GPL-3.0-or-later WITH Classpath-exception-2.0". The whole license stuff is explained here: https://substrate.stackexchange.com/questions/594/how-should-we-choose-a-license - Apache for pallets and other runtime stuff and GPL with classpath for client (also relay). I may be wrong, though - just think that we don't want to restrict license on already available code.

serban300: Ok, thanks for the explanation ! Reverted the licenses to the previous version.

PR #2943, opened 2024-04-11
[dependabot] ignore migrated crates

serban300:


.github/dependabot.yml
```diff
@@ -8,7 +8,23 @@ updates:
...
   ignore:
-  # Substrate (+ Polkadot/Cumulus pallets) dependencies
+  # Bridges polkadot-sdk dependencies
+  - dependency-name: bp-*
+    versions:
+      - ">= 0"
+  - dependency-name: bridge-runtime-common
+    versions:
+      - ">= 0"
+  - dependency-name: parachains-relay
```
svyatonik, line 18: We have no other `relay-*` dependencies here? `finality-relay`, `messages-relay`, ...?

serban300: You're right. Good catch. Added the other relays as well.

svyatonik (approved):

Issue #2419, opened 2023-05-03
[P-K Bridge] Add monitoring + alarms for BHP<>BHK


EmmanuellNorbertTulbure:
- [x] https://github.com/paritytech/parity-bridges-common/issues/2199

Need to adjust metrics specifically for Polkadot<> Kusama

bkontur:
@svyatonik do you want to take? I could take a look next week

svyatonik:
> @svyatonik do you want to take? I could take a look next week

I was planning to do that, because it is a bummer that we have it for test bridge and not for prod bridge :) But I'll probably won't have much time for that next week :/ Let's sync here - when one of us start work on that, let's just leave here a comment

svyatonik:
@bkontur I'm picking it - will try to finish today

PR #2960, opened 2024-04-18
P<>K dashboards and alerts

svyatonik:
closes https://github.com/paritytech/parity-bridges-common/issues/2419

Spend some time this morning to understand why our relayer has no registered rewards (and metrics are always `None`). Turned out they've been already claimed - totally forgot about that :facepalm:


acatangiu (approved):
Didn't review the JSONs, but deployed dashboards look good! 👍

PR #3059, opened 2024-10-04
Update latest `polkadot-sdk` master with permissionless lanes feature support

bkontur:
This PR updates dependencies to the latest polkadot-sdk master and regenerates codegen for all testnets.

## TODO

- [x] finish fellows testing
  - [x] wait just for https://github.com/paritytech/polkadot-sdk/pull/5921 then one more Cargo.lock bump here


deny.toml
```diff
@@ -35,16 +36,8 @@ targets = [
...
-# The lint level for security vulnerabilities
-vulnerability = "deny"
-# The lint level for unmaintained crates
-unmaintained = "warn"
 # The lint level for crates that have been yanked from their source registry
 yanked = "warn"
-# The lint level for crates with security notices. Note that as of
-# 2019-12-17 there are no security notice advisories in
-# https://github.com/rustsec/advisory-db
-notice = "warn"
```
serban300, line 47: Hmmm, I am not very familiar with this logic, but I think we should keep these.

acatangiu (approved):

deny.toml
bkontur: @serban300 me neither, but the actual documentation: https://embarkstudios.github.io/cargo-deny/checks/advisories/cfg.html#the-version-field-optional says:
```
vulnerability - Removed, all vulnerability advisories now emit errors.
unmaintained - Removed, all unmaintained advisories now emit errors.
notice - Removed, all notice advisories now emit errors.
```
iiuc, it should be even more strict now and ok

serban300: Oh, ok. Sounds good then ! Thanks !

Issue #3109, opened 2025-02-10
P/K substrate-relayer deployment


bkontur:
The actual deployment is triggered by a "tag trigger" (https://github.com/paritytech/parity-bridges-common/pull/3083).  
However, we need to manually trigger this for P/K deployment because:
1. We need to wait for actual Runtimes enactment live, but we want to prepare/tests docker image beforehand
2. We are also using the Docker image from the tag for Polkadot-SDK Rococo/Westend Zombienet testing.
3. Soon, we will also need to include Polkadot-Fellows CI.
4. As a result, we cannot use partial or release candidate (RC) tags at this time.  

I suggest creating a GitHub Action for manual P/K relayer deployment, where we simply select or enter the actual tag and click a "Deploy" button. This button would trigger an ArgoCD deployment as follows:  

```
deploy-chains:
    name: Deploy Kusama/Polkadot Relay
```

Access to this button should be restricted to specific Parity teams, including at least `Bridges Core`, `DevOps`, `CI`, and `Core Devs`.

BulatSaif:
you can start deployment here: https://github.com/paritytech/parity-bridges-common/actions/workflows/deploy.yml

![Image](https://github.com/user-attachments/assets/98fd812e-a524-4cf0-b41c-0c4e594145f2)

Please contact me when you will try to deploy. I will support first runs.

bkontur:
> Please contact me when you will try to deploy. I will support first runs.

cool, thank you, will do, in 1-2 weeks: https://kusama.subsquare.io/referenda/494

PR #3111, opened 2025-02-10
Manual Deploy

BulatSaif:
Fixes https://github.com/paritytech/parity-bridges-common/issues/3109

Changes:

1. Split `deploy.yml`  for two files: `deploy.yml`  and  `build-tag.yml`
2.  add workflow_dispatch  to `deploy.yml` 

@alvicsam

### DIFF
--- a/.github/workflows/build-tag.yml
+++ b/.github/workflows/build-tag.yml
@@ -1,0 +1,108 @@
+name: Build tag
+
+on:
+  push:
+    tags:
+      - v*
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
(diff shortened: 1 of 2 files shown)


BulatSaif:
@alvicsam, can you check how we can limit the users who can start workflow_dispatch ?

If it is hard to do at the GitHub level, I can add a check step similar to `Check if Docker image exists on Docker Hub`.

> Access to this deploy should be restricted to specific Parity teams, including at least Bridges Core, DevOps, CI, and Core Devs.

alvicsam (approved):
Added suggested teams to reviewers of the deploy workflow

bkontur (approved):

bkontur:
@BulatSaif so it is ready? can we merge now?

PR #3147, opened 2025-03-27
Added exceptions to stop CI checks from failing

antkve:
Added exceptions for licenses and advisories to fix CI pipeline. Also fixed some shell scripts to work on mac.

Note that we are suppressing the fact that parity-wasm is officially deprecated but still used in wasm-instrument and parts of polkadot-sdk

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -41,17 +41,19 @@ yanked = "warn"
 # A list of advisory IDs to ignore. Note that ignored advisories will still
 # output a note when they are encountered.
 ignore = [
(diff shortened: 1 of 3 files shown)


serban300 (approved):

bkontur (approved):
I am just thinking we can add more and more exceptions (I did also previously :D),
but at some point we should also start to remove them when things are getting fix, right?

antkve:
> I am just thinking we can add more and more exceptions (I did also previously :D),
> but at some point we should also start to remove them when things are getting fix, right?

Yeah, for some of them we just need to update the packages that have them as a dependency, gave it a go for a couple of them but resolving the conflicts looked daunting

PR #3218, opened 2026-02-17
Add release Skill and corresponding sub-skills together with the prepare script

x3c41a:
## Summary

Split the monolithic `/release` skill into composable sub-skills backed by a standalone `prepare-bridge-hub-release.sh` script. Enables the automated release flow from mobile or CI:

`/release-prepare` → `/release-pr` → human merge → `/release-finalize`

Related: Rohit's release workflow in #3215.

## Skills

| Skill | Description |
|-------|-------------|
| `/release` | Orchestrate the full release lifecycle, chaining sub-skills |
| `/release-prepare` | Update chain spec_version, regenerate codegen, bump relay version |
| `/release-pr` | Create release branch, commit, and PR with `A-release` label |
| `/release-finalize` | Create annotated git tag, wait for Docker build, publish GitHub Release |

## New files

- `scripts/runtime-matrix.json` — chain metadata for BridgeHub WASM downloads
- `scripts/prepare-bridge-hub-release.sh` — automates spec_version update, codegen regeneration, BlakeTwo256 fix, and relay version bump
- `.claude/skills/release-prepare/skill.md`
- `.claude/skills/release-pr/skill.md`
- `.claude/skills/release-finalize/skill.md`
- `.claude/skills/release/skill.md` (updated)

## Follow-up PRs

- #3219 — `/deploy` and `/rollback` skills (depends on this PR)
- #3220 — `/debug-ci` and `/pipeline-status` skills (independent)

## Test plan

- [x] Script argument validation tested (no args, invalid chain, bad version all error correctly)
- [x] Version parsing verified: `v1.1.6` → `1_001_006`, `v2.0.6` → `2_000_006`
- [x] JSON validated with `jq`
- [x] All skill frontmatter matches existing patterns
- [x] macOS compatibility: portable `sed`/`grep` (no `grep -oP`, `sed -i'' -e` form)
- [ ] End-to-end test with a real BridgeHub runtime upgrade

PR #3219, opened 2026-02-17
Add /deploy and /rollback skills

x3c41a:
## Summary

- `/deploy [version]` — trigger and monitor ArgoCD deployment via `gh workflow run deploy.yml`
- `/rollback [version]` — roll back to a previous version with commit diff visibility

Updates `/release` and `/release-finalize` to reference `/deploy` instead of manual workflow instructions.

Depends on #3218.

## Test plan

- [x] `gh workflow run deploy.yml --ref v<VERSION>` syntax matches deploy.yml (`github.ref_name`)
- [x] Cross-references: `/release-finalize` → `/deploy` → `/check-health`
- [x] Frontmatter matches existing skill patterns

### DIFF
--- a/.claude/skills/deploy/skill.md
+++ b/.claude/skills/deploy/skill.md
@@ -1,0 +1,109 @@
+---
+name: deploy
+description: Deploy substrate-relay to production via ArgoCD
+metadata:
+  argument-hint: "[version]"
+---
+
+# Deploy
+
+Deploy a substrate-relay version to production via the ArgoCD deploy workflow.
+
+## Usage
+
+```
+/deploy              # Auto-detect version from Cargo.toml or latest git tag
+/deploy v1.8.16      # Deploy a specific version
+```
+
+## Prerequisites
+
+- The version must have a corresponding git tag (created by `/release-finalize`)
+- The Docker image `paritytech/substrate-relay:v<VERSION>` must exist on Docker Hub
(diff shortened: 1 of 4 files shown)


dr333ws:
> (TL;DR - please don't do this before resolving my worries !! giving an AI deploy/rollback access is incredibly risky)
> 
> I'm highly concerned by a lack of human in the loop here - what does an autonomous "/deploy" or "/rollback" actually mean in practice?
> 
> Let me check I understand:
> 
> Claude reads local code/logs/PRs -> decides to use a deployment tool -> executes infrastructure changes
> 
> What prevents a prompt injection from a poisoned dependency or malicious PR description? - an attacker could hide instructions telling the agent to trigger an unauthorised deployment or teardown.
> 
> How do you know Claude is going to do what you expect it to do, like deploying the correct branch to the correct environment? (its non deterministic)
> 
> What are the accesses for the cloud/deployment credentials used - how are they isolated from the agent so they can't be stolen if the agent's context window is compromised?
> 
> Are these skills going to be wired into automated CI, or strictly run locally by engineers?

Actually, there is a human in the loop for those skills; all the commands are manual and designed for human to execute, main goal is to automate the repeated processes. 

Double checking with the team whether we need the skills...

dr333ws:
Deprecated in favour of github workflows

Issue #3110, opened 2025-02-10
Investigate how to run Bridges Zombienet tests (Polkadot SDK at least, Polkadot Fellows) within PRs as part of this CI


bkontur:
When upgrading or updating the relayer's code here (e.g., new runtime version, Polkadot SDK bump, etc.), we need to manually test the new `substrate-relay` binary with different bridges. While we can do this locally, I’m sure we can also integrate it into the CI as part of PRs.

PR #3276, opened 2026-06-24
Migrate bridge integration tests to zombienet-sdk

### LINKED ISSUE #7665: Investigate and rewrite bridges zombienet tests with zombienet-sdk
Actually, we are doing `*.zdnsl` zombienet testing, which are not so flexible and requires pretty heavy setup (node-js, two zombienets, binaries, lots of scripts, e.g. for HRMP):
https://github.com/paritytech/polkadot-sdk/tree/master/bridges/testing
https://github.com/polkadot-fellows/runtimes/tree/main/integration-tests/bridges

With new zombienet-sdk we should be able just to do it in single rust test, e.g.:
```
        // config and env
	let spawn_fn = get_spawn_fn();
	let polkadot_config = polkadot_network_config().unwrap();
	let kusama_config = kusama_network_config().unwrap();

	// spawn Kusama and Polkadot
	let polkadot_network = spawn_fn(config).await.unwrap();
	let kusama_network = spawn_fn(config).await.unwrap();

	// access the nodes
	let polkadot_alice = polkadot_network.get_node("alice").unwrap();
	let kusama_alice = kusama_network.get_node("alice").unwrap();
```

We should be able to configure storage overrides, genesis overrides, hrmps and so on.

So hopefully, we should be able to cover several other issues with this:
- Add bridges tests to the fellows CI https://github.com/polkadot-fellows/runtimes/issues/544
- Handle initialization without sudo https://github.com/paritytech/parity-bridges-common/issues/3016

## Tasks

- [ ] prepare one easy test which just spawns the whole networ setup for Rococo/Westend
- [ ] investigate how to set_storage for pallet_bridge_granda pallet_owner (alternative to root)
- [ ] rewrite step by ste

rosarp:
Closes: https://github.com/paritytech/polkadot-sdk/issues/7665
Relates: https://github.com/paritytech/parity-bridges-common/issues/3110

### Summary

Adds a Rust-based `zombienet-sdk` integration test suite for the local `Rococo ↔ Westend bridge`, replacing the older shell/native-script zombienet (v1) approach, moved from polkadot-sdk to this repo. The tests spawn both relay-chain networks together with their Bridge Hub and Asset Hub parachains, drive the substrate-relay binary as a subprocess, and assert on-chain state via subxt.

### What's added

testing/zombienet-sdk-tests/ — the test crate (bridges-zombienet-sdk-tests):
- `asset_transfer` — transfers assets across the bridge (both directions, native and wrapped), checks they arrive while free-header relayers' balances stay constant.
- `free_headers` — checks that free relay-chain and parachain headers sync to the remote Bridge Hub while the relayer is otherwise idle.
- `tests/environment.rs` — shared harness: network spawning, bridge initialization, the substrate-relay driver, and subxt query/extrinsic helpers.
- Tests are gated behind the zombie-ci feature so a plain cargo check/cargo test of the workspace stays cheap.

`testing/metadata-gen/` — a self-contained helper to (re)generate the committed subxt metadata:
- `generate.sh` reads the pinned polkadot-sdk revision from Cargo.lock, checks polkadot-sdk out at that exact commit, builds the six runtime WASM blobs (Rococo, Westend, and their Asset Hub / Bridge Hub system parachains), extracts their metadata, copies the .scale files into the test crate, and cleans up.
- `runner/` is a tiny extraction crate injected into the polkadot-sdk checkout so it uses the checkout's own substrate crates at the pinned revision — avoiding version skew and avoiding pulling heavy substrate git deps into this repo's workspace.

### Notes for reviewers

- The `metadata-files/*.scale` blobs are read at compile time by #[subxt::subxt(...)] and validated against each node's live runtime at runtime — they must match the runtimes embedded in the polkadot/polkadot-parachain binaries under test. Regenerate via testing/metadata-gen/generate.sh whenever the pinned polkadot-sdk revision changes.
- `Cargo.lock` changes are the bulk of the diff — they pull in zombienet-sdk and its transitive dependencies.

To run & test locally refer `testing/metadata-gen/README.md`


rosarp:
Moved zombienet-sdk tests from https://github.com/paritytech/polkadot-sdk/pull/12257 to bridges.
Other PR will be obsolete.

testing/zombienet-sdk-tests/tests/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+//! These are the Rust port of the legacy bash + `zndsl` tests that used to live under
+//! `bridges/testing/tests/`. They spawn two relay-chain networks (Rococo and Westend), each with
+//! a Bridge Hub and an Asset Hub parachain, drive the external `substrate-relay` binary as a set
+//! of subprocesses and assert on-chain state via `subxt`.
+//!
+//! The tests are gated behind the `zombie-ci` feature so that a plain `cargo check` of the
+//! workspace stays cheap and does not require the generated metadata files.
+
+#[cfg(feature = "zombie-ci")]
+#[subxt::subxt(runtime_metadata_path = "metadata-files/rococo-local.scale")]
```
bkontur: @rosarp hmm, I am just thinking, can we also reuse our existing codegens? like https://github.com/paritytech/parity-bridges-common/blob/master/relay-clients/client-rococo/src/codegen_runtime.rs 

This way, the test would not need to generate scales, just reuse what we have, naybe this could also force us to maintain codegens, wdyt?

testing/zombienet-sdk-tests/Cargo.toml
```diff
@@ -0,0 +1,53 @@
...
+[dependencies]
+anyhow = { workspace = true }
+codec = { workspace = true }
+env_logger = { workspace = true }
+log = { workspace = true }
+serde_json = { workspace = true, default-features = true }
+# `subxt`/`subxt-signer` are pinned here (not inherited from the workspace's 0.40.1) to match the
+# version that `zombienet-sdk` uses internally, so `node.wait_client()` hands back a client of the
+# same type these tests use — no second subxt version is introduced beyond zombienet's own.
+subxt = { version = "0.44", default-features = false, features = ["native"] }
```
bkontur: @rosarp can we move all the versions to the workspace Cargo.toml?

testing/zombienet-sdk-tests/tests/environment.rs
```diff
@@ -0,0 +1,1640 @@
...
+				client: &OnlineClient<PolkadotConfig>,
+				account: [u8; 32],
+			) -> Result<u128, anyhow::Error> {
+				free_balance_at(client, account).await
+			}
+		}
+	};
+}
+
+relay_ops!(relay_rococo, rococo, rococo_runtime);
```
bkontur: @rosarp I think we could/should split:
- configuration + generic stuff - e.g. submodule zombienet-sdk-tests/lib ("zombienet-sdk-tests-lib")
- zombienet-sdk-tests/rococo-westend/tests
- zombienet-sdk-tests/polkadot-kusama-tests

Maybe if create good `zombienet-sdk-tests-lib`, we could just configure and run the tests in Fellows CI easily or something like that

testing/zombienet-sdk-tests/tests/environment.rs
```diff
@@ -0,0 +1,1640 @@
...
+/// Runs the idempotent `init-bridge` command until the bridge is confirmed operational.
+///
+/// The relay's `init-bridge` decides it is done by reading `is_initialized` at the **best**
+/// (non-finalized) header. On a loaded host the bridge hub reorgs ~tens of blocks deep, so the
+/// init tx can land on a branch that is briefly best — fooling that check into reporting success —
+/// and is then retracted, leaving the canonical chain `Halted`. We therefore drive init from the
+/// **finalized** state: before each attempt we check `operating_mode == Normal` at a finalized
+/// block, and only stop once that holds. Each attempt re-submits (the relay no-ops fast when it
+/// believes it is already initialized), so once a submission finalizes the bridge is operational.
+async fn init_bridge_confirmed(
```
bkontur: @rosarp I would put all these helpers to separate `helpers/utils.rs` file

testing/zombienet-sdk-tests/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+The tests are gated behind the `zombie-ci` feature so that a plain `cargo check`/`cargo test` of the
+workspace stays cheap. The committed `metadata-files/*.scale` blobs are used by the `subxt` codegen,
+so the tests run without any extra setup.
+
+Select the **native** zombienet provider so the tests spawn the local `polkadot` / `polkadot-parachain`
+binaries on your `PATH` directly, instead of containers (Docker/Podman) or Kubernetes:
+
+```bash
+export ZOMBIE_PROVIDER=native
```
bkontur: @rosarp for CI, I would like to use ZOMBIE_PROVIDER=docker, when we use - `debug` images, they should be build with `--fast-runtime` so no need to build anything from PolakdotSDK - this would definitely speed up everything:
```
POLKADOT_IMAGE    │ docker.io/paritypr/polkadot-debug:9601363f
CUMULUS_IMAGE     │ docker.io/paritypr/polkadot-parachain-debug:9601363f
```

testing/zombienet-sdk-tests/tests/environment.rs
```diff
@@ -0,0 +1,1640 @@
...
+					n.with_name("asset-hub-rococo-collator2").with_args(ah_args.clone())
+				})
+		})
+		.with_global_settings(global_settings)
+		.build()
+		.map_err(config_errs)
+}
+
+fn westend_network_config() -> Result<NetworkConfig, anyhow::Error> {
+	let images = zombienet_sdk::environment::get_images_from_env();
```
bkontur: @rosarp please, bump here the images, so we see what actually is used, relates to: https://github.com/paritytech/parity-bridges-common/pull/3276/changes#r3466532467

testing/zombienet-sdk-tests/tests/environment.rs
```diff
@@ -0,0 +1,1632 @@
...
+	pub async fn spawn(init: bool, start_relayer: bool) -> Result<Self, anyhow::Error> {
+		let _ = env_logger::try_init_from_env(
+			env_logger::Env::default().filter_or(env_logger::DEFAULT_FILTER_ENV, "info"),
+		);
+
+		let spawn_fn = get_spawn_fn();
+		log::info!("Spawning Rococo network");
+		let rococo = spawn_fn(rococo_network_config()?).await?;
+		log::info!("Spawning Westend network");
+		let westend = spawn_fn(westend_network_config()?).await?;
```
bkontur: Would this optimize?

```suggestion
        // Rococo and Westend are independent networks; spawn them concurrently instead of serially
        // (saves ~one full network spawn — ~35s in local runs).
        log::info!("Spawning Rococo and Westend networks concurrently");
        let (rococo, westend) = tokio::try_join!(
                async { spawn_fn(rococo_network_config()?).await.map_err(|e| anyhow!("Rococo spawn failed: {e}")) },
                async { spawn_fn(westend_network_config()?).await.map_err(|e| anyhow!("Westend spawn failed: {e}")) },
        )?;
```

testing/zombienet-sdk-tests/tests/lib.rs
rosarp: @bkontur 
The relay-client codegen is runtime_types_only() — it has no tx()/storage() APIs
WDYT, if i modify runtime codegeneration to also include tx() & storage()
What are the consequences?
To run these tests tx()/storage() is necessary. if not available then using codegen_runtime.rs is not possible.

rosarp: Done. Removed .scale generator and used codegen tool instead.
But reusing existing codegen_runtime.rs is not possible as mentioned above.

testing/zombienet-sdk-tests/README.md
rosarp: Added CI to manually trigger zombienet tests.

testing/zombienet-sdk-tests/tests/environment.rs
```diff
@@ -0,0 +1,1664 @@
...
+	NodeImages {
+		polkadot: std::env::var("POLKADOT_IMAGE")
+			.unwrap_or_else(|_| DEFAULT_POLKADOT_IMAGE.to_string()),
+		cumulus: std::env::var("CUMULUS_IMAGE")
+			.unwrap_or_else(|_| DEFAULT_CUMULUS_IMAGE.to_string()),
+	}
+}
+
+fn rococo_network_config() -> Result<NetworkConfig, anyhow::Error> {
+	let images = node_images();
```
bkontur: @rosarp I think this was ok like this, we should be able just to set env var like POLKADOT_IMAGE / CUMULUS_IMAGE (or what exactly is get_images_from_env expecting), right? No need for custom NodeImages

```suggestion
	let images = zombienet_sdk::environment::get_images_from_env();
```

.github/workflows/zombienet.yml
```diff
@@ -0,0 +1,114 @@
...
+        uses: actions/checkout@v4
+
+      - name: Set variables
+        id: vars
+        run: |
+          cat .github/env >> "$GITHUB_OUTPUT"
+          REV="$(grep -oE 'polkadot-sdk\?branch=master#[0-9a-f]+' Cargo.lock | cut -d'#' -f2 | sort -u | head -1)"
+          [ -n "$REV" ] || { echo "ERROR: could not find the polkadot-sdk revision in Cargo.lock" >&2; exit 1; }
+          # The paritypr `*-debug` images are tagged with the short (8-char) commit.
+          echo "POLKADOT_SDK_HASH=${REV:0:8}" >> "$GITHUB_OUTPUT"
```
bkontur, line 38: @rosarp I think this POLKADOT_SDK_HASH version is used just for downloaded docker images for polkadot/polkadot-parachain, it does (probably should) not need to be parsed from PolkadotSDK used for relayer. It is orthogoval. Usually, the binaries are just for running chain. I would just put POLKADOT_IMAGE/CUMULUS_IMAGE=*/polkadot-debug:<required-hash>` to the `.env` and just read it from there

testing/zombienet-sdk-tests/tests/environment.rs
rosarp: @bkontur 
as per https://github.com/paritytech/zombienet-sdk/blob/7b2809fae14a8ddcce1ae06eb69c1fae208c410e/crates/sdk/src/environment.rs#L51
it defaults to 
```
const DEFAULT_POLKADOT_IMAGE: &str = "docker.io/parity/polkadot:latest";
```
which is not fast-runtime. Tests will timeout right?
It needs to be from `docker.io/paritypr/polkadot-debug` , but it doesnt have latest.
its format is either `#pr` or `master:#pr`. 
So using `#pr` is better choice right? Then why not use same polkadot-sdk hash from Cargo.lock?

Also which version to tag in env for POLKADOT_IMAGE / CUMULUS_IMAGE? 
if its not latest, which means, it needs to be manually bumped every now and then (if not derived automatically)

rosarp: As discussed, for now need to pin version to Cargo.lock hash of polkadot-sdk

.github/workflows/zombienet.yml
```diff
@@ -0,0 +1,127 @@
...
+          cache-on-failure: true
+          cache-all-crates: true
+
+      - name: Build substrate-relay
+        run: |
+          cargo fetch
+          CARGO_NET_OFFLINE=true SKIP_WASM_BUILD=1 cargo build --release -p substrate-relay
+          mkdir -p ./artifacts
+          mv -v ./target/release/substrate-relay ./artifacts/
+          cp -v /usr/lib/x86_64-linux-gnu/libssl.so.1.1 /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1 ./artifacts/
```
bkontur: @rosarp hmm, I've never done this hack, why? What is failing without it? Can you point me to the CI failed job with this?  I think we just need to instlal openssl or something, no?

I had also similar issue with standalone W3S provider binary with openssl: https://github.com/paritytech/web3-storage/pull/91 maybe also worth check (as separate follow up)

.github/workflows/zombienet.yml
```diff
@@ -0,0 +1,127 @@
...
+      - name: Upload substrate-relay
+        uses: actions/upload-artifact@v4
+        with:
+          name: substrate-relay
+          path: ./artifacts
+          retention-days: 2
+
+  # Runs on a Docker-capable runner (no container) so the docker provider can use the host Docker
+  # daemon and reach the spawned nodes on 127.0.0.1.
+  zombienet:
```
bkontur:
```suggestion
  run-tests:
```

.github/workflows/zombienet.yml
bkontur: @rosarp I think maybe the problem is that the bellow step is missing setting image:
```
container:
      image: ${{ needs.set-variables.outputs.CI_IMAGE }}
```

.github/workflows/zombienet.yml
```diff
@@ -0,0 +1,127 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          name: substrate-relay
+          path: ./bin
+
+      - name: Point the tests at the relay binary
+        run: |
+          chmod +x "$GITHUB_WORKSPACE/bin/substrate-relay"
+          echo "SUBSTRATE_RELAY_BINARY=$GITHUB_WORKSPACE/bin/substrate-relay" >> "$GITHUB_ENV"
+          echo "LD_LIBRARY_PATH=$GITHUB_WORKSPACE/bin${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}" >> "$GITHUB_ENV"
```
bkontur:
```suggestion
          # TODO: investiate
          # echo "LD_LIBRARY_PATH=$GITHUB_WORKSPACE/bin${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}" >> "$GITHUB_ENV"
```

.github/workflows/zombienet.yml
bkontur:
```suggestion
          # TODO: investiagate
          # cp -v /usr/lib/x86_64-linux-gnu/libssl.so.1.1 /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1 ./artifacts/
```

testing/zombienet-sdk-tests/tests/environment.rs
rosarp: @bkontur 
are you suggesting to put tests (`zombienet-sdk-tests/polkadot-kusama/tests`) in fellows CI?

bkontur: > @bkontur are you suggesting to put tests (`zombienet-sdk-tests/polkadot-kusama/tests`) in fellows CI?

not for now, we need to make them stable here, so I would add those tests directly here as a part of Bridges CI - basically, we just need chain-spec-builder and released wasm from Fellows repo to download or build here. Also I have a feeling that we will extract `zombienet-sdk-tests-lib` to `zombienet-sdk-bridges-tests-lib` and move them to the PolkadotSDK , so it will take some time to get that to the stable releases and bump in Fellows.

substrate-relay/Cargo.toml
```diff
@@ -21,6 +21,9 @@ log = { workspace = true }
 num-format = { workspace = true }
 num-traits = { workspace = true }
 rbtag = { workspace = true }
+# Pulled transitively by `relay-utils` via libcurl; declared here only so the `static-ssl` feature
+# (below) can statically link OpenSSL into the binary.
+isahc = { version = "1.7.2", optional = true }
```
bkontur, line 26: @rosarp nice, can we fix and simplify this in the PolkadotSDK relay-utils module?

.github/workflows/zombienet.yml
```diff
@@ -0,0 +1,125 @@
...
+          name: substrate-relay
+          path: ./artifacts
+          retention-days: 2
+
+  # Runs on a Docker-capable runner (no container) so the docker provider can use the host Docker
+  # daemon and reach the spawned nodes on 127.0.0.1.
+  run-tests:
+    name: Run zombienet tests
+    runs-on: parity-large
+    needs: [set-variables, build-relay]
```
bkontur: https://github.com/paritytech/parity-bridges-common/pull/3276#discussion_r3490262856

@rosarp can we add here just this to avouid that ssl issue?

```suggestion
    needs: [set-variables, build-relay]
    container:
      image: ${{ needs.set-variables.outputs.CI_IMAGE }}
```

rosarp: Running relayer in separate image will have issues. 
job container can't reach the host-daemon's published node ports on localhost.

rosarp: Because of this, needed to add isahc dep.

bkontur: > Because of this, needed to add isahc dep.

ok, let's go with this fix and we remove this feature flag hack when bumping PolkadotSDK including this fix: https://github.com/paritytech/polkadot-sdk/pull/12528 (this stuff is causing the ssl error, we don't need isahc at the end)

substrate-relay/Cargo.toml
bkontur: > [@rosarp](https://github.com/rosarp) nice, can we fix and simplify this in the PolkadotSDK relay-utils module?

https://github.com/paritytech/polkadot-sdk/pull/12528

substrate-relay/Cargo.toml
```diff
@@ -60,6 +63,13 @@ frame-support = { workspace = true }
 sp-core = { workspace = true }
 sp-runtime = { workspace = true }
 
+[features]
+# Statically link OpenSSL into the binary (via isahc/libcurl) so it has no `libssl.so` runtime
```
bkontur, line 68:
```suggestion
# TODO: remove when PolkadotSDK bump contains: https://github.com/paritytech/polkadot-sdk/pull/12528
# Statically link OpenSSL into the binary (via isahc/libcurl) so it has no `libssl.so` runtime
```

bkontur (approved):

PR #3280, opened 2026-07-02
Fix Deny CI job: install recent cargo-deny

bkontur:
## Problem

The `Deny` CI job has been failing on master ([example run](https://github.com/paritytech/parity-bridges-common/actions/runs/28587328627/job/84762090470)) with:

```
[ERROR] failed to load advisory database: parse error: error parsing .../libcrux-chacha20poly1305/RUSTSEC-2026-0124.md: TOML parse error at line 8, column 8
```

That advisory carries a **CVSS 4.0** vector, and the cargo-deny bundled in the pinned `ci-unified` image predates CVSS v4 support in the rustsec/cvss crates, so it aborts while loading the advisory database.

## Changes

- **`.github/workflows/ci.yml`**: install a current cargo-deny (0.19.9, SHA-pinned `taiki-e/install-action`) in the `deny` job instead of relying on the stale binary from the CI image. The gating `deny-licenses` job is deliberately untouched — cargo-deny 0.19.9 rejects the deprecated `GPL-3.0 WITH Classpath-exception-2.0` allowance in `deny.toml`, so bumping it there would turn the required job red.
- **`Cargo.lock`**: bump `tracing-subscriber` to 0.3.20 (+ its tracing chain) to fix [RUSTSEC-2025-0055](https://rustsec.org/advisories/RUSTSEC-2025-0055) — the only advisory in the tree fixable by a semver update.

## Notes

The parse failure was masking ~29 other advisories (wasmtime 8.0.1 via polkadot-sdk's `sc-executor-wasmtime`, rustls-webpki 0.102 via `rustls-platform-verifier`/jsonrpsee, hickory-proto 0.24 via libp2p, plus unmaintained-crate notices) that are not fixable from this repo. The `deny` job (which is `continue-on-error`) will now report them meaningfully instead of crashing; adding reasoned `ignore` entries for them in `deny.toml` can be done as a follow-up if we want the job fully green.

Verified: `cargo deny check advisories` and `check bans sources` run clean with 0.19.9 against this lockfile, and `SKIP_WASM_BUILD=1 cargo check --locked --workspace` passes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3221,8 +3221,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6e24cb5a94bcae1e5408b0effca5cd7172ea3c5755049c5f3af4cd283a165298"
 dependencies = [
  "bit-set",
- "regex-automata 0.4.8",
- "regex-syntax 0.8.5",
+ "regex-automata",
+ "regex-syntax",
 ]
 
 [[package]]
@@ -6139,11 +6139,11 @@ checksum = "ffbee8634e0d45d258acb448e7eaab3fce7a0a467395d4d9f228e3c1f01fb2e4"
 
 [[package]]
 name = "matchers"
-version = "0.1.0"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8263075bb86c5a1b1427b5ae862e8889656f126e9f77c484496e8b47cf5c5558"
+checksum = "d1525a2a28c7f4fa0fc98bb91ae755d1e2d1505079e05539e35bc876b5d65ae9"
 dependencies = [
- "regex-automata 0.1.10",
+ "regex-automata",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


Cargo.lock
```diff
@@ -6117,11 +6117,11 @@ checksum = "ffbee8634e0d45d258acb448e7eaab3fce7a0a467395d4d9f228e3c1f01fb2e4"
 
 [[package]]
 name = "matchers"
-version = "0.1.0"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8263075bb86c5a1b1427b5ae862e8889656f126e9f77c484496e8b47cf5c5558"
+checksum = "d1525a2a28c7f4fa0fc98bb91ae755d1e2d1505079e05539e35bc876b5d65ae9"
```
rosarp, line 6122: Do we need Cargo.lock updated in this PR?

bkontur: > Do we need Cargo.lock updated in this PR?

Yeah, probably not needed, but if tests pass, there shouldn't be a problem with this

rosarp (approved):

PR #3293, opened 2026-07-07
Enables the bridges zombienet-sdk integration tests

rosarp:
Enables the bridges zombienet-sdk integration tests (Rococo↔Westend asset transfer) to run against current runtimes. Two independent fixes are needed:

1. relay-substrate-client: use the caller's ambient Tokio runtime instead of creating a nested tokio::runtime::Runtime per client.
2. asset-hub-westend genesis: populate per-asset reserves for the bridged foreign assets that the local_testnet preset pre-registers.

Neither is zombienet-specific in nature — (1) is required for any Tokio-hosted relay, and (2) fixes a genesis/migration inconsistency for freshly-genesis'd Asset Hub Westend chains.

#### ntegration

- relay-substrate-client (behavioural). RpcClient no longer owns a dedicated tokio::runtime::Runtime; build_client now calls tokio::runtime::Handle::current(). Consumers must construct clients from within a Tokio runtime context. Binaries that previously drove the relay on a non-Tokio executor (e.g. async_std::task::block_on) must switch to a Tokio runtime (#[tokio::main]), otherwise Handle::current() panics with "there is no reactor running…". The upside: the caller's runtime is reused, and dropping a client no longer panics with "Cannot drop a runtime in a context where blocking is not allowed" on reconnect/restart.
- asset-hub-westend genesis (additive). Only the presets that pass foreign_assets (i.e. local_testnet) are affected; a reserves entry is now derived for each pre-registered foreign asset. No public API change; genesis JSON gains a foreignAssets.reserves array.

#### Review Notes

1. bridges/relays/client-substrate/src/client/rpc.rs — ambient runtime handle

RpcClient's ClientData stored an owned Arc<tokio::runtime::Runtime> created per-connection in build_client. That design assumes the relay itself does not run on Tokio (historically it ran on async-std). Two problems arise once the relay is Tokio-hosted:
- the caching client's background task uses ambient tokio::spawn, which needs an ambient runtime; and
- dropping the owned Runtime inside a Tokio task (on reconnect / loop restart) panics ("Cannot drop a runtime in a context where blocking is not allowed").

Switching to Handle::current() fixes both — a Handle is cheap to clone and safe to drop in async context, and data.tokio.spawn(...) keeps working unchanged.

2. cumulus/.../asset-hub-westend/src/genesis_config_presets.rs — genesis reserves

Asset Hub Westend moved its XCM reserve trust from hardcoded rules to per-asset pallet-assets::Reserves (type IsReserve = NonTeleportableAssetFromTrustedReserve<_, ForeignAssets>, no static bridging fallback). That storage is populated by ForeignAssetsReservesMigration on runtime upgrade. A chain booted from the local_testnet genesis preset runs no migration, yet the preset pre-registers the bridged ROC foreign asset — so its reserve is empty and inbound bridged reserve-transfers fail with UntrustedReserveLocation.

The preset now seeds reserves using the same provider the migration uses (AssetHubWestendForeignAssetsReservesProvider::reserves_for), keeping genesis and the upgrade migration consistent:


testing/zombienet-sdk-tests/tests/codegen/asset_hub_rococo.rs
```diff
@@ -16,10 +16,9 @@
 
 //! Autogenerated runtime API
 //! THIS FILE WAS AUTOGENERATED USING parity-bridges-common::runtime-codegen
-//! EXECUTED COMMAND: tools/runtime-codegen/target/debug/runtime-codegen --full --from-wasm-file
-//! testing/metadata-gen/polkadot-sdk/target/debug/build/
-//! bridges-zombienet-metadata-gen-f447c7795cc1eb3f/out/runtimes/x86_64-unknown-linux-gnu/release/
-//! wbuild/asset-hub-rococo-runtime/asset_hub_rococo_runtime.wasm
+//! EXECUTED COMMAND:
+//! tools/runtime-codegen/target/release/runtime-codegen --full --from-wasm-file
+//! target/release/wbuild/asset-hub-rococo-runtime/asset_hub_rococo_runtime.wasm
```
bkontur: @rosarp can you also regenerate relay-clients for rococo/westend here?

testing/zombienet-sdk-tests/tests/rococo_westend/environment.rs
```diff
@@ -439,22 +418,17 @@ impl BridgeTestEnv {
 				200_000_000,
 				12_000,
 			),
-			relay_westend::governance_transact_call(
```
bkontur, line 442: @rosarp just askin, can we modify genesis config with zombienet-sdk directly here, so we don't need to depend on PolkadotSDK config like this: https://github.com/paritytech/polkadot-sdk/pull/12587/changes#diff-1749e9f433677b2a4a80d6aea9a6103cbcead15fe6540d2bbcb75542b7332639R94 ?

rosarp: Fixed!

testing/zombienet-sdk-tests/tests/rococo_westend/environment.rs
```diff
@@ -73,6 +73,38 @@ fn relay_async_backing_override() -> serde_json::Value {
...
+	});
+	let asset_hub_rococo = serde_json::json!({
+		"parents": 2,
+		"interior": { "X2": [
+			{ "GlobalConsensus": { "ByGenesis": ROCOCO_GENESIS_HASH } },
+			{ "Parachain": ASSET_HUB_PARA_ID },
+		] },
+	});
+	serde_json::json!({
+		"foreignAssets": {
```
bkontur, line 103: @rosarp nice nice, please add here also registering foreign asset basically what genesis preset does or `force_create_foreign_asset_call` - so we will have this under control here by test, don't we need this for opposite side westend -> rococo?

testing/zombienet-sdk-tests/tests/rococo_westend/ops.rs
```diff
@@ -253,10 +253,11 @@ macro_rules! asset_hub_ops {
...
 			/// SCALE-encoded `ForeignAssets::force_create(bridged_asset, owner, is_sufficient,
-			/// min_balance)` call, wrapped in a relay-chain governance `Transact` (root). At this runtime
-			/// revision the bridged asset is not pre-registered at genesis, so it is created here before
-			/// the bridge can mint it; reserve trust is static in the runtime's XCM config, so no
-			/// per-asset reserve registration is needed.
+			/// min_balance)` call, wrapped in a relay-chain governance `Transact` (root).
+			// Unused at this runtime revision: the bridged asset is pre-registered at genesis (owned by
+			// the bridged network's sovereign account), so `force_create` fails with `InUse`. Kept for
+			// runtime revisions that don't pre-register it.
+			#[allow(dead_code)]
```
bkontur, line 260: @rosarp I would either make them as a part of the `common` and reuse for P/K and R/W setup or remove (for R/W we don't need them, because we create them in genesis and for P/K, we can add them just there)

testing/zombienet-sdk-tests/tests/codegen/asset_hub_rococo.rs
rosarp: Regenerated & updated spec_versions

Cargo.lock
```diff
@@ -800,33 +747,6 @@ dependencies = [
  "windows-sys 0.52.0",
 ]
 
-[[package]]
-name = "async-std"
```
bkontur, line 804: @rosarp looks like we can also remove some deps: async-std, signal-hook, signal-hook-async-std from the main Cargo.toml

tools/runtime-codegen/Cargo.toml
```diff
@@ -18,7 +18,7 @@ color-eyre = "0.6.1"
 indoc = "2.0.5"
 prettyplease = "0.2.20"
 proc-macro2 = "1.0.56"
-subxt-codegen = { version = "0.50.1" }
+subxt-codegen = { version = "0.44.0" }
```
bkontur, line 21: @rosarp why downgrade?

substrate-relay/Cargo.toml
```diff
@@ -11,7 +11,6 @@ workspace = true
 
 [dependencies]
 anyhow = { workspace = true }
-async-std = { workspace = true }
```
bkontur, line 14: @rosarp this is what I meant by https://github.com/paritytech/parity-bridges-common/pull/3293/changes#r3557781534, we could remove from main Cargo.toml, right?

tools/runtime-codegen/Cargo.toml
rosarp: zombienet-sdk is using 0.44. 0.44 is not compatible with 0.50.1. 
and our test's calls depend on zombienet-sdk's api. so both need to be on same version

.github/workflows/ci.yml
```diff
@@ -183,6 +183,14 @@ jobs:
       - name: Checkout sources
         uses: actions/checkout@v4
 
+      # the cargo-deny bundled in the CI image is too old to parse the modern SPDX
+      # license identifiers (e.g. `GPL-3.0-only`) that polkadot-sdk crates now declare,
+      # so install the same recent one the `Deny` job uses
+      - name: Install cargo-deny
```
bkontur, line 189: @rosarp we are doing this twice for `deny` step also, maybe deduplicate with common step, but as follow-up if make sense

bkontur (approved):

PR #3299, opened 2026-07-13
Migrate the Polkadot↔Kusama asset-transfer bridge test to zombienet-sdk

rosarp:
Ports the manual, shell-script-driven Polkadot↔Kusama bridge integration test (formerly in `polkadot-fellows-runtimes`, `integration-tests/bridges/tests/0001-polkadot-kusama-asset-transfer`) to a native Rust zombienet-sdk integration test under `testing/zombienet-sdk-tests`, alongside the existing Rococo↔Westend suite.

The new `kusama_polkadot::asset_transfer` test spawns both relay-chain networks (each with a Bridge Hub and Asset Hub), drives `substrate-relay` as subprocesses, and asserts that DOT/KSM transfer across the bridge in both directions (wrapped out, unwrapped back), the message relayers are rewarded, and the free-header relayers' balances stay constant.

In the process the shared test infrastructure was generalized so both bridge pairs reuse one implementation: the per-runtime typed-operation macros were extracted from `rococo_westend/ops.rs` into `common/ops.rs` (generalized over the remote consensus `NetworkId`), and the pair-agnostic network-spawn / genesis-override helpers were hoisted into `common/utils.rs`.

### Integration

Running the new test is native-only and requires the fellows chain-spec-generator (one binary per relay) plus the `polkadot/polkadot-parachain/substrate-relay` binaries in `~/local_bridge_testing/bin` (or via the *_BINARY env vars). See the updated `testing/zombienet-sdk-tests/README.md` for build/run steps and for regenerating the Polkadot/Kusama codegen from fellows runtime WASM:

```
ZOMBIE_PROVIDER=native \
  cargo test -p bridges-zombienet-sdk-tests --features zombie-ci kusama_polkadot -- --nocapture
```

### Review Notes

The Polkadot/Kusama runtimes live in polkadot-fellows-runtimes (not polkadot-sdk) and have no sudo pallet, which drives the two main differences from the Rococo↔Westend pair:

- **Native provider + chain-spec-generator.** The env spawns nodes with the native provider and generates chain specs via the fellows chain-spec-generator (`with_chain_spec_command`), rather than polkadot-sdk docker images.
- **Sudo-free bootstrap — genesis where possible, permissionless signed calls otherwise.** Set at genesis via `with_genesis_overrides`: the local safe + remote XCM versions (`polkadotXcm.safeXcmVersion` and `supportedVersion` — the sudo-free equivalent of `force_xcm_version`), the trusted reserve for the bridged foreign asset (`foreignAssets.reserves`, standing in for the owner's `set_reserves`), the bridge-GRANDPA pallet owner (`//Alice`, so `init-bridge` is owner-signed), and the funded sovereign/reward accounts. Two steps run post-spawn as ordinary signed extrinsics (still no sudo): the HRMP channels via the permissionless `Hrmp::establish_system_channel` (`//Alice`; both endpoints are system paras) and the `//Bob`-signed asset-conversion pools. HRMP is opened post-spawn rather than via genesis `preopenHrmpChannels`, which left the Bridge Hub with a downward-message-queue MQC-head mismatch that stalled block import.

Other pair-specific details, all verified against the freshly generated codegen: lane 0x00000001; DOT=10 / KSM=12 decimals; named NetworkId::Polkadot/Kusama (not ByGenesis); and asymmetric reward keys — Kusama BH uses a plain RewardsAccountParams, Polkadot BH wraps it in BridgeReward::PolkadotKusamaBridge.

Shared-infrastructure refactors (both behavior-preserving; R↔W call sites updated accordingly):

- `rococo_westend/ops.rs` → `common/ops.rs`: the ops take the remote network as a macro argument (`NetworkId::ByGenesis(..)` for R↔W, `NetworkId::Polkadot`/`Kusama` for K↔P), which let me drop the per-call genesis-hash arguments.
- `common/utils.rs`: `spawn_with_retry` (now generic over the provider spawn fn — docker `get_spawn_fn()` vs native `Provider::Native.get_spawn_fn()`), `bridge_hub_balances`, `global_settings`, `config_errs` and `wait_for_native_increase` are now shared by both pairs; three dead helpers with no callers were removed.

New codegen: six --full subxt clients under tests/codegen/ (polkadot, kusama, and their Asset/Bridge Hubs), generated from fellows runtime WASM via scripts/generate-codegen.sh --wasm-dir.

### Validation

`cargo check`, `cargo clippy`, nightly `rustfmt --check` and `cargo spellcheck` all pass under `--features zombie-ci` (R↔W + K↔P together). The K↔P `asset_transfer` test has been run **end-to-end on native binaries and passes** (both directions: wrapped out, unwrapped back, relayers rewarded, free-header relayers unchanged); both pairs are wired into the zombienet CI workflow.

The two runtime-behavior questions from the initial port are now resolved on that live run:

- genesis `safeXcmVersion = 5` alone does **not** suffice — the remote `supportedVersion` must also be registered at genesis, otherwise the bridged transfer fails with `SendFailure`; and
- the `-local` Asset Hub preset does **not** pre-register the bridged asset's reserve, so `foreignAssets.reserves` is seeded at genesis to avoid inbound reserve-transfers failing with `UntrustedReserveLocation`.


bkontur:
@rosarp you need to add those P/K tests here: https://github.com/paritytech/parity-bridges-common/blob/64005cc8cd2fe698640d8665949f7bacf8722a91/.github/workflows/zombienet.yml#L144-L147

.github/workflows/zombienet.yml
```diff
@@ -129,24 +129,80 @@ jobs:
...
+  build-chain-spec-generator:
+    name: Build fellows chain-spec-generator
+    runs-on: parity-large
+    needs: [set-variables]
+    container:
+      image: ${{ needs.set-variables.outputs.CI_IMAGE }}
+    env:
+      # Fellows release the committed Kusama <> Polkadot codegen was generated from. Bump this (and
+      # regenerate the codegen — see testing/zombienet-sdk-tests/README.md) together.
+      FELLOWS_RUNTIMES_TAG: v2.3.1
```
bkontur: please, move to the .github/env

.github/env
```diff
@@ -1,3 +1,4 @@
 IMAGE=docker.io/paritytech/ci-unified:bullseye-1.88.0-2025-06-27-v202507112050
 IMAGE_REGISTRY=docker.io/paritypr
 ZOMBIE_PROVIDER=docker
+FELLOWS_RUNTIMES_TAG=v2.3.1
```
bkontur, line 4: @rosarp I am just thinking how this will work, for release workflow, where we create per BHK or BHP, for example when we do a release branch just for BHK, this tests should be running just with BHK v2.3.3 version and BHP v2.3.1 version, please, create a follow up issue/PR for this, maybe we just need to split this version for both: BHP_FELLOWS_RUNTIMES_TAG / BHK_FELLOWS_RUNTIMES_TAG?

testing/zombienet-sdk-tests/tests/kusama_polkadot/environment.rs
```diff
@@ -0,0 +1,709 @@
...
+	let remote_asset_hub = serde_json::json!({ "parents": 2, "interior": { "X2": [
+			{ "GlobalConsensus": remote_network },
+			{ "Parachain": ASSET_HUB_PARA_ID },
+		] } });
+	serde_json::json!({
+		"polkadotXcm": {
+			"safeXcmVersion": XCM_VERSION,
+			"supportedVersion": supported_remote_version(remote_network, ASSET_HUB_PARA_ID),
+		},
+		"foreignAssets": {
```
bkontur, line 145: @rosarp please, as an follow-up, clean up also R/W not to use governance call, but just modify genesis on startup according to the P/K setting (we don't need sudo) + remove all the sudo batch_calls stuff

testing/zombienet-sdk-tests/tests/lib.rs
```diff
@@ -26,6 +26,13 @@ runtime_codegen!(asset_hub_westend_codegen, "codegen/asset_hub_westend.rs");
 runtime_codegen!(bridge_hub_rococo_codegen, "codegen/bridge_hub_rococo.rs");
 runtime_codegen!(bridge_hub_westend_codegen, "codegen/bridge_hub_westend.rs");
 
+runtime_codegen!(polkadot_codegen, "codegen/polkadot.rs");
```
bkontur, line 29: @rosarp please, make sure that we generate this zombienet codegen when doing release (update release pipeline to generate those, if not already done)

bkontur (approved):

